PR #21, opened 2020-07-29
Support path scanning repos

Jguer:
- Add support for scanning repositories already cloned

### DIFF
--- a/ggshield/dev_scan.py
+++ b/ggshield/dev_scan.py
@@ -76,7 +76,7 @@ def repo_cmd(ctx: click.Context, repository: str) -> int:  # pragma: no cover
             shell(["git", "clone", repository, tmpdirname])
             return scan_repo_path(client, config, tmpdirname)
 
-    raise click.ClickException(f"{repository} is neither a valid path or git URL")
+    raise click.ClickException(f"{repository} is neither a valid path nor a git URL")
 
 
 @click.command()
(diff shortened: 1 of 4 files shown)


ggshield/dev_scan.py
```diff
@@ -25,37 +26,50 @@ def cd(newdir):
...
+
+    ggshield scan repo /repositories/gg-shield
     """
-    config = ctx.obj["config"]
-    try:
+    config: Config = ctx.obj["config"]
+    client: GGClient = ctx.obj["client"]
+    if os.path.isdir(repository):
+        return scan_repo_path(client, config, repository)
+    else:
```
eugenenelou, line 69: We could check if the argument is a valid URI before trying to clone, to give the user a more explicit error message, like:
`Given argument is neither a valid path nor a valid URL`

ggshield/dev_scan.py
```diff
@@ -66,11 +70,14 @@ def repo_cmd(ctx: click.Context, repository: str) -> int:  # pragma: no cover
...
     if os.path.isdir(repository):
         return scan_repo_path(client, config, repository)
-    else:
+
+    if REGEX_GIT_URL.match(repository):
         with tempfile.TemporaryDirectory() as tmpdirname:
             shell(["git", "clone", repository, tmpdirname])
             return scan_repo_path(client, config, tmpdirname)
 
+    raise click.ClickException(f"{repository} is neither a valid path or git URL")
```
eugenenelou, line 79: nit: 
```suggestion
    raise click.ClickException(f"{repository} is neither a valid path nor a git URL")
```

PR #27, opened 2020-09-03
fix(git): do not use cwd git

### LINKED ISSUE #24: Remote Code execution on windows
Run **ggshield scan** in a directory/Repository with malicious git.exe in it on Windows

### Actual Behavior
It executes malicious git.exe in the repo while scanning

### Expected Behavior
It should not execute code/binary from repo unintentionally

Jguer:
Closes #24

### DIFF
--- a/ggshield/git_shell.py
+++ b/ggshield/git_shell.py
@@ -1,11 +1,26 @@
+import os
 import subprocess
+from functools import lru_cache
+from shutil import which
 from typing import Any, List
 
 import click
 
 
 COMMAND_TIMEOUT = 45
-GIT_PATH = "git"
+
+
+@lru_cache(None)
+def get_git_path(cwd: str) -> str:
+    git_path = str(which("git"))
+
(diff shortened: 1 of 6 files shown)


ggshield/git_shell.py
```diff
@@ -1,11 +1,26 @@
...
 
 COMMAND_TIMEOUT = 45
-GIT_PATH = "git"
+
+
+@lru_cache(None)
+def get_git_path(cwd: str) -> str:
+    git_path = str(which("git"))
+
+    if cwd in git_path:
```
dcRUSTy, line 17: `if cwd in git_path && cwd not in env PATH variable` would solve edge case as well when cwd is also in PATH explicitly

eugenenelou (approved):

ggshield/git_shell.py
Jguer: Added that small edge case in the latest commit, thanks

PR #37, opened 2021-01-21
feat(config): Accept dict as matches ignore

boblefrag:


Jguer:
Thanks @boblefrag , LGTM

PR #49, opened 2021-03-25
feat(mode_header): add mode header

SofienM:


Jguer:
Merged with a few modifications in #50

PR #53, opened 2021-04-27
Proposed fix for Issue 52

rgajason:
This is in reference to https://github.com/GitGuardian/gg-shield/issues/52

Don't throw exception when CWD is /
Do throw an exception when git is not found
Update exception message to indicate situation (not found vs rejected)
Enhance PATH/CWD comparisons to avoid symlinks from causing issues

### DIFF
--- a/ggshield/git_shell.py
+++ b/ggshield/git_shell.py
@@ -21,7 +21,9 @@ def get_git_path(cwd: str) -> str:
     # insensitive filesystems but detection is problematic
     git_path = os.path.abspath(git_path)
     cwd = os.path.abspath(cwd)
-    path_env = list(map(os.path.abspath, os.environ.get("PATH", "").split(os.pathsep)))
+    path_env = [
+        os.path.abspath(p) for p in os.environ.get("PATH", "").split(os.pathsep)
+    ]
 
     # git was found - ignore git in cwd if cwd not in PATH
     if cwd == os.path.dirname(git_path) and cwd not in path_env:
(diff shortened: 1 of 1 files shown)


rgajason:
Regarding the lint fail, that appears to be a mypy bug.

Jguer (changes requested):
Hey @rgajason , I've pulled it locally and tried out a few scenarios and everything seems to be working fine, I think we only need to fix that small issue with the linter and it should be fine

ggshield/git_shell.py
```diff
@@ -12,11 +12,21 @@
...
 
-    if cwd in git_path and cwd not in os.environ.get("PATH", "").split(os.pathsep):
+    if git_path is None:
         raise Exception("unable to find git executable in PATH/PATHEXT")
 
+    # lower()ing these would provide additional coverage on case-
+    # insensitive filesystems but detection is problematic
+    git_path = os.path.abspath(git_path)
+    cwd = os.path.abspath(cwd)
+    path_env = list(map(os.path.abspath, os.environ.get("PATH", "").split(os.pathsep)))
```
Jguer, line 24: It does seem to be a linter issue (https://github.com/python/mypy/issues/9864) 

I see two solutions
```py
path_env = list(map(os.path.abspath, os.environ.get("PATH", "").split(os.pathsep)))  # type: ignore # noqa
```
mypy ignore it and then ignore the too long line

or

```py
path_env = [
        os.path.abspath(path) for path in os.environ.get("PATH", "").split(os.pathsep)
    ]
```

use a list comprehension, I'm not too worried about the performance impact of either since this is not bottleneck sensitive code.

Up to you how you want to change it :slightly_smiling_face:

rgajason: Updated and pushed - looks like mypy is happy now.

Jguer (approved):

Jguer:
All good now, merging it and letting it steep a bit in `next`. Thanks for your help

PR #57, opened 2021-05-05
Azure Pipelines Integration

er-vin:
This heavily relies on the BUILD_BUILDID environment variable to detect the use of Azure Pipelines. Also use BUILD_SOURCEVERSION to know the head commit. This will work nicely to detect ranges in PRs, not so much for pushes in main branches (there only the latest commit will be scanned).


Jguer (approved):
Seems to be all fine. Thanks for the effort @er-vin

PR #74, opened 2021-06-21
Jeremy/ /quota overview cmd

jeremyds:


ggshield/quota.py
```diff
@@ -0,0 +1,49 @@
...
+
+@click.command()
+@click.pass_context
+def quota(ctx: click.Context, json_output: bool) -> int:
+    """Command to show quotas overview."""
+    api_key: Optional[str] = os.getenv("GITGUARDIAN_API_KEY")
+    base_uri: str = os.getenv("GITGUARDIAN_API_URL", ctx.obj["config"].api_url)
+
+    if not api_key:
+        raise click.ClickException("GitGuardian API Key is needed.")
```
Jguer, line 22: this has so much in common with the opening statement of scan we could refactor both into calls of a method like:

```py
def retrieve_client(ctx: click.Context):
    api_key = os.getenv("GITGUARDIAN_API_KEY")
    base_uri = os.getenv("GITGUARDIAN_API_URL", ctx.obj["config"].api_url)
    if not api_key:
        raise click.ClickException("GitGuardian API Key is needed.")

    ctx.obj["client"] = GGClient(
        api_key=api_key, base_uri=base_uri, user_agent="ggshield", timeout=60
    )
```

Jguer (commented):
Overall very good, I think no change besides that small refactoring is needed. The typing is excellent.

Jguer (approved):

PR #75, opened 2021-06-22
Make dockerfile/buildargs specific

Jguer:
- Make dockerfile/buildargs specific in the detection
- Easiest way without changing output logic

Example:

docker-scan

### DIFF
--- a/ggshield/scan/docker.py
+++ b/ggshield/scan/docker.py
@@ -83,7 +83,7 @@ def _get_config(archive: tarfile.TarFile) -> Tuple[Dict, Dict, File]:
         json.loads(config_file_content),
         File(
             config_file_content,
-            filename=os.path.join(archive.name, config_file_path),  # type: ignore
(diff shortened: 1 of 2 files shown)


ggshield/scan/docker.py
```diff
@@ -83,7 +83,7 @@ def _get_config(archive: tarfile.TarFile) -> Tuple[Dict, Dict, File]:
         json.loads(config_file_content),
         File(
             config_file_content,
-            filename=os.path.join(archive.name, config_file_path),  # type: ignore
+            filename="Dockerfile or build-args",  # noqa: E501
```
jeremyds, line 86: I confident it works like that but it seems so hacky to me to put an expression inside what is expected to be a filename. What is the build-args here?

jeremyds (approved):

ggshield/scan/docker.py
Jguer: I know, we'll make an issue to also make files found in a layer easier to link to the command that generated them

PR #80, opened 2021-06-29
Jeremy/62/app secrets engine version

jeremyds:


Pipfile.lock
```diff
@@ -1,734 +1,734 @@
 {
```
Jguer, line 1: :art: format this file please, I should add this one into the prettier config

Jguer (changes requested):

Jguer (approved):

PR #85, opened 2021-07-02
fix(json): fix line display on JSON output

### LINKED ISSUE #83: Scan line is offset by one in json scanning

- [x] I can reproduce this bug in the latest version

**Command executed**

`ggshield scan --json path x.py`

**Describe the bug**

line is offset by 1 :facepalm: 

**Expected behavior**

line is not offset by 1 :shrug:

**Screenshots**

image

Jguer:
Fix #83 line offset

### DIFF
--- a/ggshield/output/json/json_output.py
+++ b/ggshield/output/json/json_output.py
@@ -106,7 +106,7 @@ class JSONHandler(OutputHandler):
             "total_occurrences": len(policy_breaks),
         }
         for policy_break in policy_breaks:
-            update_policy_break_matches(policy_break.matches, lines, is_patch)
+            update_policy_break_matches(policy_break.matches, lines, is_patch, True)
             flattened_dict["occurrences"].extend(policy_break.matches)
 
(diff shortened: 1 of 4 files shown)


ggshield/output/json/json_output.py
```diff
@@ -106,7 +106,7 @@ def flattened_policy_break(
             "total_occurrences": len(policy_breaks),
         }
         for policy_break in policy_breaks:
-            update_policy_break_matches(policy_break.matches, lines, is_patch)
+            update_policy_break_matches(policy_break.matches, lines, is_patch, True)
```
jeremyds, line 109: I don't get why you did not need to set this parameter to True for non json output

jeremyds:
Is not it worth to add a test?

ggshield/output/json/json_output.py
jeremyds: ok, seen with you

jeremyds (approved):

PR #90, opened 2021-07-13
fix(docker): handle incomplete docker config

jeremyds:


ggshield/scan/docker.py
```diff
@@ -117,7 +117,7 @@ def _should_scan_layer(layer_info: Dict) -> bool:
     Only COPY and ADD layers should be scanned.
     """
     cmd = layer_info["created_by"]
-    return LAYER_TO_SCAN_PATTERN.search(cmd) is not None
+    return LAYER_TO_SCAN_PATTERN.search(cmd) is not None if cmd else True
```
jeremyds, line 120: @Jguer this part is important. I think we want to scan the layer all the time when we re not able to look after a COPY or SCAN in the empty `created_by` field. Are you ok with that or you would prefer to skip them?

Jguer: @jeremyds it should be ok since the banlists will catch a lot of the unimportant files, but we'll have to wait and see if there's certain images that become to slow to scan

Jguer (approved):

PR #98, opened 2021-08-13
feat(output): display a spinner while saving docker image

jeremyds:


ggshield/docker.py
```diff
@@ -50,12 +51,13 @@ def docker_save_to_tmp(image_name: str, temporary_path: str) -> Path:
...
-            stderr=subprocess.PIPE,
-            timeout=DOCKER_COMMAND_TIMEOUT,
-        )
+        with yaspin(text="Saving docker image"):
+            subprocess.run(
+                command,
+                check=True,
+                stderr=subprocess.PIPE,
+                timeout=DOCKER_COMMAND_TIMEOUT,
+            )
```
jeremyds, line 60: If we want to persist "Saving docker image" on output and add a "ok" once done, we can do:
```
with yaspin(text="Saving docker image") as spinner:
    subprocess.xxx
    spinner.ok()
```
I preferred to make this text disappear once done, but you can try it if you want.

Jguer (approved):

Issue #65, opened 2021-05-20
Add --exclude as a repeatable option to ignore files


Jguer:
**Is your feature request related to a problem? Please describe.**

Add --exclude as a repeatable option to ignore files from a scan without using `.gitguardian.yaml`

**Describe the solution you'd like**

- [  ] Add --exclude as a repeatable option to `scan` command
- [  ] Populate banlist patterns with these (merge with config's if available)
- [  ] Test that both are applied

PR #111, opened 2021-10-05
Add --exclude option

agateau-gg:
## Description

This PR adds a `--exclude <PATH>` option to `ggshield scan`, making it possible to exclude files from the command-line. This implements #65.

## What has been done

- Made a fix in Config to ensure keys of type `set` are not turned into `list`.
- Changed path_cmd() to get its set of paths to ignore from the context instead of directly from the Config instance, making it possible for its caller to add new paths to the ignore set.
- Added the `--exclude` option
- Actually unrelated: added `TEST_GITGUARDIAN_API_KEY` to `.env.sample` because at some point I added a vcr-based test and I couldn't understand why the test would not pass. I eventually removed that test, but I think the fix is still useful

## Questions

1. Would it make sense to call the option `--ignore` or `--paths-ignore` to match more closely with the config option?

2. The fact that `path_filter_set()` is called earlier changes the way ggshield ignores files. Here is what happens:

Given a repository with this content:
```
dir/
    secret.yaml
secret.yaml
.gitguardian.yaml
```

and a .gitguardian.yaml with this content:

```yaml
paths-ignore:
  - secret.yaml
  - .cache_ggshield
  - .gitguardian.yaml
```

Using ggshield from main and from this PR the behavior differs:

| branch | command | result |
| - | - | - |
| main |`ggshield scan -v path -r .` | scans dir/secret.yaml |
| main |`ggshield scan -v path -r dir` | ignores dir/secret.yaml |
| this branch |`ggshield scan -v path -r .` | scans dir/secret.yaml |
| this branch |`ggshield scan -v path -r dir` | scans dir/secret.yaml |

I think the new behavior makes it easier to have the same behavior regarding of which subset of the source tree you scan, but it might break things for some users. Should it be kept that way?


.env.example
```diff
@@ -1,2 +1,3 @@
 GITGUARDIAN_API_KEY=__________FILL ME__________
+TEST_GITGUARDIAN_API_KEY=__________FILL ME__________
```
Jguer, line 2: this is supposed to be copied by an user and not by a developer. 

I think it may be better to add it to the https://github.com/GitGuardian/ggshield/blob/main/CONTRIBUTING.md

PR #137, opened 2021-11-09
fix(patch_parsing): lines starting with backslash

sapristi:
Handles bug encountered due to the presence of `\ No newline at end of file` in patches, that were not handled

### DIFF
--- a/ggshield/utils.py
+++ b/ggshield/utils.py
@@ -147,6 +147,9 @@ def get_lines_from_patch(content: str, filemode: Filemode) -> Iterable[Line]:
             line_pre_index = pre_index
             line_content = line[1:]
             category = LineCategory.deletion
+        elif line_type == "\\":
+            # This type of line should'nt contain any secret; no need to set indices
+            line_content = line[1:]
 
         if line_type and line_content is not None:
             yield Line(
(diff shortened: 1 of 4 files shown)


ggshield/utils.py
```diff
@@ -147,6 +147,8 @@ def get_lines_from_patch(content: str, filemode: Filemode) -> Iterable[Line]:
             line_pre_index = pre_index
             line_content = line[1:]
             category = LineCategory.deletion
+        elif line_type == "\\":
+            line_content = line[1:]
```
agateau-gg, line 151: This makes sense, but can you add a test for it?

sapristi: Added a supplementary test case for the `make_indices`.
There is no test yet for the `get_lines_from_file` function.

agateau-gg: Looks good to me :+1:.

sapristi:
@jeremyds can you review and merge if appropriate ?

ggshield/utils.py
agateau-gg: Actually something is odd: codecov says [your code is not reached by the test suite](https://app.codecov.io/gh/GitGuardian/ggshield/compare/137/diff?src=pr&el=tree). Can you look into this?

sapristi: Seems fixed with latest changes

agateau-gg (approved):

PR #139, opened 2021-11-17
fix(filter): stop to list FS to exclude files to scan

jeremyds:
ggshield supports to exclude some files from scan using exclude
patterns.  Those patterns can be set using cli arguments, configuration
file and there are default excluded patterns as well.

We used to list all the files matching those patterns using glob in
order to filter out files to scan. We recently added some default
patterns to ignore and it leds us to realize the performance issue with
this approach: FS listing is very slow and it does not scale.

We now use regex to check if a file path is included in the exclude
patterns.

### DIFF
--- a/ggshield/filter.py
+++ b/ggshield/filter.py
@@ -3,13 +3,20 @@ import math
 import operator
 import re
 from collections import OrderedDict
-from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Set
 
+import click
 from pygitguardian.models import Match, PolicyBreak, ScanResult
 
 
 REGEX_MATCH_HIDE = re.compile(r"[^+\-\s]")
+REGEX_SPECIAL_CHARS = set(".^$+*?{}()[]\\|")
+INVALID_PATTERNS_REGEX = re.compile(
+    r"(\*\*\*)"  # the "***" sequence is not valid
+    r"|(\*\*[^/])"  # a "**" sequence must be immediately followed by a "/"
+    r"|([^/]\*\*)"  # a "**" sequence must be either at the start of the string or
+    # immediately preceded by a "/"
+)
 
 MAXIMUM_CENSOR_LENGTH = 60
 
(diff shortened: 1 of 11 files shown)


ggshield/filter.py
```diff
@@ -118,19 +125,57 @@ def leak_dictionary_by_ignore_sha(
...
+            raise click.ClickException(f"{path} is not a valid exclude pattern.")
+        res.add(translate_user_pattern(path))
+    return res
+
+
+def is_filepath_excluded(filepath: str, excluded_patterns: Set[str]) -> bool:
+    for regex in excluded_patterns:
+        if re.search(regex, filepath):
+            return True
+    return False
```
agateau-gg, line 178: nit: can be simplified to:
```suggestion
    return any(re.search(r, filepath) for r in excluded_patterns)
```

agateau-gg (commented):
To make the code easier to read I suggest renaming `excluded_patterns` to `excluded_regexes` when we are working with regexes.

Also, it might make sense to compile the regexes and then store them as `Set[re.Pattern]` instead of `Set[str]`. Would be faster and a bit more type safe as a bonus.

ggshield/filter.py
jeremyds: I don't like the `any` syntax, I feel it is not more readable. But it is clearly more _pythonic_, let's do that ;)

jeremyds:
@agateau-gg about the renaming, `excluded_patterns` to `excluded_regexes`:
`excluded_regexes` makes me think we are excluding regexes, whereas we re excluding files based on patterns. Then I agree with you `excluded_patterns` is not clear enough. Would `excluded_patterns_regex` makes everything clearer?

> Also, it might make sense to compile the regexes and then store them as Set[re.Pattern] instead of Set[str]. Would be faster and a bit more type safe as a bonus.

:+1:

jeremyds:
Joao input about variable naming: `exclusion_patterns`

ggshield/filter.py
```diff
@@ -118,19 +125,54 @@ def leak_dictionary_by_ignore_sha(
...
+def translate_user_pattern(pattern: str) -> str:
+    """
+    Translate the user pattern into a regex. This function assumes that the given
+    pattern is valid and has been normalized beforehand.
+    """
+
+    # Escape each special character
+    pattern = "".join(
+        f"\\{char}" if char in REGEX_SPECIAL_CHARS else char for char in pattern
+    )
```
agateau-gg, line 137: We can use `re.escape()` here I think (and remove REGEX_SPECIAL_CHARS):
```suggestion
    pattern = re.escape(pattern)
```

ggshield/filter.py
```diff
@@ -118,19 +125,54 @@ def leak_dictionary_by_ignore_sha(
...
+    if pattern[-1] != "/":
+        pattern += "$"
+    if pattern[0] == "/":
+        pattern = "^" + pattern[1:]
+    else:
+        pattern = "(^|/)" + pattern
+
+    # Replace * and ** sequences
+    pattern = re.sub(r"\\\*\\\*/", "([^/]+/)*", pattern)
+    pattern = re.sub(r"\\\*", "([^/]+)", pattern)
```
agateau-gg, line 149: We can use plain `str.replace()` calls here, to save a few back-slashes and a few msec :)
```suggestion
    pattern = pattern.replace(r"\*\*/", "([^/]+/)*")
    pattern = pattern.replace(r"\*", "([^/]+)")
```

agateau-gg (changes requested):
Made a few minor suggestions, but looks good :+1: .

Jguer (approved):

PR #143, opened 2021-11-22
Only create/update cache if there are new entries

Jguer:
- ggshield always creates/updates cache even if there are no results. Modify this behavior so `.cache_ggshield` is only written if there are entries in the cache

- also fix tests writing to the same file in the fs. they should run in isolation now with `pyfakefs`


tests/conftest.py
```diff
@@ -454,3 +455,12 @@ def cli_runner():
...
     with cli_runner.isolated_filesystem():
         yield cli_runner
+
+
+@pytest.fixture(scope="function")
+def isolated_fs(fs):
+    # isolate fs but include CA bundle for https validation
+    fs.add_real_directory(os.path.dirname(extract_zipped_paths(DEFAULT_CA_BUNDLE_PATH)))
+    # add cassetes dir
+    cassetes_dir = join(dirname(realpath(__file__)), "cassettes")
```
agateau-gg, line 465: typo: casset**t**es

tests/test_config.py
```diff
@@ -172,12 +172,20 @@ def test_save_cache(self, cli_fs_runner):
             file_content = json.load(file)
             assert file_content == {"last_found_secrets": ["XXX"]}
 
+    @pytest.mark.parametrize("with_entry", [True, False])
     @patch("ggshield.config.Config.CONFIG_LOCAL", [".gitguardian.yml"])
-    def test_save_cache_first_time(self, cli_fs_runner):
-        os.remove(".cache_ggshield")
+    def test_save_cache_first_time(self, isolated_fs, with_entry):
+        """
+        GIVE no existing cache
```
agateau-gg, line 179: typo
```suggestion
        GIVEN no existing cache
```

ggshield/config.py
```diff
@@ -280,6 +280,10 @@ def to_dict(self) -> Dict[str, Any]:
         return _cache
 
     def save(self) -> bool:
+        if len(self.last_found_secrets) == 0:
```
agateau-gg, line 283: PEP 8 says this should be `if not self.last_found_secrets:`

agateau-gg (changes requested):
Minor remarks, looks good otherwise.

tests/conftest.py
Jguer: ? seems to be ok here

agateau-gg: I meant the variable name (and matching comment)

Jguer: fair point, didn't notice it

agateau-gg (approved):

PR #145, opened 2021-11-24
Update jobs to push ggshield formula on both GitGuardian's taps

pierrelalanne:
### Context
In this MR, we want to lay the ground for migrating to a central tap for all GitGuardian's tool.  

### What we did
This MR simply modifies the `tag` workflow to push ggshield formula to two taps. 
**Warning :** We should wait for `GitGuardian/homebrew-tap` to be public before merging.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -108,7 +108,7 @@ jobs:
 
   push_to_tap:
     needs: pypi
-    name: Push GitGuardian tap
+    name: Push to GitGuardian taps
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/tag.yml
```diff
@@ -123,22 +123,41 @@ jobs:
...
           token: ${{ secrets.PAT_GITHUB }}
-          path: ./brew
+          path: ./ggshield
 
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install ggshield homebrew-pypi-poet
-          poet -f ggshield | sed 's/Shiny new formula/Detect secret in source code, scan your repos and docker images for leaks/g' | sed '7a\  license "MIT"' > brew/Formula/ggshield.rb
+          poet -f ggshield | sed 's/Shiny new formula/Detect secrets in source code, scan your repos and docker images for leaks/g' | sed '7a\  license "MIT"' | tee brew/Formula/ggshield.rb tap/Formula/ggshield.rb
```
agateau-gg, line 145: nit: can you wrap that line? It was already super long before, now it's worse :)

I haven't tested it, but something like this would be more readable:

```suggestion
          poet -f ggshield \
              | sed 's/Shiny new formula/Detect secrets in source code, scan your repos and docker images for leaks/g' \
              | sed '7a\  license "MIT"' \
              | tee brew/Formula/ggshield.rb tap/Formula/ggshield.rb
```

.github/workflows/tag.yml
```diff
@@ -108,7 +108,7 @@ jobs:
 
   push_to_tap:
     needs: pypi
-    name: Push GitGuardian tap
+    name: Push GitGuardian taps
```
agateau-gg, line 111:
```suggestion
    name: Push to GitGuardian taps
```

pierrelalanne:
Thanks for the review Aurélien, should be better now.

agateau-gg (approved):
Looks good now, thanks! :+1:

.github/workflows/tag.yml
```diff
@@ -123,22 +123,44 @@ jobs:
...
+          token: ${{ secrets.PAT_GITHUB }}
+          path: ./tap
+
       - name: Checkout Homebrew-ggshield
         uses: actions/checkout@master
         with:
           repository: GitGuardian/homebrew-ggshield
           token: ${{ secrets.PAT_GITHUB }}
-          path: ./brew
+          path: ./ggshield
```
jeremyds, line 139: I am surprised, don't you want to checkout it in the brew directory like it was done before?

.github/workflows/tag.yml
```diff
@@ -123,22 +123,44 @@ jobs:
...
 
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install ggshield homebrew-pypi-poet
-          poet -f ggshield | sed 's/Shiny new formula/Detect secret in source code, scan your repos and docker images for leaks/g' | sed '7a\  license "MIT"' > brew/Formula/ggshield.rb
+          poet -f ggshield \
+              | sed 's/Shiny new formula/Detect secrets in source code, scan your repos and docker images for leaks/g' \
+              | sed '7a\  license "MIT"' \
+              | tee brew/Formula/ggshield.rb tap/Formula/ggshield.rb
```
jeremyds, line 148: it seems to me you are relying on something at `brew` path here

jeremyds (changes requested):
.

.github/workflows/tag.yml
pierrelalanne: We checkout this repo to `./ggshield` now.

.github/workflows/tag.yml
pierrelalanne: 🤦  Good catch, thanks.

.github/workflows/tag.yml
jeremyds: there is already a ggshield directory in the repo root, don't you want to use another non existent one?

pierrelalanne: Should be good now.

jeremyds (approved):

PR #146, opened 2021-12-03
fix(tests): support macos tmp dir to make tests pass

phililippe:
Tests are failing on Mac OS because the tmp directory is not /tmp. Also /tmp is a symlink for /private/tmp

This PR proposes:
- to use the tempfile package to get the current tmp directory
- to test against /private/tmp if Darwin platform is detected

### DIFF
--- a/tests/test_scan.py
+++ b/tests/test_scan.py
@@ -1,5 +1,4 @@
-import platform
-from os import getcwd
+import os
 from unittest.mock import ANY, Mock, patch
 
 from ggshield.config import Cache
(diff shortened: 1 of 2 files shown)


tests/test_scan.py
```diff
@@ -11,7 +12,10 @@
 def test_cd_context_manager():
     prev = getcwd()
     with cd("/tmp"):  # nosec
-        assert getcwd() == "/tmp"  # nosec
+        if platform.system() == "Darwin":
+            assert getcwd() == "/private/tmp"  # nosec
+        else:
+            assert getcwd() == "/tmp"  # nosec
```
agateau-gg, line 18: I think this can be reworked to avoid platform-specific code like this (untested):
```python
target = os.path.realpath("/tmp")
with cd(target):
    assert getcwd() == target
assert getcwd() == prev
```

agateau-gg (changes requested):
Minor remark, looks good otherwise.

tests/test_scan.py
SofienM: Agree with Agateau-gg. Otherwise it looks good to me :+1:

agateau-gg (approved):
Great!

Issue #116, opened 2021-10-15
"No secret have been found" in ouput when secret are detected


jeremyds:
**GitGuardian Shield Version**

- [ ] I can reproduce this bug in the latest version

**Command executed**

`ggshield scan -v ci`

**Describe the bug**

When running `ggshield scan -v ci` using its docker image in gitlab CI, the output looks like that:

>  $ ggshield scan -v ci
> CI_COMMIT_BEFORE_SHA: 0000000000000000000000000000000000000000
> CI_COMMIT_SHA: xxxx
> Commits to scan: 1
> Scanning Commits
> secrets-engine-version: 2.52.1
> No secrets have been found
> commit xxxx
> Author: xxx xx <xxx@xxxx.com>
> Date: Fri Oct 15 14:53:41 2021 +0200
> 🛡️  ⚔️  🛡️  1 incident has been found in file slack_bot.py
> >>> Incident 1(Secrets detection): Slack Bot Token (Ignore with SHA: xxxx) (1 occurrence)
>     | @@ -0,0 +1,2 @
>   1 | def main():
>   2 |     slack_key = "xxxx"

"No secrets have been found" appears in ggshield output.
 
**Expected behavior**

"No secrets have been found" should not appear when a secret is detected.

PR #147, opened 2021-12-03
fix(ci) no secret found no more displayed if secrets found

phililippe:
This PR fixes https://github.com/GitGuardian/ggshield/issues/116

A scan can be composed of a list of subscans. In this case, found secrets are not listed in the main scan.results list, but in sub scans results. This PR proposes to test if any sub scan has a result in order to trigger the "no secret found" message.

### DIFF
--- a/ggshield/output/text/text_output.py
+++ b/ggshield/output/text/text_output.py
@@ -54,10 +54,7 @@ class TextHandler(OutputHandler):
             if self.verbose:
                 has_results = False
                 if scan.scans:
-                    for sub_scan in scan.scans:
-                        if sub_scan.results:
-                            has_results = True
-                            break
+                    has_results = any(x.results for x in scan.scans)
 
                 if not has_results:
                     scan_buf.write(no_leak_message())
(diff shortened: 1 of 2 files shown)


ggshield/output/text/text_output.py
```diff
@@ -52,7 +52,15 @@ def process_scan(self, scan: ScanCollection, top: bool = True) -> Tuple[str, int
                 scan_buf.write(self.process_result(result))
         else:
             if self.verbose:
-                scan_buf.write(no_leak_message())
+                has_results = False
+                if scan.scans:
+                    for sub_scan in scan.scans:
+                        if sub_scan.results:
+                            has_results = True
+                            break
```
agateau-gg, line 60: I think this can be simplified to:
```suggestion
                    has_results = any(x.results for x in scan.scans)
```

agateau-gg (changes requested):
Minor remark. Looks good, thanks for the fix!

agateau-gg (approved):

PR #156, opened 2021-12-17
Refactor output code

### LINKED ISSUE #140: Factorize export handler code
## The problem

There is some duplicated code between the [text](https://github.com/GitGuardian/ggshield/blob/2097e18b9c9b3c14efa73a089b97e01412ea4873/ggshield/output/text/text_output.py) and [json](https://github.com/GitGuardian/ggshield/blob/2097e18b9c9b3c14efa73a089b97e01412ea4873/ggshield/output/json/json_output.py) output handler implementations.

By refactoring the existing handlers to reduce the amount of duplicated code before implementing new output handlers we can avoid spreading the problem.

### Line creation

The code creating lines from content and taking into account the `--show-secrets` option can be factorized. Move it to a helper function in `OutputHandler`, or as a standalone function if it does not need the handler instance.

### Factorize final output handling

The code to write to either the destination file or to stdout is duplicated. This can be fixed by splitting `OutputHandler.process_scan()` in two: move the common code to the currently empty `OutputHandler.process_scan()` and make `OutputHandler.process_scan()` call a `_process_scan_impl()` method, implemented by the concrete output handler implementation.

This would look roughly like this:

```py
class OutputHandler:
    def process_scan(self, scan: ScanCollection) -> int:
        """
        Returns:
            exit code
        """
        content, return_code = self._process_scan_impl(scan)
        # handle output
        return return_code
        
    def _process_scan_impl(self, scan: ScanCollection, top: bool=True) -> Tuple[str, int]
        """
        Returns:
            tuple of (handler text, exit code)
        """
        pass
```

This has the additional benefit of hiding the `top` argument, which is an implemention detail.

### Remove `OutputHandler.process_result()`

This method should not be part of `OutputHandler` API: it's only called by `OutputHandler` implementations, not by `OutputHandler` itself or `OutputHandler` users. Implementation classes are free to create one, but it's not a requirement.

This is not a real factorization, but it makes the API cleaner.

agateau-gg:
This PR refactors the code in the ggshield.module package to reduce code duplication.

Summary of the changes:
- Remove unnecessary empty methods from OutputHandler
- Move code writing to stdout or to the specified file to OutputHandler
- Move code computing the exit code to OutputHandler
- Rename concrete classes from `${FORMAT}Handler` to `${FORMAT}OutputHandler` and their files from `${format}_output.py` to `${format}_output_handler.py`
- Add docstrings to OutputHandler methods

Fixes #140

### DIFF
--- a/ggshield/output/output_handler.py
+++ b/ggshield/output/output_handler.py
@@ -1,6 +1,6 @@
 from typing import Any, Optional, Tuple
 
-from ggshield.scan import Result, ScanCollection
+from ggshield.scan import ScanCollection
 
 
 class OutputHandler:
@@ -21,7 +21,4 @@ class OutputHandler:
         self.output = output
 
     def process_scan(self, scan: ScanCollection, top: bool = True) -> Tuple[Any, int]:
-        pass
-
-    def process_result(self, result: Result) -> Any:
         pass
--- a/ggshield/output/output_handler.py
+++ b/ggshield/output/output_handler.py
@@ -40,11 +40,12 @@ class OutputHandler(ABC):
         """Implementation of scan processing,
         called by :meth:`OutputHandler.process_scan`
 
+        Must return a string for the report.
+
         :param scan: The scan collection to process
         :return: The content
         """
         raise NotImplementedError()
-
 
     @staticmethod
     def _get_exit_code(scan: ScanCollection) -> int:
--- a/ggshield/output/output_handler.py
+++ b/ggshield/output/output_handler.py
@@ -1,9 +1,12 @@
-from typing import Any, Optional, Tuple
+from abc import ABC, abstractmethod
+from typing import Optional, Tuple
+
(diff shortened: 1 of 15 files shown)


ggshield/output/output_handler.py
```diff
@@ -13,15 +16,41 @@ def __init__(
...
+    def _process_scan_impl(self, scan: ScanCollection) -> str:
+        """Implementation of scan processing,
+        called by :meth:`OutputHandler.process_scan`
+
+        Must return a string for the report.
+
+        :param scan: The scan collection to process
+        :return: The content
+        """
+        return ""
```
jeremyds, line 48: minor: you might raise `NotImplementedError` to emphasize it should be implemented.

jeremyds:
Nice rework :+1:

jeremyds (approved):

Issue #151, opened 2021-12-07
ggshield pre-receive messages are not shown when committing from GitLab web UI


agateau-gg:
**GitGuardian Shield Version**

- [x] I can reproduce this bug in the latest version

**Describe the bug**

- install ggshield pre-receive hook on a GitLab repository, make sure it is called with the `--web` option
- edit a document using GitLab web UI, add a secret to it
- try to commit
- a generic error message appears, saying "Could not commit. An unexpected error occurred."

**Expected behavior**

The message should display ggshield pre-receive output, similar to what happens on the command-line when the pre-receive hook rejects a commit.

**Screenshots**

![image](https://user-images.githubusercontent.com/91945295/145049562-6960a4eb-8efb-4eb0-ae6f-ac8730366cb7.png)

**Technical details**

This is caused by GitLab filtering the hook output before showing it. GitLab only shows lines starting with the `GL-HOOK-ERR:` prefix (see <https://docs.gitlab.com/ee/administration/server_hooks.html#example-custom-error-message>).

When called from GitLab web UI without the `--web` option, ggshield just disables itself. This is a surprising behavior which should be changed.

What should be done is to deprecate the `--web` option, and instead detect when being called from GitLab web UI (this can be done by checking the `GL_PROTOCOL` environment variable is set to `web`, this is what ggshield uses to disable itself). When this is the case, prefix all error messages with the `GL-HOOK-ERR:` prefix.

This currently causes some formatting issues which needs to be addressed: on our test instance a message passed this way looks like this:

![image](https://user-images.githubusercontent.com/91945295/145052356-06622133-e7a8-4822-9eef-6857afe55747.png)

Update: I filed an issue on GitLab issue tracker regarding newline handling: https://gitlab.com/gitlab-org/gitlab/-/issues/350349.

PR #158, opened 2021-12-23
Fix pre-receive hook error handling with GitLab Web UI

agateau-gg:
## Description

This PR detects if the pre-receive hook is running from GitLab Web UI, and if so, prefix the error messages so that they show up in the UI instead of disabling the hook completely.

It also deprecates the `--web` option, which is no longer useful.

## Issue

#151 

## Possible improvements

The output has room for improvement, though, as you can see:

**Using the "Edit" button**

image-edit-button

Issues:
- The ASCII art formatting for the secret indicator is broken by the use of a variable-width font.
- The URL is not turned into a link.
- The "rewrite history" part of the message is not relevant: there is no history to rewrite since the commit does not exist yet: when using the web UI, the pre-receive hook behaves more like a pre-commit hook.

**Using the "Web IDE"**

image-webide

In addition to the "Edit" button issues, we see GitLab turns new-lines characters into `<br>`, but shows them directly. This is an issue on GitLab side, similar to [this one](https://gitlab.com/gitlab-org/gitlab/-/issues/284640) (but it's not it, need to file an issue for it)

I think this could be avoided by using much simpler messages for the web UI use case. I did a PoC on this in the `unbreak-gitlab-web-pre-receive-short-messages`, but it's not ready yet.

### DIFF
--- a/ggshield/output/output_handler.py
+++ b/ggshield/output/output_handler.py
@@ -21,6 +21,13 @@ class OutputHandler(ABC):
         self.verbose = verbose
         self.output = output
 
+    def echo(self, message: str, err: bool = False) -> None:
+        """Print a message.
+
+        Default implementation forwards the arguments to click.echo().
+        """
+        click.echo(message, err=err)
+
     def process_scan(self, scan: ScanCollection) -> int:
         """Process a scan collection, write the report to :attr:`self.output`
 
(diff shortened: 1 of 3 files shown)


agateau-gg:
Closing this PR, #172 replaces it.

PR #162, opened 2022-01-20
fix(ci): stop to tag every Docker images as latest

jeremyds:
We tagged every Docker images we built as latest. It makes every merges
we did a release for our customers using latest Docker images.

This commit now makes the CI to tag images built on a merge into the
main branch as unstable instead of latest. We continue to tag Docker
images built for a release with the corresponding git tag and we add the
latest tag.


agateau-gg (approved):
Looks good!

PR #166, opened 2022-01-26
Fix uncaught Exception when cache is created on a read-only FS

CharlesB2:
This PR fixes an issue when ggshield runs on a read-only filesystem, by catching more general `OSError` during cache write.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -172,6 +172,22 @@ class TestCache:
             file_content = json.load(file)
             assert file_content == {"last_found_secrets": ["XXX"]}
 
+    def test_read_only_fs(self):
+        """
+        GIVEN a read-only file-system
+        WHEN save is called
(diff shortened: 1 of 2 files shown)


agateau-gg (approved):
Thanks, and the test is nice!

tests/test_config.py
```diff
@@ -172,6 +172,22 @@ def test_save_cache(self, cli_fs_runner):
             file_content = json.load(file)
             assert file_content == {"last_found_secrets": ["XXX"]}
 
+    def test_read_only_fs(self):
+        """
+        GIVEN a read-only files-system
```
agateau-gg, line 177: nit: small typo
```suggestion
        GIVEN a read-only file-system
```

agateau-gg (commented):
Just a small typo to fix, then I can merge it

PR #167, opened 2022-01-27
chore(docs): add GITGUARDIAN_API_URL to pre-receive hook template

CharlesB2:


doc/pre-receive.sample
```diff
@@ -9,7 +9,10 @@
 # either set an environment variable machine wide `GITGUARDIAN_API_KEY`
 # or uncomment and insert your key in the following line
 
-#export GITGUARDIAN_API_KEY=<INSERT YOUR KEY>
+# export GITGUARDIAN_API_KEY=<INSERT YOUR KEY>
+
+# (set this if you are using a self-hosted GitGuardian instance)
```
agateau-gg, line 14: nit: wording.
```suggestion
# Set this if you use a self-hosted GitGuardian instance
```

doc/pre-receive.sample
```diff
@@ -9,7 +9,10 @@
...
+# export GITGUARDIAN_API_URL="https://dashboard.gitguardian.mycorp.local/exposed/"
```
agateau-gg, line 15: Is the "exposed/" part always there for self-hosted instances? If not I would use a simpler URL for the example.

CharlesB2: Yes, after checking the `exposed` part is necessary for self-hosted

agateau-gg: OK, all good then :+1:

agateau-gg (approved):

PR #172, opened 2022-02-17
Fix GitLab Web UI showing empty error messages

agateau-gg:
## Description

This PR detects if the pre-receive hook is running from GitLab Web UI. If it is the case, it uses a shorter error message, prefixed with what GitLab expects so that the message shows up in the UI instead of disabling the hook completely.

It also deprecates the --web option, which is no longer useful.

## Issue

https://github.com/GitGuardian/ggshield/issues/151

## Screenshots

Web UI:

webui

Web IDE:

image

Not the pinnacle of design and user-friendliness, but GitLab UI does not allow any better.

### DIFF
--- a/ggshield/output/gitlab_webui/gitlab_webui_output_handler.py
+++ b/ggshield/output/gitlab_webui/gitlab_webui_output_handler.py
@@ -1,0 +1,61 @@
+from pygitguardian.models import PolicyBreak
+
+from ggshield.filter import censor_match
+from ggshield.output.output_handler import OutputHandler
+from ggshield.scan import ScanCollection
+from ggshield.text_utils import pluralize, translate_validity
+
+
+def format_policy_break(policy_break: PolicyBreak) -> str:
+    """Returns a string with the policy name, validity and a comma-separated,
+    double-quoted, censored version of all `policy_break` matches.
+
+    Looks like this:
+
+    PayPal OAuth2 Keys (Validity: Valid, id="aa*******bb", secret="cc******dd")
+    """
+    match_str = ", ".join(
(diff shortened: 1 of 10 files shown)


ggshield/output/gitlab_webui/gitlab_webui_output_handler.py
```diff
@@ -0,0 +1,64 @@
...
+class GitLabWebUIOutputHandler(OutputHandler):
+    """
+    Terse OutputHandler optimized for GitLab Web UI, because GitLab Web UI only shows
+    lines starting with GL-HOOK-ERR.
+
+    See https://docs.gitlab.com/ee/administration/server_hooks.html#custom-error-messages
+    """
+
+    def __init__(self) -> None:
+        super().__init__(show_secrets=False, verbose=False)
```
jeremyds, line 33: `show-secret` parameter default value is None. Any reason I miss we don't want to show secrets if our users explicitly set it to True?

jeremyds (approved):

ggshield/output/gitlab_webui/gitlab_webui_output_handler.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def echo(self, message: str, err: bool = False) -> None:
```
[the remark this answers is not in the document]
agateau-gg: Oups, that's a left-over from a more invasive change :). Removed.

ggshield/output/gitlab_webui/gitlab_webui_output_handler.py
agateau-gg: Good catch, fixed this.

PR #173, opened 2022-02-17
Minor fixes

agateau-gg:
This PR does some minor fixes on issues I found while working on #172. Since they ended up not being required for the GitLab work I moved them to a separate PR.


ggshield/cmd.py
```diff
@@ -164,7 +164,7 @@ def scan(
 scan = cast(click.Group, json_output_option_decorator(scan))
 
 
-@scan.resultcallback()
+@scan.result_callback()
```
jeremyds, line 167: As discussed with you, I did it in `f0035d559e` and finally had to revert it in `a607b26d45`. If I remember well, system wide installation using pip can be broken if an older version of `click` is already installed.....

agateau-gg: This new syntax was added in Click 8.0.0, released in may 2021. What do you think of defining this as the minimum required version?

jeremyds (approved):

Issue #113, opened 2021-10-11
Input: Pypi packages


ericfourrier:
**Is your feature request related to a problem? Please describe.**

I want to scan packages (before publishing them to a registry but also be able to download from a registry and scan the package for secrets) 

**Describe alternatives you've considered**

Maybe scanning a tar.gz since a package is a path. Maybe path should support compressed directory and ggshield should handle the decompression.

## Commands

- `ggshield scan pypi <package-name>`

Given a package-name, ggshield will attempt to pull from pip the wheel and then scan the full archive

- `ggshield scan archive <archive_path>`

Given an archive path, ggshield will attempt to download the archive if an url and then scan the full archive.

agateau-gg:
This has been implemented in https://github.com/GitGuardian/ggshield/pull/176.

PR #176, opened 2022-02-24
Jeremyds/113/scan pypi and archive

jeremyds:
This PR adds 2 new commands to be able to scan:
- python package (wheel),
- archive.

#113

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ GITGUARDIAN_API_KEY=<GitGuardian API Key>
 # Installation
 
 ## On MacOS
+
 ### Using Homebrew
 
 You can install ggshield using Homebrew by running the following command:
@@ -93,6 +94,7 @@ $ brew install gitguardian/tap/ggshield
 ```
 
 ## On other Operating Systems
+
 ### Using pip
 
 Install and update using `pip`:
@@ -250,6 +252,13 @@ Commands:
   You can use `pip` environment variables or configuration files to set `pip download` parameters as explained in [pip documentation](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables)
   For example, you can set `pip` `--index-url` parameter by setting `PIP_INDEX_URL` environment variable.
 
+- `archive`: scan an archive files.
+
+  ```
+  scan archive <PATH>
+  ```
+
+  No options or arguments.
 
 ## Install command
 
--- a/README.md
+++ b/README.md
@@ -238,6 +238,19 @@ Commands:
     -h, --help  Show this message and exit.
   ```
 
+- `pypi`: scan a pypi package.
+
+  ```
+  ggshield scan pypi PACKAGE_NAME
+  ```
+
+  No options or arguments.
+
+  Under the hood this command uses the `pip download` command to download the python package.
+  You can use `pip` environment variables or configuration files to set `pip download` parameters as explained in [pip documentation](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables)
+  For example, you can set `pip` `--index-url` parameter by setting `PIP_INDEX_URL` environment variable.
+
+
 ## Install command
 
 The `install` command allows you to use ggshield as a pre-commit or pre-push hook
--- a/ggshield/pypi.py
+++ b/ggshield/pypi.py
@@ -1,0 +1,117 @@
+import re
+import shutil
+import subprocess
+import tempfile
+from pathlib import Path
+from typing import Any, Dict, List, Set
+
+import click
+
+from ggshield.config import Config
+from ggshield.output import OutputHandler
+from ggshield.path import get_files_from_paths
+from ggshield.scan import Files, Result, ScanCollection
+from ggshield.utils import SupportedScanMode
+
+
+PYPI_DOWNLOAD_TIMEOUT = 30
+
+
+def save_package_to_tmp(temp_dir: str, package_name: str) -> None:
+    command: List[str] = [
+        "pip",
+        "download",
+        package_name,
+        "--dest",
+        temp_dir,
+        "--no-deps",
+    ]
+
+    try:
+        click.echo("Downloading pip package... ", nl=False)
+        subprocess.run(
+            command,
+            check=True,
+            stderr=subprocess.PIPE,
+            timeout=PYPI_DOWNLOAD_TIMEOUT,
+        )
+        click.echo("OK")
+
+    except subprocess.CalledProcessError:
+        raise click.ClickException(f'Failed to download "{package_name}"')
+
+    except subprocess.TimeoutExpired:
+        raise click.ClickException('Command "{}" timed out'.format(" ".join(command)))
+
+
+def get_files_from_package(
+    archive_dir: str,
+    package_name: str,
+    exclusion_regexes: Set[re.Pattern],
+    verbose: bool,
+) -> Files:
+    archive_dir_path = Path(archive_dir)
+    archive: Path = next(archive_dir_path.iterdir())
+    unpack_kwargs: Dict[str, str] = (
+        {"format": "zip"} if archive.suffix == ".whl" else {}
+    )
+
+    try:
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -238,6 +240,26 @@ Commands:
...
+- `pypi`: scan a pypi package.
+
+  ```
+  ggshield scan pypi PACKAGE_NAME
+  ```
+
+  No options or arguments.
+
+  Under the hood this command uses the `pip download` command to download the python package.
+  You can use `pip` environment variables to set `pip download` parameter as explained in [pip documentation](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables)
```
jeremyds, line 252:
```suggestion
  You can use `pip` environment variables to set `pip download` parameters as explained in [pip documentation](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables)
```

ggshield/pypi.py
```diff
@@ -0,0 +1,109 @@
...
+        subprocess.run(
+            command,
+            check=True,
+            stderr=subprocess.PIPE,
+            timeout=PYPI_DOWNLOAD_TIMEOUT,
+        )
+        click.echo("OK")
+
+    except subprocess.CalledProcessError:
+        raise click.ClickException(f'Package "{package_name}" not found')
```
agateau-gg, line 41: Download could fail for other reasons, like network issues. Maybe use a more generic message like `'Failed to download "{package_name}"'`?

ggshield/pypi.py
```diff
@@ -0,0 +1,109 @@
...
+    unpack_kwargs: Dict[str, str] = (
+        {"format": "zip"} if archive.suffix == ".whl" else {}
+    )
+
+    try:
+        shutil.unpack_archive(str(archive), extract_dir=temp_dir_path, **unpack_kwargs)
+    except Exception as exn:
+        raise click.ClickException(f'Failed to unpack package "{package_name}": {exn}.')
+
+    exclusion_regexes.add(re.compile(archive.name))
```
agateau-gg, line 64: better use `re.escape` in case the archive contains funny chars.

ggshield/pypi.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def get_files_from_package(
+    temp_dir: str,
+    package_name: str,
+    exclusion_regexes: Set[re.Pattern],
+    verbose: bool,
+) -> Files:
+    temp_dir_path: Path = Path(temp_dir)
+    archive: Path = list(temp_dir_path.glob("*"))[0]
```
agateau-gg, line 54: could be `next(temp_dir_path.iterdir())`, but I think it would be nicer to have `save_package_to_tmp()` do this and return the path.

agateau-gg (changes requested):
Minor remarks, looks good!

ggshield/pypi.py
```diff
@@ -0,0 +1,109 @@
...
+        raise click.ClickException('Command "{}" timed out'.format(" ".join(command)))
+
+
+def get_files_from_package(
+    temp_dir: str,
+    package_name: str,
+    exclusion_regexes: Set[re.Pattern],
+    verbose: bool,
+) -> Files:
+    temp_dir_path: Path = Path(temp_dir)
```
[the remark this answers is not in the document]
jeremyds: I would have to make it a string in `save_package_to_tmp` function in this case :sweat_smile:

ggshield/pypi.py
jeremyds: Done for the use of `next...`. 

I feel the current code split into the 2 functions is clearer and easier to test like that, `save_package_to_tmp` doing only the command call in subprocess and `get_files_from_package` handling the "logic".

ggshield/pypi.py
```diff
@@ -0,0 +1,109 @@
...
+        results: List[Result] = files.scan(
+            client=ctx.obj["client"],
+            cache=ctx.obj["cache"],
+            matches_ignore=config.matches_ignore,
+            banlisted_detectors=config.banlisted_detectors,
+            all_policies=config.all_policies,
+            verbose=config.verbose,
+            mode_header=SupportedScanMode.PYPI.value,
+        )
+        scan: ScanCollection = ScanCollection(
```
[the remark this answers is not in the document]
jeremyds: Done but honestly I prefer to type hint all the time and not bring exceptions :sweat_smile:

README.md
```diff
@@ -238,6 +240,26 @@ Commands:
...
+- `pypi`: scan a pypi package.
+
+  ```
+  ggshield scan pypi PACKAGE_NAME
+  ```
+
+  No options or arguments.
+
+  Under the hood this command uses the `pip download` command to download the python package.
+  You can use `pip` environment variables to set `pip download` parameters as explained in [pip documentation](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables)
```
CharlesB2, line 252:
```suggestion
  You can use `pip` environment variables or configuration files to set `pip download` parameters as explained in [pip documentation](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables)
```

README.md
```diff
@@ -238,6 +240,26 @@ Commands:
...
+  For example, you can set `pip` `--index-url` parameter by setting `PIP_INDEX_URL` environment variable.
+
+- `archive`: scan an archive files.
```
CharlesB2, line 255:
```suggestion
- `archive`: scan an archive file, either a wheel package (`.whl`) or source distribution archive (`.tar.gz`)
```

ggshield/pypi.py
```diff
@@ -0,0 +1,107 @@
...
+
+    except subprocess.CalledProcessError:
+        raise click.ClickException(f'Failed to download "{package_name}"')
+
+    except subprocess.TimeoutExpired:
+        raise click.ClickException('Command "{}" timed out'.format(" ".join(command)))
+
+
+def get_files_from_package(
+    temp_dir: str,
```
CharlesB2, line 48: nit: rename `temp_dir` to more general `package_dir`, in the function's context the directory doesn't have to be temporary

ggshield/pypi.py
```diff
@@ -0,0 +1,107 @@
...
+        save_package_to_tmp(temp_dir=temp_dir, package_name=package_name)
+
+        files: Files = get_files_from_package(
+            temp_dir=temp_dir,
+            package_name=package_name,
+            exclusion_regexes=ctx.obj["exclusion_regexes"],
+            verbose=config.verbose,
+        )
+
+        results: List[Result] = files.scan(
```
CharlesB2, line 96: Same remark here

ggshield/pypi.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    scan a pypi package <NAME>.
+    """
+    config: Config = ctx.obj["config"]
+    output_handler: OutputHandler = ctx.obj["output_handler"]
+
+    with tempfile.TemporaryDirectory(suffix="ggshield") as temp_dir:
+        save_package_to_tmp(temp_dir=temp_dir, package_name=package_name)
+
+        files: Files = get_files_from_package(
```
CharlesB2, line 89: `get_files_from_package ` already type-hints the return type, so I don't see why type-hint the variable here

CharlesB2 (changes requested):
Looks good, just a few remarks and docs suggestion

ggshield/pypi.py
jeremyds: It seems to me that it becomes too complex to know whether we should type hint or not. Can type hint a variable be a problem?

ggshield/pypi.py
jeremyds: same

README.md
jeremyds: The `scan pypi` should be used for that. `shutils.unpack_archive` does not handle wheel format.

jeremyds:
@CharlesB2 good point, fixed!

CharlesB2 (approved):

PR #185, opened 2022-03-24
Jchaumont/3979/auth login token

julienc91:
Implementation of the `auth login --method token` flow.


agateau-gg:
I know it's a draft but I had a look at it anyway :).

Looks good overall, but I am worried about the syntax: `ggshield auth login --method token --token $TOKEN` means the token can end up in shell history files and other places. Have we considered reading it from an environment variable or from stdin (like [gh does](https://cli.github.com/manual/gh_auth_login)) instead?

julienc91:
@agateau-gg Done, I replaced it with a `click.prompt` :slightly_smiling_face:

agateau-gg:
> @agateau-gg Done, I replaced it with a `click.prompt` slightly_smiling_face

Looks good :+1:

PR #186, opened 2022-03-25
Draft: Create rpm and deb packages

agateau-gg:
Draft


scripts/build-packages/build-packages
```diff
@@ -0,0 +1,77 @@
...
+#
+# It expects [nfpm][] and [shiv][] to be installed.
+#
+# Packages are created in the `packages` directory.
+#
+# [nfpm]: https://github.com/goreleaser/nfpm/
+# [shiv]: https://github.com/linkedin/shiv
+
+cd $(dirname $0)
+NFPM_YAML_TMPL=$PWD/nfpm.yaml.tmpl
```
jeremyds, line 14:
```suggestion
NFPM_YAML_TMPL=$(dirname $0)/nfpm.yaml.tmpl
```
To remove the need to `cd`

setup.cfg
```diff
@@ -5,4 +5,4 @@ ignore = E203, W503
 exclude = **/snapshots/*.py, .venv
 
 [metadata]
-description-file = README.md
+description_file = README.md
```
jeremyds, line 8: I don't understand this change :sweat: 

from [setup.cfg doc](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) it seems the correct syntax is:
```
long_description = file: README.rst, CHANGELOG.rst, LICENSE.rst
```

jeremyds (approved):

scripts/build-packages/build-packages
agateau-gg: I do it this way to ensure the variable contains the absolute full path. That avoids potential issues.

setup.cfg
agateau-gg: That was strange indeed. What was even stranger was that even modifying the file to something like `description_file = foo` or `long_description = bar` did not change anything: no build error and the generated package still contained the README.md. That's only when I removed the README.md file that I realized it was read from setup.py and set as `long_description` from there... :upside_down_face:

I just removed the "metadata" section from setup.cfg for now. Converting from setup.py to setup.cfg or pyproject.toml is out-of-scope for this PR.

PR #187, opened 2022-03-25
Draft: Auth protected config file

eugenenelou:
This PR refactors the config class to  handle different config files format for the config.

It also implements the Auth config file that is a new config file to store the authentication configuration (tokens, hosts...) for potentially several GitGuardian dashboard instances.

The existing config has been renamed `UserConfig`, a better name proposition is welcome.

TODO:
- [x] Implement priority between the source of GITGUARDIAN_API_TOKEN values
- [x] Implement priority between the source of GITGUARDIAN_HOST values
- [x] Handle the new GITGUARDIAN_DASHBOARD_URL and handles retrocompatibility to infer API URL from Dashboard URL or vice versa if only one is defined.
- [x] Add tests for the Auth Config.
- [x] Relock the pipfile once https://github.com/GitGuardian/ggshield/pull/184 is merged.

### DIFF
--- a/ggshield/utils.py
+++ b/ggshield/utils.py
@@ -35,7 +35,7 @@ EMPTY_TREE = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
 # We try and anticipate that so we can control the return code
 PRERECEIVE_TIMEOUT = 4.5
 
-IGNORED_DEFAULT_PATTERNS = [
+IGNORED_DEFAULT_WILDCARDS = [
     "**/.git/**/*",
     "**/.pytest_cache/**/*",
     "**/.mypy_cache/**/*",
--- a/ggshield/cmd.py
+++ b/ggshield/cmd.py
@@ -21,7 +21,7 @@ from .pypi import pypi_cmd
 from .quota import quota
 from .status import status
 from .text_utils import display_error
-from .utils import IGNORED_DEFAULT_PATTERNS, json_output_option_decorator, load_dot_env
+from .utils import IGNORED_DEFAULT_WILDCARDS, json_output_option_decorator, load_dot_env
 
 
 def get_max_commits_for_hook() -> Optional[int]:
@@ -128,7 +128,7 @@ def scan(
         paths_ignore.update(exclude)
 
     if not ignore_default_excludes and not ctx.obj["config"].ignore_default_excludes:
-        paths_ignore.update(IGNORED_DEFAULT_PATTERNS)
+        paths_ignore.update(IGNORED_DEFAULT_WILDCARDS)
 
     ctx.obj["exclusion_regexes"] = init_exclusion_regexes(paths_ignore)
     config: Config = ctx.obj["config"]
--- a/tests/test_prepush_cmd.py
+++ b/tests/test_prepush_cmd.py
@@ -6,7 +6,7 @@ from click.testing import CliRunner
 
 from ggshield.cmd import cli
 from ggshield.filter import init_exclusion_regexes
-from ggshield.utils import EMPTY_SHA, EMPTY_TREE, IGNORED_DEFAULT_PATTERNS
+from ggshield.utils import EMPTY_SHA, EMPTY_TREE, IGNORED_DEFAULT_WILDCARDS
 
 
 class TestPrepush:
@@ -75,9 +75,11 @@ class TestPrepush:
         input: Optional[str],
     ):
         """
-        GIVEN a prepush range with a 20 commits through the old env vars of the framework
+        GIVEN a prepush range with 20 commits provided by the pre-commit framework
         WHEN the command is run
         THEN it should pass onto scan and return 0
+        AND the `exclusion_regexes` argument of the scan_commit_range() call should
+        match IGNORED_DEFAULT_WILDCARDS
         """
         scan_commit_range_mock.return_value = 0
         commit_list = ["a"] * 20
@@ -106,7 +108,7 @@ class TestPrepush:
         assert "Commits to scan: 20" in result.output
         assert result.exit_code == 0
 
-        expected_exclusion_regexes = init_exclusion_regexes(IGNORED_DEFAULT_PATTERNS)
+        expected_exclusion_regexes = init_exclusion_regexes(IGNORED_DEFAULT_WILDCARDS)
         expected_exclusion_patterns = [r.pattern for r in expected_exclusion_regexes]
         result_exclusion_regexes = scan_commit_range_mock.call_args_list[0][1][
             "exclusion_regexes"


ggshield/cache.py
```diff
@@ -1,126 +1,73 @@
...
-                    for elem in item:
-                        if elem not in attr:
-                            attr.append(elem)
-                else:
-                    setattr(self, key, item)
-
-                setattr(self, key, item)
-            else:
-                click.echo("Unrecognized key in cache: {}".format(key))
+        if "last_found_secrets" in kwargs:
```
agateau-gg, line 43: Maybe you can define a module constant for `"last_found_secrets"`?

ggshield/config.py
```diff
@@ -23,139 +24,198 @@ def replace_in_keys(data: Dict, old_char: str, new_char: str) -> None:
...
-            Attribute("show_secrets", False),
-            Attribute("verbose", False),
-            Attribute("ignore_default_excludes", False),
-        ]
-
-        for attr in self.attributes:
-            setattr(self, attr.name, attr.default)
-        self.load_configs(self.CONFIG_GLOBAL)
-        self.load_configs(self.CONFIG_LOCAL)
+def load_yaml(path: str, raise_exc: bool = False) -> Optional[Dict[str, Any]]:
```
agateau-gg, line 31: Maybe this function should always return an empty dict, to simplify calling code (or it should not do a fallback on empty dict when calling `safe_load()`)

ggshield/config.py
```diff
@@ -23,139 +24,198 @@ def replace_in_keys(data: Dict, old_char: str, new_char: str) -> None:
...
+    def from_dict(cls, data: Dict) -> "YamlFileConfig":
+        replace_in_keys(data, old_char="-", new_char="_")
+        return cls(**data)
+
+    def update_config(self, data: Dict[str, Any]) -> bool:
         """
-        Save config in the first CONFIG_LOCAL file.
-        If no local config file, creates a local .gitguardian.yaml
+        Update the current config
+        raise an error if trying to set a field that does not exist
```
agateau-gg, line 72: This method does not do what it says in the doc: it never raises. Also, it always return True. Either the `continue` should be replaced with a `return False` or the return type should be changed to None.

ggshield/config.py
```diff
@@ -36,287 +28,266 @@ def replace_in_keys(data: Dict, old_char: str, new_char: str) -> None:
...
+    max_commits_for_hook: int = 50
+    banlisted_detectors: Set[str] = field(default_factory=set)
+    ignore_default_excludes: bool = False
+
+    def __post_init__(self) -> None:
+        self.api_url = os.getenv("GITGUARDIAN_API_URL", DEFAULT_API_URL)
+
+    def save(self) -> None:
+        """
+        Save config in the first CONFIG_LOCAL or the path it the config was loaded from
```
agateau-gg, line 120:
```suggestion
        Save config in the first CONFIG_LOCAL or the path it was loaded from
```

ggshield/config.py
```diff
@@ -36,287 +28,266 @@ def replace_in_keys(data: Dict, old_char: str, new_char: str) -> None:
...
-                if isinstance(item, list):
-                    attr = getattr(self, key)
-                    for elem in item:
-                        if elem not in attr:
-                            attr.append(elem)
-                else:
-                    setattr(self, key, item)
+def get_auth_config_dir() -> str:
+    dir_path: str = user_config_dir(appname="ggshield", appauthor="GitGuardian")
+    return dir_path
```
agateau-gg, line 206: Can be simplified to:
```suggestion
    return user_config_dir(appname="ggshield", appauthor="GitGuardian")
```

ggshield/config.py
```diff
@@ -170,6 +174,70 @@ def add_ignored_match(self, secret: Dict) -> None:
...
+    id: int
+    url: str
+    token: str
+    type: str
+    token_name: str
+    raw_expire_at: InitVar[str]
+    expire_at: Optional[datetime] = None
+
+    def __post_init__(self, raw_expire_at: str) -> None:
+        self.expire_at = datetime.fromisoformat(raw_expire_at)
```
agateau-gg, line 188: Does this mean the key in the file must be named `raw_expire_at`? If so I think we should reverse and name the file key `expire_at` and the parsed key something like `expire_at_datetime`: this avoid propagating current limitations of our implementations to file formats. If in the future we have a way to parse datetime fields without the need of an intermediate one then we can make the switching without having to care about converting config files.

agateau-gg (changes requested):
Minor remarks, looks good!

+1 for the `write_*()` helpers in test_config.py :+1:

ggshield/config.py
```diff
@@ -23,139 +24,198 @@ def replace_in_keys(data: Dict, old_char: str, new_char: str) -> None:
...
-        # Convert all sets into more human readable lists
-        for key in self.get_attributes_keys():
-            value = _config[key]
-            if type(value) is set:
-                _config[key] = list(value)
-        replace_in_keys(_config, "_", "-")
-        return _config
-
-    def save(self) -> bool:
+class YamlFileConfig:
```
martinfaucheux, line 53: why not using abstract class?

ggshield/config.py
eugenenelou: Indeed, I changed the implementation because it was breaking existing tests.

I don't understand about the return type, it always return, the `continue` can just make it leave the loop and go to `return True`.

ggshield/config.py
eugenenelou: I was able to workaround that using only the `expire_at` field.

ggshield/config.py
eugenenelou: why should we use it here? it's useful to expose abstraction from a module, but I don't really find it useful inside one module.

ggshield/config.py
agateau-gg: What I mean with the return type is that this function can never return False, so the return value is not useful. What to do depends on whether callers need to know if there was a problem:

- If they need to know, then we should replace the `continue` with a `return False`.
- If they don't need to know, then the code can be simplified by changing the return type to None (and callers need to be updated to not check the return value).

ggshield/config.py
agateau-gg: Not a blocker for me, but if I prefer using exceptions for error management than relying on the difference between `None` and `{}`: loading an empty YAML document returns `None` after all.

ggshield/config.py
eugenenelou: @agateau-gg the purpose I guess (this was a pre-existing function) was being able to do `if result` I guess, but yes without a `return False` it will either return true or raise so it's not needed.

However replacing `continue` with a `return False` breaks the implementation.

ggshield/config.py
eugenenelou: no because mypy fails, is there a better way to do it? (mypy just feels like a waste of time).

agateau-gg: mypy is annoying sometimes, but it regularly catches mistakes, I find it quite useful.

In this case that's probably because the `appdirs` package does not ship with type hints. Someone created a package for this though: https://pypi.org/project/types-appdirs/, I think adding this to our packages (and to mypy config in the .pre-commit-config.yaml) should help.

martinfaucheux (approved):

ggshield/utils.py
```diff
@@ -279,3 +256,58 @@ def handle_exception(e: Exception, verbose: bool) -> int:
...
+
+def dashboard_to_api_url(dashboard_url: str) -> str:
+    parsed_url = urlparse(dashboard_url)
+    if parsed_url.scheme != "https":
+        raise ValueError(
+            f"Invalid scheme for dashboard URL '{dashboard_url}', expected HTTPS"
+        )
+    if parsed_url.netloc.endswith(".gitguardian.com"):  # SaaS
+        if parsed_url.path:
+            raise ValueError(
```
julienc91, line 290: This is introducing a regression, as URLs with path being `/` used to be accepted.

eugenenelou: good find, fixed by cleaning the trailing slashes

ggshield/config.py
```diff
@@ -153,8 +151,32 @@ class UserConfig(YamlFileConfig):
     banlisted_detectors: Set[str] = field(default_factory=set)
     ignore_default_excludes: bool = False
 
-    def __post_init__(self) -> None:
-        self.api_url = os.getenv("GITGUARDIAN_API_URL", DEFAULT_API_URL)
+    def set_env_gitguardian_urls(self) -> None:
```
agateau-gg, line 154: (naming) I thought this was updating environment variables. I would rename it to something like `update_from_env()`.

agateau-gg (changes requested):

ggshield/config.py
```diff
@@ -1,322 +1,470 @@
...
+            return obj.to_dict()  # type: ignore
+        result = []
+        for f in fields(obj):
+            value = custom_asdict(getattr(obj, f.name))
+            result.append((f.name, value))
+        return dict(result)
+    elif isinstance(obj, (list, tuple)):
+        return type(obj)(custom_asdict(v) for v in obj)  # type: ignore
+    elif isinstance(obj, dict):
+        return type(obj)((custom_asdict(k), custom_asdict(v)) for k, v in obj.items())
```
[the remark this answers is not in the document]
eugenenelou: I took inspiration from dataclases.asdict that does, but I agree.

ggshield/config.py
```diff
@@ -1,322 +1,470 @@
...
+        return copy.deepcopy(obj)  # type: ignore
+
+
+class YamlFileConfig:
+    """Helper class to define configuration object loaded from a YAML file"""
+
+    def __init__(self, **kwargs: Any) -> None:
+        raise NotImplementedError
+
+    def to_dict(self) -> Union[List, Dict]:
```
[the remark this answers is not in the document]
eugenenelou: I'll rename it `to_json`

ggshield/config.py
```diff
@@ -1,322 +1,470 @@
...
+    data: Union[List, Dict], old_char: str, new_char: str, recursive: bool = True
+) -> None:
     """Replace old_char with new_char in data keys."""
-    for key in list(data):
-        if old_char in key:
-            new_key = key.replace(old_char, new_char)
-            data[new_key] = data.pop(key)
+    if recursive:
+        if isinstance(data, dict):
+            for key, value in list(data.items()):
```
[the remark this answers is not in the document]
eugenenelou: no, because we are modifying the keys during iteration, it causes an error otherwise:

```py
RuntimeError: dictionary keys changed during iteration
```

ggshield/config.py
```diff
@@ -1,322 +1,475 @@
...
+    def to_dict(self) -> Union[List, Dict]:
+        data: Dict = custom_asdict(self, root=True)  # type: ignore
+        data["accounts"] = [data.pop("account")]
+        return data
+
+    @property
+    def expired(self) -> bool:
+        return (
+            self.account.expire_at is not None
+            and self.account.expire_at <= datetime.now()
```
martinfaucheux, line 289:
```suggestion
            and self.account.expire_at >= datetime.now()
```
also this assumes the datetime stored is unaware. By default when fetching the dt from the API they will be signed. This should either
* expect aware datetimes
* expect unaware datetimes
* expect anything but then utc-sign incoming unaware dt and compare it to `datetime.now(timezone.utc)`

julienc91: I disagree when the sign change, but agree with the rest :slightly_smiling_face:

eugenenelou: Agreed, I was waiting to see if the timezone awareness was requested.

martinfaucheux (approved):

tests/test_config.py
```diff
@@ -1,124 +1,182 @@
...
+    Config,
+    InstanceConfig,
+    get_auth_config_filepath,
+    replace_in_keys,
+)
+from ggshield.utils import dashboard_to_api_url
+
+
+@pytest.fixture
+def tmp_dir_name():
```
agateau-gg, line 24: pytest already provides tmp_dir and tmp_path (which is even nicer) fixtures. Maybe we can use tmp_path instead?

agateau-gg (approved):

PR #190, opened 2022-03-30
feat(headers): Compute headers based on click context

Walz:
Changes:
- Upgrade black to avoir click deprecation issues
- Compute headers based on click context

I've removed the param `mode_header` and now generate the headers using `click.get_current_context(silent=True)`. I've also changed the header to something more explicit, from `mode` to `GGShield-HeaderName`. The current headers are:
- `GGShield-Version` ggshield version
- `GGShield-Command-Path` full path of the scan, ie: `ggshield scan path`
- ~~`GGShield-Command-Name` name of the command, ie: `path-cmd`~~

### DIFF
--- a/ggshield/scan/scannable.py
+++ b/ggshield/scan/scannable.py
@@ -7,6 +7,7 @@ from pygitguardian import GGClient
 from pygitguardian.config import MULTI_DOCUMENT_LIMIT
 from pygitguardian.models import ScanResult
 
+from ggshield import __version__
 from ggshield.config import CPU_COUNT, MAX_FILE_SIZE, Cache
 from ggshield.config_types import IgnoredMatch
 from ggshield.filter import (
@@ -107,6 +108,19 @@ class Files:
     def scannable_list(self) -> List[Dict[str, Any]]:
         return [entry.scan_dict for entry in self.files.values()]
 
+    @property
+    def extra_headers(self) -> Dict[str, str]:
+        # get_current_context returns None if outside a click command.
+        # It happens in the tests and if gg-shield is used as a library.
+        context = click.get_current_context(silent=True)
+
+        command_path = context.command_path if context is not None else "external"
+
+        return {
+            "GGShield-Version": __version__,
+            "GGShield-Command-Path": command_path,
+        }
+
     def scan(
         self,
         client: GGClient,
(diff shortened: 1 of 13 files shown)


ggshield/scan/scannable.py
```diff
@@ -107,14 +108,27 @@ def files(self) -> Dict[str, File]:
...
+        # get_current_context returns None if outside a click command
+        context = click.get_current_context(silent=True)
+
+        command_path = context.command_path if context is not None else "external"
+        command_name = context.command.name if context is not None else "external"
+
+        return {
+            "GGShield-Version": __version__,
+            "GGShield-Command-Path": command_path,
+            "GGShield-Command-Name": command_name,
```
agateau-gg, line 122: If I understand correctly `command_name` is `command_path[-1]`, if this is correct then maybe we only need `GGShield-Command-Path`?

agateau-gg (changes requested):
Looks good overall, just some minor remarks.

Nice to see the code is smaller after the changes :)

ggshield/scan/scannable.py
Walz: The `command_name` is the name of the function and the `command_path` is the series of commands used in the CLI.   
For example the headers sent by `ggshield scan path setup.py` are:
```json
{
  "GGShield-Version": "1.11.0",
  "GGShield-Command-Path": "ggshield scan path",
  "GGShield-Command-Name": "path-cmd"
}
```

ggshield/scan/scannable.py
```diff
@@ -107,14 +108,27 @@ def files(self) -> Dict[str, File]:
     def scannable_list(self) -> List[Dict[str, Any]]:
         return [entry.scan_dict for entry in self.files.values()]
 
+    @property
+    def extra_headers(self) -> Dict[str, str]:
+        # get_current_context returns None if outside a click command
```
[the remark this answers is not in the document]
Walz: I've expanded the comment. It can also happen if someone uses gg-shield as a library.

ggshield/scan/scannable.py
agateau-gg: OK, but then we can still deduce the command name from the other, and it's more of an implementation detail: we could rename the function and the app would still work the same, as long as the Click decorators are kept unchanged. I think we don't need it.

Walz: I've removed `GGShield-Command-Name`.

agateau-gg: Thanks, going to merge it when CI is :green_circle:

agateau-gg (approved):

PR #198, opened 2022-04-06
fix(config): fix api/dashboard URLs parsing not handling params and queries

eugenenelou:
The config refactoring break an Integration test in docker due to the GITGUARDIAN_API_URL format that highlighted a parsing issue. The trailing / being only cleaned if the URL had no params nor queries.

This MR fixes it and adds unit tests for the `dashboard_to_api_url` and `api_to_dashboard_url` to test these edge cases.

Also cleans a potential /v1 and warns about it.


agateau-gg (approved):
Looks good!

PR #199, opened 2022-04-06
Jchaumont/3979/auth login token

julienc91:
Implementation of the auth login --method token flow.


ggshield/auth.py
```diff
@@ -10,21 +13,65 @@
...
+    if not instance:
+        instance = config.auth_config.default_instance
+
     if method == "token":
         token = click.prompt("Enter your GitGuardian API token", hide_input=True)
         if not token:
             raise click.ClickException("No API token was provided.")
 
-        config = ctx.obj["config"]
+        config.auth_config.current_instance = instance
```
martinfaucheux, line 43: nit: this will be used even it the method is not "token", this can go outside of the if block (before)

ggshield/auth.py
```diff
@@ -10,21 +13,65 @@
...
 
+        instance_config = config.auth_config.instances.setdefault(
+            instance,
+            InstanceConfig(
+                # account is initialized as None because the instance must exist in
+                # the config before using the client
+                account=None,  # type: ignore
+                url=instance,
+            ),
+        )
```
martinfaucheux, line 54: nit: this also doesn't depend on the method, this can go before the if block as well

martinfaucheux (approved):

martinfaucheux (approved):

ggshield/auth.py
```diff
@@ -10,21 +13,65 @@
...
     """Authenticate to your GitGuardian account."""
+    config = ctx.obj["config"]
+
+    if instance:
+        if not instance.startswith("https://"):
+            raise click.BadParameter("Instance must be a valid https URL.")
+
+        # Normalize instance name by removing the path, if it is set
+        parsed_url = urlparse(instance)
+        instance = f"https://{parsed_url.netloc}"
```
eugenenelou, line 33: nit: maybe you can reuse the `clean_url` method I added in [this PR](https://github.com/GitGuardian/ggshield/pull/198)

eugenenelou (approved):

tests/test_auth_cli.py
```diff
@@ -0,0 +1,154 @@
...
+            ("test_auth_login_token_invalid_scope", False),
+            ("test_auth_login_token_valid", True),
+        ],
+    )
+    def test_auth_login_token(self, cli_fs_runner, cassette, expect_success):
+        """
+        GIVEN an API token, valid or not
+        WHEN the auth login command is called with --method=token
+        THEN the validity of the token should be checked, and if valid, the user should be logged in
+        """
```
agateau-gg, line 30: I don't understand how this test works: did you edit the cassettes by hand to produce the expected responses? If so, that sounds dangerous: I would expect to be able to rm all cassettes to update them.

ggshield/auth.py
```diff
@@ -0,0 +1,82 @@
...
+                expire_at=api_token_data.get("expire_at"),
+                token_name=api_token_data.get("name", ""),
+                type=api_token_data.get("type", ""),
+            )
+
+            instance_config.account = account_config
+            config.save()
+            click.echo("Authentication was successful.")
+        else:
+            raise click.ClickException("Authentication failed with token.")
```
agateau-gg, line 76: nit: you can reduce indentation by raising early:

```python
if not response.ok:
    raise click.ClickException("...)
# rest of the code
```

ggshield/auth.py
julienc91: Ok I'll wait for your PR to be merged then

tests/test_auth_cli.py
julienc91: Yes, but I don't see how I could do it like you suggest without using valid tokens. I could do what I did for the other tests however, which is mocking the entire API call without relying to cassettes.

tests/test_git_shell.py
```diff
@@ -1,6 +1,10 @@
+import os
+from unittest.mock import patch
+
 from ggshield.git_shell import GIT_PATH, is_git_dir, shell
 
 
+@patch.dict(os.environ, {"LANG": "C"})
```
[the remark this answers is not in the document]
julienc91: I tried doing it in `shell` by adding `["LANG", "C"]` before `["git", ...]`, but it didn't work. Plus, I'm not sure what the impact would be if we enforced the `LANG` everytime we call the git process (especially in CI environments).

tests/test_auth_cli.py
agateau-gg: I think it would be better to mock, indeed. Maybe you can use a requests mocker, like [requests-mock](https://github.com/jamielennox/requests-mock)?

I am also starting to think maybe the details of the network calls should be in pygitguardian, what do you think? :thinking:

julienc91: I did a PR on pygitguardian ([here](https://github.com/GitGuardian/py-gitguardian/pull/30)) but the idea was rejected because the endpoint is not supposed to be public (it won't be documented in the public documentation).

I'll reuse the monkeypatch solution I already did in the other tests for now :slightly_smiling_face:

tests/test_git_shell.py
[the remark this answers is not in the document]
agateau-gg: That is most likely because the `LANG=C git...` syntax is interpreted by your shell (as in, bash, zsh...). `shell()` uses `subprocess.run()` without going through the shell (which is a good thing) so you have to force the env with something like this:

```python
env = os.environ.copy()
env["LANG"] = "C"
subprocess.run(..., env=env)
```

I don't think it would be a problem on CI environments, where LANG is most likely to be set to C or en_US, but it could avoid issues when running on users machines, where things are more... exotic (I remember a bug in a tool parsing numbers which failed on Windows Thai machines because numbers were localized...)

tests/test_auth_cli.py
agateau-gg: Ah OK, I am not sure I agree with the reasoning, but monkeypatching sounds good for now :+1:.

tests/test_git_shell.py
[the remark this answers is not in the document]
julienc91: Ok I'll add that, thanks for your input!

agateau-gg (approved):
Looks good to me :+1:

ggshield/auth.py
agateau-gg: I just merged Eugène PR, you can now use `clean_url`.

eugenenelou (approved):

PR #200, opened 2022-04-08
Implement ggshield web oauth flow

martinfaucheux:
Implement ggshield login auth "web" method which implement the OAuth2 flow.
The core logic is implemented in `OAuthClient.oauth_process()`. In short it does
* redirect to GitGuarian login page (with ggshield options)
* open a local server to start listening
* retrieve the code from the call to localhost
* validate the token and retrieve its data
* save the token to the config
* open a tab to the "authenticated" page

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,12 +4,13 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [packages]
-ggshield = { editable = true, path = "." }
+appdirs = ">=1.4.4,<1.5.0"
 click = ">=8.0,<8.1"
+ggshield = { editable = true, path = "." }
+oauthlib = "~=3.2"
 pygitguardian = ">=1.3.4,<1.4.0"
 pyyaml = ">=6.0,<6.1"
 python-dotenv = ">=0.19.1,<0.20.0"
-appdirs = ">=1.4.4,<1.5.0"
 
 [dev-packages]
 black = "==22.3.0"
@@ -28,5 +29,7 @@ snapshottest = "*"
 types-PyYAML = "*"
 types-requests = "*"
 types-appdirs = "*"
+types-oauthlib = "*"
+typing-extensions = "*"
 vcrpy = "*"
 pyfakefs = "*"
--- a/ggshield/auth.py
+++ b/ggshield/auth.py
@@ -2,6 +2,7 @@ import click
 
 from .client import retrieve_client
 from .config import AccountConfig, InstanceConfig
+from .oauth import OAuthClient
 from .utils import clean_url
 
 
@@ -9,7 +10,7 @@ from .utils import clean_url
 @click.option(
     "--method",
     required=True,
-    type=click.Choice(["token"]),
+    type=click.Choice(["token", "web"]),
     help="Authentication method.",
 )
 @click.option(
@@ -74,7 +75,8 @@ def login_cmd(ctx: click.Context, method: str, instance: str) -> int:
         instance_config.account = account_config
         config.save()
         click.echo("Authentication was successful.")
-
+    elif method == "web":
+        OAuthClient(config, instance).oauth_process()
     return 0
 
 
--- a/ggshield/auth.py
+++ b/ggshield/auth.py
@@ -1,3 +1,5 @@
+import os
+
 import click
 
 from .client import retrieve_client
@@ -21,7 +23,14 @@ from .utils import clean_url
 )
 @click.pass_context
 def login_cmd(ctx: click.Context, method: str, instance: str) -> int:
-    """Authenticate to your GitGuardian account."""
+    """
+    Authenticate to your GitGuardian account.
+
+    Use `--method token` to authenticate using an existing token.
+
+    Use `--method web` to let ggshield authenticate through your web browser and
+    generate a token for you. Note: This is experimental for now.
+    """
     config = ctx.obj["config"]
 
     if instance:
@@ -76,7 +85,10 @@ def login_cmd(ctx: click.Context, method: str, instance: str) -> int:
         config.save()
         click.echo("Authentication was successful.")
     elif method == "web":
-        OAuthClient(config, instance).oauth_process()
+        if os.getenv("IS_WEB_AUTH_ENABLED", False):
+            OAuthClient(config, instance).oauth_process()
+        else:
+            raise click.ClickException("The web auth login method is not enabled.")
     return 0
 
 
--- a/ggshield/oauth.py
+++ b/ggshield/oauth.py
@@ -17,7 +17,12 @@ from .config import AccountConfig, Config
 
 CLIENT_ID = "ggshield_oauth"
 SCOPE = "scan"
-REDIRECT_PORT = 1234
+
+# potential port range to be used to run local server
+# to handle authorization code callback
+# this is the largest band of not commonly occupied ports
+# https://stackoverflow.com/questions/10476987/best-tcp-port-number-range-for-internal-applications
+USABLE_PORT_RANGE = (29170, 29998)
 
 
 class OAuthClient:
@@ -35,6 +40,7 @@ class OAuthClient:
 
         self._handler_wrapper = RequestHandlerWrapper(oauth_client=self)
         self._access_token: Optional[str] = None
+        self._port = USABLE_PORT_RANGE[0]
         self.server: Optional[HTTPServer] = None
 
         self._generate_pkce_pair()
@@ -107,15 +113,20 @@ class OAuthClient:
         webbrowser.open_new_tab(request_uri)
 
     def _prepare_server(self) -> None:
-        try:
-            self.server = HTTPServer(
-                # only consider requests from localhost on the predetermined port
-                ("127.0.0.1", REDIRECT_PORT),
-                # attache the wrapped request handler
-                self._handler_wrapper.request_handler,
-            )
-        except OSError:
-            raise click.ClickException(f"Port {REDIRECT_PORT} is already in use.")
+        for port in range(*USABLE_PORT_RANGE):
+            try:
+                self.server = HTTPServer(
+                    # only consider requests from localhost on the predetermined port
+                    ("127.0.0.1", port),
+                    # attach the wrapped request handler
+                    self._handler_wrapper.request_handler,
+                )
+                self._port = port
+                break
+            except OSError:
+                continue
+        else:
+            raise click.ClickException("Could not find unoccupied port.")
 
     def _wait_for_callback(self) -> None:
         """
@@ -206,7 +217,7 @@ class OAuthClient:
 
     @property
     def redirect_uri(self) -> str:
-        return f"http://localhost:{REDIRECT_PORT}"
+        return f"http://localhost:{self._port}"
 
     @property
     def dashboard_url(self) -> str:
--- a/ggshield/oauth.py
+++ b/ggshield/oauth.py
@@ -1,0 +1,290 @@
+import os
+import urllib.parse as urlparse
+import webbrowser
+from base64 import urlsafe_b64encode
+from datetime import datetime
+from hashlib import sha256
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from typing import Any, Dict, Optional, Type
+
+import click
+import requests
+from oauthlib.oauth2 import OAuth2Error, WebApplicationClient
+
+from .client import retrieve_client
+from .config import AccountConfig, Config
+
+
+CLIENT_ID = "ggshield_oauth"
+SCOPE = "scan"
+REDIRECT_PORT = 1234
+
+
+class OAuthClient:
+    """
+    Helper class to handle the OAuth authentication flow
+    the logic is divided in 2 steps:
+    - open the browser on GitGuardian login screen and run a local server to wait for callback
+    - handle the oauth callback to exchange an authorization code against a valid access token
+    """
+
+    def __init__(self, config: Config, instance: str) -> None:
+        self.config = config
+        self.instance = instance
+        self._oauth_client = WebApplicationClient(CLIENT_ID)
+
+        self._handler_wrapper = RequestHandlerWrapper(oauth_client=self)
+        self._access_token: Optional[str] = None
+        self.server: Optional[HTTPServer] = None
+
+        self._generate_pkce_pair()
+
+    def oauth_process(self) -> None:
+        """
+        Handle the whole oauth process which includes
+        - opening the user's webbrowser to GitGuardian login page
+        - open a server and wait for the callback processing
+        """
+        # enable redirection to http://localhost
+        os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = str(True)
+
+        self._token_name = "ggshield token " + datetime.today().strftime("%Y-%m-%d")
+
+        self._prepare_server()
+        self._redirect_to_login()
+        self._wait_for_callback()
+        click.echo(f"Created Personal Access Token {self._token_name}")
+
+    def process_callback(self, callback_url: str) -> None:
+        """
+        This function runs within the request handler do_GET method.
+        It takes the url of the callback request as argument and does
+        - Extract the authorization code
+        - Exchange the code against an access token with GitGuardian's api
+        - Validate the new token against GitGuardian's api
+        - Save the token in configuration
+        Any error during this process will raise a OAuthError
+        """
+        authorization_code = self._get_code(callback_url)
+        self._claim_token(authorization_code)
+        token_data = self._validate_access_token()
+        self._save_token(token_data)
+
+    def _generate_pkce_pair(self) -> None:
+        """
+        Generate a code verifier (random string) and its sha encoded version to be used
+        for the pkce checking process
+        """
+        self.code_verifier = self._oauth_client.create_code_verifier(128)  # type: ignore
+        self.code_challenge = (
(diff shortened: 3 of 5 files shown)


ggshield/oauth.py
```diff
@@ -0,0 +1,240 @@
...
+    def _wait_for_code(self) -> str:
+        """
+        Open a local server on the predetermined port to receive the post-login redirection
+        including the authorazation code.
+        It uses a special request handler which is able to kill the server once the code
+        is received
+        return the authorization code or None if none could be fetched.
+        """
+        handler_wrapper = RequestHandlerWrapper(oauth_client=self)
+        server = HTTPServer(("", REDIRECT_PORT), handler_wrapper.request_handler)
```
julienc91, line 113: You're opening HTTPServer to "0.0.0.0", meaning that it will be exposed to all your interfaces. You should restrict to localhost.

ggshield/oauth.py
```diff
@@ -0,0 +1,240 @@
...
+        handler_wrapper.server = server
+
+        try:
+            server.serve_forever()
+        except OSError:
+            raise click.ClickException(f"Port {REDIRECT_PORT} is already in used.")
```
julienc91, line 119:
```suggestion
            raise click.ClickException(f"Port {REDIRECT_PORT} is already in use.")
```

tests/test_auth_cli.py
```diff
@@ -7,6 +7,15 @@
 from ggshield.config import Config
 
 
+_TOKEN_RESPONSE_PAYLOAD = {
```
julienc91, line 10: You can slightly reduce code redundancy in `TestAuthLoginToken` by doing so:

```
VALID_TOKEN_PAYLOAD = {**_TOKEN_RESPONSE_PAYLOAD}
VALID_TOKEN_INVALID_SCOPE_PAYLOAD = {**_TOKEN_RESPONSE_PAYLOAD, "scope": ["read:incident", "write:incident", "share:incident", "read:member"]}
```

tests/test_auth_cli.py
```diff
@@ -7,6 +7,15 @@
...
+    "account_id": 17,
+    "name": "key",
+    "scope": ["scan"],
+    "expire_at": None,
+}
+
+
 @pytest.fixture(autouse=True)
 def tmp_config(monkeypatch, tmp_path):
     monkeypatch.setattr("ggshield.config.get_auth_config_dir", lambda: str(tmp_path))
```
julienc91, line 21: The `TestAuthLoginToken` class was supposed to be restricted to the `--method=token` flow. You should either rename it or create a `TestAuthLoginWeb` class.

julienc91 (changes requested):

ggshield/auth.py
```diff
@@ -74,7 +75,10 @@ def login_cmd(ctx: click.Context, method: str, instance: str) -> int:
         instance_config.account = account_config
         config.save()
         click.echo("Authentication was successful.")
-
+    elif method == "web":
+        OauthClient(config, instance).oauth_process()
+    else:
+        raise click.ClickException("Authentication failed with token.")
```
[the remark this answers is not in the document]
martinfaucheux: yes you're right I'll simply remove it

ggshield/oauth.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        Handle the whole oauth process which includes
+        - opening the user's webbrowser to GitGuardian login page
+        - open a server and wait from the browser redirection
+        - retrieve the oauth authorization code from the request uri
+        - exchange the authorization code against an access token user gg's api
+        - save the access token
+        """
+        # enable redirection to http://localhost
+        os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = str(True)
```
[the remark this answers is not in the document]
martinfaucheux: as we discussed, this variable is set only for the running script. If the script fails, it won't persist afterwards

ggshield/oauth.py
```diff
@@ -0,0 +1,242 @@
...
+
+                    self.stop_server()
+
+        return RequestHandler
+
+    def stop_server(self) -> None:
+        # The server cannot be killed from the main thread
+        killer = threading.Thread(target=self.server.shutdown)  # type: ignore
+        killer.daemon = True
+        killer.start()
```
agateau-gg, line 242: I am wondering how this is going to perform on Windows. Going to give it a try.

agateau-gg (changes requested):

tests/test_auth_cli.py
```diff
@@ -161,3 +154,191 @@ def test_auth_login_token_update_existing_config(self, monkeypatch, cli_fs_runne
...
+                ok=is_token_valid,
+                json=lambda: (_TOKEN_RESPONSE_PAYLOAD if is_token_valid else {}),
+            )
+        )
+        monkeypatch.setattr("ggshield.client.GGClient.get", client_get_mock)
+
+        # run cli command
+        cmd = ["auth", "login", "--method=web"]
+        result = cli_fs_runner.invoke(cli, cmd, color=False, catch_exceptions=True)
+        print(result.output)
```
agateau-gg, line 245: This can be removed

tests/test_auth_cli.py
```diff
@@ -161,3 +154,191 @@ def test_auth_login_token_update_existing_config(self, monkeypatch, cli_fs_runne
...
+            client_post_mock.assert_called_once()
+            self._assert_post_payload(client_post_mock, token_name)
+            client_get_mock.assert_called_once()
+
+            assert result.output.endswith(
+                f"\nCreated Personal Access Token {token_name}\n"
+            )
+            self._assert_config(token)
+
+    def _assert_config_is_empty(self):
```
agateau-gg, line 298: This method can be static

tests/test_auth_cli.py
```diff
@@ -161,3 +154,191 @@ def test_auth_login_token_update_existing_config(self, monkeypatch, cli_fs_runne
...
+        config = Config()
+        assert len(config.auth_config.instances) == 0
+
+    def _assert_config(self, token=None):
```
agateau-gg, line 302: This method can be static

ggshield/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+    def oauth_process(self) -> None:
+        """
+        Handle the whole oauth process which includes
+        - opening the user's webbrowser to GitGuardian login page
+        - open a server and wait for the callback processing
+        """
+        # enable redirection to http://localhost
+        os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = str(True)
+
+        # TODO: get the token name from init params
```
agateau-gg, line 44: Do we need this to be done before merging?

ggshield/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+                    self_.send_header("Location", redirect_url)
+                self_.end_headers()
+
+        return RequestHandler
+
+
+class OAuthError(Exception):
+    """
+    Exception raised during the authorization exchange code process
+    Its message is catched and will be raise again as a click Exception
```
agateau-gg, line 270:
```suggestion
    Its message is caught and will be raised again as a click Exception
```

But actually maybe you can make OAuthError inherit from ClickException? Then you don't have to raise it again?

ggshield/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+        request_uri = self._oauth_client.prepare_request_uri(
+            uri=urlparse.urljoin(self.dashboard_url, "auth/login"),
+            redirect_uri=self.redirect_uri,
+            scope=SCOPE,
+            code_challenge=self.code_challenge,
+            code_challenge_method="S256",
+            **static_params,
+        )
+        click.echo(
+            "Complete the login via GitGuardian client. Launching browser to:\n"
```
agateau-gg, line 98: I think "client" is not the right word here. I would say something like this:

> To complete the login process, follow the instructions from {request_uri}.
> Opening your web browser now...

agateau-gg (changes requested):
Some minor remarks, but it's getting good.

ggshield/oauth.py
martinfaucheux: I can't really do that because OAuthError is raised withing the request handler which wraps the exception. I have to pass the message through the RequestHandlerWrapper to re-raise the exception out of its initial scope

ggshield/oauth.py
martinfaucheux: I copied this one from Vault auth client, I can change it

ggshield/oauth.py
agateau-gg: Ah OK, makes sense. Can you just fix the typos, then? :)

Pipfile
```diff
@@ -20,6 +21,7 @@ flake8-quotes = "*"
 ipdb = "*"
 mock = "*"
 mypy = "*"
+types-oauthlib = "*"
```
eugenenelou, line 24: you should reorder this as well

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+import requests
+from oauthlib.oauth2 import OAuth2Error, WebApplicationClient
+
+from .client import retrieve_client
+from .config import AccountConfig, Config
+
+
+CLIENT_ID = "ggshield_oauth"
+SCOPE = "scan"
+REDIRECT_PORT = 1234
```
eugenenelou, line 21: we should look for an open port, and optionally to allow the user to choose the port manually.

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+class OAuthClient:
```
eugenenelou, line 24: nit: :book: a short docstring could be useful

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+        """
+        # enable redirection to http://localhost
+        os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = str(True)
+
+        self._token_name = "ggshield token " + datetime.today().strftime("%Y-%m-%d")
+
+        self._prepare_server()
+        self._redirect_to_login()
+        self._wait_for_code()
+        click.echo(f"Created Personal Access Token {self._token_name}")
```
eugenenelou, line 49: It does not feel very explicit what this function does, both reading the docstring and the code.

I don't expect `_wait_for_code` to do much but it appears to even create the token. There are some names to improve.

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+        self._claim_token(authorization_code)
+        token_data = self._validate_access_token()
+        self._save_token(token_data)
+
+    def _generate_pkce_pair(self) -> None:
+        """
+        Generate a code verifier (random string) and its sha encoded version to be used
+        for the pkce checking process
+        """
+        self.code_verifier = token_urlsafe(96)[:128]
```
eugenenelou, line 71: :question: what's the point of `[:128]`? it's a fixed number of bytes, it should be the same character length for the accepted characters, no?

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+        self.code_challenge = (
+            urlsafe_b64encode(sha256(self.code_verifier.encode()).digest())
+            .decode()
+            .rstrip("=")
```
eugenenelou, line 75: :question: why?

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+        except OSError:
+            raise click.ClickException(f"Port {REDIRECT_PORT} is already in use.")
+
+    def _wait_for_code(self) -> None:
+        """
+        Open a local server on the predetermined port to receive the post-login redirection
+        including the authorazation code.
+        It uses a special request handler which is able to kill the server once the code
+        is received
+        return the authorization code or None if none could be fetched.
```
eugenenelou, line 116: this function does not return anything.

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+                self.server.handle_request()  # type: ignore
+        except KeyboardInterrupt:
+            raise click.ClickException("Aborting")
+
+        if self._handler_wrapper.error_message is not None:
+            raise click.ClickException(self._handler_wrapper.error_message)
+
+    def _get_code(self, uri: str) -> str:
+        """
+        extract the authorization from the incoming request uri
```
eugenenelou, line 129: `uri` -> `URI` in every comments/docstring

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+        )
+
+        response = requests.post(
+            urlparse.urljoin(self.api_url, "oauth/token"),
+            request_body,
+            headers={"Content-Type": "application/x-www-form-urlencoded"},
+        )
+
+        if not response.ok:
+            raise OAuthError("Cannot retrieve token.")
```
eugenenelou, line 164: :question: this is not a retrieve operation, it's a creation

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+
+        self._access_token = response.json()["key"]
+        self.config.auth_config.current_token = self._access_token
+
+    def _validate_access_token(self) -> Dict[str, Any]:
+        """
+        Validate the token using GitGuardian public api.
+        If the token is not valid, exit the authentication process with an error message.
+        """
+        client = retrieve_client(self.config)
```
eugenenelou, line 174: :art: the client keeps a session, it would be better to reuse the same client everywhere in this class.

a `client` cached_property should do.

ggshield/oauth.py
```diff
@@ -0,0 +1,273 @@
...
+                If not the response will be a user error 400
+                """
+                status_code = 301 if redirect_url is not None else 400
+                self_.send_response(status_code)
+
+                if redirect_url is not None:
+                    self_.send_header("Location", redirect_url)
+                self_.end_headers()
+
+        return RequestHandler
```
eugenenelou, line 263: :question: I'm not very experienced with this server handling, but I have a question: why couldn't we just open the server in the main thread? killing the main process would kill the server as we want, and it would make the code a lot simpler to read and understand.
Right now, the code in the OAuth just contains half of the flow.

ggshield/oauth.py
agateau-gg: Ah right. Vault uses "your OIDC provider" instead of "GitGuardian client", but I don't think it's super clear either. I'd prefer if we changed it.

ggshield/oauth.py
martinfaucheux: In fact I'll revert to the `oauthlib`'s `create_code_verifier` method
https://github.dev/oauthlib/oauthlib/blob/b123283ba3d41acb3e787fdf68bd5907972b4bad/oauthlib/oauth2/rfc6749/clients/base.py#L468

ggshield/oauth.py
```diff
@@ -102,31 +109,37 @@ def _redirect_to_login(self) -> None:
     def _prepare_server(self) -> None:
         try:
             self.server = HTTPServer(
-                ("127.0.0.1", REDIRECT_PORT), self._handler_wrapper.request_handler
+                # only consider requests from localhost on the predetermined port
+                ("127.0.0.1", REDIRECT_PORT),
+                # attache the wrapped request handler
```
eugenenelou: :eyes: attach

ggshield/oauth.py
```diff
@@ -107,15 +113,21 @@ def _redirect_to_login(self) -> None:
...
+                self.server = HTTPServer(
+                    # only consider requests from localhost on the predetermined port
+                    ("127.0.0.1", port),
+                    # attach the wrapped request handler
+                    self._handler_wrapper.request_handler,
+                )
+                self._port = port
+                break
+            except OSError:
+                click.echo(f"Port {port} already in use, try port {port+1}")
```
eugenenelou, line 127: nit: I'm not sure this is required, it may spam the console.

eugenenelou (approved):

eugenenelou:
port searching is good :+1:

ggshield/oauth.py
```diff
@@ -0,0 +1,302 @@
...
+        self._claim_token(authorization_code)
+        token_data = self._validate_access_token()
+        self._save_token(token_data)
+
+    def _generate_pkce_pair(self) -> None:
+        """
+        Generate a code verifier (random string) and its sha encoded version to be used
+        for the pkce checking process
+        """
+        self.code_verifier = self._oauth_client.create_code_verifier(43)  # type: ignore
```
eugenenelou, line 84: :question: why the smallest possible? and not 128? it does not cost much, no?

agateau-gg (approved):
Looks good to me. The only question I have is about the availability of the server-side part. If it's not ready yet, we may want to hide this behind a feature flag.

tests/test_auth_cli.py
```diff
@@ -161,3 +170,251 @@ def test_auth_login_token_update_existing_config(self, monkeypatch, cli_fs_runne
             config.auth_config.instances[second_instance].account.token
             == second_instance_token
         )
+
+
+class TestAuthLoginWeb:
+    def test_auth_login_web_not_enabled(self, cli_fs_runner):
+        cmd = ["auth", "login", "--method=web"]
```
julienc91, line 177: GIVEN/WHEN/THEN

julienc91 (approved):

tests/test_auth_cli.py
julienc91, line 177: GIVEN/WHEN/THEN

PR #206, opened 2022-04-19
Fix Azure Pipelines example script

agateau-gg:
Azure Pipelines expects the Docker container to start as root. They then run `useradd -m -u 1001 vsts_azpcontainer` in the container to create a matching non-root user.

This fails with ggshield Docker container because it starts as its own non-privileged user, which is not allowed to run `useradd`.

Tell Azure Pipelines to start the container as root (with `-u 0`) to avoid that issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -963,7 +963,9 @@ jobs:
   - job: GitGuardianShield
     pool:
       vmImage: 'ubuntu-latest'
-    container: gitguardian/ggshield:latest
+    container:
+      image: gitguardian/ggshield:latest
+      options: -u 0
     steps:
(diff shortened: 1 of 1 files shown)


jeremyds:
:pray: 
Would you mind to add an azure pipeline config yaml file to the repo?

jeremyds (approved):

agateau-gg:
> pray Would you mind to add an azure pipeline config yaml file to the repo?

As discussed on Slack, I think it's better to do this when we have an Azure Pipelines account setup for ggshield.

PR #208, opened 2022-04-21
Add auth login optional params for web method

martinfaucheux:
- refactor tests into separate test functions
- handle optional parameters for the auth login web method
  - token name
  - lifetime of the token (in days)
- make sure that these value are included in the state passed to the authentication server and that they are returned unchanged in the callback url
- make the server silent so it doesn't log incoming http requests
- remove the feature flag for more clarity (as requested by @agateau-gg )

### DIFF
--- a/ggshield/cmd/auth/login.py
+++ b/ggshield/cmd/auth/login.py
@@ -1,5 +1,3 @@
-import os
-
 import click
 
 from ggshield.cmd.auth.utils import check_instance_has_enabled_flow
@@ -90,8 +88,5 @@ def login_cmd(ctx: click.Context, method: str, instance: str) -> int:
     check_instance_has_enabled_flow(config=config)
 
     if method == "web":
-        if os.getenv("IS_WEB_AUTH_ENABLED", False):
-            OAuthClient(config, instance).oauth_process()
-        else:
-            raise click.ClickException("The web auth login method is not enabled.")
+        OAuthClient(config, instance).oauth_process()
     return 0
--- a/ggshield/cmd/auth/login.py
+++ b/ggshield/cmd/auth/login.py
@@ -20,8 +20,14 @@ from ggshield.core.utils import clean_url
     type=str,
     help="URL of the instance to authenticate to.",
 )
+@click.option(
+    "--token-name",
+    required=False,
+    type=str,
+    help="Name of new token.",
+)
 @click.pass_context
-def login_cmd(ctx: click.Context, method: str, instance: str) -> int:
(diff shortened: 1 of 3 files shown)


ggshield/cmd/auth/login.py
```diff
@@ -21,8 +21,27 @@
...
+    required=False,
+    type=str,
+    help="Name of new token.",
+)
+@click.option(
+    "--lifetime",
+    required=False,
+    type=None,
+    default=None,
+    help="Amount of days before the token expires. 0 corresponds to no expiry.",
```
agateau-gg, line 35:
```suggestion
    help="Amount of days before the token expires. 0 means the token never expires.",
```

ggshield/cmd/auth/login.py
```diff
@@ -27,8 +28,21 @@
     type=str,
     help="Name of new token.",
 )
+@click.option(
+    "--lifetime",
+    required=False,
+    type=None,
```
agateau-gg, line 33: You can use `IntRange(0)` for type. This makes Click validate more things for us, and make `--help` output more precise.

tests/cmd/auth/test_login.py
```diff
@@ -175,58 +171,68 @@ def test_auth_login_token_update_existing_config(self, monkeypatch, cli_fs_runne
...
+            # all ports already used
+            ("test token", 1000, "some_valid_authorization_code", True, True, True),
+            # can't get token from server
+            ("test token", 0, "some_valid_authorization_code", True, False, False),
+            # invalid token from server
+            ("test token", 0, "some_valid_authorization_code", True, True, False),
+            # invalid authorization code from callback
+            ("test token", 0, None, True, True, True),
+            # invalid state code from callback
+            ("test token", 0, None, False, True, True),
```
agateau-gg, line 199: This test method is getting a bit large. Maybe we need to think about factorizing the setup so that it can be split? Not sure what the best split would be, though :thinking:

Not a blocker to merge, just something to think about.

agateau-gg (changes requested):

tests/cmd/auth/test_login.py
martinfaucheux: I thought about it. The thing is there are so many things to mock that we must keep a reference to. I'll give it a try.

agateau-gg (approved):
Nice work on splitting the big test! :+1:

agateau-gg (changes requested):
Minor remarks: looks good overall

agateau-gg (approved):

PR #209, opened 2022-04-22
feat(auth): add a check that the instance handles the web flow before starting the flow

eugenenelou:
Add a check with several cases:
- if the /metadata endpoint returns 404: say the dashboard version is too low
- if the /metadata endpoint returns a version of `2022.04.x`: say the dashboard version is too low (:warning: the version is currently not well injected in the image, so it's a commit SHA, the next error message will be sent instead)
- if the /metadata returns the preference being off: say the flow is disabled.

If the preference is removed, we'll assume the feature is on by default.


ggshield/cmd/auth/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    client = retrieve_client(config)
+    response = client.get(endpoint="metadata")
+    if response.status_code == 404:
+        raise click.ClickException(VERSION_TOO_LOW_MESSAGE)
+    data = response.json()
+    if data["version"].startswith("2022.04"):
+        raise click.ClickException(VERSION_TOO_LOW_MESSAGE)
+
+    flow_enabled = data["preferences"].get(
+        "public_api__ggshield_auth_flow_enabled", True
```
agateau-gg, line 29: This looks like we are peeking at the internals. Are we confident this will remain available over time?

tests/cmd/auth/test_login.py
```diff
@@ -426,3 +452,57 @@ def _wait_for_callback(self, *args, **kwargs):
...
+                    json=lambda: {
+                        "version": version,
+                        "preferences": {
+                            "public_api__ggshield_auth_flow_enabled": preference_enabled
+                        }
+                        if preference_enabled is not None
+                        else {},
+                    },
+                )
+            raise NotImplementedError
```
agateau-gg, line 500: Should be `raise NotImplementedError()`, but maybe you can replace the `if endpoint == "metadata"` with `assert endpoint == "metadata"` instead?

agateau-gg (changes requested):

ggshield/cmd/auth/utils.py
eugenenelou: @agateau-gg no, a feature flag is made to be removed at some point, hence the `.get` when the flag is removed it should be considered always on.

tests/cmd/auth/test_login.py
agateau-gg: Oh nice, didn't know about raising the exception class.

I don't have strong opinions on `raise` vs `assert`, so LGTM.

agateau-gg (approved):

PR #214, opened 2022-04-29
Configuration file fixes

agateau-gg:
## Description

This PR fixes some issues I found in the new configuration files:
- Remove `dashboard-url` key: we do not want users to have to know the difference between the API URL and the dashboard URL.
- Introduce `instance` key: going to replace `api-url` in the future. Does the same job as `api-url` except it stores a dashboard URL.
- Do not store `config-path` in the configuration file when saving
- Keep storing `paths-ignore` and `banlisted-detectors` keys as lists, not as unordered sets
- If local config file is `.gitguardian` or `.gitguardian.yml`, save to this file, not to `.gitguardian.yaml`

For reference, here is how the main branch currently saves a config file (`matches-ignore` key removed for brevity):

```yaml
all-policies: false
allow-self-signed: false
banlisted-detectors: !!set {}
exit-zero: false
ignore-default-excludes: false
instance: https://dashboard.gitguardian.com
max-commits-for-hook: 50
paths-ignore: !!set
  '**/README.md': null
  .dockerignore: null
  .env: null
  .gitignore: null
  LICENSE: null
  doc/*: null
show-secrets: false
verbose: false
```

And here is how this branche saves:

```yaml
all-policies: false
allow-self-signed: false
banlisted-detectors: []
exit-zero: false
ignore-default-excludes: false
instance: https://dashboard.gitguardian.com
max-commits-for-hook: 50
paths-ignore:
- '**/README.md'
- .env
- .dockerignore
- doc/*
- .gitignore
- LICENSE
show-secrets: false
verbose: false

```

## What has been done

Changes are quite extensive, sorry for that :/. Hopefully the commit messages explain them clearly enough. Best reviewed commit-by-commit.


martinfaucheux (approved):

ggshield/core/config.py
```diff
@@ -156,29 +156,36 @@ def save(self, config_path: str) -> None:
...
+
+        for local_config_path in LOCAL_CONFIG_PATHS:
+            local_data = load_yaml(local_config_path)
+            if local_data and user_config.update_config(local_data):
+                config_path = local_config_path
+                break
+
+        if config_path is None:
+            config_path = DEFAULT_LOCAL_CONFIG_PATH
+        return user_config, config_path
```
eugenenelou, line 192: isn't it a breaking change to load global+local after the specified config file? the specified config file should stop the others from loading.

agateau-gg: No because in the case of the specified config file there is an early return (line 176). But your comment makes me think it's probably not-obvious, I can remove the return and use an `else` block like the original code.

ggshield/core/config.py
```diff
@@ -332,22 +307,8 @@ def __init__(self, instance: str):
 
 @dataclass
 class AuthConfig(YAMLFileConfig):
-    current_token: Optional[str] = None
-    current_instance: Optional[str] = None
-    default_instance: str = DEFAULT_DASHBOARD_URL
-    default_token_lifetime: Optional[int] = None
```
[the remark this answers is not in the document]
agateau-gg: Because it's never used: the OAuth code uses the value from InstanceConfig.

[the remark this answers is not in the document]
eugenenelou: @agateau-gg That's a problem, the default_token_lifetime is useful. But it cannot be used as you say because a None value in the instanceConfig will be considered as a "no expiry". 

I propose to use 0 as "no expiry" and disallowing `None` for default_lifetime_config, but allowing it for the instance Config lifetime.

ggshield/core/config.py
eugenenelou: Yes i looked for it quickly but didn't find it. Not used to review on GitHub.

I'm ok with it do as you wish.

eugenenelou (approved):

PR #217, opened 2022-05-02
Skip auth if existing token

martinfaucheux:
When using the web auth login method, if a token is found in the config for the current instance, skip the process with an explicit message.

This also include refactoring how the instance config are stored. instead of using a dict we know use a list.
This is not a real performance issue as the end users are not supposed to have many instances at once.
All the tests have been adapted to suit this new configuration file.

### DIFF
--- a/ggshield/core/config.py
+++ b/ggshield/core/config.py
@@ -236,7 +236,8 @@ class AccountConfig:
 
 @dataclass
 class InstanceConfig:
-    account: AccountConfig  # Only handle 1 account per instance for the time being
+    # Only handle 1 account per instance for the time being
+    account: Optional[AccountConfig]
     url: str
     name: Optional[str] = None
     # The token lifetime. If it's set it overrides AuthConfig.default_token_lifetime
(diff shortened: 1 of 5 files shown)


ggshield/core/config.py
```diff
@@ -258,7 +258,8 @@ def to_dict(self) -> Dict:
 
 @dataclass
 class InstanceConfig:
-    account: AccountConfig  # Only handle 1 account per instance for the time being
+    # Only handle 1 account per instance for the time being
+    account: Optional[AccountConfig]
```
agateau-gg, line 262: Why do we need `account` to become optional?

martinfaucheux: in `login.py` Julien instanciates the config with no account with this comment
> account is initialized as None because the instance must exist in
> the config before using the client

agateau-gg: Ah right, I think I can fix this in my refacto PR by using the lower-level [`create_client()`](https://github.com/GitGuardian/ggshield/blob/agateau/config-fixes/ggshield/core/client.py#L29) function. Going to give it a try.

martinfaucheux:
Note: this needs to wait to rebase on https://github.com/GitGuardian/ggshield/pull/214

agateau-gg (approved):
Minor remark, looks good!

PR #219, opened 2022-05-02
Allow to interrupt auth flow from auth server

martinfaucheux:
For ggshield auth login web method, if an error occur on the server side, we want to be able to stop the client from listening for callback.

What this PR do:
* if the callback contains an "error" parameter, it will interrupt the flow directly and output the error message given by the server

Note: this can't really be tested as the local server logic can only be mocked.

### DIFF
--- a/ggshield/core/oauth.py
+++ b/ggshield/core/oauth.py
@@ -38,6 +38,16 @@ def get_pretty_date(dt: datetime) -> str:
     return dt.strftime(f"%B {dt.day}{month_suffix} %Y")
 
 
+def get_error_param(parsed_url: urlparse.ParseResult) -> Optional[str]:
+    """
+    extract the value of the 'error' url param. If not present, return None.
+    """
+    params = urlparse.parse_qs(parsed_url.query)
+    if "error" in params:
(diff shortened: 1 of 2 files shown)


agateau-gg (approved):
It's a "I trust you" approve :)

martinfaucheux:
@agateau-gg in fact we will expect the error by the same route -> GET `http://localhost:2143/?error=some+message`

ggshield/core/oauth.py
```diff
@@ -363,27 +363,49 @@ def do_GET(self_) -> None:
...
+            def _get_error_param(
+                self_, parsed_url: urlparse.ParseResult
+            ) -> Optional[str]:
+                """
+                extract the value of the 'error' url param. If not present, return None.
+                """
+                params = urlparse.parse_qs(parsed_url.query)
+                if "error" in params:
+                    return params["error"][0]
+                return None
```
eugenenelou, line 399: You cannot do integration test due to the local server, but you could unit test functions like this to avoid simple mistakes.

eugenenelou (approved):

PR #221, opened 2022-05-04
Add auth logout command

martinfaucheux:
Add the `logout` command to ggshield client. It will delete the local configuration and attempt to revoke the token at the same time.

- implement the command `ggshield auth logout`
- it should remove the local configuration
- if `--instance` is not specified, it will resort to the default one
- it should revoke the old token by default unless `--no-revoke` option is passed
- if the `--all` option is passed, it will iterate over all found configs (`--instance` is ignored in this case)

### DIFF
--- a/ggshield/cmd/auth/logout.py
+++ b/ggshield/cmd/auth/logout.py
@@ -25,30 +25,37 @@ REVOKE_FAIL_MESSAGE = (
     default=True,
     help="Whether the token should be revoked before being removed from the config.",
 )
+@click.option("--all", "all_", is_flag=True, help="Iterate over every saved tokens.")
 @click.pass_context
-def logout_cmd(ctx: click.Context, instance: str, revoke: bool) -> int:
+def logout_cmd(ctx: click.Context, instance: str, revoke: bool, all_: bool) -> int:
     """
     Delete saved authentication details for the specified instance (or default instance if not specified)
     By default, this will also try to revoke found tokens unless --no-revoke is specified.\n
     If --all is specified, it will iterate over all instances.
     """
-
     config = ctx.obj["config"]
 
-    if not instance:
-        instance = config.instance_name
-
-    check_account_config_exists(config, instance)
-    if revoke:
-        revoke_token(config, instance)
-    logout(config, instance)
+    if all_:
+        for _instance in config.auth_config.instances:
+            logout(config, _instance.url, revoke=revoke)
+    else:
+        if not instance:
+            instance = config.instance_name
+        logout(config, instance, revoke=revoke)
     return 0
+
+
+def logout(config: Config, instance_url: str, revoke: bool) -> None:
+    check_account_config_exists(config, instance_url)
+    if revoke:
+        revoke_token(config, instance_url)
+    delete_account_config(config, instance_url)
 
 
 def check_account_config_exists(config: Config, instance_url: str) -> None:
     instance = config.auth_config.get_instance(instance_url)
     if instance.account is None:
-        raise click.ClickException(f"No token found for instance {instance_url}.")
+        raise LogoutError(f"No token found for instance {instance_url}.")
 
 
 def revoke_token(config: Config, instance_url: str) -> None:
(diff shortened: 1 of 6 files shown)


tests/cmd/auth/test_logout.py
```diff
@@ -0,0 +1,206 @@
...
+                + expected_output
+            )
+
+        assert output == expected_output
+
+    def test_logout_revoke_timeout(self, monkeypatch, cli_fs_runner):
+        """
+        GIVEN a saved instance configuration
+        WHEN running the logout command (with implied token revokation)
+        AND the revoke request gets a tiemout
```
agateau-gg, line 88: nit: typo
```suggestion
        AND the revoke request gets a timeout
```

tests/cmd/auth/test_logout.py
```diff
@@ -0,0 +1,206 @@
...
+            cmd.append("--instance=" + instance)
+        if no_revoke:
+            cmd.append("--no-revoke")
+        if all_tokens:
+            cmd.append("--all")
+        result = cli_fs_runner.invoke(cli, cmd, color=False)
+        return result.exit_code, result.output
+
+
+# TODO: mutualize this with test_login.py
```
agateau-gg, line 176: I think you forgot to remove this function and uses the one you added to utils.py :)

ggshield/cmd/auth/logout.py
```diff
@@ -0,0 +1,104 @@
...
+    "--all", "all_tokens", is_flag=True, help="Iterate over every saved tokens."
+)
+@click.pass_context
+def logout_cmd(
+    ctx: click.Context, instance: str, no_revoke: bool, all_tokens: int
+) -> int:
+    """
+    Delete local auth config for the specified instance (or default instance if not specified)
+    By default, this will also try to revoke found tokens unless --no-revoke is specified
+    if --all is specified, it will iterate over all instances.
```
agateau-gg, line 37:
```suggestion
    By default, this will also try to revoke found tokens unless --no-revoke is specified/
    If --all is specified, it will iterate over all instances.
```

ggshield/cmd/auth/logout.py
```diff
@@ -0,0 +1,104 @@
...
+)
+@click.option(
+    "--all", "all_tokens", is_flag=True, help="Iterate over every saved tokens."
+)
+@click.pass_context
+def logout_cmd(
+    ctx: click.Context, instance: str, no_revoke: bool, all_tokens: int
+) -> int:
+    """
+    Delete local auth config for the specified instance (or default instance if not specified)
```
agateau-gg, line 35: "local auth config" sounds a bit too ggshield-internal to me.

```suggestion
    Delete saved authentication details for the specified instance (or default instance if not specified).
```

ggshield/cmd/auth/logout.py
```diff
@@ -0,0 +1,104 @@
...
+
+@click.command()
+@click.option(
+    "--instance",
+    required=False,
+    type=str,
+    help="URL of the instance to logout from.",
+)
+@click.option(
+    "--no-revoke",
```
agateau-gg, line 23: It's better to use the `--revoke/--no-revoke` syntax here (see https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags), it's the recommended way to define boolean flags in Click. As a bonus, it avoids double negatives in the code: `no_revoke` becomes `revoke` (with a default value of False) and `if not no_revoke` becomes `if revoke`, which is more readable.

ggshield/cmd/auth/logout.py
```diff
@@ -0,0 +1,104 @@
...
+
+    client = create_client(
+        token,
+        dashboard_to_api_url(instance_url),
+        allow_self_signed=config.allow_self_signed,
+    )
+    try:
+        response = client.post(endpoint="token/revoke")
+    except ConnectionError:
+        raise click.ClickException(REVOKE_FAIL_MESSAGE)
```
agateau-gg, line 80: Should this be a LogoutError?

agateau-gg (changes requested):

agateau-gg (approved):

PR #222, opened 2022-05-04
fix: send verbose messages to stderr

### LINKED ISSUE #157: `ggshield scan --verbose --json path <some_file> > out.json` produces invalid JSON

- [x] I can reproduce this bug in the latest version

**Command executed**

`ggshield scan --verbose --json path some_file > out.json`

**Describe the bug**

When scanning a file or a directory with both the `--json` and the `--verbose` flags set, but without `--output`, then the produced JSON is invalid because stdout contains the verbose messages in addition to the json output.

```console
$ ggshield scan --json --verbose path setup.cfg > ~/tmp/out.json 

$ cat ~/tmp/out.json 
- /home/agateau/src/ggshield/setup.cfg
{"id": "/home/agateau/src/ggshield/setup.cfg", "type": "path_scan", "total_incidents": 0, "total_occurrences": 0}
```

**Expected behavior**

Verbose messages should be sent to stderr.

zhangmatthieu:
Send verbose messages to stderr to fix `--json` and `--verbose` behavior.

Closes #157

### DIFF
--- a/ggshield/cmd/scan/archive.py
+++ b/ggshield/cmd/scan/archive.py
@@ -1,4 +1,5 @@
 import shutil
+import sys
 import tempfile
 from pathlib import Path
 from typing import Any, Dict, List
@@ -37,7 +38,7 @@ def archive_cmd(ctx: click.Context, path: str) -> int:  # pragma: no cover
         )
 
         with click.progressbar(
-            length=len(files.files), label="Scanning"
+            length=len(files.files), label="Scanning", file=sys.stderr
         ) as progressbar:
 
             def update_progress(chunk: List[Dict[str, Any]]) -> None:
--- a/ggshield/cmd/scan/ci.py
+++ b/ggshield/cmd/scan/ci.py
@@ -25,7 +25,8 @@ def jenkins_range(verbose: bool) -> List[str]:  # pragma: no cover
 
     if verbose:
         click.echo(
-            f"\tGIT_COMMIT: {head_commit}" f"\nGIT_PREVIOUS_COMMIT: {previous_commit}"
+            f"\tGIT_COMMIT: {head_commit}" f"\nGIT_PREVIOUS_COMMIT: {previous_commit}",
+            err=True,
         )
 
     if previous_commit:
@@ -51,7 +52,8 @@ def travis_range(verbose: bool) -> List[str]:  # pragma: no cover
 
     if verbose:
         click.echo(
-            f"TRAVIS_COMMIT_RANGE: {commit_range}" f"\nTRAVIS_COMMIT: {commit_sha}"
+            f"TRAVIS_COMMIT_RANGE: {commit_range}" f"\nTRAVIS_COMMIT: {commit_sha}",
+            err=True,
         )
 
     if commit_range:
@@ -74,7 +76,7 @@ def travis_range(verbose: bool) -> List[str]:  # pragma: no cover
 def bitbucket_pipelines_range(verbose: bool) -> List[str]:  # pragma: no cover
     commit_sha = os.getenv("BITBUCKET_COMMIT", "HEAD")
     if verbose:
-        click.echo(f"BITBUCKET_COMMIT: {commit_sha}")
+        click.echo(f"BITBUCKET_COMMIT: {commit_sha}", err=True)
 
     commit_list = get_list_commit_SHA("{}~1...".format(commit_sha))
     if commit_list:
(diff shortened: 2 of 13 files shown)


agateau-gg (changes requested):
Looks good, but I would be more confident the fail won't come back if we had tests for this. Ideally all scan commands should be tested for this. The relevant test files are in tests/cmd/scan/.

I made the changes to some tests to get you started (tests are failing on main, it's up to you to make them pass ;):

```diff
diff --git a/tests/cmd/scan/test_docker.py b/tests/cmd/scan/test_docker.py
index 2b2236b..bd3397f 100644
--- a/tests/cmd/scan/test_docker.py
+++ b/tests/cmd/scan/test_docker.py
@@ -1,3 +1,4 @@
+import json
 from pathlib import Path
 from unittest.mock import Mock, patch
 
@@ -68,11 +69,13 @@ class TestDockerCMD:
     @pytest.mark.parametrize(
         "image_path", [DOCKER_EXAMPLE_PATH, DOCKER__INCOMPLETE_MANIFEST_EXAMPLE_PATH]
     )
+    @pytest.mark.parametrize("json_output", (False, True))
     def test_docker_scan_archive(
         self,
         get_files_mock: Mock,
         cli_fs_runner: click.testing.CliRunner,
         image_path: Path,
+        json_output: bool,
     ):
         assert image_path.exists()
 
@@ -80,15 +83,22 @@ class TestDockerCMD:
             files=[File(document=_SIMPLE_SECRET, filename="file_secret")]
         )
         with my_vcr.use_cassette("test_scan_file_secret"):
+            json_arg = ["--json"] if json_output else []
+            cli_fs_runner.mix_stderr = False
             result = cli_fs_runner.invoke(
                 cli,
                 [
                     "-v",
                     "scan",
+                    *json_arg,
                     "docker-archive",
                     str(image_path),
                 ],
             )
             get_files_mock.assert_called_once()
-            assert "1 incident has been found in file file_secret" in result.output
             assert result.exit_code == 1
+
+            if json_output:
+                json.loads(result.output)
+            else:
+                assert "1 incident has been found in file file_secret" in result.output
diff --git a/tests/cmd/scan/test_path.py b/tests/cmd/scan/test_path.py
index 988c11d..6d5e961 100644
--- a/tests/cmd/scan/test_path.py
+++ b/tests/cmd/scan/test_path.py
@@ -1,3 +1,4 @@
+import json
 import os
 from pathlib import Path
 
@@ -74,6 +75,7 @@ class TestPathScan:
 
         cassette_name = f"test_scan_file_secret-{validity}"
         with my_vcr.use_cassette(cassette_name):
+            cli_fs_runner.mix_stderr = False
             result = cli_fs_runner.invoke(
                 cli, ["-v", "scan", "--json", "path", "file_secret"]
             )
@@ -84,6 +86,7 @@ class TestPathScan:
             assert '"validity": "valid"' in result.output
         else:
             assert '"validity": "valid"' not in result.output
+        json.loads(result.output)
 
     def test_scan_file_secret_exit_zero(self, cli_fs_runner):
         Path("file_secret").write_text(UNCHECKED_SECRET)
```

zhangmatthieu:
I was checking how to add tests for prepush, prereceive and pypi and it seems that for the two first, there is no json output while for pypi, pip prints text to stdout so it won't be possible to pipe the json output for pypi. Also, a lot of the prints are also done in non-verbose mode. Maybe an alternative to allowing to pipe the result in a file with `>` would be to provide that as an argument (`ggshield -v scan --json=out.json ...`), this could prevent having to print everything to stderr in the future, what do you think? If that would solve the issue, we could add an issue about that

agateau-gg:
> EDIT: I saw there was already the `--output` flag for that. Maybe we can just accept that json piping won't work for pypi then I was checking how to add tests for prepush, prereceive and pypi and it seems that for the two first, there is no json output while for pypi, pip prints text to stdout so it won't be possible to pipe the json output for pypi. Also, a lot of the prints are also done in non-verbose mode. Maybe an alternative to allowing to pipe the result in a file with `>` would be to provide that as an argument (`ggshield -v scan --json=out.json ...`), this could prevent having to print everything to stderr in the future, what do you think? If that would solve the problem, we could add an issue about that

Oh, pip is annoying! You can force its output to go to stderr like this:

```diff
diff --git a/ggshield/cmd/scan/pypi.py b/ggshield/cmd/scan/pypi.py
index 75648f3..6be49a4 100644
--- a/ggshield/cmd/scan/pypi.py
+++ b/ggshield/cmd/scan/pypi.py
@@ -32,7 +32,8 @@ def save_package_to_tmp(temp_dir: str, package_name: str) -> None:
         subprocess.run(
             command,
             check=True,
-            stderr=subprocess.PIPE,
+            stdout=sys.stderr,
+            stderr=sys.stderr,
             timeout=PYPI_DOWNLOAD_TIMEOUT,
         )
         click.echo("OK", err=True)
```

agateau-gg (approved):
Nice!

PR #223, opened 2022-05-04
feat: print "No secrets have been found" when no secrets are found 

### LINKED ISSUE #117: Print a message when no secrets are detected
**Is your feature request related to a problem? Please describe.**
when ggshield does not detect secret and the verbose mode is not set, there is no output and the cli only returns a zero no error code. Although it is compliant with linux tool philosophy, it is not always clear for our users / customers that the scanning was successful and that there are no leaks. 

**Describe the solution you'd like**
This issue goal is more to track the output issue and think about it. To tackle the lack of information described above, we can either always output a clear message like "No leak found" even if the verbose mode is not set, or we can make the verbose mode set by default.

zhangmatthieu:
Print "No secrets have been found" when no secrets are found even in non verbose mode.

### Testing
- Added a test to check the output without the verbose tag

### Issue
- Closes #117

### DIFF
--- a/ggshield/output/text/text_output_handler.py
+++ b/ggshield/output/text/text_output_handler.py
@@ -49,13 +49,12 @@ class TextOutputHandler(OutputHandler):
             for result in scan.results:
                 scan_buf.write(self.process_result(result))
         else:
-            if self.verbose:
-                has_results = False
-                if scan.scans:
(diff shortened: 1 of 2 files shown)


agateau-gg (approved):
Looks good!

PR #224, opened 2022-05-05
Add auth config manipulation commands

martinfaucheux:
add the following commands to ggshield
* `ggshield auth config list` list the config  
* `ggshield auth config set <field> <value>`  -> set the value for the specific field for the default instance, for the specified instance (`--instance`) or for all configs (`--all`)
* `ggshield auth config unset <field>`  -> remove the value for the specific field for the default instance, for the specified instance (`--instance`) or for all configs (`--all`)
* `ggshield auth config get <field>`  -> get the value of the specific field for the default instance or for the specified instance (`--instance`)


ggshield/cmd/auth/auth_config.py
```diff
@@ -0,0 +1,67 @@
...
+        [default]
+        default_token_lifetime: 14d
+        workspace: 23:
+        url: *dashboard.gitguardian.com*
+        token: *62890f237c703c92fbda8236ec2a055ac21332a46115005c976d68b900535fb5*
+        token_name: *my_token*
+        token_type: *personal_access_token*
+        expiry: *2022-02-23T12:34:56Z*
+
+        """
```
agateau-gg, line 48: I think this can be removed

ggshield/cmd/auth/auth_config.py
```diff
@@ -0,0 +1,67 @@
...
+        else:
+            account_id = token = token_name = expiry = "not set"  # type: ignore
+
+        message_lines = [
+            f"[{instance_name}]",
+            f"default_token_lifetime: {instance.default_token_lifetime}",
+            f"workspace: {account_id}:",
+            f"url: {instance.url}:",
+            f"token: {token}:",
+            f"token_name: {token_name}:",
```
agateau-gg, line 70: Why are there `:` at the end of some fields?

tests/cmd/auth/test_auth_config.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import Tuple
+
+import pytest
+
+from ggshield.cmd.main import cli
+
+from .utils import prepare_config
+
+
+expected_output = """[https://dashboard.gitguardian.com]
```
agateau-gg, line 14: style: should be `EXPECTED_OUTPUT` since it's a top-level constant. Or move it to TestAuthConfigList, since it's related to the `list` command.

agateau-gg (changes requested):

ggshield/cmd/auth/auth_config.py
```diff
@@ -0,0 +1,195 @@
...
+                if expire_at is not None
+                else "never"
+            )
+        else:
+            account_id = token = token_name = expiry = "not set"  # type: ignore
+
+        message_lines = [
+            f"[{instance_name}]",
+            f"default_token_lifetime: {instance.default_token_lifetime}",
+            f"workspace: {account_id}:",
```
agateau-gg, line 67: I think this field should be called `account_id`.

martinfaucheux: I am following the tech specs https://www.notion.so/gitguardian/ggshield-auth-9ffc8074e97e4c439805246867380074#c989f6c9170a4f1b8762fb734153fbb9

but I will check

ggshield/cmd/auth/auth_config.py
martinfaucheux: that's a mistake

ggshield/cmd/auth/auth_config.py
agateau-gg: Indeed, I am curious if the different name is there on purpose.

martinfaucheux: from Eric:
> account is not something understandable for the end user.
we must use workspace(_id?) unfortunately

agateau-gg: Uh, OK.

ggshield/cmd/auth/auth_config.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+@click.command()
+@click.pass_context
+def config_list_cmd(ctx: click.Context) -> int:
+    """
+    List all auth configurations. Example of output
+
+    [default]
+    default_token_lifetime: 14d
```
eugenenelou, line 25: it should be `14` as a number to match the auth config format. There's a mistake in the specs for this particular command.

tests/cmd/auth/test_auth_config.py
```diff
@@ -0,0 +1,442 @@
...
+token_name: not set
+expiry: not set
+"""
+
+
+@pytest.fixture(autouse=True)
+def tmp_config(monkeypatch, tmp_path):
+    monkeypatch.setattr(
+        "ggshield.core.config.get_auth_config_dir", lambda: str(tmp_path)
+    )
```
eugenenelou, line 44: :question: why do you need this? most (all?) tests are run with a temporary filesystem already

martinfaucheux: otherwise it uses the local config so tests data is not truly isolated

martinfaucheux:
All comments are addressed. Please accept and merge

ggshield/cmd/auth/auth_config.py
```diff
@@ -0,0 +1,169 @@
...
+@click.option(
+    "--instance",
+    required=False,
+    type=str,
+    help="URL of the instance to set the config.",
+)
+@click.pass_context
+def config_set_command(
+    ctx: click.Context, field_name: str, value: str, instance: Optional[str]
+) -> int:
```
agateau-gg, line 74: Since we are exposing `account_id` as `workspace`, we must translate it here. But I think we should actually rename the field from `account_id` to `workspace` in the config file, would save us some trouble.

agateau-gg (changes requested):

eugenenelou (approved):

agateau-gg:
Did some manual testing and I found a few minor issues with the `--help` output.

```
$ ggshield auth config get --help

Usage: ggshield auth config get [OPTIONS] {default_token_lifetime}

  Get the value of the specified parameter. If --instance is passed, retrieve
  the value for this specific instance

Options:
  --instance TEXT  URL of the instance to set the config.
  -h, --help       Show this message and exit.
```

It's weird to see `{default_token_lifetime}` here, but there's not much we can do because if we set `metavar="FIELD_NAME"` then possible values are not listed anymore :(

The `--instance` option mentions setting the config, but we do a `get` here. Maybe make this a little generic and create a helper function to centralize the option definition (bonus point if you replace `TEXT` with `URL`, using `metavar`)

```
$ ggshield auth config list --help

Usage: ggshield auth config list [OPTIONS]

  List all auth configurations. Example of output

  [default] default_token_lifetime: 14 workspace_id: 23 url:
  *dashboard.gitguardian.com* token:
  *62890f237c703c92fbda8236ec2a055ac21332a46115005c976d68b900535fb5*
  token_name: *my_token* token_type: *personal_access_token* expiry:
  *2022-02-23T12:34:56Z*

Options:
  -h, --help  Show this message and exit.
```

The example is wrapped here, you can use `\b` to avoid that (see https://click.palletsprojects.com/en/8.1.x/documentation/#preventing-rewrapping)

```
$ ggshield auth config set --help

Usage: ggshield auth config set [OPTIONS] {default_token_lifetime} VALUE

  Set the value of a specific field in the auth config.

Options:
  --instance TEXT  URL of the instance to set the config.
  -h, --help       Show this message and exit.
```

```
$ ggshield auth config unset --help

Usage: ggshield auth config unset [OPTIONS] {default_token_lifetime}

  This command removes a parameter value from the protected config. if --all
  is passed, it will iterate over all protected configs.

Options:
  --instance TEXT  URL of the instance to set the config.
  --all            Iterate over every saved tokens.
  -h, --help       Show this message and exit.
```

This command talks about "protected config", the other use the "auth config" term instead. I think we should use "auth" everywhere.

ggshield/cmd/config/config_unset.py
```diff
@@ -0,0 +1,44 @@
...
+    help="URL of the instance to unset the config.",
+)
+@click.option("--all", "all_", is_flag=True, help="Iterate over every saved tokens.")
+@click.pass_context
+def config_unset_command(
+    ctx: click.Context, field_name: str, instance_url: Optional[str], all_: bool
+) -> int:
+    """
+    This command removes a parameter value from the auth config.
+    if --all is passed, it will iterate over all protected configs.
```
agateau-gg, line 27:
```suggestion
    if --all is passed, it iterates over all auth configs.
```

agateau-gg (changes requested):
Minor remark. Looks good otherwise.

PR #226, opened 2022-05-09
feat(auth): implement the --sso-url parameter to specify a SSO page to use for authentication

eugenenelou:
This adds an optional `--sso-url` parameter to the `ggshield auth login` command. to use with `--method=web`.

- it handles the collision with a `--instance` declaring another host.
- it handles deducing the instance from the sso url if the instance is not specified in the command.

### DIFF
--- a/ggshield/cmd/auth/login.py
+++ b/ggshield/cmd/auth/login.py
@@ -1,11 +1,51 @@
-from typing import Optional
+import re
+from typing import Optional, Tuple
 
 import click
 
 from ggshield.cmd.auth.utils import check_instance_has_enabled_flow
 from ggshield.core.client import create_client_from_config
-from ggshield.core.config import AccountConfig, InstanceConfig, UnknownInstanceError
+from ggshield.core.config import AccountConfig, Config
 from ggshield.core.oauth import OAuthClient
+from ggshield.core.utils import clean_url
+
+
+def validate_login_path(
+    config: Config, instance: Optional[str], sso_url: Optional[str]
+) -> Tuple[Optional[str], Optional[str]]:
+    """
+    Validate that the SSO URL and the instance refer to the same instance if they are both defined,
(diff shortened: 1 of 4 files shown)


tests/cmd/auth/test_login.py
```diff
@@ -479,12 +481,12 @@ def prepare_mocks(
 
         # run cli command
 
-    def run_cmd(self, cli_fs_runner):
+    def run_cmd(self, cli_fs_runner, method="web", expect_check_ff=True):
```
agateau-gg, line 484: What does "ff" mean in `expect_check_ff`?

agateau-gg (commented):
Just one question, looks good otherwise.

ggshield/cmd/auth/login.py
```diff
@@ -99,7 +145,17 @@ def login_cmd(
     check_instance_has_enabled_flow(config=config)
```
martinfaucheux, line 145: I tested in local when passing --instance, the instance is set in the config later so `check_instance_has_enabled_flow` won't check for the correct instance.

I think `check_instance_has_enabled_flow` should not take config as arg but instance instead.

Note: I am not sure that you can easily test sso but you can check the normal flow with staging instance by passing `--instance https://dashboard.staging.gitguardian.com`

martinfaucheux (changes requested):

tests/cmd/auth/test_login.py
eugenenelou: feature_flag, I could juste write it.

ggshield/cmd/auth/login.py
eugenenelou: ok, @martinfaucheux I'll move this in the `web` if block

agateau-gg (approved):

martinfaucheux (approved):
all good

PR #229, opened 2022-05-12
Plalanne/ /bring back mode header

pierrelalanne:
In this PR, we re-introduce the "mode" header for API calls.  
This header can take several values depending on the type of scan launched.

**Note :** We also took advantage of this PR to remove a useless verbose parameter.


agateau-gg (approved):
Looks good :+1:

PR #231, opened 2022-05-13
feat: Adding a filter for IAC files

zhangmatthieu:
- Adding a filter for IAC files based on the extension of the file, on excluded patterns and on the git status of the file (based on the ignore_git flag)

### Testing
- Added unit tests

### DIFF
--- a/ggshield/scan/scannable.py
+++ b/ggshield/scan/scannable.py
@@ -1,5 +1,6 @@
 import concurrent.futures
 import re
+from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, List, NamedTuple, Optional, Set
 
 import click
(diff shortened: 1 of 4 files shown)


ggshield/scan/scannable.py
```diff
@@ -87,6 +87,10 @@ def scan_dict(self) -> Dict[str, Any]:
             "filemode": self.filemode,
         }
 
+    def has_extensions(self, extensions: Iterable[str]) -> bool:
+        """Returns True iff the file has one of the given extensions."""
+        return any(re.search(rf"\.{ext}$", self.filename) for ext in extensions)
```
agateau-gg, line 92: Instead of using a regex, you can make `extensions` a `Set[str]` and then the function can become something like this:
```python
_, extension = os.path.splitext(self.filename)
return extension in extensions
```

agateau-gg (changes requested):

ggshield/scan/scannable.py
zhangmatthieu: Yes, I was doing something like that at first but saw what was already done in one of the internal projects. No preference on my side, I will do this since it introduces less dependencies

agateau-gg (approved):

PR #234, opened 2022-05-13
Various fixes related to ggshield auth and config commands

martinfaucheux:
* handle ConnectionError (unresolved host) when checking if ggshield flow is enabled. e.g. `gghsield auth login --method web --instance https://fake-host.com"
* recognize gitguardian.tech domain as gitguardian own domain (to allow logging in with gghsield on https://dashboard.staging.gitguardian.tech)
* fix `auth config` command documentation
* remove unused code

### DIFF
--- a/ggshield/cmd/auth/utils.py
+++ b/ggshield/cmd/auth/utils.py
@@ -1,4 +1,5 @@
 import click
+from requests import ConnectionError
 
 from ggshield.core.client import create_session
 from ggshield.core.config import Config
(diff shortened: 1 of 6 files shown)


ggshield/cmd/auth/utils.py
```diff
@@ -21,7 +22,11 @@ def check_instance_has_enabled_flow(config: Config) -> None:
 
     # don't use gitguardian client because the request should not include the token
     session = create_session(config.allow_self_signed)
-    response = session.get(urljoin(config.api_url, "/v1/metadata"), timeout=30)
+
+    try:
+        response = session.get(urljoin(config.api_url, "/v1/metadata"), timeout=30)
+    except ConnectionError:
+        raise click.ClickException("Invalid GitGuardian instance url.")
```
agateau-gg, line 29: This looks good, but I wonder if it should not be caught upper: are we sure this call is always going to be the first API call?

martinfaucheux: for the login command yes, but you're right, I should handle logout as well

martinfaucheux: In fact it's already taken care of for logout:
```python
    try:
        response = client.post(endpoint="token/revoke")
    except ConnectionError:
        raise LogoutError(REVOKE_FAIL_MESSAGE)
```

agateau-gg (approved):

PR #239, opened 2022-05-18
feat: add method to create a tar stream from Files

zhangmatthieu:
- Add a method which transforms a Files object in a BytesIO stream of a tar.gz of the files it represents

### Testing
- Added unit tests. Testing both for relative and absolute paths

### DIFF
--- a/ggshield/core/constants.py
+++ b/ggshield/core/constants.py
@@ -4,7 +4,7 @@ import os
 # max file size to accept
 MAX_FILE_SIZE = 1048576
 # max files size to create a tar from
-MAX_TAR_CONTENT_SIZE = 30 * 1024 * 1024  # 30MB
+MAX_TAR_CONTENT_SIZE = 30 * 1024 * 1024
 
 CPU_COUNT = os.cpu_count() or 1
 
--- a/ggshield/core/constants.py
+++ b/ggshield/core/constants.py
@@ -3,6 +3,8 @@ import os
 
 # max file size to accept
 MAX_FILE_SIZE = 1048576
+# max files size to create a tar from
+MAX_TAR_CONTENT_SIZE = 30 * 1024 * 1024  # 30MB
 
 CPU_COUNT = os.cpu_count() or 1
 
--- a/ggshield/scan/scannable.py
+++ b/ggshield/scan/scannable.py
@@ -1,5 +1,7 @@
 import concurrent.futures
 import re
+import tarfile
+from io import BytesIO
 from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, List, NamedTuple, Optional, Set
 
@@ -135,6 +137,13 @@ class Files:
 
     def apply_filter(self, filter_func: Callable[[File], bool]) -> "Files":
         return Files([file for file in self.files.values() if filter_func(file)])
+
+    def get_tar_stream(self) -> BytesIO:
+        tar_stream = BytesIO()
+        with tarfile.open(fileobj=tar_stream, mode="w:gz") as tar:
+            for filename in self.files:
+                tar.add(filename)
+        return tar_stream
 
     def scan(
         self,
--- a/ggshield/scan/scannable.py
+++ b/ggshield/scan/scannable.py
@@ -1,4 +1,5 @@
 import concurrent.futures
+import os
 import re
 import tarfile
 from io import BytesIO
@@ -12,7 +13,7 @@ from pygitguardian.models import ScanResult
 
 from ggshield import __version__
 from ggshield.core.cache import Cache
-from ggshield.core.constants import CPU_COUNT, MAX_FILE_SIZE
+from ggshield.core.constants import CPU_COUNT, MAX_FILE_SIZE, MAX_TAR_CONTENT_SIZE
 from ggshield.core.filter import (
     is_filepath_excluded,
     remove_ignored_from_result,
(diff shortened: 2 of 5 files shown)


ggshield/scan/scannable.py
```diff
@@ -127,6 +129,13 @@ def extra_headers(self) -> Dict[str, str]:
             **extra_headers,
         }
 
+    def get_tar_stream(self) -> BytesIO:
+        tar_stream = BytesIO()
```
agateau-gg, line 133: This is going to create the whole archive in memory, are we sure this won't be a problem? An alternative would be to pass an opened file object in which the method would write. This way it's up to the caller to decide if the archive should be created in memory or not.

tests/scan/test_scannable.py
```diff
@@ -179,3 +183,53 @@ def test_patch_max_size():
     files = list(c.get_files())
 
     assert len(files) == 0
+
+
+@pytest.mark.parametrize("absolute_path", (True, False))
```
agateau-gg, line 188: Instead of handling absolute paths, I think it would be simpler and safer to always store paths relative to the base scan dir. Given this file hierarchy:

```
/foo
- /bar
  - /baz
    - file1.txt
  - file2.py
```

Then if a command is run on /foo/bar, the archive contains:

- baz/file1.txt
- file2.py

tests/scan/test_scannable.py
```diff
@@ -179,3 +183,53 @@ def test_patch_max_size():
...
+        else "test_get_tar_stream.tar.gz"
+    )
+
+    # Create files
+    try:
+        with open(file1_path, "w") as file1:
+            file1.write(file1_content)
+        os.makedirs(dir_path, exist_ok=True)
+        with open(file2_path, "w") as file2:
+            file2.write(file2_content)
```
agateau-gg, line 212: nit: can be simplified to:

```suggestion
        file1.write_text(file1_content)
        dir_path.mkdir(parents=True, exist_ok=True)
        file2.write_text(file2_content)
```

tests/scan/test_scannable.py
```diff
@@ -179,3 +183,53 @@ def test_patch_max_size():
...
+        # Create tar archive from BytesIO stream
+        with open(tar_path, "wb") as tmp_file:
+            tmp_file.write(tar_stream.getvalue())
+        with tarfile.open(tar_path, "r:gz") as tar:
+            tar.extractall()
+
+        with open(file1_path) as extracted_file1:
+            assert file1_content == extracted_file1.read()
+        with open(file2_path) as extracted_file2:
+            assert file2_content == extracted_file2.read()
```
agateau-gg, line 228: Similarly here:
```suggestion
        assert file1_content == file1_path.read_text()
        assert file2_content == file2_path.read_text()
```

agateau-gg (changes requested):
Some questions about the new API.

Also, I think this PR should target the `iac` branch.

tests/scan/test_scannable.py
zhangmatthieu: Yes, we discussed that with Henri. I will make a follow-up issue to remove the unnecessary parent directories at some point before sending the archive (either when building the archive or even before)

ggshield/scan/scannable.py
zhangmatthieu: Discussed with Henri, one of the reasonings behind putting it in memory is also that it safeguards us as we will put it in memory on our side. One thing we will add in a next issue though is a mechanism to prevent creating archives which are too big

ggshield/core/constants.py
```diff
@@ -3,6 +3,8 @@
 
 # max file size to accept
 MAX_FILE_SIZE = 1048576
+# max files size to create a tar from
+MAX_TOTAL_FILES_SIZE = 31457280  # 30MB
```
agateau-gg, line 7: Naming: `MAX_TAR_CONTENT_SIZE` would be more self-explanatory.
Also: instead of writing ` = 31457280  # 30MB` you can write ` = 30 * 1024 * 1024`

ggshield/scan/scannable.py
agateau-gg: OK. If we really want to generate the archive in memory, then there is no reason to return a BytesIO: we might as well rename this method `create_tar()` and make it return `bytes`.

ggshield/scan/scannable.py
```diff
@@ -140,8 +141,14 @@ def apply_filter(self, filter_func: Callable[[File], bool]) -> "Files":
 
     def get_tar_stream(self) -> BytesIO:
         tar_stream = BytesIO()
+        current_dir_size = 0
         with tarfile.open(fileobj=tar_stream, mode="w:gz") as tar:
             for filename in self.files:
+                current_dir_size += os.path.getsize(filename)
+                if current_dir_size > MAX_TOTAL_FILES_SIZE:
+                    raise click.ClickException(
+                        "The total size of the files processed exceeds 30MB, please try again with less files"
```
agateau-gg, line 150: The message should reuse the constant (but divided by 1024 * 1024) to ensure it does not get out of sync.

tests/scan/test_scannable.py
```diff
@@ -231,6 +233,24 @@ def test_get_tar_stream(tmp_path, absolute_path):
             os.rmdir(dir_path)
 
 
+def test_get_tar_stream_exceeds_max_total_files_size(tmp_path):
```
agateau-gg, line 236: nit: `exceeds` → `cannot_exceed` :)

agateau-gg (changes requested):

zhangmatthieu:
Addressed the comments with a new revision

tests/scan/test_scannable.py
zhangmatthieu: I changed the signature to use a root_path for the base scan dir and expect the files to have a relative path to this root_path. Also moved it out of the Files class since it now needs a tuple of data Path/Files which need to be consistent

gg-mmill (approved):
LGTM

tests/scan/test_scannable.py
```diff
@@ -181,6 +187,71 @@ def test_patch_max_size():
...
+        dir_path.mkdir(parents=True, exist_ok=True)
+        file2_path.write_text(file2_content)
+
+        file1 = File(file1_content, str(file1_path))
+        file2 = File(file2_content, str(file2_path))
+        files = Files([file1, file2])
+        tar_stream = files.get_tar_stream()
+
+        # Create tar archive from BytesIO stream
+        with open(tar_path, "wb") as tmp_file:
```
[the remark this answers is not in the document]
zhangmatthieu: Updated the test

agateau-gg (approved):

PR #240, opened 2022-05-18
Switch to Pydantic for configuration classes

agateau-gg:
## Description

This PR adds a dependency on [Pydantic][] and makes all configuration classes use it.

[Pydantic]: https://github.com/samuelcolvin/pydantic

## What has been done

- Some preliminary work: splitting test files for configuration classes.
- Removed YAMLFileConfig, it is replaced with ConfigBaseModel.
- Made all configuration classes inherit from ConfigBaseModel and adjust things to make tests happy.

Note that this PR targets the iac branch, to avoid disrupting the release.


ggshield/core/config/auth_config.py
```diff
@@ -71,33 +42,71 @@ def expired(self) -> bool:
         )
 
 
-@dataclass
-class AuthConfig(YAMLFileConfig):
+def prepare_auth_config_dict_for_parse(data: Dict[str, Any]) -> Dict[str, Any]:
```
zhangmatthieu, line 45: I'm lacking a bit of context here but any reason why we support multiple accounts in the yaml file and only map it to a single account ? 
Would it make more sense to only support a single account in the YAML or multiple accounts in the InstanceConfig but only use the first one for now ?

zhangmatthieu, line 45: I'm lacking a bit of context here but any reason why we support multiple accounts in the yaml file and only map it to a single account ? 
Would it make more sense to only support a single account in the YAML or multiple accounts in the InstanceConfig but only use the first one for now ?

ggshield/core/config/user_config.py
```diff
@@ -62,26 +57,44 @@ def load(cls, config_path: Optional[str] = None) -> Tuple["UserConfig", str]:
...
+            user_config._update_from_file(config_path)
             return user_config, config_path
 
         for global_config_filename in GLOBAL_CONFIG_FILENAMES:
             global_config_path = get_global_path(global_config_filename)
-            global_data = load_yaml(global_config_path)
-            if global_data and user_config.update_config(global_data):
+            if os.path.exists(global_config_path):
+                user_config._update_from_file(global_config_path)
                 break
```
zhangmatthieu, line 67: nit: not related to this PR but we could reorder the local and the global and return `user_config, config_path` instead of breaking, that would prevent having to run the global part if a local config is set

zhangmatthieu, line 67: nit: not related to this PR but we could reorder the local and the global and return `user_config, config_path` instead of breaking, that would prevent having to run the global part if a local config is set

ggshield/core/config/utils.py
```diff
@@ -80,61 +65,42 @@ def get_global_path(filename: str) -> str:
...
-        field_names = {field_.name for field_ in fields(self)}
-        for key, item in data.items():
-            if key not in field_names:
-                click.echo("Unrecognized key in config: {}".format(key))
-                continue
-            if isinstance(getattr(self, key), list):
-                getattr(self, key).extend(item)
-            elif isinstance(getattr(self, key), set):
-                getattr(self, key).update(item)
+        assert self.__class__ is other.__class__
```
zhangmatthieu, line 77: For now we don't have subconfigs but could it make sense to check if the `other` class is a parent of the `self` class?

zhangmatthieu, line 77: For now we don't have subconfigs but could it make sense to check if the `other` class is a parent of the `self` class?

ggshield/core/config/user_config.py
zhangmatthieu: Ha, I didn't realize the global config was still taken but only overriden for the fields which are defined in the local config. Disregard this comment

zhangmatthieu: Ha, I didn't realize the global config was still taken but only overriden for the fields which are defined in the local config. Disregard this comment

ggshield/core/config/auth_config.py
agateau-gg: I agree with you, but it's been decided to add this in case it's need later.

ggshield/core/config/utils.py
agateau-gg: Good idea, going to do this.

agateau-gg:
I am closing this PR: it's replaced by #245.

PR #241, opened 2022-05-19
Save only auth config for auth interactions

martinfaucheux:
when logging in (disregarding the method) or saving new auth related data, only save the AuthConfig instance rather than the whole Config.

* this will prevent the project's `.gitguardian` file from being modified when developing in local
* this is cleaner


agateau-gg (approved):
Looks good. I wonder if we should completely remove `Config.save()`. I don't think it's used anymore and I don't see a use-case for it. What do you think?

PR #242, opened 2022-05-19
Fix SSO wrongly not matching with instance url

martinfaucheux:
* Fix SSO wrongly not matching witn instance url (it happened if the url of the instance was not the default one)
* update logout error message if no network

### DIFF
--- a/ggshield/cmd/auth/login.py
+++ b/ggshield/cmd/auth/login.py
@@ -11,7 +11,7 @@ from ggshield.core.utils import clean_url
 
 
 def validate_login_path(
-    config: Config, instance: Optional[str], sso_url: Optional[str]
+    instance: Optional[str], sso_url: Optional[str]
 ) -> Tuple[Optional[str], Optional[str]]:
     """
     Validate that the SSO URL and the instance refer to the same instance if they are both defined,
@@ -39,7 +39,7 @@ def validate_login_path(
     if instance is None:
         return sso_instance, sso_login_path
 
-    config_parsed_url = clean_url(config.dashboard_url)
+    config_parsed_url = clean_url(instance)
     if (
         config_parsed_url.scheme != sso_parsed_url.scheme
         or config_parsed_url.netloc != sso_parsed_url.netloc
@@ -66,7 +66,7 @@ def validate_login_path(
     "--sso-url",
     required=False,
     type=str,
-    help="Instance URL to authenticate with on a pre-selected SAML SSO page.",
+    help="URL of your SSO login page to force the authentication flow through your workspace SSO.",
 )
 @click.option(
     "--token-name",
@@ -147,9 +147,7 @@ def login_cmd(
         return 0
 
     if method == "web":
-        instance, login_path = validate_login_path(
-            config, instance=instance, sso_url=sso_url
-        )
+        instance, login_path = validate_login_path(instance=instance, sso_url=sso_url)
         if instance:
             config.set_cmdline_instance_name(instance)
         defined_instance = config.instance_name
--- a/ggshield/cmd/auth/logout.py
+++ b/ggshield/cmd/auth/logout.py
@@ -5,12 +5,7 @@ from ggshield.core.client import create_client
 from ggshield.core.config import Config
 from ggshield.core.utils import dashboard_to_api_url
 
-
-REVOKE_FAIL_MESSAGE = (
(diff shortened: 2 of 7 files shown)


ggshield/cmd/auth/logout.py
```diff
@@ -87,10 +80,16 @@ def revoke_token(config: Config, instance_url: str) -> None:
     try:
         response = client.post(endpoint="token/revoke")
     except ConnectionError:
-        raise LogoutError(REVOKE_FAIL_MESSAGE)
+        raise click.ClickException(
+            "Could not connect to GitGuardian.\nCheck your internet connection."
```
elacaille18, line 84: @martinfaucheux do we agree that this error will also be displayed when trying to run the `login` command without network as well (not only the `logout` command)?

ggshield/cmd/auth/login.py
```diff
@@ -66,7 +66,7 @@ def validate_login_path(
     "--sso-url",
     required=False,
     type=str,
-    help="Instance URL to authenticate with on a pre-selected SAML SSO page.",
+    help="URL of your SSO login page to force the authentication flow through your workspace SSO.",
```
elacaille18, line 69: ✅

elacaille18 (commented):
@martinfaucheux I allowed myself to review a part of the MR.

I have a question about the wording connection error => it should be applied to all commands requiring internet access. Is it the case?

Aurélien is OOO today so you might want to pick another tech reviewer

ggshield/cmd/auth/logout.py
martinfaucheux: In fact I just realized this error can be due to 2 things
* invalid url
* no internet

How can I combine both messages? I propose
"Could not connect to GitGuardian.\nCheck the URL and your internet connection."

I'll apply the same thing for the login error.

elacaille18: invalid url looks too broad no? 

eg: you want a specific error for an invalid SSO url for example no?

martinfaucheux: I can't know if the problem is because of internet connection or invalid url. I want an message that cover both case

elacaille18: @martinfaucheux then "Could not connect to GitGuardian.\nPlease check your internet connection or if the specified URL is correct."

ggshield/cmd/auth/login.py
```diff
@@ -39,7 +39,7 @@ def validate_login_path(
     if instance is None:
         return sso_instance, sso_login_path
 
-    config_parsed_url = clean_url(config.dashboard_url)
+    config_parsed_url = clean_url(instance)
```
julienc91, line 42: The `config` parameter becomes unused, should we remove it entirely from the function's signature?

ggshield/cmd/config/config_list.py
```diff
@@ -32,7 +33,7 @@ def config_list_cmd(ctx: click.Context) -> int:
 
         message_lines = [
             f"[{instance_name}]",
-            f"default_token_lifetime: {instance.default_token_lifetime}",
+            f"default_token_lifetime: {instance.default_token_lifetime or default_token_lifetime}",
```
julienc91, line 36: Not sure about this: isn't `0` a valid value that is considered as "no expiration"? In that case we would need:

```
            f"default_token_lifetime: {instance.default_token_lifetime if instance.default_token_lifetime is not None else default_token_lifetime}",
```

ggshield/cmd/auth/utils.py
```diff
@@ -8,6 +8,10 @@
 
 VERSION_TOO_LOW_MESSAGE = "Your GitGuardian dashboard version is too low, upgrade to be able to use this command."
 DISABLED_FLOW_MESSAGE = "Your GitGuardian dashboard does not authorize this command, contact your administrator."
+CONNECTION_ERROR_MESSAGE = (
+    "Could not connect to GitGuardian.\n"
+    "Please check your internet connection or if the specified URL is correct."
```
julienc91, line 13: If I were to nit pick on logical conditions, I'd say you want to use a "and" connector instead of "or", since you want the user to check both conditions and not only one of them :angel:

elacaille18: correct. My bad for the imperfection of my english ;)

julienc91 (approved):

PR #243, opened 2022-05-20
feat(iac): add new classes to handle response from the client endpoint

zhangmatthieu:
In order to be able to call the API and handle the results, we:
- inherit the GGClient, adding a method to call the iac/scan endpoint
- add classes for an IaC match, IaC policy break, IaC scan result which inherits the current classes, adding some fields specific to IaC

### Testing
- Added a test to check deserialization of the introduced classes
- Currently missing a test for the client, will be added when we define the API response and get cassettes

### DIFF
--- a/ggshield/iac/models/iac_match.py
+++ b/ggshield/iac/models/iac_match.py
@@ -1,0 +1,35 @@
+from typing import Any, Dict, Optional
+
+from marshmallow import fields, post_load
+from pygitguardian.models import Match, MatchSchema
+
+
+class IaCMatchSchema(MatchSchema):
+    filename = fields.String(required=True)
+
+    @post_load
+    def make_match(self, data: Dict[str, Any], **kwargs: Any) -> "IaCMatch":
+        return IaCMatch(**data)
+
+
+class IaCMatch(Match):
+    SCHEMA = IaCMatchSchema()
+
+    def __init__(
+        self,
+        filename: str,
+        match: str,
+        match_type: str,
+        line_start: Optional[int] = None,
+        line_end: Optional[int] = None,
+        index_start: Optional[int] = None,
+        index_end: Optional[int] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(
+            match, match_type, line_start, line_end, index_start, index_end, **kwargs
+        )
+        self.filename = filename
+
+    def __repr__(self) -> str:
+        return f"filename:{self.filename}, {super().__repr__()}"
--- a/ggshield/iac/models/iac_policy_break.py
+++ b/ggshield/iac/models/iac_policy_break.py
@@ -1,0 +1,41 @@
+from typing import Any, Dict, List
+
+from marshmallow import fields, post_load
+from pygitguardian.models import PolicyBreak, PolicyBreakSchema
+
+from ggshield.iac.models.iac_match import IaCMatch, IaCMatchSchema
+
+
+class IaCPolicyBreakSchema(PolicyBreakSchema):
+    matches = fields.List(fields.Nested(IaCMatchSchema), required=True)
+    policy_details = fields.Dict(
+        fields.String(required=True), fields.String(required=True)
+    )
+    ignore_sha = fields.String(required=True)
+
+    @post_load
+    def make_policy_break(
+        self, data: Dict[str, Any], **kwargs: Any
+    ) -> "IaCPolicyBreak":
+        return IaCPolicyBreak(**data)
+
+
+class IaCPolicyBreak(PolicyBreak):
+    SCHEMA = IaCPolicyBreakSchema()
+
+    def __init__(
+        self,
+        break_type: str,
+        policy: str,
+        validity: str,
+        matches: List[IaCMatch],
+        policy_details: Dict[str, str],
+        ignore_sha: str,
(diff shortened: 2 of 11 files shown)


gg-mmill (commented):
We will also need a model for the scan parameters, and pass them with the request.

ggshield/iac/models/iac_policy_break.py
```diff
@@ -0,0 +1,42 @@
...
+    def make_iac_policy_break(
+        self, data: Dict[str, Any], **kwargs: Any
+    ) -> "IaCPolicyBreak":
+        return IaCPolicyBreak(**data)
+
+
+class IaCPolicyBreak(PolicyBreak):
+    SCHEMA = IaCPolicyBreakSchema()
+
+    def __init__(
```
gg-mmill, line 29: We will need the `sha` of the policy break as well.

ggshield/iac/models/iac_scan_result.py
```diff
@@ -0,0 +1,24 @@
...
+
+from ggshield.iac.models.iac_policy_break import IaCPolicyBreakSchema
+
+
+class IaCScanResultSchema(ScanResultSchema):
+    policy_breaks = fields.List(fields.Nested(IaCPolicyBreakSchema), required=True)
+
+    # Override parent method to prevent using the parent deserializer
+    def make_scan_result(self, data: Dict[str, Any], **kwargs: Any) -> Dict[str, Any]:
+        return data
```
gg-mmill, line 14: We should raise if the method is not intended to be used.

tests/iac/test_models.py
```diff
@@ -0,0 +1,71 @@
...
+                    ],
+                },
+            ),
+        ],
+    )
+    def test_schema_loads(self, schema_klass, expected_klass, instance_data):
+        """
+        GIVEN the right kwargs  and an extra field in dict format
+        WHEN loading using the schema
+        THEN the extra field should be excluded
```
gg-mmill, line 63: Not very clear, do you mean here "the extra field is not taken into account" ? Otherwise I don't see a test that the `extra` field is excluded from the models ?

ggshield/iac/models/iac_scan_result.py
zhangmatthieu: Ok, I will check if it's used by the preload of the parent but it shouldn't

tests/iac/test_models.py
zhangmatthieu: This test is the same as the one in pygitguardian. Basically, it doesn't add the extra field to the instance when deserializing the object

ggshield/iac/models/iac_policy_break.py
zhangmatthieu: Should we have it as such or can it be added in the policy_details?

gg-mmill: Well the sha would be computed both from policy and matches, so it does not really belong in policy_details IMO.

zhangmatthieu: Ha, yes, didn't understand you were talking about the identifier. I called it policy_break_id for now, we can change the name after your work on determining a vulnerability ID if there is a better match

agateau-gg (changes requested):

ggshield/iac/models/iac_match.py
```diff
@@ -0,0 +1,38 @@
...
+
+from marshmallow import fields, post_load
+from pygitguardian.models import Match, MatchSchema
+
+
+class IaCMatchSchema(MatchSchema):
+    filename = fields.String(required=True)
+
+    def make_match(self, data: Dict[str, Any], **kwargs: Any) -> Dict[str, Any]:
+        raise NotImplementedError("Should not use the Match deserializer")
```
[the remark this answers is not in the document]
zhangmatthieu: Yes, good point

ggshield/iac/models/iac_scan_parameters.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any, Dict, List, Optional
+
+from marshmallow import fields, post_load
+from pygitguardian.models import Base, BaseSchema
+
+
+class IaCScanParametersSchema(BaseSchema):
+    excluded_policies = fields.List(fields.String())
```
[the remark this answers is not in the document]
zhangmatthieu: Will rename

ggshield/iac/client.py
```diff
@@ -0,0 +1,37 @@
...
+
+from pygitguardian import GGClient
+from pygitguardian.client import is_ok, load_detail
+from pygitguardian.models import Detail
+
+from ggshield.iac.models.iac_scan_parameters import IaCScanParameters
+from ggshield.iac.models.iac_scan_result import IaCScanResult
+
+
+class IaCGGClient(GGClient):
```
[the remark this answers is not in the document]
zhangmatthieu: Yes, the goal was to still instantiate it with the create_client functions. I will move the class for consistency

zhangmatthieu:
Inheriting the match and policy break classes introduce complexity in the data handling which doesn't simplify/help where those are used (for output handlers for example). We will create specific classes for the IaC to match a simpler API contract

PR #244, opened 2022-05-20
Add invalid saml url error handling

martinfaucheux:
Handle login error code when the given SAML URL is invalid


martinfaucheux:
Included in https://github.com/GitGuardian/ggshield/pull/242

PR #245, opened 2022-05-24
Use Marshmallow to load/save config files

agateau-gg:
## Description

This PR is the Marshmallow version of #240. I switched to Marshmallow after realizing ggshield already indirectly depends on it, via pygitguardian.

## What has been done

- Some preliminary work: splitting test files for configuration classes.
- Add Marshmallow schema classes for our configuration classes.

Note that this PR targets the iac branch, to avoid disrupting the release.


ggshield/core/config/auth_config.py
```diff
@@ -137,3 +152,6 @@ def get_instance_token(self, instance_name: str) -> str:
         if instance.account is None:
             raise MissingTokenError(instance=instance_name)
         return instance.account.token
+
+
+AuthConfigSchema = marshmallow_dataclass.class_schema(AuthConfig)
```
zhangmatthieu, line 157: I think we could use the decorators `@dataclass` https://pypi.org/project/marshmallow-dataclass/ to simplify the code

zhangmatthieu: Ha, we already have a dataclass decorator, not sure how those two would interact with each other

ggshield/core/config/user_config.py
```diff
@@ -32,23 +41,12 @@ class UserConfig(YAMLFileConfig):
...
-        try:
-            api_url = data.pop("api_url")
-        except KeyError:
-            pass
-        else:
-            if "instance" not in data:
-                data["instance"] = api_to_dashboard_url(api_url, warn=True)
-        return super(UserConfig, self).update_config(data)
-
     def save(self, config_path: str) -> None:
```
zhangmatthieu, line 44: For my own understanding, are we writing on config files when calling save on the config?

agateau-gg: Yes, that's what is happening.

ggshield/core/config/utils.py
```diff
@@ -80,60 +65,37 @@ def get_global_path(filename: str) -> str:
...
-        p.parent.mkdir(parents=True, exist_ok=True)
-        with p.open("w") as f:
-            try:
-                stream = yaml.dump(data, indent=2, default_flow_style=False)
-                f.write(stream)
-            except Exception as e:
-                raise click.ClickException(
-                    f"Error while saving config in {path}:\n{str(e)}"
-                ) from e
+def update_from_other_instance(dst: Any, src: Any) -> None:
```
zhangmatthieu, line 68: nit: we don't have a test for this method, should we add one ?

zhangmatthieu (approved):

PR #247, opened 2022-05-25
Introduce new config file format

agateau-gg:
## Description

This PR introduces a new format for ggshield configuration file, opening the way to new tools.

## What has been done

- Add support for versionized file formats
- Move secret-specific configuration keys to a SecretConfig class

## TODO

- [x] Document the new format in the README
- [x] Finish some remaining bits of configuration changes


README.md
```diff
@@ -411,45 +412,102 @@ You can also use the option `--config-path` on the main command to set another c
...
-show-secrets: false # default: false
+banlisted-detectors:
+  - Generic Password
+```
 
-# Set to true if the desired exit code for the CLI is always 0,
-# otherwise the exit code will be 1 if incidents are found.
-# the environment variable GITGUARDIAN_EXIT_ZERO=true can also be used toggle this behaviour.
-exit-zero: false # default: false
+And here is the equivalent v2 file:
```
zhangmatthieu, line 490: Since we already have a way to serialize/deserialize configs, did we think about having an automatic generation of v2 config files from v1 config files?

tests/cmd/scan/test_path.py
```diff
@@ -42,6 +41,7 @@ def test_scan_file(self, cli_fs_runner, verbose):
             result = cli_fs_runner.invoke(cli, ["-v", "secret", "scan", "path", "file"])
         else:
             result = cli_fs_runner.invoke(cli, ["secret", "scan", "path", "file"])
+        assert result.exit_code == 0, result.output
```
zhangmatthieu, line 44: nit: since we use this assertion a lot, we might want to extract a method `assert_successful_scan()` and a method `assert_failed_scan(exception)`
```
def assert_successful_scan():
    assert result.exit_code == 0, result.output
    assert not result.exception
    ...
```

tests/core/config/test_user_config.py
```diff
@@ -81,21 +86,100 @@ def test_accumulation_matches(
...
+                  match: dbca
+            """,
+        )
+        config, _ = UserConfig.load(config_path)
+
+        assert config.secret.ignored_matches == [
+            {"name": "", "match": "abcd"},
+            {"name": "", "match": "dbca"},
+        ]
+
```
zhangmatthieu, line 172: We could add a test comparing a v1 and a v2 config making sure we have the correspondance between the first one and the second one

ggshield/core/config/errors.py
```diff
@@ -47,3 +50,25 @@ def __init__(self, instance: str):
...
+
+
+def format_validation_error(exc: ValidationError) -> str:
+    """
+    Take a Marshmallow ValidationError and turn it into a more user-friendly message
+    """
+    message_dct = exc.normalized_messages()
+    lines = []
+
+    def format_items(dct: Dict[str, Any], indent: int) -> None:
```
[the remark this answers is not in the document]
agateau-gg: I haven't tried, but I suspect the output would be less user-friendly, since it would add curly braces, commas and quotes everywhere.

tests/core/config/test_user_config.py
agateau-gg: There is already `test_load_v1`. Do you think we need more?

tests/cmd/scan/test_path.py
agateau-gg: Here it is: https://github.com/GitGuardian/ggshield/issues/253

zhangmatthieu (approved):

PR #248, opened 2022-05-25
feat(iac): handle config file for iac

zhangmatthieu:
- Allow setting iac settings by adding iac parameters to the user config file as described in the config file format v2 doc

### Testing
- Added unit tests for the loading of the config, global or local

### DIFF
--- a/ggshield/core/config/user_config.py
+++ b/ggshield/core/config/user_config.py
@@ -23,12 +23,28 @@ from ggshield.core.utils import api_to_dashboard_url
 
 
 @dataclass
+class IaCConfig:
+    """
+    Holds the iac config as defined .gitguardian.yaml files
+    (local and global).
+    """
+
+    ignored_paths: Set[str] = field(default_factory=set)
+    ignored_policies: Set[str] = field(default_factory=set)
+    minimum_severity: str = "LOW"
+
+
+IaCConfigSchema = marshmallow_dataclass.class_schema(IaCConfig)
+
+
+@dataclass
 class UserConfig:
     """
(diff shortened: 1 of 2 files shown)


ggshield/core/config/user_config.py
```diff
@@ -22,13 +22,29 @@
...
+@dataclass
+class IaCConfig:
+    """
+    Holds the iac config as defined .gitguardian.yaml files
+    (local and global).
+    """
+
+    ignored_paths: Set[str] = field(default_factory=set)
+    ignored_policies: Set[str] = field(default_factory=set)
+    minimum_severity: str = "LOW"
```
gg-mmill, line 34: I'm wondering if we should validate severity levels here:
 - they will be validated at API call anyway, but validating a config file is always a nice to have
 - this also raises the question of validating ignored policies, even though in this case we cannot fully validate the input without an API call. WDYT @GG-HH  ?

agateau-gg: Validating when reading the config file sounds like a good idea because it helps catching errors early, but you have to be sure the client can never be out of sync with the server (ie: if the server starts accepting a new value, the client should not reject it when reading it from the config file)

gg-mmill (approved):

ggshield/core/config/user_config.py
zhangmatthieu: Agreed, if we don't have a single source for the severities which is used both on the client side and on the server side, we need to make sure there are mechanisms to update every clients after a server update. Since releases are manual, it's hard to do in the current state. Even in this case, I'm not sure we would want to throw an error with non-existing severity or policy filters, maybe add a warning and ignore the filter

PR #250, opened 2022-05-31
feat(iac): handle API result with text and json output handlers

zhangmatthieu:
## feat(iac): handle API result with text and json output handlers

- Update iac scan command to call the API through the client, using the context passed as an argument (from local and global config files) and overriding it with cli arguments.
- Call the API to scan the path passed as a directory. Only support one path
- Update the filter method to remove the root path from the files name
- Introduce text and JSON output handlers for the iac scan, with the format as defined in the tech specs

### Testing
- Added unit tests and updated current unit tests
- Need to add some cassettes when the API is ready

## feat(iac): update classes to match api model

- Update classes to match {"id": "", type": "", "iac_engine_version": "","entities_with_incidents": [{"filename": "filename","incidents": [...]}]}

### DIFF
--- a/ggshield/cmd/iac/scan.py
+++ b/ggshield/cmd/iac/scan.py
@@ -1,9 +1,22 @@
 from pathlib import Path
-from typing import Any, Sequence
+from typing import Any, Optional, Sequence, Type
 
 import click
 
-from ggshield.iac.utils import POLICY_ID_PATTERN, validate_policy_id
+from ggshield import __version__
+from ggshield.core.client import create_client_from_config
+from ggshield.core.config import Config
+from ggshield.core.extra_headers import get_extra_headers
+from ggshield.core.filter import init_exclusion_regexes
+from ggshield.iac.filter import get_iac_files_from_paths
+from ggshield.iac.models import IaCScanResult
+from ggshield.iac.models.iac_scan_parameters import IaCScanParameters
+from ggshield.iac.utils import POLICY_ID_PATTERN, create_tar, validate_policy_id
+from ggshield.output import OutputHandler
+from ggshield.output.json.iac_json_output_handler import IaCJSONOutputHandler
+from ggshield.output.text.iac_text_output_handler import IaCTextOutputHandler
+from ggshield.scan import ScanCollection
+from ggshield.scan.scannable_errors import handle_scan_error
 
 
 def validate_exclude(_ctx: Any, _param: Any, value: Sequence[str]) -> None:
@@ -23,9 +36,10 @@ def validate_exclude(_ctx: Any, _param: Any, value: Sequence[str]) -> None:
     help="Always return 0 (non-error) status code.",
 )
 @click.option(
-    "--level",
+    "--minimum-severity",
+    "minimum_severity",
     type=click.Choice(("LOW", "MEDIUM", "HIGH", "CRITICAL")),
-    help="Level of the blocking alerts.",
+    help="Minimum severity of the policies",
     default="LOW",
 )
 @click.option("--verbose", "-v", is_flag=True, help="Verbose display mode.")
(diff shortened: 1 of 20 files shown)


ggshield/output/text/iac_text_output_handler.py
```diff
@@ -0,0 +1,69 @@
...
+                )
+
+        return scan_buf.getvalue()
+
+    def process_iac_file_result(
+        self, file_path: Path, file_result: IaCFileResult
+    ) -> str:
+        """
+        Build readable message on the found incidents.
+
```
gg-mmill, line 45: Can you add a description for  the `file_path` parameter as well ?

ggshield/output/text/iac_text_output_handler.py
```diff
@@ -0,0 +1,69 @@
...
+    def _process_scan_impl(self, scan: ScanCollection) -> str:
+        scan_buf = StringIO()
+        if scan.optional_header and (scan.iac_result or self.verbose):
+            scan_buf.write(scan.optional_header)
+
+        if scan.iac_result:
+            scan_buf.write(iac_engine_version(scan.iac_result.iac_engine_version))
+
+        if scan.iac_result:
```
gg-mmill, line 30:
```suggestion
```

ggshield/output/text/iac_text_output_handler.py
```diff
@@ -0,0 +1,69 @@
...
+
+        :param file_result: The file results from the IaC scanning API
+        :return: The formatted message to display
+        """
+        result_buf = StringIO()
+
+        result_buf.write(file_info(file_result.filename, len(file_result.incidents)))
+        with open(file_path, "rb") as buffered_file:
+            content = buffered_file.read()
+        file = File.from_bytes(content, str(file_path))
```
gg-mmill, line 54:
```suggestion
        file = File.from_bytes(file_path.read_bytes(), str(file_path))
```

ggshield/cmd/iac/scan.py
```diff
@@ -46,15 +61,98 @@ def validate_exclude(_ctx: Any, _param: Any, value: Sequence[str]) -> None:
...
+
+    ignored_paths = config.user_config.iac.ignored_paths
+    if ignore_paths is not None:
+        ignored_paths.update(ignore_paths)
+
+    ctx.obj["exclusion_regexes"] = init_exclusion_regexes(ignored_paths)
+
+    ignored_policies = config.user_config.iac.ignored_policies
+    if ignore_policies is not None:
+        ignored_policies.update(ignore_policies)
```
gg-mmill, line 107:
```suggestion
    if ignore_policies is not None:
        config.user_config.iac.ignored_policies.update(ignore_policies)
```

Would make it easier to understand what is going on (`ignore_policies` and `ignored_policies` are difficult to distinguish). Same for `ignored_paths`.

ggshield/output/text/text_output_handler.py
```diff
@@ -18,20 +18,7 @@
...
-    return max(len(str(lines[-1].pre_index or 0)), len(str(lines[-1].post_index or 0)))
-
-
-def get_offset(padding: int, is_patch: bool = False) -> int:
-    """Return the offset due to the line display."""
-    if is_patch:
-        return len(LINE_DISPLAY["patch"].format("0" * padding, "0" * padding))
-
-    return len(LINE_DISPLAY["file"].format("0" * padding))
+from .utils import get_offset, get_padding
```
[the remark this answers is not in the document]
zhangmatthieu: I'm using them in a different file. They are not specific to this specific file but still specific to text outputs so I also wouldn't want to move them to `core/text_utils.py`

ggshield/scan/scannable.py
```diff
@@ -68,6 +70,9 @@ def __init__(self, document: str, filename: str):
         self.filename = filename
         self.filemode = Filemode.FILE
 
+    def remove_root_path(self, root_path: Path) -> "File":
```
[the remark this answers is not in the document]
zhangmatthieu: will rename

ggshield/output/text/iac_text_output_handler.py
zhangmatthieu: Sure

GG-HH (approved):

PR #258, opened 2022-06-06
refactor: turn IgnoredMatch into a dataclass

Anonymous-fox:
- make Cache, is_ignored() and SecretConfig compatible
- make tests compatible

### DIFF
--- a/ggshield/core/cache.py
+++ b/ggshield/core/cache.py
@@ -7,13 +7,14 @@ from pygitguardian.models import PolicyBreak
 
 from ggshield.core.constants import CACHE_FILENAME
 from ggshield.core.filter import get_ignore_sha
+from ggshield.core.types import IgnoredMatch, IgnoredMatchSchema
 
 
 SECRETS_CACHE_KEY = "last_found_secrets"
 
 
 class Cache:
-    last_found_secrets: List
+    last_found_secrets: List[IgnoredMatch]
 
     def __init__(self, cache_filename: str = CACHE_FILENAME) -> None:
         self.cache_filename = cache_filename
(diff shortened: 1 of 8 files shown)


ggshield/core/cache.py
```diff
@@ -44,13 +45,20 @@ def load_cache(self) -> bool:
 
     def update_cache(self, **kwargs: Any) -> None:
         if SECRETS_CACHE_KEY in kwargs:
-            self.last_found_secrets = kwargs.pop(SECRETS_CACHE_KEY)
+            self.last_found_secrets = [
+                IgnoredMatchSchema().load(data=secret)
```
agateau-gg, line 49: Here the code is going to create an `IgnoredMatchSchema` instance for each secret. It would be more efficient to do something like this:
```python
   schema = IgnoredMatchSchema()
   self.last_found_secrets = [schema.load(data=secret)
       for secret in kwargs.pop(SECRETS_CACHE_KEY)]
```

agateau-gg (changes requested):
Some minor optimization remark, but this looks great! 👍🏻

Anonymous-fox:
My code fails against four tests because it doesn't support v1 configuration file format. Should I change those tests?
`TestCache.test_save_cache`
`TestCache.test_read_only_fs`
`TestCache.test_save_cache_first_time[True]`
`TestUserConfig.test_load_v1`

agateau-gg:
> I could change those TestCache tests, so they call update_cache() with new v2 secrets format

This sounds good.

> but I don't know what to do about test_load_v1

It is important that our code continues to be able to load v1 configuration files, so the test must pass unchanged.

I think you need to change the type definition of `UserV1Config.matches_ignore` to accept both the old and the new formats. The code is already in place to convert matches to the new format, so it might be enough to replace `List[IgnoredMatch]` with `List[Dict[str, Optional[str]]` (the old definition of IgnoredMatch).

agateau-gg (approved):
Looks good, thanks!

PR #261, opened 2022-06-08
Small fixes for iac

zhangmatthieu:
fix(iac): fix ignore-policy click option

fix(iac): add default behavior in case files are not found to display for the text output handler, to prevent complete failure

fix(iac): remove default CLI input to minimum-severity to avoid overriding config files

### DIFF
--- a/ggshield/output/text/iac_text_output_handler.py
+++ b/ggshield/output/text/iac_text_output_handler.py
@@ -13,6 +13,7 @@ from .message import (
     iac_engine_version,
     iac_vulnerability_header,
     iac_vulnerability_location,
+    iac_vulnerability_location_failed,
     no_iac_vulnerabilities,
 )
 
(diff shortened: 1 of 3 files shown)


ggshield/output/text/iac_text_output_handler.py
```diff
@@ -51,19 +52,31 @@ def process_iac_file_result(
 
         result_buf.write(file_info(file_result.filename, len(file_result.incidents)))
 
-        file = File.from_bytes(file_path.read_bytes(), str(file_path))
-        lines: List[Line] = get_lines_from_content(file.document, Filemode.FILE, False)
+        try:
```
gg-mmill, line 55: I'm not sure this is really useful, something causing this kind of error (where a file was read and sent to the API, but could not be read afterwards) would probably have much worse consequences than having ggshield crash anyway.

gg-mmill (approved):

ggshield/output/text/iac_text_output_handler.py
jd4625988: It can happen if we have a change in the API in the file format for example and not only if the file does not exist. I would prefer not to fail unexpectedly when there are results from the API, especially since errors are not fully wrapped in ggshield.

zhangmatthieu: It can happen if we have a change in the API in the file format for example and not only if the file does not exist. I would prefer not to fail unexpectedly when there are results from the API, especially since errors are not fully wrapped in ggshield.

Issue #253, opened 2022-06-02
Add assert helpers to factorize cli_fs_runner calls in tests


agateau-gg:
## Description

Our test suite uses `cli_fs_runner.invoke()` a lot. A common idiom is to write the test like this:

```python
# Test success
result = cli_fs_runner.invoke(cli, ["secret", "scan", "path", "file"])
assert result.exit_code == 0, result.output

# Test failures
result = cli_fs_runner.invoke(cli, ["secret", "scan", "path", "file"])
assert result.exit_code == 1, result.output
```

The `result.output` argument of the `assert` is used to show what ggshield printed in case the exit code does not match what we expected.

The goal of this issue is to factorize these calls.

## Requirements

- The helpers must support testing success and failure
- In case of unexpected result, the helpers must print:
    - the actual exit code
    - the expected exit code
    - both stdout and stderr, to take into account the case where `cli_fs_runner.mix_stderr` has been set to false

## API proposal

Add these to tests/conftest.py:

```python
def assert_invoke_exited_with(result: Click.Result, exit_code: int):
    # ...

def assert_invoke_ok(result: Click.Result):
    assert_invoke_exited_with(result, 0)
```

Then the two examples can be rewritten to:

```python
# Test success
result = cli_fs_runner.invoke(cli, ["secret", "scan", "path", "file"])
assert_invoke_ok(result)

# Test failures
result = cli_fs_runner.invoke(cli, ["secret", "scan", "path", "file"])
assert_invoke_exited_with(result, 1)
```

PR #268, opened 2022-06-15
Add test method in conftest to factorize cli_fs_runner calls

amascia-gg:
Closes issue #253

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,7 +5,7 @@ from pathlib import Path
 
 import pytest
 import vcr
-from click.testing import CliRunner
+from click.testing import CliRunner, Result
 from pygitguardian import GGClient
 from pygitguardian.models import ScanResult
 from requests.utils import DEFAULT_CA_BUNDLE_PATH, extract_zipped_paths
@@ -502,3 +502,20 @@ def isolated_fs(fs):
     # add cassettes dir
     cassettes_dir = join(dirname(realpath(__file__)), "cassettes")
     fs.add_real_directory(cassettes_dir)
+
+
+def assert_invoke_exited_with(result: Result, exit_code: int):
+    msg = f"""
+    Expected code {exit_code}, got {result.exit_code}.
+
+    stdout:
+    {result.stdout}
+
+    stderr:
+    {result.stderr if result.stderr_bytes is not None else ""}
+    """
+    assert result.exit_code == exit_code, msg
+
+
+def assert_invoke_ok(result: Result):
+    assert_invoke_exited_with(result, 0)
--- a/tests/cmd/auth/test_login.py
+++ b/tests/cmd/auth/test_login.py
@@ -20,6 +20,7 @@ from ggshield.core.oauth import (
     get_error_param,
     get_pretty_date,
 )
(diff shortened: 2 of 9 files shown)


tests/conftest.py
```diff
@@ -502,3 +503,15 @@ def isolated_fs(fs):
...
     cassettes_dir = join(dirname(realpath(__file__)), "cassettes")
     fs.add_real_directory(cassettes_dir)
+
+
+def assert_invoke_exited_with(result: Result, exit_code: int):
+    msg = f"Expected code {exit_code}, got {result.exit_code}"
+    if result.exit_code != exit_code:
+        sys.stderr.write(msg)
+        sys.stdout.write(msg)
+    assert result.exit_code == exit_code, msg
```
agateau-gg, line 513: Oh, I misunderstood your screenshot. This is not what I had in mind.

Originally, tests running ggshield commands would do so like this:

```python
result = cli_fs_runner.invoke(cli, cmd)
assert result.exit_code == 0
```

The problem with this is that if the command fails, we don't know what it printed, because the stdout and the stderr of the command is in `result.output`. This makes the failure hard to debug.

To work around that, we started replacing the `assert` call with this:

```python
assert result.exit_code == 0, result.output
```

This way, if a command fails pytest prints what the failed command printed on stdout.

The goal of this new `assert_invoke_exited_with()` function is to systematize this idea of printing what the command printed in case of errors. To do so you must add to `msg` the content of `result.stdout` and `result.stderr` (for the cases where stderr is not merged into `result.stderr`)

agateau-gg (changes requested):

tests/conftest.py
amascia-gg: Thanks for @agateau-gg for the clarification. I pushed some changes that I believe correspond to what you were expecting, tell me if that's not the case!

tests/conftest.py
```diff
@@ -506,11 +505,11 @@ def isolated_fs(fs):
...
-    msg = f"Expected code {exit_code}, got {result.exit_code}"
-    if result.exit_code != exit_code:
-        sys.stderr.write(msg)
-        sys.stdout.write(msg)
-    assert result.exit_code == exit_code, msg
+    msg = f"Expected code {exit_code}, got {result.exit_code}."
+    result_msg = f"\n\n Details of the error: \n {result.stdout}" + (
+        "\n" + result.stderr if result.stderr_bytes is not None else ""
+    )
+    assert result.exit_code == exit_code, msg + result_msg
```
agateau-gg, line 512: Looks good! Maybe you can use a triple-quote string instead of concatenating msg and result_msg?

Also, it would be nice if the message had "stdout:" and "stderr:" prefixes. The message would look like this (this was not in the issue, but I thought about it after trying your branch):

```
Expected code {exit_code}, got {result.exit_code}.

stdout:
{result.stdout}

stderr:
{result.stderr}
```

agateau-gg (changes requested):

tests/conftest.py
amascia-gg: I've changed the format as you requested, thanks for the suggestion!
If every thing seems good with you, I'll squash the last 2 commits into the initial one.

PR #270, opened 2022-06-17
amascia/iac modify scan output message and style

amascia-gg:
Three changes:
- changed colour of policy title in output message
- added policy GG ID in output message
- click raises error when file is given to iac scan command

### DIFF
--- a/ggshield/cmd/iac/scan.py
+++ b/ggshield/cmd/iac/scan.py
@@ -62,7 +62,8 @@ def validate_exclude(_ctx: Any, _param: Any, value: Sequence[str]) -> Sequence[s
 )
 @click.option("--json", is_flag=True, help="JSON output.")
 @click.argument(
-    "directory", type=click.Path(exists=True, readable=True, path_type=Path)
+    "directory",
+    type=click.Path(exists=True, readable=True, path_type=Path, file_okay=False),
 )
 @click.pass_context
 def scan_cmd(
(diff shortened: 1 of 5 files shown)


ggshield/cmd/iac/scan.py
```diff
@@ -78,6 +79,7 @@ def scan_cmd(
     update_context(
         ctx, exit_zero, minimum_severity, verbose, ignore_policies, ignore_paths, json
     )
+
```
gg-mmill, line 82: Nit: avoid commiting meaningless changes.

ggshield/output/text/message.py
```diff
@@ -222,7 +222,9 @@ def iac_vulnerability_header(issue_n: int, vulnerability: IaCVulnerability) -> s
         issue_n,
         format_text("IaC", STYLE["detector"]),
         format_text(vulnerability.component, STYLE["detector"]),
-        format_text(vulnerability.policy, STYLE["detector"]),
+        format_text(
```
gg-mmill, line 225: I think it'd be better to change the format  template: add ` ({})` to the templatre, as well as an argument dedicated to policy_id (we might want to change the color of the id only, plus having white parenthesis allow to make it stand out a bit more.

ggshield/cmd/iac/scan.py
amascia-gg: Yes sorry, the PR was not ready...

ggshield/output/text/message.py
amascia-gg: Much better like this indeed, thanks

gg-mmill (approved):
LGTM !

PR #271, opened 2022-06-20
Fix IaC git handling

zhangmatthieu:
## fix(iac): Only send git files if the repo scanned is a git repository

Only send git files if the repo scanned is a git repository

### Testing
- Tested with command line
- Updated tests for ignore_git


ggshield/cmd/iac/scan.py
```diff
@@ -131,7 +131,8 @@ def iac_scan(ctx: click.Context, directory: Path) -> Optional[IaCScanResult]:
         path=directory,
         exclusion_regexes=ctx.obj["exclusion_regexes"],
         verbose=ctx.obj["config"].verbose,
-        ignore_git=True,
+        # If the repository is a git repository, ignore non-tracked files
```
amascia-gg, line 134: nit: untracked

tests/iac/test_filter.py
```diff
@@ -66,7 +69,10 @@ def test_get_iac_files_from_paths_ignore_git(tmp_path):
     subprocess.run(["git", "init"], cwd=tmp_path_str)
     subprocess.run(["git", "add", "."], cwd=tmp_path_str)
 
-    files = get_iac_files_from_paths(tmp_path, set(), True)
-    assert len(files.files) == 9
-    assert str("file2.json") not in files.files
-    assert str("file3.yaml") in files.files
+    files = get_iac_files_from_paths(tmp_path, set(), True, ignore_git)
+    if ignore_git:
+        assert len(files.files) == 10
```
amascia-gg, line 74: nit: maybe add `assert "file2.json" in files.files` here to make the comparison with the `else` case even clearer, if you feel it's redundant, it's not very important of course.

amascia-gg (commented):
some minor comments but otherwise lgtm!

amascia-gg:
The issue mentions the `.git` folder as well as the `.gitignore` file. When setting `ignore_git` to `True` does it also ignore everything that is is in `.git`? If so, is it possible to add a test to confirm it?

tests/iac/test_filter.py
```diff
@@ -2,8 +2,9 @@
 import subprocess
 from pathlib import Path
 
-from ggshield.iac.filter import get_iac_files_from_paths
+import pytest
```
amascia-gg, line 6: I think the issue with flake8 is with the imports here

amascia-gg (commented):
There is a minor issue with flake8 but otherwise LGTM!

PR #277, opened 2022-06-27
Improve docker filesystem banlist

### LINKED ISSUE #193: ggshield scan docker misses some files

- 1.11.0

**Describe the bug**

1. Unpack this archive: [missed-secret.zip](https://github.com/GitGuardian/ggshield/files/8415919/missed-secret.zip)
2. Create a Docker image from it:
```
cd missed-secret
docker build -t missed-secret:1 .
```
3. Scan the image:
```
ggshield scan docker missed-secret:1
```

→ No secrets are found.

4. Change the `COPY` line in the Dockerfile to:
```
COPY index.py /index.py
```

5. Rebuild the Docker image:
```
docker build -t missed-secret:2 .
```
6. Scan the image:

```
ggshield scan docker missed-secret:2
```

→ Secret is found.

pierrelalanne:
In this PR, we do mainly two things :  
- Improve the existing way we "pre-validate" filepath before sending them to the API.  
- Modify a bit the banlist to fix #193.  
  - We now allow some very specific cases of `/usr/` filepath when the file relates to nginx. 
  - We allow `/gems` filepath, as it introduces a low volume of results with some interesting results.  
  - I tried to identify some other rules that we could change : I think this is outside the scope of this issue, and it would require some more work.

### DIFF
--- a/ggshield/scan/docker.py
+++ b/ggshield/scan/docker.py
@@ -134,6 +134,19 @@ def _get_layers_files(
         yield from _get_layer_files(archive, layer_info)
 
 
+def _validate_filepath(
+    filepath: str,
+    extension_banlist: Iterable[str],
+    filepath_banlist: Iterable[str],
+) -> bool:
+    if any(banned_filepath in "/" + filepath for banned_filepath in filepath_banlist):
+        return False
+
+    if any(filepath.endswith(extension) for extension in extension_banlist):
+        return False
+    return True
+
+
 def _get_layer_files(archive: tarfile.TarFile, layer_info: Dict) -> Iterable[File]:
     """
     Extracts File objects to be scanned for given layer.
@@ -148,13 +161,11 @@ def _get_layer_files(archive: tarfile.TarFile, layer_info: Dict) -> Iterable[Fil
         if not file_info.isfile():
             continue
 
-        if file_info.size > MAX_FILE_SIZE * 0.95:
-            continue
-
-        if any(dir in "/" + file_info.path for dir in DEFAULT_FS_BANLIST):
-            continue
-
-        if any(file_info.path.endswith(ext) for ext in DEFAULT_EXTENSION_BANLIST):
+        if not _validate_filepath(
+            file_info.path,
+            extension_banlist=DEFAULT_EXTENSION_BANLIST,
+            filepath_banlist=DEFAULT_FS_BANLIST,
+        ):
             continue
 
         file = layer_archive.extractfile(file_info)
--- a/ggshield/scan/docker.py
+++ b/ggshield/scan/docker.py
@@ -18,24 +18,26 @@ from ggshield.scan import ScanCollection
 from ggshield.scan.scannable import File, Files
 
 
-DEFAULT_FS_BANLIST = {
-    "/usr/",
-    "/lib/",
-    "/share/",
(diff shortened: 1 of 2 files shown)


ggshield/scan/docker.py
```diff
@@ -18,21 +18,23 @@
...
+    r"^\/?bin\/",
+    r"^\/?sbin\/",
+    r"^\/?node_modules\/",
+    r"^\/?include\/",
+    r"^\/?vendor\/",
+    r"^\/?texlive\/",
+    r"^\/?var\/",
+    r"^\/?fonts\/",
+    r"^\/?npm\/",
+    r"^\/?site-packages\/",
```
agateau-gg, line 34: The regexes can be simplified a bit: `/` does not need to be escaped.

ggshield/scan/docker.py
```diff
@@ -134,6 +136,22 @@ def _get_layers_files(
         yield from _get_layer_files(archive, layer_info)
 
 
+def _validate_filepath(
+    filepath: str,
+    extension_banlist: Iterable[str],
+    filepath_banlist_patterns: Iterable[re.Pattern],
+) -> bool:
+    if any(
+        re.search(banned_pattern, filepath)
```
agateau-gg, line 145: nit: can be simplified to `banned_pattern.search(filepath)`.

ggshield/scan/docker.py
```diff
@@ -148,13 +166,11 @@ def _get_layer_files(archive: tarfile.TarFile, layer_info: Dict) -> Iterable[Fil
         if not file_info.isfile():
             continue
 
-        if file_info.size > MAX_FILE_SIZE * 0.95:
-            continue
```
agateau-gg, line 152: I think this test should be brought back, to avoid scanning large files.

ggshield/scan/docker.py
```diff
@@ -134,6 +136,22 @@ def _get_layers_files(
         yield from _get_layer_files(archive, layer_info)
 
 
+def _validate_filepath(
+    filepath: str,
+    extension_banlist: Iterable[str],
+    filepath_banlist_patterns: Iterable[re.Pattern],
```
agateau-gg, line 142: nit: since this function is always called with `DEFAULT_EXTENSION_BANLIST` and `FILEPATH_BANLIST_PATTERNS` as arguments, maybe these two arguments can be removed?

agateau-gg (changes requested):

ggshield/scan/docker.py
pierrelalanne: I thought we may change the banlists at some point, but I think you are right. 👍

ggshield/scan/docker.py
agateau-gg: Oh good catch, but then I think it's better to remove the other line, because the first check does not read the file content, whereas the second does.

pierrelalanne: Hmm we have to read it anyway for the `File.from_bytes` call, no ? 🤔

agateau-gg: Yes, but the early check avoid reading the content of large files we are going to ignore.

pierrelalanne: Alright, should be better now. Thanks.

PR #280, opened 2022-06-30
feat(iac): handle exceptions more gracefully for IaC

zhangmatthieu:
- Handle error validation for detail if detail is not in the response
- Remove display of chunks when there are errors
- Handle ReadTimeout errors from requests

### Testing
- Added unit tests

Pair-programming done with Laurent Tramoy


ggshield/scan/scannable_errors.py
```diff
@@ -43,3 +43,10 @@ def handle_scan_error(detail: Detail, chunk: List[Dict[str, str]]) -> None:
         )
 
         display_error(str(detail))
+
+
+def handle_scan_error(detail: Detail) -> None:
+    if detail.status_code == 401:
+        raise click.UsageError(detail.detail)
+    display_error("\nError scanning. Results may be incomplete.")
```
gg-mmill, line 51: Will there ever be any result ?

If not, the `Results may be incomplete.` can be removed I think.

ggshield/scan/scannable_errors.py
```diff
@@ -7,7 +7,7 @@
 from ggshield.core.text_utils import STYLE, display_error, format_text, pluralize
 
 
-def handle_scan_error(detail: Detail, chunk: List[Dict[str, str]]) -> None:
+def handle_scan_chunked_error(detail: Detail, chunk: List[Dict[str, str]]) -> None:
```
gg-mmill, line 10: Nit:
replace name by `handle_scan_chunk_error` ( or maybe `handle_chunked_scan_error` ?)

gg-mmill (commented):
Some nits,  otherwise good to me !

PR #283, opened 2022-07-05
Fix ggshield secret scan repo crashes on repositories containing files with a 0 bytes in them

### LINKED ISSUE #264: `ggshield secret scan repo` crashes on repositories containing files with a `0` bytes in them
## Description

`ggshield secret scan repo` crashes on repositories containing files with '0' bytes in them. This is a `secret scan repo` variant of #155.

## How to reproduce

- Create a test repository.
- Commit the test file provided in #155 to it.
- Run `ggshield secret scan repo` on it.
- ggshield crashes with this error:

```
$ ggshield secret scan repo .
Scanning Commits  [------------------------------------]    0%
Error: {0: {'document': ['document has null characters']}}
```

## WIP fix

The `agateau/verbose-scan-repo` branch contains a fix for this. See commit 7c4e263c46e05513e9278e7d5fa016b528d9b03f.

agateau-gg:
Since version 1.3.5, py-gitguardian takes care of cleaning input documents by replacing `0` bytes. no need to do it ourselves.

Also, don't replace replacement characters: they do not cause problems and are a better fit than a space.

Fixes #264

### DIFF
--- a/tests/core/test_file_utils.py
+++ b/tests/core/test_file_utils.py
@@ -9,12 +9,12 @@ from ggshield.core.file_utils import generate_files_from_paths
 @pytest.mark.parametrize(
     ["filename", "input_content", "expected_content"],
     [
-        ("normal.txt", "Normal", "Normal"),
-        ("0-inside.txt", "Ins\0de", "Ins de"),
+        ("normal.txt", b"Normal", "Normal"),
+        ("invalid-utf8-start-byte.txt", b"Hello\x81World", "Hello\uFFFDWorld"),
     ],
 )
 def test_generate_files_from_paths(
-    tmp_path, filename: str, input_content: str, expected_content: str
+    tmp_path, filename: str, input_content: bytes, expected_content: str
 ):
     """
     GIVEN a file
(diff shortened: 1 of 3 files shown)


tests/core/test_file_utils.py
```diff
@@ -9,12 +9,12 @@
 @pytest.mark.parametrize(
     ["filename", "input_content", "expected_content"],
     [
-        ("normal.txt", "Normal", "Normal"),
-        ("0-inside.txt", "Ins\0de", "Ins de"),
+        ("normal.txt", b"Normal", "Normal"),
+        ("invalid-utf8-start-byte.txt", b"Hello\x81World", "Hello\uFFFDWorld"),
```
pierrelalanne, line 13: Nit : We could add a test to verify that `\0` is indeed not replaced

pierrelalanne (approved):
One minor remark. Looks good otherwise :+1:

tests/core/test_file_utils.py
agateau-gg: Good point, just added a case for it

PR #286, opened 2022-07-06
fix(user_config): #278 Don't fail on unknown keys for user config

### LINKED ISSUE #278: Only warn on unknown configuration keys, do not fail
## Description

Given a configuration file with an unknown configuration key:

```
$ cat > .gitguardian.yaml <<EOF
heredoc> version: 2
heredoc> unknown: bla
heredoc> EOF
```

ggshield from the iac branch currently fails:

```
$ ggshield api-status                                                     
Error: Error in ./.gitguardian.yaml:
unknown: ['Unknown field.']
```

Compare with the behavior of ggshield from the main branch:

```
$ ggshield api-status
Unrecognized key in config: version
Unrecognized key in config: unknown
Status: healthy
App version: v2.8.0
Secrets engine version: 2.70.0
```

We need to bring back this behavior: it is critical that users of ggshield vX.Y can continue to run ggshield even if the configuration file contains keys used by ggshield vX.Y+1.

GG-HH:
- Add a base config class to handle pre_load and output message with unknown keys in config
- Add marshmallow_dataclass plugin for mypy

Please note that we don't raise a warning for additional fields in IgnoredMatches. I thought those are less important since they don't change the behaviour of the scan and can be a very long output in the case of a faulty copy/paste. This can be changed if you think it is better to raise the warning.

 Close #278

### DIFF
--- a/ggshield/core/config/user_config.py
+++ b/ggshield/core/config/user_config.py
@@ -20,6 +20,7 @@ from ggshield.core.constants import (
     GLOBAL_CONFIG_FILENAMES,
     LOCAL_CONFIG_PATHS,
 )
+from ggshield.core.text_utils import display_warning
 from ggshield.core.types import IgnoredMatch, IgnoredMatchSchema
 from ggshield.core.utils import api_to_dashboard_url
 from ggshield.iac.utils import POLICY_ID_PATTERN, validate_policy_id
@@ -34,7 +35,7 @@ class BaseConfig:
     @pre_load(pass_many=False)
     def filter_fields(cls, data: Dict, **kwargs: Any) -> Dict:
         """
-        Remove and alert on unknown field.
+        Remove and alert on unknown fields.
         """
         field_names = {field_.name for field_ in fields(cls)}
         filtered_fields = {}
@@ -42,7 +43,7 @@ class BaseConfig:
             if key in field_names:
                 filtered_fields[key] = item
             else:
-                click.echo("Unrecognized key in config: {}".format(key))
+                display_warning("Unrecognized key in config: {}".format(key))
 
         return filtered_fields
 
--- a/ggshield/core/config/user_config.py
+++ b/ggshield/core/config/user_config.py
@@ -1,10 +1,11 @@
 import os
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, fields
 from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
 import click
 import marshmallow_dataclass
 from marshmallow import ValidationError
+from marshmallow.decorators import pre_load
 
 from ggshield.core.config.errors import ParseError, format_validation_error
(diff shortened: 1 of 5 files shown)


ggshield/core/config/user_config.py
```diff
@@ -27,8 +28,27 @@
...
+        """
+        Remove and alert on unknown field.
+        """
+        field_names = {field_.name for field_ in fields(cls)}
+        filtered_fields = {}
+        for key, item in data.items():
+            if key in field_names:
+                filtered_fields[key] = item
+            else:
+                click.echo("Unrecognized key in config: {}".format(key))
```
agateau-gg, line 45: Can you use `display_warning()` here instead of `click.echo()`? `display_warning()` applies coloring and does not break pipe redirection because it prints to stderr instead of stdout.

ggshield/core/config/user_config.py
```diff
@@ -27,8 +28,27 @@
...
 
-@dataclass
-class SecretConfig:
+@marshmallow_dataclass.dataclass
+class BaseConfig:
+    @classmethod
+    @pre_load(pass_many=False)
+    def filter_fields(cls, data: Dict, **kwargs: Any) -> Dict:
+        """
+        Remove and alert on unknown field.
```
agateau-gg, line 37: nit
```suggestion
        Remove and alert on unknown fields.
```

ggshield/core/types.py
```diff
@@ -1,11 +1,14 @@
...
 
 import marshmallow_dataclass
+from marshmallow import EXCLUDE
 
 
-@dataclass
+@marshmallow_dataclass.dataclass
 class IgnoredMatch:
+    class Meta:
+        unknown = EXCLUDE
```
agateau-gg, line 10: This is going to silently ignore unknown keys in matches. I think you can fix this by moving BaseConfig here so that IgnoredMatch can use it.

Also, can you rename BaseConfig to something like FilteredConfig?

agateau-gg (changes requested):

ggshield/core/types.py
GG-HH: I know, I was wondering if we wanted to alert in that case:
- extra fields won't change anything in the behaviour of ggshield (typos in the field name as well)
- it can result in a lot of warning if an extra field is present with too many occurences (if you ignore 100 secrets, you get 100 warnings)
- some users may want to add extra information to the ignored match and it can be useful to allow other keys
- it is the current behaviour in the main branch
- 
On the other hand,
- It is a different behaviour than the remaining config
- it may be harder to spot a typo in the "match" key since it will only show an error on a missing key but not the extra one.

WDYT ?

GG-HH: Ok, I have done it.
It must be noted that this changes the behaviour of V1 config where it will now alert on extra keys.

agateau-gg (approved):
Looks great, thanks!

PR #289, opened 2022-07-07
fix: fix docker output breaking json redirection

agateau-gg:
Print progress info on stderr so that we can pipe ggshield output to tools consuming json from stdin.

Before:
```
$ ggshield secret scan --json docker ubuntu:20.04 | jq
parse error: Invalid numeric literal at line 1, column 7
Scanning  [####################################]  100%          
Error: [Errno 32] Broken pipe
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

After:
```
$ ggshield secret scan --json docker ubuntu:20.04 | jq      
Saving docker image... OK
Scanning  [####################################]  100%          
{
  "id": "ubuntu:20.04",
  "type": "scan_docker_archive",
  "total_incidents": 0,
  "total_occurrences": 0
}
```

### DIFF
--- a/ggshield/core/cache.py
+++ b/ggshield/core/cache.py
@@ -7,6 +7,7 @@ from pygitguardian.models import PolicyBreak
 
 from ggshield.core.constants import CACHE_FILENAME
 from ggshield.core.filter import get_ignore_sha
+from ggshield.core.text_utils import display_warning
 
 
 SECRETS_CACHE_KEY = "last_found_secrets"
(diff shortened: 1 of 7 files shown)


Walz (commented):
I looked at where `click.echo` is used and I found two places where error messages are printed in stdin:
- `ggshield.core.config.utils.load_yaml`: when the yaml isn't properly loaded and `raise_exc` is `False`.
- `ggshield.core.cache.Cache.update_cache`: When called with an unrecognized key.

This may not be relevant as I'm not sure in which exact context those errors can occur.

PR #290, opened 2022-07-07
Add --debug option

agateau-gg:
To help with customer support, this PR adds a `--debug` option to log the following info:

- path of loaded config files
- command line arguments
- filename and filemode (added, removed, modified) of scanned documents
- sha of scanned commits
- scan exit code


pierrelalanne (approved):
I just have one minor remark : while testing the `--debug` option, we should note that we will catch logs from other libs. For instance `urllib3`

PR #298, opened 2022-07-15
Fix diff parser failing to extract filenames containing spaces

### LINKED ISSUE #273: pre-receive does not print filenames with spaces correctly

- 1.12.0

**Command executed**

`ggshield secret scan pre-receive`

**Describe the bug**

- Unpack this [zip file][zip], it contains a git repository with this content:
    - `biggerpython_server copie 11.py`: a file with spaces in its name. Contains a secret.
    - `startscan.sh`: a script to reproduce the bug. It calls `ggshield secret scan pre-receive` with the right stdin content.
- Run `startscan.sh`

[zip]: https://github.com/GitGuardian/ggshield/files/8957755/prereceive-filename-bug-example.zip

The output looks like this:

```
$ ./startscan.sh 
Scanning Commits  [####################################]  100%

commit 9e0b342414884fc68e8939504545c479524c38dd
Author: Aurelien Gateau <aurelien.gateau@gitguardian.com>
Date: Wed Jun 22 14:52:35 2022 +0200

🛡️  ⚔️  🛡️  1 incident has been found in file pie

(...)
```

As you can see the file name is wrong.

**Expected behavior**

File name should be printed correctly, like `ggshield secret scan path 'biggerpython_server copie 11.py'` does:

```
secrets-engine-version: 2.69.0

🛡️  ⚔️  🛡️  1 incident has been found in file /home/agateau/tmp/prereceive-filename-bug-example/biggerpython_server copie 11.py
(...)
```

agateau-gg:
This bug caused crashes in `scan pre-receive` and `scan repo`.

Refactors the parser to use a stricter way to extract the filename. As a result, test data in test_scannable.py had to be adjusted because their diff files were wrong.

Fixes #273

### DIFF
--- a/ggshield/scan/scannable.py
+++ b/ggshield/scan/scannable.py
@@ -33,6 +33,14 @@ logger = logging.getLogger(__name__)
 _MATCH_FILENAME_RX = re.compile(r"^a/(.*) b/\1$")
 
 
+class PatchParseError(Exception):
+    """
+    Raised by Commit.get_files() if it fails to parse its patch.
+    """
+
+    pass
+
+
 class Result(NamedTuple):
     """
     Return model for a scan which zips the information
@@ -290,7 +298,7 @@ class Commit(Files):
         if line.startswith("old"):
             return Filemode.PERMISSION_CHANGE
 
-        raise click.ClickException(f"Filemode is not detected:{line}")
+        raise PatchParseError("Filemode is not detected")
 
     @staticmethod
     def _parse_diff_header_lines(lines: List[str]) -> Tuple[str, Filemode]:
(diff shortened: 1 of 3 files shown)


ggshield/scan/scannable.py
```diff
@@ -298,6 +292,52 @@ def get_filemode(line: str) -> Filemode:
...
+        # PERMISSION_CHANGE and RENAME without changes do not have these lines.
+        filemode = Commit._get_filemode(lines[1])
+        if filemode == Filemode.RENAME:
+            prefix = "rename to "
+            for line in lines[2:]:
+                if line.startswith(prefix):
+                    filename = line[len(prefix) :]
+                    break
+            else:
+                raise click.ClickException(f"Could not extract filename from {lines}")
```
pierrelalanne, line 333: From what I understand, this should almost never happen : but we will end up logging code and possibly secrets ? Shouldn't we log the first lines only instead ?

ggshield/scan/scannable.py
```diff
@@ -298,6 +292,52 @@ def get_filemode(line: str) -> Filemode:
...
+        else:
+            match = _MATCH_FILENAME_RX.match(lines[0])
+            if not match:
+                raise click.ClickException(f"Could not extract filename from {lines}")
```
pierrelalanne, line 337: Same remark as above.

pierrelalanne (commented):
Looks nice. I had one remark about exceptions.

ggshield/scan/scannable.py
agateau-gg: Good point. Going to rewrite these.

pierrelalanne (approved):

PR #301, opened 2022-07-19
feat(scan): Do not abort repo scan if a commit fails to scan

### LINKED ISSUE #267: Do not abort repo scan if a commit fails to scan
## The problem

When `ggshield secret scan repo` fails to scan a commit, it stops. This is annoying because it forces users to restart the full scan.

## Proposed solution

The command should note all failures and carry on. At the end it should summarize all failures and exit with an error.

To be decided: in the case of JSON output, should the failure summary be part of the JSON document or should it be sent directly to stderr?

Walz:
closes #267 

When an error occurs when scanning a commit, `ggshield` will no longer stop. The error is printed in stderr and included in the JSON output.

I changed the timeout to 0.3s to force errors during the scan, here an example of the JSON output:
```json
{
  "id": "hello-gg",
  "type": "commit-range",
  "scans": [
    {
      "id": "5be496155e6c6a86199a744903a42419983d2592",
      "type": "commit",
      "errors": [
        {
          "files": [
            {
              "mode": "NEW",
              "filename": "test.py"
            }
          ],
          "description": "HTTPSConnectionPool(host='api.gitguardian.com', port=443): Read timed out. (read timeout=0.3)"
        }
      ],
      "extra_info": {
        "author": "Samuel Guillaume",
        "email": "sguillaume@dev.dev",
        "date": "Mon Aug 16 16:41:06 2021 +0200"
      },
      "total_incidents": 0,
      "total_occurrences": 0
    }
  ],
  "total_incidents": 0,
  "total_occurrences": 0
}
```

### DIFF
--- a/ggshield/cmd/secret/scan/archive.py
+++ b/ggshield/cmd/secret/scan/archive.py
@@ -10,7 +10,7 @@ from ggshield.core.config import Config
 from ggshield.core.file_utils import get_files_from_paths
 from ggshield.core.utils import SupportedScanMode
 from ggshield.output import OutputHandler
-from ggshield.scan import Files, Result, ScanCollection
+from ggshield.scan import Files, ScanCollection
 
 
 @click.command()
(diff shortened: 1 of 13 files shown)


agateau-gg (approved):
Looks good!

Issue #306, opened 2022-07-20
Do not create ggshield cache when running `secret scan pre-receive`


agateau-gg:
## GitGuardian Shield Version

- 1.12.0

## Command executed

`ggshield secret scan pre-receive`

## Describe the bug

When ggshield is used in pre-receive mode, ggshield cache should be skipped as it's not actionable after. This can also avoid trying to save on a read-only file-system.

## Expected behavior

`ggshield secret scan pre-receive` should not save its cache.

## Technical details

This is already done in `secret/scan/ci.py` by using a ReadOnlyCache. This ReadOnlyCache could be moved to Cache and reused in the pre-receive code.

PR #317, opened 2022-08-09
feat(scan): do not create ggshield cache when scan pre-receive

guillaume-malle:
Issue #306 
Like it is done for `scan ci`, we do not want to create cache when running `scan pre-receive`

Tested manually using the doc/pre-receive.sample script :
After committing a secret, run `echo "HEAD <recent-commit-ID> <unused-arg>" | sh doc/pre-receive.sample`
No cache file .cache_ggshield is created


jeremyds (approved):
lgtm

PR #325, opened 2022-08-18
fix(iac): fix iac exit code to return a non zero code in case of an error

zhangmatthieu:
fix(iac): fix iac exit code to return a non zero code in case of an error

Updated unit tests

### DIFF
--- a/tests/output/test_iac_text_output.py
+++ b/tests/output/test_iac_text_output.py
@@ -8,32 +8,50 @@ from tests.conftest import (
     _IAC_MULTIPLE_VULNERABILITIES,
     _IAC_NO_VULNERABILITIES,
     _IAC_SINGLE_VULNERABILITY,
+    assert_invoke_exited_with,
+    assert_invoke_ok,
     my_vcr,
 )
 
 
 @my_vcr.use_cassette("test_iac_scan_single_vulnerability")
-def test_display_single_vulnerabilities(cli_fs_runner: CliRunner):
-    Path("tmp/").mkdir(exist_ok=True)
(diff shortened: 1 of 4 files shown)


tests/output/test_iac_text_output.py
```diff
@@ -25,11 +26,26 @@ def test_display_single_vulnerabilities(cli_fs_runner: CliRunner):
...
         ],
     )
-
     assert_iac_version_displayed(result)
     assert_file_single_vulnerability_displayed(result)
 
 
+@my_vcr.use_cassette("test_iac_scan_single_vulnerability")
+def test_exit_zero_single_vulnerability(cli_fs_runner: CliRunner):
+    Path("tmp/").mkdir(exist_ok=True)
```
agateau-gg, line 35: You can use pytest `tmp_path` fixture here (https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture), to avoid leaving temporary directories behind.

tests/output/test_iac_text_output.py
```diff
@@ -25,11 +26,26 @@ def test_display_single_vulnerabilities(cli_fs_runner: CliRunner):
...
+    result = cli_fs_runner.invoke(
+        cli,
+        [
+            "iac",
+            "scan",
+            "--exit-zero",
+            "tmp",
+        ],
+    )
+    assert_invoke_exited_with(result, 0)
```
agateau-gg, line 46: You can use `assert_invoke_invoke_ok(result)` here.

agateau-gg (changes requested):

agateau-gg (approved):
Looks good!

PR #326, opened 2022-08-19
Update action.yml

ziadghalleb:
Update action.yml for a new listing on GH Marketplace.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,8 +1,25 @@
-name: 'ggshield Local Action'
-description: 'Scan commits for security incidents (using repository image)'
+name: 'ggshield scan'
+description: 'Scan commits for hardcoded secrets and security policy breaks.'
+author: GitGuardian <support@gitguardian.com>
+
 branding:
   icon: 'shield'
   color: 'blue'
+  
+inputs:
+  args:
+    description: |
+      Arguments to be passed to ggshield secret scan
+      Options:        
+        --json                        Output results in JSON format  [default: False]
+        --show-secrets                Show secrets in plaintext instead of hiding them.
+        --all-policies                Present fails of all policies (Filenames, FileExtensions,
+                                      Secret Detection). By default, only Secret Detection is shown.
+        --exit-zero                   Always return a 0 (non-error) status code, even if incidents are found.
+        -v, --verbose                 Verbose display mode.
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -12,3 +29,4 @@ runs:
     - secret
     - scan
     - ci
+    - ${{ inputs.args }}
```
agateau-gg, line 32: Because of the way ggshield argument parser works, arguments like `--json` or `--banlist-detector` must be placed between `scan` and `ci`, so I think `args` must be changed to something like this:

```yaml
args:
  - ggshield
  - -v
  - secret
  - scan
  - ${{ inputs.args }}
  - ci
```

But I haven't verified if it works (and I am actually not sure how to try it)

action.yml
```diff
@@ -1,8 +1,25 @@
...
+  args:
+    description: |
+      Arguments to be passed to ggshield secret scan
+      Options:        
+        --json                        Output results in JSON format  [default: False]
+        --show-secrets                Show secrets in plaintext instead of hiding them.
+        --all-policies                Present fails of all policies (Filenames, FileExtensions,
+                                      Secret Detection). By default, only Secret Detection is shown.
+        --exit-zero                   Always return a 0 (non-error) status code, even if incidents are found.
+        -v, --verbose                 Verbose display mode.
```
agateau-gg, line 19: Since `runs.args` already contains `-v`, the output is always going to be verbose. This means listing `--verbose` in the arguments is confusing because it won't affect the output.

agateau-gg (changes requested):

action.yml
ziadghalleb: Agreed, I realized this after I created the PR.

action.yml
ziadghalleb: Alright. Is it a good idea if I create the same PR on this (soon to be deprecated) [repository](https://github.com/GitGuardian/ggshield-action) to try your suggested change?

agateau-gg: Since we want to deprecate it, I think we should leave it as is for now. When we are confident the action in this repository works as expected we can add a deprecation message to the ggshield-action repository (both in the README and in the action code itself).

Regarding testing the changes: if I am not mistaken you should be able to test your branch using `uses: GitGuardian/ggshield@list-gh-action-mktplace` (but maybe you already knew that :) )

PR #327, opened 2022-08-19
Fix diff parser failing on some patches

### LINKED ISSUE #322: diff parser fails on some commits
## GitGuardian Shield Version

1.13.0

## Command executed

`ggshield secret scan repo .`

## Describe the bug

When scanning some repositories, the scan sometimes fail with the message:

```
Error parsing commit 3415eaed89dd06289ec6e581d6fcc563b98f27ab: Could not extract filename
```

This happens on commits which are both a rename and a permission change. ggshield mis-identifies the commit as a permission change and the regex used in this case to identify the file name does not match.

## Possible fixes

- Identify rename + permission change as a rename, not a permission change => could work, but still fragile
- Use a sentinel value for the a/ and b/ prefix and fix the name regex => could cause other issues down stream
- Use `--patch -z --raw` to generate diffs and parse the header => no fragile regex, more robust solution

agateau-gg:
## Problem

A git diff looks like this:

```diff
diff --git a/ggshield/scan/scannable.py b/ggshield/scan/scannable.py
index d155cf1..02634cd 100644
--- a/ggshield/scan/scannable.py
+++ b/ggshield/scan/scannable.py
@@ -29,8 +29,66 @@ from .scannable_errors import handle_scan_chunk_error
 
 logger = logging.getLogger(__name__)
```

The current diff parser uses the `diff --git` line to extract the name of the affected file, it assumes that in `diff --git a/<old> b/<new>` `old` and `new` are the same unless the affected file has also been renamed. It must do this to be able to correctly extract file names with spaces (see #298).

Unfortunately it gets confused by complex diffs combining a permission change + rename + edit, because it identifies the diff as a permission change and thus fails to extract the filename (because the diff is also a rename, `old` is different from `new`).

## What has been done

- Rewrite the parser to use `git show --raw -z --patch`. This produces a more robust header at the beginning of the diff, which follows this format: https://git-scm.com/docs/git-diff-tree#_raw_output_format
- Add a set of patch files in `tests/data/patches` to run unit-tests on. The directory also contains the script used to generate said patches, making it easy to regenerate them or add new test patches.
- Update existing test patches stored in `tests/conftest.py` so they are formatted as `git show --raw -z --patch` would have generated them.

## Review hints

The bulk of the work is in `ggshield/scan/scannable.py` and `tests/scan/test_commit.py`. The rest is just adapting to the changes.

## Issue

Fixes #322


tests/scan/test_commit.py
```diff
@@ -1,109 +1,222 @@
...
+# diff --git a/old.txt b/new.txt
+# similarity index 71%
+# index 0000000..b80e3df
+# """
+#
+#
+# INVALID_MODIFIED_FILE = """
+# diff --git a/test1.txt b/test2.txt
+# index e965047..802992c 100644
+# """
```
Walz, line 115: The old code is in comments.

Walz (changes requested):

tests/scan/test_commit.py
agateau-gg: Oups, that's embarrassing 😅 

Just fixed it and also reworked the parser a bit as I noticed it was eating trailing `\n`.

Walz (approved):

PR #330, opened 2022-08-22
feat(iac): add iac ggshield action

zhangmatthieu:
Move ggshield action + add iac action


actions/entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#! /usr/bin/env bash
+
+args=("$@")
+ggshield secret scan -v ${args[@]} ci
```
agateau-gg, line 4: I don't know the syntax used here for `args`. What does it bring compared to the simpler `ggshield secret scan -v "$@" ci`?

agateau-gg (commented):
The file hierarchy in the PR is:

```
/
  actions  # secret action
    iac    # iac action
```

Is it possible to have this instead?

```
/
  actions
    iac    # iac action
    secret # secret action
```

There is no reason for secret to be above iac.

actions/entrypoint.sh
zhangmatthieu: Sure, will update the file hierarchy if it doesn't make sense to have secret above iac. Unsure about the syntax. This was the working syntax in ggshield-actions, I will check

zhangmatthieu: It doesn't seem to work without the args trick. The help command is displayed

ziadghalleb (approved):

actions/entrypoint.sh
agateau-gg: Strange, it works here. Another Bash mystery I guess.

agateau-gg (changes requested):
One last thing before this can go in: Can you update the README section about the action? (https://github.com/GitGuardian/ggshield#github)

zhangmatthieu:
> Let me know if you are okay with the last change on commit [4957fdb](https://github.com/GitGuardian/ggshield/pull/330/commits/4957fdbb141225e884cc011d42c047ead1696baf).

Each type doesn't seem really descriptive. Other than that, looks good to me, let me add the commit in the PR

agateau-gg (approved):
Looks good, thanks!

Issue #313, opened 2022-07-26
Bug: scan first commit in GitHub action fails


GG-HH:
**GitGuardian Shield Version**

- [x] I can reproduce this bug in the latest version (bug occured with gg-shield-action@master)

**Command executed**

I use the GitGuardian/gg-shield-action@master in my GitHub CI.

**Describe the bug**

When pushing the first commit of a repo, the action fails.

**Expected behavior**

I expect this commit to be scanned and the action to finish.

**Traceback (if available)**

```
Error: Unable to get commit range. Please submit an issue with the following info:
  Repository URL: <Fill if public>
github_push_before_sha: 0000000000000000000000000000000000000000
github_push_base_sha: 
github_pull_base_sha: github_default_branch: main
github_head_sha: REDACTED
```

** Action configuration**
```yaml
 ggshield:
    name: Check for secrets with ggshield
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0 # fetch all history so multiple commits can be scanned

      - name: GitGuardian scan
        uses: GitGuardian/gg-shield-action@master
        env:
          GITHUB_PUSH_BEFORE_SHA: ${{ github.event.before }}
          GITHUB_PUSH_BASE_SHA: ${{ github.event.base }}
          GITHUB_PULL_BASE_SHA: ${{ github.event.pull_request.base.sha }}
          GITHUB_DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
          GITGUARDIAN_API_KEY: ${{ secrets.GITGUARDIAN_API_KEY }}
```

PR #338, opened 2022-08-25
fix(ci): Fix commit range computation when on a root commit

Walz:
fixes https://github.com/GitGuardian/ggshield/issues/313

Reproduce the bug by creating a repository with only one commit and running:

```
$ CI=1 GITHUB_ACTIONS=1 GITHUB_PUSH_BEFORE_SHA=0000000000000000000000000000000000000000 GITHUB_DEFAULT_BRANCH=master GITHUB_SHA="$(git rev-parse HEAD )" ggshield secret scan ci
Error: Unable to get commit range. Please submit an issue with the following info:
  Repository URL: <Fill if public>
github_push_before_sha: 0000000000000000000000000000000000000000
github_push_base_sha: None
github_pull_base_sha: None
github_default_branch: master
github_head_sha: 5dacd7de7867c8781fb2edbb7d107de2a28bdb18
```

`ggshield` will execute the following command:

```
$ git rev-list --reverse 5dacd7de7867c8781fb2edbb7d107de2a28bdb18~1...
fatal: argument '5dacd7de7867c8781fb2edbb7d107de2a28bdb18~1...' ambigu : révision inconnue ou chemin inexistant.
Utilisez '--' pour séparer les chemins des révisions, comme ceci :
'git <commande> [<révision>...] -- [<chemin>...]'
```

The real error is hidden and can be seen if we add `--`
```
$ git rev-list --reverse 5dacd7de7867c8781fb2edbb7d107de2a28bdb18~1... --
fatal: bad revision '5dacd7de7867c8781fb2edbb7d107de2a28bdb18~1...'
```

`5dacd7de7867c8781fb2edbb7d107de2a28bdb18` is the first commit, it has no parent thus `5dacd7de7867c8781fb2edbb7d107de2a28bdb18~1` is invalid.

To fix this bug, we catch `bad revision` errors when using `~1...` and we return the root commit.


ggshield/core/git_shell.py
```diff
@@ -130,10 +135,17 @@ def get_list_commit_SHA(commit_range: str) -> List[str]:
...
     """
+    try:
+        commit_list = shell_split(
+            [GIT_PATH, "rev-list", "--reverse", *commit_range.split(), "--"],
+            raise_on_process_error=True,
+        )
+    except subprocess.CalledProcessError as e:
+        if b"bad revision" in e.stderr and commit_range.endswith("~1..."):
+            # Tried to access the parent of a root commit
+            return [commit_range.removesuffix("~1...")]
```
agateau-gg, line 146: This is dangerous: if the user makes an error while using `secret scan commit-range`, the exception is raised.

With the main branch:
```
$ ggshield secret scan commit-range hello~1...
Error: invalid commit range
```

With this branch:
```
$ ggshield secret scan commit-range hello~1...        
Scanning Commits  [####################################]  100%

No secrets have been found
```

I think it would be safer to do something like this in case of failure:
- check the bottom ref is a valid commit, Stack Overflow tells me to use `git cat-file -e $sha^{commit}` (https://stackoverflow.com/a/31780867/20107)
- if it is the case then we can return `[$sha] + get_commit_range($sha...)`

(also: we can't use `str.removesuffix()`, it requires Python 3.9)

agateau-gg (changes requested):

ggshield/core/git_shell.py
```diff
@@ -125,15 +130,38 @@ def git_ls(wd: Optional[str] = None) -> List[str]:
...
     :param commit_range: A range of commits (ORIGIN...HEAD)
     """
+    try:
+        commit_list = shell_split(
+            [GIT_PATH, "rev-list", "--reverse", *commit_range.split(), "--"],
+            raise_on_process_error=True,
+        )
+    except subprocess.CalledProcessError as e:
+        if b"bad revision" in e.stderr and commit_range.endswith("~1..."):
+            commit_ref = commit_range[:-5]
```
agateau-gg, line 160: Can you add a comment explaining we handle the case where `commit_ref` has no parent here?

ggshield/cmd/secret/scan/ci.py
```diff
@@ -205,7 +205,7 @@ def github_actions_range(verbose: bool) -> List[str]:  # pragma: no cover
         "  Repository URL: <Fill if public>\n"
         f"github_push_before_sha: {push_before_sha}\n"
         f"github_push_base_sha: {push_base_sha}\n"
-        f"github_pull_base_sha: {pull_req_base_sha}"
+        f"github_pull_base_sha: {pull_req_base_sha}\n"
```
agateau-gg, line 208: Good catch 👍🏻

agateau-gg (changes requested):

ggshield/core/git_shell.py
```diff
@@ -100,7 +104,8 @@ def shell(command: List[str], timeout: int = COMMAND_TIMEOUT) -> str:
         )
         return result.stdout.decode("utf-8", errors="ignore").rstrip()
     except subprocess.CalledProcessError:
-        pass
+        if raise_on_process_error:
+            raise
```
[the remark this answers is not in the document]
Walz: We need to know when an error occurs. If `shell` doesn't raise an exception, it needs to return the process return code. An empty output is not enough for `is_valid_git_commit_ref` for example.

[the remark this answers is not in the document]
agateau-gg: I agree, but with the example I gave `is_valid_git_commit_ref()` would call `shell(cmd, check=True)` and it would work just the same, no?

agateau-gg (approved):

PR #345, opened 2022-08-30
Do not duplicate pygitguardian upload-limit constant

agateau-gg:
Use pygitguardian `DOCUMENT_SIZE_THRESHOLD_BYTES` constant instead of duplicating its value in core.constants.

Avoids head-scratching when one increases this value and wonders why the changes have no effects...


pierrelalanne (approved):
👍

PR #348, opened 2022-09-02
Modify multi-thread value for ggshield secret scan commands

amascia-gg:
To speed up the `ggshield secret scan repo` and `path` commands, we want to modify the number of threads that are used:
- for `ggshield secret scan path` we will use 10 `scan_thread`s to scan chunks of files in parallel
- for `ggshield secret scan repo` we will use 10 `commit_thread`s to parallelize commits and 4 `scan_thread`s to scan chunks.

### DIFF
--- a/ggshield/cmd/secret/scan/path.py
+++ b/ggshield/cmd/secret/scan/path.py
@@ -8,6 +8,9 @@ from ggshield.output import OutputHandler
 from ggshield.scan import ScanCollection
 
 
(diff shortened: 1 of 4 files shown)


pierrelalanne:
I am closing this one as it will be included in https://github.com/GitGuardian/ggshield/pull/356.

agateau-gg:
As suggested by Pierre, I am closing this PR since #356 is an evolution of it.

Issue #344, opened 2022-08-30
Move scanning code outside of scannable.py


agateau-gg:
## Description

The `scan.scannable` module contains deals two topics: scanning documents and loading documents from files or commits.

The scanning code should be moved in a separate, secret-specific, module to make `scannable` simpler. This would improve the architecture of our code, for example the IAC commands use `scannable` to read files, but they do not need the secret-oriented `scan()` method.

## Proposed reorganization

1. Rename the `scan` package to `secret`. This matches the existing `iac` package.
2. Move secret scanning code from `scannable` to `secret`:
    - `Files.scan()` to a `secret.scan_files(files: Files)` function.
    - `Result`, `Error`, `Results` as `secret` classes.
3. Move `ScanCollection` to `output`: it can't go to `secret` because it contains `iac_result`, so it's not secret-only
4. Move the rest of `scannable` (`File`, `Commit`, `Files`, `CommitFile` and related parsing code) to `core.scannable`
5. (Bonus) Move `utils.ScanMode` to `secret`, rename it `SecretScanMode`
6. (Bonus) Move `utils.Filemode` to `core.scannable`, or remove it if it turns out we don't need it

## Definition of Done
- `scan` package does not exist anymore
- `core.scannable` does not depend on any secret-related code (no import of pygitguardian, no network-related code)
- `core.scannable` does not depend on any iac-related code

pierrelalanne:
A suggestion : should we rename `ScanMode` to `SecretScanMode` for clarity and in case we introduce different modes for other scanning capabilities ?

agateau-gg:
> A suggestion : should we rename `ScanMode` to `SecretScanMode` for clarity and in case we introduce different modes for other scanning capabilities ?

Good idea, amending the description.

PR #352, opened 2022-09-08
Improve scan repo perfs

pierrelalanne:
## Context  
The `ggshield secret scan repo` command used to make one API call per commit. This implementation was not efficient and could result in thousands of call for important repos. In most cases, one commit contains modifications to only a couple of files. This implies a lot of very small API calls.   

We propose an implementation to group commits together as long as the total number of files sent to the API is below a fixed limit (20 at the time of writing).  

### Part 1 - Improve scan repo logic

The implementation modifies `scan_commit_range` as follows : we create an intermediary steps where commits are iterated over and grouped to form batches of at most 20 files. These are ultimately sent to the API using the `Files.scan` interface.   
We also adapted the results parsing logic : now that one API call contains files from several commits, we have to re-create the link between the results and the commit they come from.  

One small drawback is that we lose this commit granularity when it comes to errors : we cannot tell which commit triggered which error.  

### Part 2 - Improve progress bars
While working on this PR, we experimented broken progress bars. Especially when it comes to scanning very big repos. We therefore propose several improvements to our progress bars :  
- We add [`rich`](https://github.com/Textualize/rich) as a dependency. It provides a nice and smooth management of progress bars. We may also use it in the future for nice output displays.  
- We introduce a callback logic in the `Scanner.scan` method : this is in order to be able to update progress bars whenever a batch of content is sent to the API.  
- We use this logic in most of our cmds : `pypi`, `docker`, `repo`, `archive`, `path`.  
- Note that the progress bar is sent to stderr not to interact with other outputs, json for instance.  

Here is an example of results :  
image

### DIFF
--- a/ggshield/cmd/secret/scan/archive.py
+++ b/ggshield/cmd/secret/scan/archive.py
@@ -1,16 +1,16 @@
 import shutil
-import sys
 import tempfile
+from functools import partial
 from pathlib import Path
-from typing import List
 
 import click
 
 from ggshield.core.config import Config
 from ggshield.core.file_utils import get_files_from_paths
+from ggshield.core.text_utils import create_progress_bar
 from ggshield.core.utils import ScanContext, ScanMode
 from ggshield.output import OutputHandler
-from ggshield.scan import File, Files, ScanCollection
+from ggshield.scan import Files, ScanCollection
 
 
 @click.command()
@@ -38,12 +38,10 @@ def archive_cmd(ctx: click.Context, path: str) -> int:  # pragma: no cover
             ignore_git=True,
         )
 
-        with click.progressbar(
-            length=len(files.files), label="Scanning", file=sys.stderr
-        ) as progressbar:
-
-            def update_progress(chunk: List[File]) -> None:
-                progressbar.update(len(chunk))
+        with create_progress_bar(doc_type="files") as progress:
+            task_scan = progress.add_task(
+                "[green]Scanning Archive...", total=len(files.files)
+            )
 
             scan_context = ScanContext(
                 scan_mode=ScanMode.ARCHIVE,
@@ -56,10 +54,10 @@ def archive_cmd(ctx: click.Context, path: str) -> int:  # pragma: no cover
                 scan_context=scan_context,
                 matches_ignore=config.secret.ignored_matches,
                 ignored_detectors=config.secret.ignored_detectors,
-                on_file_chunk_scanned=update_progress,
+                progress_callback=partial(progress.update, task_scan),
             )
 
-            scan = ScanCollection(id=path, type="archive_scan", results=results)
+        scan = ScanCollection(id=path, type="archive_scan", results=results)
 
-            output_handler: OutputHandler = ctx.obj["output_handler"]
-            return output_handler.process_scan(scan)
+        output_handler: OutputHandler = ctx.obj["output_handler"]
+        return output_handler.process_scan(scan)
--- a/ggshield/core/text_utils.py
+++ b/ggshield/core/text_utils.py
@@ -1,7 +1,16 @@
+import sys
 from enum import Enum, auto
 from typing import Any, Dict, NamedTuple, Optional, Union
 
 import click
+from rich.console import Console
+from rich.progress import (
+    BarColumn,
+    Progress,
+    TaskProgressColumn,
+    TextColumn,
+    TimeRemainingColumn,
+)
 
 
 LINE_DISPLAY = {"file": "{} | ", "patch": "{} {} | "}
@@ -132,3 +141,14 @@ def translate_validity(validity_id: Optional[str]) -> str:
     # If we don't have a text for the validity_id, return it as is. We assume the text
     # of the ID is more valuable than a generic "Unknown" string
     return _VALIDITY_TEXT_FOR_ID.get(validity_id, validity_id)
+
+
+def create_progress_bar(doc_type: str) -> Progress:
+    return Progress(
+        TextColumn("[progress.description]{task.description}"),
+        BarColumn(),
+        TaskProgressColumn(),
+        TextColumn(f"{{task.completed}} {doc_type} scanned out of {{task.total}}"),
+        TimeRemainingColumn(),
+        console=Console(file=sys.stderr),
+    )
--- a/ggshield/scan/repo.py
+++ b/ggshield/scan/repo.py
@@ -1,8 +1,8 @@
-import concurrent.futures
 import itertools
 import os
 import re
 import sys
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
 from typing import Iterable, Iterator, List, Optional, Set
 
@@ -158,30 +158,25 @@ def scan_commit_range(
         )
     )
 
-    with concurrent.futures.ThreadPoolExecutor(
-        max_workers=MAX_WORKERS
-    ) as executor:
-        future_to_process = [
-            executor.submit(
-                scan_commits_content,
-                commits,
-                client,
-                cache,
-                matches_ignore,
-                scan_context,
-                ignored_detectors,
-            )
-            for commits in commits_batch
-        ]
-
-        scans: List[ScanCollection] = []
+    with ThreadPoolExecutor(max_workers=MAX_WORKERS) as executor:
         with click.progressbar(
-            iterable=concurrent.futures.as_completed(future_to_process),
-            length=len(future_to_process),
+            iterable=(
+                executor.submit(
+                    scan_commits_content,
+                    commits,
+                    client,
+                    cache,
+                    matches_ignore,
+                    scan_context,
+                    ignored_detectors,
+                )
+                for commits in commits_batch
+            ),
             label=format_text("Scanning Commits", STYLE["progress"]),
             file=sys.stderr,
-        ) as completed_futures:
-            for future in completed_futures:
+        ) as futures:
+            scans: List[ScanCollection] = []
+            for future in as_completed(futures):
                 scan_collection = future.result()
                 for scan in scan_collection.scans:
                     if scan.results and scan.results.errors:
--- a/ggshield/scan/repo.py
+++ b/ggshield/scan/repo.py
@@ -1,10 +1,10 @@
 import itertools
 import os
 import re
-import sys
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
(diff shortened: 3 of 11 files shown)


ggshield/scan/repo.py
```diff
@@ -75,13 +77,56 @@ def scan_commit(
...
+        )
+
+    return ScanCollection(id=scan_context.command_id, type="commit-ranges", scans=scans)
+
+
+def get_commits_content_by_batch(
+    commit_list: List[str],
+    exclusion_regexes: Set[re.Pattern],
+    batch_max_size: int = MULTI_DOCUMENT_LIMIT,
+) -> Iterator[Tuple[Files, List[Commit]]]:
```
agateau-gg, line 107: The signature of this function is surprising: it returns the same information twice. Reading more of the code I guess this is because we need a single instance of `Files` to run `Files.scan()`. That's another justification for the refactoring in #344: if `Files.scan` were to become a `scan(List[File], ...)` function then this would not be a problem.

Until this is done, I think it would be better to hide that issue inside `scan_commits_content()`:

1. make `get_commits_content_by_batch()` return a `Iterator[List[Commit]]`
2. in `scan_commits_content()`, create a Files instance like this:

```python
commit_files = list(itertools.chain.from_iterable(c.files for c in commits))
files = Files(commit_files)
```

ggshield/scan/repo.py
```diff
@@ -75,13 +77,56 @@ def scan_commit(
...
+    """
+    Given a list of commit shas yield the commit files
+    by biggest batches possible of length at most MULTI_DOCUMENT_LIMIT
+    """
+    current_files_batch: List[File] = []
+    current_commits_batch: List[Commit] = []
+    call_counter = 0
```
agateau-gg, line 114: I think this can be removed

pierrelalanne: Yep, this was not exactly ready yet 😅   
I added a test and cleaned this. Should be better.

ggshield/scan/repo.py
agateau-gg: The patch is only parsed once: `commit.files` caches the results, so this should not be a problem.

ggshield/scan/repo.py
agateau-gg: Oups... 🚀

ggshield/scan/repo.py
```diff
@@ -75,13 +78,55 @@ def scan_commit(
...
+    batch_max_size: int = MULTI_DOCUMENT_LIMIT,
+) -> Iterator[List[Commit]]:
+    """
+    Given a list of commit shas yield the commit files
+    by biggest batches possible of length at most MULTI_DOCUMENT_LIMIT
+    """
+    commits_with_num_files = ((commit, len(commit.files)) for commit in commit_list)
+    current_count = 0
+    batch = []
+    for commit, num_files in commits_with_num_files:
```
agateau-gg, line 118: I think we can get rid of `commits_with_num_files`:
```suggestion
    current_count = 0
    batch = []
    for commit in commit_list:
        num_files = len(commit.files)
```

ggshield/scan/repo.py
```diff
@@ -102,21 +147,26 @@ def scan_commit_range(
     :param verbose: Display successful scan's message
     """
 
+    commits_batch = get_commits_by_batch(
+        commit_list=[
+            Commit(sha=sha, exclusion_regexes=exclusion_regexes) for sha in commit_list
+        ]
```
agateau-gg, line 153: `commit_list` argument can be turned into a generator to avoid creating a list:

```suggestion
        commit_list=(
            Commit(sha=sha, exclusion_regexes=exclusion_regexes) for sha in commit_list
        )
```

tests/scan/test_scan_repo.py
```diff
@@ -0,0 +1,71 @@
+from typing import List
+from unittest.mock import MagicMock
+
+import pytest
+
+from ggshield.scan import Commit, File
+from ggshield.scan.repo import get_commits_by_batch
+
+
+class TestScanRepo:
```
agateau-gg, line 10: nit: I don't think we need a class here.

agateau-gg (changes requested):
minor remarks. Looks good!

pierrelalanne:
Should be good now.  
Note : I modified a bit the progress bar. On big repos, we used to not have a progress bar at all. Now, it gets displayed, but does not move until a very long time after the start of the scan. I could not find the reason of the behavior.

ggshield/scan/repo.py
```diff
@@ -153,37 +153,33 @@ def scan_commit_range(
...
+            ):
                 scan_collection = future.result()
-                for scan in scan_collection.scans:
+                for scan in scan_collection.scans_with_results:
                     if scan.results and scan.results.errors:
                         for error in scan.results.errors:
                             # Prefix with `\n` since we are in the middle of a progress bar
                             display_error(f"\n{error.description}")
                     scans.append(scan)
+                bar.update(len(scan_collection.scans_with_results))
```
agateau-gg, line 182: Regarding the progress bar not moving at startup, maybe the `for` loop above takes a long time to execute? How many items are in `scan_collection.scans_with_results`?

I think you can do two changes here which may help:
- Move the `bar.update()` call before the `for` loop
- Remove the `scans.append(scan)` inside the loop with a `scans.extend(scan_collection.scans_with_results)` *outside* the loop

pierrelalanne: Unfortunately I tested both approaches you suggest. I also had a look into using tqdm, I had the same problem.  
I will try to find another solution in the meantime.

ggshield/scan/scannable.py
```diff
@@ -321,12 +324,18 @@ def _scan_chunk(
...
+        files: Iterable[File],
+        progress_callback: Callable[..., None],
+        progress_advance: Optional[int] = None,
     ) -> Tuple[Dict[Future, List[File]], List[File]]:
         """
         Start all scans, return a tuple containing:
         - a mapping of future to the list of files it is scanning
         - a list of files which we did not send to scan because we could not decode them
+
+        on_file
```
agateau-gg, line 338: typo, I guess :)

ggshield/scan/scannable.py
```diff
@@ -400,21 +415,22 @@ def _collect_results(
...
         ) as executor:
-            chunks_for_futures, skipped_chunk = self._start_scans(executor, files)
-
-            if skipped_chunk:
-                # Report skipped_chunk as progress. If we only report files really sent
-                # for scanning then the progress bar won't reach 100%.
-                on_file_chunk_scanned(skipped_chunk)
-
-            return self._collect_results(chunks_for_futures, on_file_chunk_scanned)
+            chunks_for_futures, _ = self._start_scans(
```
agateau-gg, line 426: Since we don't need the `skipped_chunk` part of the returned tuple, `_start_scans()` can be simplified to only return `chunks_for_futures`.

ggshield/cmd/secret/scan/archive.py
```diff
@@ -38,12 +46,17 @@ def archive_cmd(ctx: click.Context, path: str) -> int:  # pragma: no cover
...
-
-            def update_progress(chunk: List[File]) -> None:
-                progressbar.update(len(chunk))
+        with Progress(
+            TextColumn("[progress.description]{task.description}"),
+            BarColumn(),
+            TaskProgressColumn(),
+            TextColumn("{task.completed} files scanned out of {task.total}"),
+            TimeRemainingColumn(),
+            console=Console(file=sys.stderr),
```
agateau-gg, line 55: This code is duplicated with a few variants in multiple places. Can you add a function to create a progress bar to `text_utils.py`? I am trying to group all output-related code there.

agateau-gg (changes requested):
The switch to `rich` looks promising. Just a few remarks and then it will be great!

ggshield/scan/scannable.py
pierrelalanne: Actually, I wanted to keep it as we may want to be more verbose about these `skipped_chunks` in the future. WDYT ?

pierrelalanne:
I integrated your remarks, thanks for the review.

ggshield/scan/scannable.py
agateau-gg: If we decide to do something about it, it will be easy to re-add it. I would remove it to simplify the code.

ggshield/core/text_utils.py
```diff
@@ -132,3 +141,15 @@ def translate_validity(validity_id: Optional[str]) -> str:
     # If we don't have a text for the validity_id, return it as is. We assume the text
     # of the ID is more valuable than a generic "Unknown" string
     return _VALIDITY_TEXT_FOR_ID.get(validity_id, validity_id)
+
+
+def create_progress_bar(type_doc: str, **kwargs: Any) -> Progress:
```
agateau-gg, line 146: I would rename `type_doc`to `doc_type` 🙃.

ggshield/scan/repo.py
```diff
@@ -110,7 +112,7 @@ def scan_commits_content(
 
 
 def get_commits_by_batch(
-    commit_list: Iterable[Commit],
+    commit_list: List[Commit],
```
agateau-gg, line 115: Out of curiosity: do we need this change? Taking an Iterable is nicer.

ggshield/scan/repo.py
```diff
@@ -152,33 +154,39 @@ def scan_commit_range(
     :param verbose: Display successful scan's message
     """
 
-    commits_batch = get_commits_by_batch(
-        commit_list=(
-            Commit(sha=sha, exclusion_regexes=exclusion_regexes) for sha in commit_list
+    with create_progress_bar(type_doc="commits", transient=True) as progress:
```
agateau-gg, line 157: Why do we need `transient` here and not on the others?

agateau-gg (changes requested):
A few questions, looks promising.

ggshield/scan/repo.py
pierrelalanne: No we don't, this was an unwanted change left after my experiments.

ggshield/scan/repo.py
pierrelalanne: Oh, my bad ! This came from the fact that the output handler was used in the context. The progress bar would be immediately "re-rendered" after the output was generated, resulting in two progress bars in the end.  
But by moving the output handler outside the context, everything is fine without the transient option.

pierrelalanne:
Thanks for the thorough review.

agateau-gg (approved):
Looks great!

PR #356, opened 2022-09-13
Modified value of multi-threads for ggshield

pierrelalanne:
Same as https://github.com/GitGuardian/ggshield/pull/348 with slight improvements.

### DIFF
--- a/ggshield/cmd/secret/scan/path.py
+++ b/ggshield/cmd/secret/scan/path.py
@@ -2,6 +2,7 @@ from typing import List
 
 import click
 
+from ggshield.core.constants import MAX_WORKERS
 from ggshield.core.file_utils import get_files_from_paths
 from ggshield.core.utils import ScanContext, ScanMode, handle_exception
 from ggshield.output import OutputHandler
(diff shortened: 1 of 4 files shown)


ggshield/cmd/secret/scan/path.py
```diff
@@ -2,12 +2,17 @@
...
 
+from ggshield.core.constants import CPU_COUNT
 from ggshield.core.file_utils import get_files_from_paths
 from ggshield.core.utils import ScanContext, ScanMode, handle_exception
 from ggshield.output import OutputHandler
 from ggshield.scan import ScanCollection
 
 
+# We add a maximal value to avoid silently consuming all threads on powerful machines
+SCAN_CMD_THREADS = min(CPU_COUNT, 32)
```
agateau-gg, line 13: I think it would be good to have the `min` logic centralized in `constants.py`, with an environment variable to override the 32. Something like this:

```python
def _get_max_workers() -> int:
    # We add a maximal value to avoid silently consuming all threads on powerful machines
    return int(os.getenv("GG_MAX_WORKERS", "32"))

MAX_WORKERS = min(CPU_COUNT, _get_max_workers())
```

Then `SCAN_CMD_THREADS` here and `COMMIT_THREADS` in repo.py can be defined as `MAX_WORKERS`.

agateau-gg (changes requested):
Minor remark, looks good otherwise.

agateau-gg (approved):

PR #358, opened 2022-09-16
Fix .env loading, and log the loaded path

agateau-gg:
This PR fixes loading .env files: ggshield was loading the parent .env file instead of the one in the current directory.

It also adds a debug log to report the path to the loaded .env file, which can be useful when debugging setup issues. To do this I had to move the load of the .env after the log setup, otherwise the path was not printed.

### DIFF
--- a/ggshield/core/utils.py
+++ b/ggshield/core/utils.py
@@ -1,10 +1,11 @@
+import logging
 import os
 import re
 import traceback
 import uuid
 from dataclasses import dataclass
 from enum import Enum
-from typing import Iterable, List, NamedTuple, Union
+from typing import Iterable, List, NamedTuple, Optional, Union
 from urllib.parse import ParseResult, urlparse
 
 import click
@@ -15,6 +16,9 @@ from ggshield.core.constants import ON_PREMISE_API_URL_PATH_PREFIX
 
 from .git_shell import get_git_root, is_git_dir
 from .text_utils import Line, LineCategory, display_error, display_warning
+
+
+logger = logging.getLogger()
 
 
 REGEX_PATCH_HEADER = re.compile(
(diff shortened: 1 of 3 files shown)


tests/core/test_utils.py
```diff
@@ -249,3 +251,86 @@ def test_unexpected_path_dashboard_url(self, dashboard_url):
...
+
+
+@patch("ggshield.core.utils.load_dotenv")
+@pytest.mark.parametrize(
+    ["cwd", "env", "expected_dotenv_path"],
+    [
+        ("sub", {}, None),
+        (".", {}, ".env"),
+        (".", {"GITGUARDIAN_DONT_LOAD_ENV": "1"}, None),
+        (".", {"GITGUARDIAN_DOTENV_PATH": ".custom-env"}, ".custom-env"),
```
pierrelalanne, line 263: We could add a test for `"GITGUARDIAN_DOTENV_PATH": "dont_exist.txt"` that should raise an exception ?

ggshield/core/utils.py
```diff
@@ -270,7 +270,7 @@ def load_dot_env() -> None:
                 "GITGUARDIAN_DOTENV_LOCATION does not point to a valid .env file"
             )
 
-    cwd_env = os.path.join("..", ".env")
+    cwd_env = ".env"
```
pierrelalanne: Open question : should we add a more detailed docstring, or do you think the code is self-explanatory ?

tests/core/test_utils.py
```diff
@@ -249,3 +251,86 @@ def test_unexpected_path_dashboard_url(self, dashboard_url):
...
+    THEN the .env file at the root of the git repository is loaded
+    """
+    is_git_dir_mock.return_value = True
+    get_git_root_mock.return_value = str(tmp_path)
+
+    sub_dir = tmp_path / "sub1" / "sub2"
+    git_root_dotenv = tmp_path / ".env"
+
+    sub_dir.mkdir(parents=True)
+    (tmp_path / "sub1" / "sub2").touch()
```
pierrelalanne, line 326: Shouldn't this be `(tmp_path / "sub1" / ".env").touch()` to reflect the hierarchy you describe in the docstring ?

pierrelalanne (changes requested):

tests/core/test_utils.py
agateau-gg: Oups, good catch!

tests/core/test_utils.py
agateau-gg: Good idea, going to add it. But the current behavior does not raise an exception: it prints an error and continues on its way.

I added a comment saying maybe we should not load any .env file found after that. That would change the existing behavior, but only for people who would have a GITGUARDIAN_DOTENV_PATH env var not pointing to an existing file. What do you think?

ggshield/core/utils.py
agateau-gg: I think it's self-explanatory, but I wrote it so I am not the right person to ask :)

Going to add something like `# look for a .env in the current directory`.

tests/core/test_utils.py
agateau-gg: Just went ahead and did it. I think it is what was intended after all.

pierrelalanne (approved):
LGTM. Not a big deal, but in order to have a clean git history, I think we should squash the two commits, there are some "fixes" following my comments on the first commit included in the second one.  
Your choice though 🙂

agateau-gg:
Squashed!

PR #363, opened 2022-09-26
fix: Fail when the server cannot be accessed

Walz:
If you run `ggshield` without an internet connection, it will print errors but not fail. Within a CI if the server is not accessible, the CI will not fail.

This PR raises an exception when the status code is missing, meaning the server did not respond.


agateau-gg (approved):
Makes sense

PR #368, opened 2022-09-30
Add functional test suite

### LINKED ISSUE #350: Add functional tests
## Description

ggshield has good unit-test coverage, but it lacks functional tests: tests which test the whole product, end-to-end, performing real scans with real API requests.

This can be implemented using pytest by making it run the ggshield command.

These tests will be slow. It's OK. To ensure this does not make it painful to run other tests while developing we can keep them in a separate directory.

There is a work-in-progress branch implementing these tests in <https://github.com/agateau-gg/ggshield/tree/agateau/functests>. Tests are in the `functests` directory. They can be run with `pytest functest`.

The goal of this issue is to integrate this branch.

## Tasks

- [ ] Measure coverage, add missing commands and/or create issues to add them.
- [ ] Make CI run functional tests. We can run them on all PRs for now, but we may have to revisit this later if they start to take too long.
- [ ] Document the functional tests in CONTRIBUTE.md.

agateau-gg:
## Description

This PR adds functional tests in addition to unit tests.

For now the tests are run for each PR. It does not add too much time to the build. We can always change this later if it causes problems.

## What has been done

- Move existing unit tests to `tests/unit`
- Add functional tests in `tests/functional`
- Update `CONTRIBUTE.md`, CI scripts and `Makefile` accordingly

## Review tips

The total number of changed files is big, but fear not! the changes in `tests/unit` are just import adjustments. The files to really review are in `tests/functional`.

## Issue

Fixes #350


tests/functional/secret/test_scan_docker.py
```diff
@@ -0,0 +1,52 @@
...
+    "image_name",
+    (
+        "docker-leaking-in-env",
+        "docker-leaking-in-layer",
+    ),
+)
+def test_scan_docker_find_secret(tmp_path: Path, image_name: str) -> None:
+    build_image(tmp_path, image_name)
+    proc = run_ggshield_scan("docker", image_name, expected_code=1)
+    print(proc.stdout)
```
pierrelalanne, line 51: Don't we want to assert that `ggtt-v-azerty1234` is part of stdout ?

tests/functional/data/docker-leaking-in-layer/foo.conf
```diff
@@ -0,0 +1 @@
+token="ggtt-v-azerty1234"
```
pierrelalanne, line 1: This may be overkill, but could we generate this file from `GG_VALID_TOKEN` in  `tests/conftest.py` ?

pierrelalanne (commented):
Minor remarks, looks cool !

tests/functional/data/docker-leaking-in-layer/foo.conf
pierrelalanne: It would avoid having multiple tokens defined in many places.

agateau-gg: Good idea, doing it.

pierrelalanne (approved):
👍

Issue #336, opened 2022-08-24
Fix tests/cmd/iac/test_scan.py tests using modified cassette output


agateau-gg:
## Description

As found out in #302, 3 tests use a modified cassette output to simulate errors:
- tests/cmd/iac/test_scan.py::test_iac_scan_error_response
- tests/cmd/iac/test_scan.py::test_iac_scan_json_error_response
- tests/cmd/iac/test_scan.py::test_iac_scan_unknown_error_response

Keeping cassettes unaltered makes it possible to detect when the test expectations diverge from the real server behavior by regularly running tests without cassettes.

If the code cannot be tested without altering the cassettes then the tests should use a different way of simulating the wanted behavior.

## TODO

Rewrite these tests so that they do not rely on modified cassettes, either by simulating the error with a real server call or by mocking at a different level, for example using MockerFixture. This is what [tests/cmd/iac/test_scan.py::test_iac_scan_error_response_read_timeout](https://github.com/GitGuardian/ggshield/blob/802de39b9c8e0183948ef3313075ef693ed150b7/tests/cmd/iac/test_scan.py#L127) does.

agateau-gg:
Fixed by #370

PR #370, opened 2022-10-05
Fix issue #336 : Rewriting tests without altered cassette

Kariss83:
I created two return mocks that are called and returns errors using the error that was previously in the casette output.
I've decided to put the mocks in `tests/conftest.py`, not sure that is the best place, let me know if something is more suitable.


tests/conftest.py
```diff
@@ -610,3 +610,29 @@ def assert_invoke_exited_with(result: Result, exit_code: int):
...
+
+def mocked_requests_scan_error(*args, **kwargs):
+    class MockResponse:
+        def __init__(self, *args, **kwargs):
+            self.headers = {"content-type": "application/json"}
+            self.status_code = 404
+            self.json_data = {"detail": "Not found (404)"}
+
+        def json(self):
+            return self.json_data
```
agateau-gg, line 636: The two `mocked_requests_*` functions are quite similar. Maybe you can factorize these by creating a customizable class to mock requests responses? Something like this:

```python
class MockRequestsResponse:
    def __init__(self, status_code, json_data):
        self.headers = {"content-type": "application/json"}
        self.status_code = status_code
        self.json_data = json_data

        def json(self):
            return self.json_data
```

Then the `test_iac_scan*` functions can be changed like this:

```python
    mocker.patch(
        "ggshield.core.client.IaCGGClient.request",
        return_value=MockRequestsResponse(404, <the_json_dict_here>)
    )
```

Note: I used `return_value` instead of `side_effect` in the call to `mocker.patch()` because we are mocking the return value of the method call here, not a raised exception.

agateau-gg (changes requested):
Thanks for looking into this :+1:. Made one remark but looks good overall.

tests/conftest.py
Kariss83: Thx for the advices !!

agateau-gg (approved):
Looks good now, thanks a lot!

agateau-gg:
Can you rebase your branch on top of `main` to solve the conflicts? Once it's done I can merge this PR.

Kariss83:
@agateau-gg I rebased and everything seems good on my side. It seems that all the checks fail due to functional tests but not mine. Let me know if I missed something.

agateau-gg:
Yes, that's because the PR comes from a fork and so it does not have access to the required credentials to run the functional tests. We need to look into this.

Your code looks good, merging it. Thanks again!

PR #371, opened 2022-10-05
feat: Check for updates and display a warning when an update is available 

### LINKED ISSUE #299: Look for updates and alert the user when available
**Is your feature request related to a problem? Please describe.**
Users are not informed when a new version is released, unless they watch the repository on GitHub.

**Describe the solution you'd like**
An optional feature to look for the version number of the latest version and print a warning when a newest version is available when ggshield is executed.

**Describe alternatives you've considered**
A new command to look for update, it would be less intrusive but only alert users who go out of their way to use it.

Walz:
When the command is called, look for updates by requesting the latest release on GitHub. Save the latest version in a local file to only query GitHub once per day.

image

closes #299

### DIFF
--- a/ggshield/cmd/main.py
+++ b/ggshield/cmd/main.py
@@ -17,6 +17,7 @@ from ggshield.cmd.secret import secret_group
 from ggshield.cmd.secret.ignore import deprecated_ignore_cmd
 from ggshield.cmd.secret.scan import scan_group
 from ggshield.cmd.status import status_cmd
+from ggshield.core import check_updates
 from ggshield.core.cache import Cache
 from ggshield.core.config import Config
 from ggshield.core.text_utils import display_warning
@@ -102,6 +103,12 @@ def setup_debug_logs(debug: bool) -> None:
     help="Ignore ssl verification.",
 )
 @click.option("--debug", is_flag=True, default=None, help="Show debug information.")
+@click.option(
+    "--check-for-updates/--no-check-for-updates",
+    is_flag=True,
+    default=True,
+    help="Check for ggshield updates.",
+)
 @click.version_option()
 @click.pass_context
 def cli(
@@ -110,6 +117,7 @@ def cli(
     verbose: bool,
     allow_self_signed: bool,
     debug: Optional[bool],
+    check_for_updates: bool,
 ) -> None:
     ctx.ensure_object(dict)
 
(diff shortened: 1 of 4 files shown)


tests/unit/core/test_utils.py
```diff
@@ -330,3 +337,67 @@ def test_load_dot_env_loads_git_root_env(
...
+    ),
+)
+def test_check_for_update(
+    request_get_mock: Mock,
+    current_version: str,
+    remote_version: str,
+    print_warning: bool,
+    capsys: CaptureFixture[str],
+    monkeypatch: MonkeyPatch,
+):
```
agateau-gg, line 362: You can use the `fs` fixture from [pyfakefs](https://pypi.org/project/pyfakefs/) in your test to ensure the test never touches the real file system. This avoids side-effects and removes the need for the optional `cache` argument in `check_for_update()`.

ggshield/core/utils.py
```diff
@@ -388,3 +393,53 @@ class ScanContext:
...
+    if (
+        latest_version is None
+        or check_at < 0
+        or (time.time() - check_at > 24 * 60 * 60)
+    ):
+        logger.debug("Checking the latest released version of ggshield...")
+        resp = requests.get(
+            "https://api.github.com/repos/GitGuardian/GGShield/releases/latest"
+        )
+        if resp.status_code == 200:
```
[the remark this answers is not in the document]
Walz: Even if the request to GitHub API fails, we may still return a warning if a new version has been cached.

agateau-gg (approved):
Looks good but... CI is not happy 😢 

I suspect the failures come from the fact that we execute the unit-test test-suite with `--disable-socket`, and since a lot of tests go through `cli()` we end up checking for updates a lot :).

This is, I think, a rare case where it makes sense for the main code to alter its behavior depending on whether it's being run by pytest. I think you can fix this by modifying `main.cli()` to not check for updates if the `PYTEST_CURRENT_TEST` environment variable is set. This looks a nice way to detect if the code is running under pytest without importing pytest in the code itself.

ggshield/cmd/main.py
```diff
@@ -140,6 +148,11 @@ def cli(
     logger.debug("args=%s", sys.argv)
     logger.debug("py-gitguardian=%s", pygitguardian.__version__)
 
+    if check_for_updates and "PYTEST_CURRENT_TEST" not in os.environ:
```
agateau-gg, line 151: Can you add a comment explaining why we added the `PYTEST_CURRENT_TEST` check? Should help future-us being less puzzled about this code.

ggshield/core/check_updates.py
```diff
@@ -0,0 +1,66 @@
...
+                {"latest_version": latest_version, "check_at": check_at}, CACHE_FILE
+            )
+
+    if latest_version is not None:
+        current_version_split = tuple(map(int, __version__.split(".")))
+        latest_version_split = tuple(map(int, latest_version.split(".")))
+        if current_version_split < latest_version_split:
+            return (
+                f"A new version of ggshield (v{latest_version}) has been released "
+                f"(https://github.com/GitGuardian/ggshield)."
```
agateau-gg, line 64: That's not exactly what I had in mind with "separation of concerns": the idea is this function does not need to concern itself with how the information is going to be presented to the user: it should return either None (no update available), or the version number of the available version. Then it's up to the caller to present the information, which in our case means including the version number in a text message.

agateau-gg (changes requested):
Minor remarks. Looks good otherwise.

agateau-gg (approved):
Looks good, thanks!

PR #378, opened 2022-10-11
Modify iac scan mode name for analytics

amascia-gg:
The modes for the IAC API analytics will be `directory` and `others` so we want to align with that.

### DIFF
--- a/ggshield/cmd/iac/scan.py
+++ b/ggshield/cmd/iac/scan.py
@@ -10,7 +10,7 @@ from ggshield.core.config import Config
 from ggshield.core.extra_headers import get_headers
 from ggshield.core.filter import init_exclusion_regexes
 from ggshield.core.text_utils import display_error
-from ggshield.core.utils import ScanContext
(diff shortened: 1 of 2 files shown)


ggshield/cmd/iac/scan.py
```diff
@@ -157,7 +157,7 @@ def iac_scan(ctx: click.Context, directory: Path) -> Optional[IaCScanResult]:
         get_headers(
             scan_context=ScanContext(
                 command_path=ctx.command_path,
-                scan_mode="external",
+                scan_mode="directory",
```
agateau-gg, line 160: We have started to add an enum for the different `scan_mode` values (see core/utils.py, ScanMode enum). Can you add an `IAC_DIRECTORY = "directory"` entry to it and use it there? (the other enums should be prefixed with `SECRET_` but they are not for now).

amascia-gg: Done!

agateau-gg (approved):

PR #379, opened 2022-10-11
Add missing packages to "install_requires"

agateau-gg:
## Description

While working on fixing Homebrew taps update, I found out our .whl would not install all required dependencies 🙀.

## What has been done
- Add the dependencies.
- Rework the `functest` Makefile target to ensure this can't go in unnoticed in the future.


pierrelalanne:
Do you think we should also add a comment in the `Pipfile` or in the `setup.py` to warn contributors to "sync" the deps ?

agateau-gg:
> Do you think we should also add a comment in the `Pipfile` or in the `setup.py` to warn contributors to "sync" the deps ?

Good idea, just did it.

PR #386, opened 2022-10-13
Do not fail when `auth login` is called with a lifetime argument

agateau-gg:
- Factorize code creating the AccountConfig instance from auth/login.py and oauth.py in a new method: InstanceConfig.init_account()

- Fix parsing of the optional "expire_at" value we receive from the API: we receive it as a string, so it must be converted into a `datetime`


martinfaucheux (approved):
I don't understand how come it's been working since I made this

agateau-gg:
> I don't understand how come it's been working since I made this

:)

I guess it's because by default the lifetime is None.

PR #387, opened 2022-10-17
Do not crash if pushed branch contains no new commits

agateau-gg:
In this case the call to `git rev-list...` returns no commits, causing
ggshield to crash because `ancestors` is empty.

### DIFF
--- a/tests/unit/cmd/scan/test_prereceive.py
+++ b/tests/unit/cmd/scan/test_prereceive.py
@@ -289,6 +289,35 @@ class TestPreReceive:
             ignored_detectors=set(),
         )
 
+    @patch("ggshield.cmd.secret.scan.prereceive.scan_commit_range")
+    def test_new_branch_without_commits(
+        self,
+        scan_commit_range_mock: Mock,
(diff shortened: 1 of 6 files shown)


pierrelalanne:
Looks good, but I think we will hit the same issue with pre-push ?

tests/unit/cmd/scan/test_prereceive.py
```diff
@@ -289,6 +289,34 @@ def test_new_branch(
...
+
+        with cd(str(repo.path)):
+            result = cli_fs_runner.invoke(
+                cli,
+                ["-v", "secret", "scan", "pre-receive"],
+                input=f"{EMPTY_SHA} {sha} refs/heads/{branch_name}",
+            )
+
+        assert_invoke_ok(result)
+        scan_commit_range_mock.assert_not_called()
```
pierrelalanne, line 318: We could assert something on the output as for the pre-push.

pierrelalanne (approved):
One small remark. Ok for me otherwise.

PR #388, opened 2022-10-17
docs(readme): update missing args in pre-push commands

pierrelalanne:
Fix missing args in the Readme.


agateau-gg (approved):
Good catch!

PR #393, opened 2022-10-18
feat: Add support for the docsets JSONL input

er-vin:
Added both functional and unit tests, they all pass. Also did some manual testing which seems to find secrets just fine.


ggshield/cmd/secret/scan/docset.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+        scans.append(ScanCollection(id=path, type="docset", results=results))
+        progress_callback()
+
+    return scans
+
+
+@click.command()
+@click.argument(
+    "paths", nargs=-1, type=click.Path(exists=True, resolve_path=True), required=True
```
agateau-gg, line 49: If I am not mistaken, this does not accept data from stdin, does it?

Not a blocker, but it would be nice to have.

ggshield/cmd/secret/scan/docset.py
```diff
@@ -0,0 +1,86 @@
...
+)
+@click.pass_context
+def docset_cmd(ctx: click.Context, paths: List[str]) -> int:  # pragma: no cover
+    """
+    scan docset JSONL files.
```
agateau-gg, line 54: At some point we will have to publish a spec of the JSONL format.

ggshield/cmd/secret/scan/docset.py
```diff
@@ -0,0 +1,86 @@
...
+
+    return scans
+
+
+@click.command()
+@click.argument(
+    "paths", nargs=-1, type=click.Path(exists=True, resolve_path=True), required=True
+)
+@click.pass_context
+def docset_cmd(ctx: click.Context, paths: List[str]) -> int:  # pragma: no cover
```
pierrelalanne, line 52: Sorry to jump in the discussion. When testing locally with a jsonl input, the progress bar does not behave correctly :(   

Also, and it is linked, I am wondering why we are using the `Scanner.scan` interface instead of the `Files.scan` interface as in other commands implementation (see pypi for instance).

ggshield/cmd/secret/scan/docset.py
```diff
@@ -0,0 +1,86 @@
...
+    config = ctx.obj["config"]
+    output_handler: OutputHandler = ctx.obj["output_handler"]
+    try:
+        with create_progress_bar(doc_type="files") as progress:
+            task_scan = progress.add_task(
+                "[green]Scanning content...", total=len(paths)
+            )
+
+            scan_context = ScanContext(
+                scan_mode=ScanMode.PATH,
```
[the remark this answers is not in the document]
agateau-gg: I think it would be a good idea, yes.

ggshield/cmd/secret/scan/docset.py
agateau-gg: Regarding `Scanner.scan()`, I think it's OK: when #344 is done, `Files.scan()` will be gone and all commands should do what this one does.

ggshield/cmd/secret/scan/docset.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def generate_files_from_docsets(path: str, verbose: bool = False) -> Iterator[File]:
+    with open(path, "r") as file:
+        for line in file:
+            obj = json.loads(line)
+            documents = obj["documents"]
+            for document in documents:
+                if verbose:
+                    click.echo(f"  * {document['id']}", err=True)
```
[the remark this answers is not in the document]
er-vin: done

ggshield/cmd/secret/scan/docset.py
er-vin: done

ggshield/cmd/secret/scan/docset.py
er-vin: Indeed, maybe worth waiting until the doc stabilizes for sure?

ggshield/cmd/secret/scan/docset.py
er-vin: The progress bar should be fixed now. It advances per input file though, not per line (as computing the total number of line before hand would be expensive).

ggshield/cmd/secret/scan/docset.py
agateau-gg: Yes, makes sense.

ggshield/cmd/secret/scan/docset.py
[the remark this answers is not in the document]
er-vin: done

Issue #377, opened 2022-10-11
Make core.config.utils.load_yaml check the type of the loaded data


agateau-gg:
## Environment

- ggshield version: 1.13.3

## Describe the bug

When a .gitguardian.yml file contains a valid YAML document, but this document is not a dictionary, the user gets a puzzling error message.

Steps to reproduce:

1. Create a .gitguardian.yml file containing `Hello` (A valid YAML document containing a *string*)
2. Run any ggshield command, for example `ggshield quota`

Actual result:

ggshield fails with an error like this:

```
$ ggshield quota
Traceback (most recent call last):
  File "/home/agateau/.local/bin/ggshield", line 8, in <module>
    sys.exit(main())
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/ggshield/cmd/main.py", line 160, in main
    return cli.main(args, prog_name="ggshield", standalone_mode=not show_crash_log)
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/click/core.py", line 1656, in invoke
    super().invoke(ctx)
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/ggshield/cmd/main.py", line 121, in cli
    config = Config(config_path)
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/ggshield/core/config/config.py", line 40, in __init__
    self.user_config, self._config_path = UserConfig.load(config_path=config_path)
  File "/home/agateau/.local/pipx/venvs/ggshield/lib/python3.8/site-packages/ggshield/core/config/user_config.py", line 144, in load
    user_config._update_from_file(local_config_path)
[...]

PR #398, opened 2022-10-20
Make core.config.utils.load_yaml check the type of the loaded data

alaabenfatma:
## :notebook: Intro
This PR fixes the problem we have when the `.gitguardian.yml` file contains a string (or anything else that is not a valid dict), we get a non-explanatory error message.

Moreover, this PR also cleans and refactors some parts of the code by:
- Removing unused imports
- Removing unused variables
- Removing unused arguments

It also changes the tests to make them compliant with the new changes.

## :red_circle: Target issue
https://github.com/GitGuardian/ggshield/issues/377

## :computer: New output
If the `configuration` is an invalid dict, we get the following output:
`The configuration in '.gitguardian.yml' is invalid. Please verify it.`

### DIFF
--- a/ggshield/core/check_updates.py
+++ b/ggshield/core/check_updates.py
@@ -8,7 +8,7 @@ import requests
 from ggshield import __version__
 from ggshield.core.dirs import get_cache_dir
 
-from .config.utils import load_yaml, save_yaml
+from .config.utils import load_yaml_dict, save_yaml_dict
 
 
 logger = logging.getLogger(__name__)
@@ -33,7 +33,11 @@ def check_for_updates() -> Optional[str]:
     """
     check_at = -1.0
     # Load the last time we checked
-    cached_data = load_yaml(CACHE_FILE)
+    try:
+        cached_data = load_yaml_dict(CACHE_FILE)
+    except ValueError:
+        # Swallow the error
+        cached_data = None
     if cached_data is not None:
         try:
             check_at = cached_data["check_at"]
@@ -49,7 +53,7 @@ def check_for_updates() -> Optional[str]:
     # Save check time now so that it is saved even if the check fails. This ensures we
     # don't try for every command if the user does not have network access.
     try:
-        save_yaml({"check_at": time.time()}, CACHE_FILE)
+        save_yaml_dict({"check_at": time.time()}, CACHE_FILE)
     except Exception as e:
         logger.error("Could not save time of version check to cache: %s", e)
         # Do not continue if we can't save check time. If we continue we are going to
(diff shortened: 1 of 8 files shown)


agateau-gg (changes requested):
This is better, but it leads to some inconsistent behavior because `load_yaml()` swallows some errors and let the app continue:

```
$ echo 'a_valid_yaml_string' > .gitguardian.yaml

$ ggshield api-status
Error: The configuration in './.gitguardian.yaml' is invalid. Please verify it.

$ echo 'invalid_yaml: [' > .gitguardian.yaml 

$ ggshield api-status                       
Parsing error while reading ./.gitguardian.yaml:
while parsing a flow node
expected the node content, but found '<stream end>'
  in "./.gitguardian.yaml", line 2, column 1
Status: healthy
App version: v2.16.2
Secrets engine version: 2.77.0
```

I think the app should stop with an error in both cases.

It's a bit more work, but the code would be saner if we made `load_yaml()` raise an exception when parsing fails or when the decoded content is not a string, then catch it in `_update_from_file()` like you did. And since `load_yaml()` and `save_yaml()` are really about dict, we should rename them to `load_yaml_dict()` and `save_yaml_dict()`.

Therefore can you made the following changes?
- Rename `load_yaml()` to `load_yaml_dict()` and `save_yaml()` to `save_yaml_dict()`
- Make `load_yaml_dict()` raise a `ValueError` if the content is invalid YAML or is not a dict
- Handle the exception wherever `load_yaml_dict()` is called. In most case this means rethrowing it as a ClickException, but in the case of `check_for_updates()` we must just swallow the error.

ggshield/core/check_updates.py
```diff
@@ -33,7 +33,11 @@ def check_for_updates() -> Optional[str]:
     """
     check_at = -1.0
     # Load the last time we checked
-    cached_data = load_yaml(CACHE_FILE)
+    try:
+        cached_data = load_yaml_dict(CACHE_FILE)
+    except ValueError:
+        # Swallow the error
+        pass
```
agateau-gg, line 40: I think you need `cached_data = None` here, otherwise code will fail on next line saying `cached_data` is not defined.

agateau-gg (changes requested):
One minor remark and I think this is good to go!

ggshield/core/check_updates.py
alaabenfatma: Very good remark @agateau-gg, I missed it. Sorry!

agateau-gg (approved):

agateau-gg:
Looks good, merging!

PR #407, opened 2022-10-27
rework outputs of pre-commit, pre-push and pre-receive

alina-tuholukova-gg:
pre-commit
Screenshot from 2022-10-28 10-46-22

pre-push
Screenshot from 2022-10-28 11-13-40

pre-receive
Screenshot from 2022-10-28 11-14-23

### DIFF
--- a/ggshield/cmd/secret/scan/precommit.py
+++ b/ggshield/cmd/secret/scan/precommit.py
@@ -5,7 +5,21 @@ import click
 from ggshield.core.git_shell import check_git_dir
 from ggshield.core.utils import handle_exception
 from ggshield.output import TextOutputHandler
+from ggshield.output.text.message import remediation_message
 from ggshield.scan import Commit, ScanCollection, ScanContext, ScanMode, SecretScanner
+
+
+REMEDIATION_STEPS = """  Since the secret was detected before the commit was made:
+  1. replace the secret with its reference (e.g. environment variable).
+  2. commit again."""
+
+BYPASS_MESSAGE = """  - if you use the pre-commit framework:
+
+     SKIP=ggshield git commit -m "<your message>"
+
+  - otherwise (warning: the following command bypasses all pre-commit hooks):
+
+     git commit -m "<your message>" --no-verify"""
 
 
 @click.command()
@@ -39,8 +53,18 @@ def precommit_cmd(
         )
         results = scanner.scan(commit.files)
 
-        return output_handler.process_scan(
+        return_code = output_handler.process_scan(
             ScanCollection(id="cached", type="pre-commit", results=results)
         )
+        if return_code:
+            click.echo(
+                remediation_message(
+                    remediation_steps=REMEDIATION_STEPS,
+                    bypass_message=BYPASS_MESSAGE,
+                    rewrite_git_history=False,
+                ),
+                err=True,
+            )
+        return return_code
     except Exception as error:
         return handle_exception(error, config.verbose)
--- a/ggshield/core/text_utils.py
+++ b/ggshield/core/text_utils.py
@@ -26,13 +26,22 @@ STYLE: Dict[str, Dict[str, Any]] = {
     "no_secret": {"fg": "white"},
     "detector": {"fg": "bright_yellow", "bold": True},
     "policy": {"fg": "cyan", "bold": True},
-    "ignore_sha": {"fg": "cyan"},
     "detector_line_start": {"fg": "cyan"},
     "line_count": {"fg": "white", "dim": True},
     "line_count_secret": {"fg": "yellow"},
     "progress": {"fg": "bright_yellow", "bold": False},
     "warning": {"fg": "yellow"},
+    "incident_validity": {"fg": "bright_yellow", "bold": True},
+    "policy_break_type": {"fg": "bright_yellow", "bold": True},
+    "occurrence_count": {"fg": "bright_yellow", "bold": True},
+    "ignore_sha": {"fg": "bright_yellow", "bold": True},
 }
+
+_DISPLAY_NAME_FOR_POLICY = {"secrets detection": "Secret detected"}
+
+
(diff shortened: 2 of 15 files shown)


ggshield/output/text/message.py
```diff
@@ -24,6 +25,35 @@
...
+
+{}
+{}
+{} [To apply with caution] If you want to bypass ggshield (false positive or other reason), run:
+{}
+    """.format(
+        format_text(">", STYLE["detector_line_start"]),
+        remediation_steps,
+        rewrite_git_history_message,
+        format_text(">", STYLE["detector_line_start"]),
```
agateau-gg, line 52: I would rework this to use f-strings. Something like this:

```python
line_start = format_text(">", STYLE["detector_line_start"])

return f"""{line_start} How to remediate

  {remediation_steps}
  {rewrite_git_history_message}
  {line_start} [To apply with caution] If you want to bypass ggshield (false positive or other reason), run:
  {bypass_message}
"""
```

Also by including the 2-spaces indent in this string, it can be removed from remediation_steps, rewrite_git_history_message and bypass_message.

ggshield/core/text_utils.py
```diff
@@ -32,8 +32,11 @@
     "line_count_secret": {"fg": "yellow"},
     "progress": {"fg": "bright_yellow", "bold": False},
     "warning": {"fg": "yellow"},
+    "incident_value": {"fg": "bright_yellow", "bold": True},
 }
 
+POLICY_TO_DISPLAY_NAME = {"secrets detection": "Secret detected"}
```
agateau-gg, line 38: Instead of exposing a dict, I think it's nicer to provide a function to return the display name, so that callers do not need to know the details. Something like this (see `translate_validity` for an example):

```python
_DISPLAY_NAME_FOR_POLICY = {
    # ...
}

get_display_name_for_policy(policy: str) -> str:
    return _DISPLAY_NAME_FOR_POLICY.get(policy.lower(), policy)
```

I am also a fan of using `<dst>_for_<src>` instead of `<src>_to_<dst>`, because of https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/ (warning, long read!)

ggshield/core/text_utils.py
```diff
@@ -32,8 +32,11 @@
     "line_count_secret": {"fg": "yellow"},
     "progress": {"fg": "bright_yellow", "bold": False},
     "warning": {"fg": "yellow"},
+    "incident_value": {"fg": "bright_yellow", "bold": True},
```
agateau-gg, line 35: Should be `incident_validity` I think.

agateau-gg (changes requested):

ggshield/core/text_utils.py
alina-tuholukova-gg: I was hesitating how to name it, it is used for all values in the incident output (not only for validity). Maybe it could be just "incident"?

ggshield/cmd/secret/scan/precommit.py
```diff
@@ -37,8 +38,25 @@ def precommit_cmd(
...
+        if return_code:
+            remediation_steps = """  Since the secret was detected before the commit was made:
+  1. replace the secret with its reference (e.g. environment variable).
+  2. commit again.
+"""
+            bypass_message = """  - if you use the pre-commit framework:
+      SKIP=ggshield git commit -m "<your message>"
+  - otherwise (warning: the following command bypasses all pre-commit hooks):
+      git commit -m "<your message>" --no-verify
+"""
```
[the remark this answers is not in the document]
alina-tuholukova-gg: I am not sure about module level, since it will be used only at this part of the code. I could do this way maybe?
```
remediation_message(
                    remediation_steps="""  Since the secret was detected before the commit was made:
  1. replace the secret with its reference (e.g. environment variable).
  2. commit again.""",
                    bypass_message="""  - if you use the pre-commit framework:
      SKIP=ggshield git commit -m "<your message>"
  - otherwise (warning: the following command bypasses all pre-commit hooks):
      git commit -m "<your message>" --no-verify
""",
                    rewrite_git_history=False,
                ),
```

ggshield/output/text/message.py
agateau-gg: Ah right, that would mean only the first line would not be indented. Never mind then.

ggshield/core/text_utils.py
agateau-gg: Good point. I think it's better to have "semantic" entries, even if they use the same colors for now, maybe you can create these?
- incident_validity
- policy_break_type
- occurrence_count
- ignore_sha

ggshield/cmd/secret/scan/precommit.py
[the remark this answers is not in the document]
agateau-gg: Using them directly looks worse I think: the first line is even more indented.

Even if it's used in only one place, I think it's OK to create a constant. That helps grouping all messages at the top instead of having them buried in the code. It also makes the code easier to read, IMO.

Having said that, it's not a blocker for me. You can leave it as is if you prefer it this way.

agateau-gg:
Looks good overall, I just a few layout remarks.

## pre-commit

image

1. I think it would look nicer if the `SKIP=ggshield git...` and `git commit...` lines were aligned with the text of the ordered list from the previous section.

2. It would be easier to read if the command had empty lines before and after them (same for the other command)

3. Can you remove these two empty lines?

## pre-receive

image

1. Same as in the pre-commit: it would look nicer IMO if the URL and the command were aligned with the text of the ordered list.

2. Can you remove one of the two empty lines here?

3. Can you insert a new line before the command, and remove the empty line after it?

Let me know if this is not clear/sounds weird to you :)

alina-tuholukova-gg:
**pre-commit**

image

***pre-push**
image

***pre-receive**
image

agateau-gg (approved):
Looks good, thanks!

PR #409, opened 2022-10-31
Rework error messages when we can't load the configuration

agateau-gg:
## Description

This PR slightly reworks error messages when we can't load the configuration. It removes some  `Error: Error in (...)` sequences and some duplication of the filename. The "Invalid YAML" error message is still a bit filled with jargon, but it's good enough for now I think.

It also reduces indentation in `load_yaml_dict()` to make code a bit easier to read.

### Before

Invalid YAML:

```
$ ggshield api-status
Error: Error in ./.gitguardian.yaml:
Parsing error while reading ./.gitguardian.yaml:
while parsing a flow node
expected the node content, but found '<stream end>'
  in "./.gitguardian.yaml", line 2, column 1
```

YAML string:

```
$ ggshield api-status
Error: Error in ./.gitguardian.yaml:
The configuration in './.gitguardian.yaml' is invalid. Configuration must be a dict.
```

### After

Invalid YAML:

```
$ ggshield api-status
Error: ./.gitguardian.yaml is not a valid YAML file:
while parsing a flow node
expected the node content, but found '<stream end>'
  in "./.gitguardian.yaml", line 2, column 1
```

YAML string:

```
$ ggshield api-status
Error: ./.gitguardian.yaml should be a dictionary.
```

### DIFF
--- a/ggshield/core/config/user_config.py
+++ b/ggshield/core/config/user_config.py
@@ -171,9 +171,9 @@ class UserConfig(FilteredConfig):
                 )
         except ValidationError as exc:
             message = format_validation_error(exc)
-            raise ParseError(f"Error in {config_path}:\n{message}") from exc
-        except ValueError as ve:
-            raise ParseError(f"Error in {config_path}:\n{str(ve)}") from ve
+            raise ParseError(message) from exc
+        except ValueError as exc:
+            raise ParseError(str(exc)) from exc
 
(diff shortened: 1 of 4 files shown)


alaabenfatma (approved):
The messages are easier to read and interpret indeed.

The `PR` looks good to me :+1:

PR #410, opened 2022-10-31
feat: add ignore known secrets option

alina-tuholukova-gg:


.gitguardian.yaml
```diff
@@ -30,6 +30,8 @@ secret:
       name: Test secret mentioned in commit 6282ceb9035c1bfcceec5f6372f69063cdd588f7
     - match: e90eaa00d17c03309b660e1879c70b815509422c2d4cf0980b2e1221dffa0381
       name: Test secret mentioned in commit 6282ceb9035c1bfcceec5f6372f69063cdd588f7
+    - match: 6a1d11d5bbbb48f044bedfc2db5749e7fde3ad7e3f6a79601db034478a95f88a
+      name: Test secrets in snap_test_text_output.py
```
agateau-gg, line 34: nitpick: can you put this one with its friend from line 22, and use a similar name?

ggshield/scan/scanner.py
```diff
@@ -68,15 +68,40 @@ def from_exception(exc: Exception) -> "Results":
...
+    ):
+        self.id = id
+        self.type = type
+        self.results = results
+        self.scans = scans
+        self.iac_result = iac_result
+        self.optional_header = optional_header
+        self.extra_info = extra_info
+
+        self.has_new_secrets = self._init_has_new_secrets()
```
agateau-gg, line 98: This one can be a property: it's only used once.

Also, I was surprised by the names of `has_new_secrets` and `has_new_secrets_all_scans`. I expected `has_new_secrets` to be true for scans and sub-scans. Plus `has_new_secrets_all_scans` does not sound English to me (but I could be wrong on that 😉). I suggest renaming them like this:

- `has_new_secrets` → `current_scan_has_new_secrets`
- `has_new_secrets_all_scans` → `has_new_secrets`

ggshield/scan/scanner.py
```diff
@@ -68,15 +68,40 @@ def from_exception(exc: Exception) -> "Results":
...
+        (
+            self.known_secrets_count,
+            self.new_secrets_count,
```
agateau-gg, line 101: Looks like new_secrets_count does not need to be a member: it's only used to set `has_new_secrets_all_scans`. Either turn it into a member variable, or turn `has_new_secrets_all_scans` into a property returning `self.new_secrets_count > 0`.

ggshield/scan/scanner.py
```diff
@@ -91,13 +116,37 @@ def has_iac_result(self) -> bool:
...
+        """
+        if self.results and self.results.results:
+            return not all(
+                policy_break.known_secret
+                for result in self.results.results
+                for policy_break in result.scan.policy_breaks
+            )
+        return False
+
+    def _init_known_new_secrets_count(self) -> Tuple[int, int]:
```
agateau-gg, line 131: nitpick: naming is surprising: I expect a `_init_something(self)` to define `self.something`, not to return it. I would replace the `_init` prefix with something like `_get`, `_compute`, or just remove it.

agateau-gg (changes requested):

ggshield/output/text/text_output_handler.py
```diff
@@ -26,49 +26,79 @@ class TextOutputHandler(OutputHandler):
...
 
         :param result: The result from scanning API
         :param nb_lines: The number of lines to display before and after a secret in the
         patch
         :param show_secrets: Option to show secrets value
+        :param show_only_known_secrets: If True, display only known secrets, and only new secrets otherwise
         :return: The formatted message to display
         """
         result_buf = StringIO()
+        result = deepcopy(result)
```
alina-tuholukova-gg, line 92: I needed to do it, because we change content, policy breaks and matches in this function. It worked before since it was always executing only once. In future we could rework functions such that they do not change what is in the result.

agateau-gg: That's a good point. Can you add it as a comment in the code? It's going to be lost if it's only in the PR.

Also, from what I see we only change policy breaks, no? If that's the case then maybe we can restrict the copy to policy breaks by deepcopying only on line 94?

Pipfile
```diff
@@ -13,7 +13,7 @@ click = ">=8.1,<8.2"
 marshmallow = ">=3.18.0,<3.19.0"
 marshmallow-dataclass = ">=8.5.8,<8.6.0"
 oauthlib = ">=3.2.1,<3.3.0"
-pygitguardian = ">=1.3.6,<1.4.0"
+pygitguardian = {git = "https://github.com/GitGuardian/py-gitguardian.git", editable = true, ref = "alina/add-parameter-to-multiscan"}
```
agateau-gg, line 16: I think you need to change the ref here to unbreak the lint step.

ggshield/scan/scanner.py
```diff
@@ -68,15 +69,40 @@ def from_exception(exc: Exception) -> "Results":
...
+        self.iac_result = iac_result
+        self.optional_header = optional_header
+        self.extra_info = extra_info
+
+        (
+            self.known_secrets_count,
+            self.new_secrets_count,
+        ) = self._get_known_new_secrets_count()
+        self.has_new_secrets = self.new_secrets_count > 0
+        self.has_known_secrets = self.known_secrets_count > 0
```
agateau-gg, line 104: `has_new_secrets` and `has_known_secrets` can be properties.

agateau-gg (changes requested):

Issue #197, opened 2022-04-06
`scan` options should be allowed later in the command


agateau-gg:
**GitGuardian Shield Version**

- I can reproduce this bug in the latest version

**Command executed**

```
ggshield scan path -v foo
```

**Describe the bug**

When the command is executed, ggshield complains like this:

```
Usage: ggshield scan path [OPTIONS] PATHS...
Try 'ggshield scan path -h' for help.

Error: No such option: -v
```

It will only work if one does: ` ggshield scan -v path foo`.

**Expected behavior**

It should be possible to use any `ggshield scan` options (see `ggshield scan --help`) after a `scan` sub-command.

PR #411, opened 2022-11-03
Allow --verbose, --debug and --allow-self-signed everywhere

agateau-gg:
## Description

Click does not allow defining options across groups, this means `ggshield -v secret scan path foo.py` works but not `ggshield secret scan path foo.py -v`.

## What has been done

This PR introduces a new module in `cmd`: `common_options`. This module provides an `add_common_options()` decorator to define Click options we want to make available anywhere on the command-line. Supported options are:

- `-v`, `--verbose`
- `--debug`
- `--allow-self-signed`

With these changes, all these commands now work and do the same thing:

```
ggshield -v secret scan path foo.py
ggshield secret -v scan path foo.py
ggshield secret scan -v path foo.py
ggshield secret scan path -v foo.py
ggshield secret scan path foo.py -v
```

## In the next ~episode~ PR

I have another branch stacked on top of this one to do something similar for `secret scan` options like `--json`, `--exit-zero`, `--show-secrets`...

## Issue

Part of #197

### DIFF
--- a/ggshield/cmd/common_options.py
+++ b/ggshield/cmd/common_options.py
@@ -64,10 +64,28 @@ _debug_option = click.option(
 )
 
 
+def allow_self_signed_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[bool]
+) -> Optional[bool]:
+    if value is not None:
+        _get_config(ctx).allow_self_signed = value
+    return value
+
+
+_allow_self_signed_option = click.option(
+    "--allow-self-signed",
+    is_flag=True,
+    default=None,
+    help="Ignore ssl verification.",
+    callback=allow_self_signed_callback,
+)
+
+
 def add_common_options() -> Callable[[AnyFunction], AnyFunction]:
     def decorator(cmd: AnyFunction) -> AnyFunction:
         _verbose_option(cmd)
         _debug_option(cmd)
+        _allow_self_signed_option(cmd)
         return cmd
 
     return decorator
--- a/ggshield/cmd/common_options.py
+++ b/ggshield/cmd/common_options.py
@@ -1,0 +1,50 @@
+"""
+This module defines options which should be available on all commands, such as the
+-v, --verbose option.
+
+To use it:
+- Add the `@add_common_options()` decorator after all the `click.option()` calls of the
+  command function.
+- Add a `**kwargs: Any` argument to the command function.
+
+The `kwargs` argument is required because due to the way click works,
+`add_common_options()` adds an argument for each option it defines.
+"""
+from typing import Any, Callable, Optional, cast
+
+import click
+
+from ggshield.core.config.user_config import UserConfig
+
+
+AnyFunction = Callable[..., Any]
+
+
+def _get_config(ctx: click.Context) -> UserConfig:
+    return cast(UserConfig, ctx.obj["config"].user_config)
+
+
+def _verbose_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[bool]
+) -> Optional[bool]:
+    if value is not None:
+        _get_config(ctx).verbose = value
+    return value
+
+
+_verbose_option = click.option(
+    "-v",
+    "--verbose",
+    is_flag=True,
+    default=None,
+    help="Verbose display mode.",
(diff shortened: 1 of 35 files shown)


ggshield/cmd/main.py
```diff
@@ -91,43 +114,43 @@ def setup_debug_logs(debug: bool) -> None:
...
-    # creation of the Config instance will be ignored
-    setup_debug_logs(debug is True)
-
     load_dot_env()
 
-    config = Config(config_path)
+    config = ctx.obj["config"]
 
     if debug is not None:
         config.debug = debug
```
pierrelalanne: Shouldn't we handle this in the debug callback ?

ggshield/cmd/main.py
```diff
@@ -136,12 +159,6 @@ def cli(
         # we still have to setup logs
```
pierrelalanne, line 105: This part on the other hand seems already handled in the callback ?

ggshield/cmd/common_options.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def debug_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[bool]
+) -> Optional[bool]:
+    # The --debug option is marked as "is_eager" so that we can setup logs as soon as
+    # possible. If we don't then log commands for the creation of the Config instance
+    # are ignored
+    if value is not None:
+        setup_debug_logs(value is True)
```
pierrelalanne, line 53: In the end, we don't update the context config as it was done L133 in `main.py`, is this normal ?

pierrelalanne (commented):
Looks great. I have two general remarks :  
- We should be careful to add these common_options to future commands when we add them : should we document this ?
- Do you think we should add some kind of functional testing on these options' positions ? I think we should at least write the exhaustive list of commands to QA (which may be super long...)
WDYT ?

ggshield/cmd/common_options.py
agateau-gg: At this point the context config may not be available because `debug_callback()` can be called before `config_callback()` (they are both marked as "is_eager"). This means we can't be sure we can set config.debug.

I could add this back to `main()` though. Going to give it a try.

ggshield/cmd/main.py
agateau-gg: You mean `debug_callback()`? It only handles the `--debug` argument, not `debug: true` being set in the configuration file.

agateau-gg:
> * We should be careful to add these common_options to future commands when we add them : should we document this ?

I agree with the idea, but I don't know a reliable place to document it, any suggestions? On the other hand cargo-culting is likely to prevent issues :)

> * Do you think we should add some kind of functional testing on these options' positions ? I think we should at least write the exhaustive list of commands to QA (which may be super long...)

Testing all possible combinations sounds overkill. Maybe modifying our tests to use different positions in different tests would be good enough?

ggshield/cmd/main.py
pierrelalanne: My comment was not clear. I mean that before the changes, we used to set `config.debug` to the value passed as an option if it was not None. This is not the case in the new version, we only setup_debug_logs but don't set the debug value in the `config`. Is this normal?  
In the end, this comment is the same as [this one](https://github.com/GitGuardian/ggshield/pull/411/commits/bc5fc94000000908ef6717a5dcdf966efe788088#r1016629305).

ggshield/cmd/common_options.py
agateau-gg: Should be good now.

pierrelalanne (approved):

pierrelalanne:
I agree with your second point : 
> Testing all possible combinations sounds overkill. Maybe modifying our tests to use different positions in different tests would be good enough?  

Adding some variations in the functional tests looks wise.

agateau-gg:
> I agree with your second point :
> 
> > Testing all possible combinations sounds overkill. Maybe modifying our tests to use different positions in different tests would be good enough?
> 
> Adding some variations in the functional tests looks wise.

Just did this. Going to merge when CI is happy.

PR #412, opened 2022-11-04
Add a benchmark script

agateau-gg:
## Description

This PR adds a `scripts/perfbench/perfbench` script.

Results show up in the summary of the "Performance benchmark" check.

TODO:

- [x] Decide test repositories
- [x] Decide test commands
- [x] Decide if the benchmark must run on each PR or on main only:
  looks fast enough to run on each PR for now
- [x] Make benchmark report a warning/error if results are worse

### DIFF
--- a/.github/workflows/perfbench.yml
+++ b/.github/workflows/perfbench.yml
@@ -1,0 +1,43 @@
+name: Performance benchmark
+
+on:
+  pull_request:
+  workflow_dispatch:
+    run_options:
+      description: 'Additional options to pass to `run`'
+
+jobs:
+  benchmark:
+    name: Run performance benchmark
+    runs-on: ubuntu-latest
+    env:
+      PYTHONUNBUFFERED: 1
+      # How many time to repeat each run
+      REPEATS: 3
+      # If a run takes more than $MAX_DELTA seconds compared to reference, fail the job
+      MAX_DELTA: 3
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade pipenv
+          pipenv install --system
+
+      - name: Setup benchmark
+        run: |
+          scripts/perfbench/perfbench setup
+
+      - name: Run benchmark
+        env:
+          GITGUARDIAN_API_KEY: ${{ secrets.GITGUARDIAN_API_KEY }}
+        run: |
+          scripts/perfbench/perfbench run --repeats $REPEATS ${{ github.event.inputs.run_options }}
+
+      - name: Generate report
+        run: |
+          scripts/perfbench/perfbench report --max-delta $MAX_DELTA >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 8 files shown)


.github/workflows/perfbench.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Setup benchmark
+        run: |
+          scripts/perfbench/perfbench setup
+
+      - name: Run benchmark
+        env:
+          GITGUARDIAN_API_KEY: ${{ secrets.GITGUARDIAN_API_KEY }}
+        run: |
+          scripts/perfbench/perfbench run --times 3 ${{ github.event.inputs.run_options }}
```
pierrelalanne, line 33: Suggestion : The 3 could be defined as an env var ?
```
env:
     NB_RUNS: 3
```

scripts/perfbench/report_cmd.py
```diff
@@ -0,0 +1,137 @@
...
+from perfbench_utils import (
+    RawReportEntry,
+    find_latest_prod_version,
+    get_raw_report_path,
+    work_dir_option,
+)
+
+
+# Do not show warn or improvement symbol if the delta is less than this percent
+MIN_DELTA_PERCENT = 1
```
pierrelalanne, line 19: Suggestion : should this be an option of the cmd instead ? So we can fine tune it in the github workflow directly ?

scripts/perfbench/report_cmd.py
```diff
@@ -0,0 +1,137 @@
...
+        duration = median(durations_for_version)
+        cell = f"{duration:.3f}"
+
+        if len(durations_for_version) > 1:
+            deviation = stdev(durations_for_version)
+            cell += f" ±{deviation:.3f}"
+
+        if reference is None:
+            reference = duration
+        else:
```
pierrelalanne, line 83: Suggestion : I would advocate for a new column with the delta instead for readability ?

pierrelalanne (commented):
Looks great ! 😍

.github/workflows/perfbench.yml
agateau-gg: What would we gain from this? Environment variables are easier to get wrong than CLI arguments: a typo in a CLI argument is reported by the tool, unlike a typo in an environment variable.

scripts/perfbench/report_cmd.py
agateau-gg: For info: I just changed it to a duration-based delta instead of a percent. Percentages were not useful for the durations we record. It would make more sense with longer operations I think.

I can turn this into an option indeed.

scripts/perfbench/report_cmd.py
agateau-gg: Good idea, just did it.

.github/workflows/perfbench.yml
pierrelalanne: What I mean was just making it explicit that the value 3 is a parameter and can be changed. Defining it as a variable at the top of the github workflow seems rather explicit.

agateau-gg: Ah I see :+1:. Just did it.

PR #418, opened 2022-11-15
Update Pipfile.lock and its generation script

agateau-gg:
This PR does two things:

- Fix a failure in the update-pipfile-lock script
- Actually update the Pipfile.lock


alaabenfatma (approved):
Looks good to me! 
Thanks for the work @agateau-gg

PR #424, opened 2022-11-18
Allow using --exit-zero, --exclude and --ignore-known-secrets anywhere on the command line

agateau-gg:
This is the second episode of #197.

First commit introduces a helper function to create Click callbacks which set config values: `create_config_callback()`. It then uses this helper for the --verbose and --allow-self-signed options.

Second commit introduces a `add_secret_scan_common_options()` decorator to define all scan options, similarly to `add_common_options()`. This decorator calls `add_common_options()`, so the commit modifies all scan commands to use `add_secret_scan_common_options()` instead of `add_common_options()`.

`add_secret_scan_common_options()` only implements the --exit-zero, --exclude  and --ignore-known-secrets options for now. Other options are coming in the next episode.

### DIFF
--- a/ggshield/cmd/common_options.py
+++ b/ggshield/cmd/common_options.py
@@ -10,7 +10,7 @@ To use it:
 The `kwargs` argument is required because due to the way click works,
 `add_common_options()` adds an argument for each option it defines.
 """
-from typing import Any, Callable, Optional, cast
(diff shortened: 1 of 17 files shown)


pierrelalanne:
Same small remark as in the first PR : we could add some variations in the functional tests.  
Otherwise looks great !

PR #427, opened 2022-11-21
feat: show header information only if there are policy breaks for the scan result

alina-tuholukova-gg:


agateau-gg (approved):
Looks good!

PR #429, opened 2022-11-22
Allow using --show-secrets, --json, --output and --banlist-detector anywhere on the command line

### LINKED ISSUE #197: `scan` options should be allowed later in the command

agateau-gg:
This is the third and last episode of #197.

First commit defers the creation of the `OutputHandler` instance: it must be created at the end so that we have the occasion of handling all output parameters.

The other commits use the same technique as before for the remaining options.

I made some changes in the tests to test using `--json` in various places, but there are no tests for the others, so no changes for them.

Fixes #197.


pierrelalanne (approved):
LGTM. I'm not against a second reviewer as these changes are not easy ones 🤯

agateau-gg:
Heh, yes that one hurts a bit :/

alina-tuholukova-gg (approved):
looks good to me, locally works well.

Issue #428, opened 2022-11-21
secrets displayed twice if same content and same file is sent to multiscan endpoint


alina-tuholukova-gg:
## Environment

- ggshield version: 1.14.1
- Operating system (Linux, macOS, Windows): Linux
- Operating system version: 20.04
- Python version: 3.9

## Describe the bug

The secret is shown twice if exactly the same file with the same content is present in the different commits. 

Steps to reproduce:

1. setup a repo, create a first commit on main branch
2. create a branch B1 from main, add file with the secret, commit
3. create a branch B2 from main, add same file with the secret (same name), commit
4. scan repo with ggshield secret scan repo .

Actual result:
1st commit
 - secret
 - secret

2nd commit
- secret
- secret

Expected result:

1st commit
 - secret

2nd commit
- secret

PR #430, opened 2022-11-23
fix: same secret should be shown once in the results output for scan command

alina-tuholukova-gg:
Fixes issue https://github.com/GitGuardian/ggshield/issues/428

### DIFF
--- a/ggshield/scan/repo.py
+++ b/ggshield/scan/repo.py
@@ -72,11 +72,9 @@ def scan_commits_content(
     ignored_detectors: Optional[Set[str]] = None,
     ignore_known_secrets: bool = False,
 ) -> ScanCollection:  # pragma: no cover
-    commit_files_tuple = []
+    commit_files_tuples = []
     try:
-        for commit in commits:
-            for file in commit.files:
-                commit_files_tuple.append((commit, file))
+        commit_files_tuples = [(c, f) for c in commits for f in c.files]
 
         progress_callback(advance=len(commits))
         scanner = SecretScanner(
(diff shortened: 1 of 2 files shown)


ggshield/scan/repo.py
```diff
@@ -73,8 +72,12 @@ def scan_commits_content(
     ignored_detectors: Optional[Set[str]] = None,
     ignore_known_secrets: bool = False,
 ) -> ScanCollection:  # pragma: no cover
+    commit_files_tuple = []
     try:
-        commit_files = list(itertools.chain.from_iterable(c.files for c in commits))
+        for commit in commits:
+            for file in commit.files:
+                commit_files_tuple.append((commit, file))
```
agateau-gg, line 79: I would rename `commit_files_tuple` to `commit_files_tuples`, since its a list of tuples and not a list.

Also, I think you can create it using a comprehension list, like this (untested):

```
commit_files_tuples = [(c, f) for c in commits for f in c.files]
```

ggshield/scan/repo.py
```diff
@@ -85,28 +88,20 @@ def scan_commits_content(
...
-    for commit in commits:
-        concerned_results = [
-            res
-            for res in results.results
-            if any(
-                res.content == file.document and res.filename == file.filename
-                for file in commit.files
-            )
-        ]
+    for (commit, file), result in zip(commit_files_tuple, results.results):
```
agateau-gg, line 98: Looks like we do not use `file` in this loop, so it can be replaced with `_`:
```suggestion
    for (commit, _), result in zip(commit_files_tuple, results.results):
```

It's interesting that this new code is probably going to be faster! 😀

agateau-gg (changes requested):

ggshield/scan/repo.py
pierrelalanne: @agateau-gg  
Side note : if I am not mistaken we do not benchmark `scan repo` in the performance benchmark ? We could have seen it here. But it may be too long to include the repo option in the benchmark ?

agateau-gg: We benchmark `scan commit-range`, which uses this code.

ggshield/scan/repo.py
alina-tuholukova-gg: indeed, thanks !

agateau-gg (approved):
Looks good, thanks!

PR #435, opened 2022-11-29
fix: the returned results from multiscan should be associated to the commits correctly

alina-tuholukova-gg:


agateau-gg (approved):
Looks good, thanks!

Nice to see you were able to remove the CommitFile class :)

PR #439, opened 2022-12-05
Improve error handling

agateau-gg:
This PR does two things:

- First it documents the fact that "--verbose" can be used to get stack traces from unexpected failures.
- Second it provides a workaround for Windows users hitting Unicode decoding issues.


pierrelalanne (commented):
One small comment, otherwise looks good.

agateau-gg:
> Looks good. I spotted two other possible locations where we `click.echo` a content that could contain problematic characters : in `scanner.py` and in `file_utils.py`. In both cases, this would be special characters in filenames. I am not sure this can really happen though. What do you think ?

Good point. I moved the check to our exception handling function, this should ensure we catch all cases.

pierrelalanne (approved):

PR #442, opened 2022-12-12
feat: Introduce new headers (OS, Python) version

alaabenfatma:
### Introduction
This MR introduces 3 new HTTP headers to the requests we submit to GG's API.

### Details
The headers include:
- `OS-Name` header: A header that contains the OS on which ggshield is running
- `OS-Version`: The version of said OS
- `Python-Version`: The python version used to run ggshield

### Output
The previously mentioned changes result in the reception of the following extra logs:
```
ggshield_os_name=ubuntu 
ggshield_os_version=22.04 
ggshield_python_version=3.10.6
``` 

### Questions
I wonder whether these metadata are okay or if we should trim them. 
For instance, instead of submitting a full verbose Python version, we may probably just send the first part (e.g., `3.10.6 `).
> I would advise against trimming the OS version though because we'll probably fail to match the **version** only (e.g., `#59-Ubuntu`) given that every OS may yield a different output.

### DIFF
--- a/ggshield/scan/scan_context.py
+++ b/ggshield/scan/scan_context.py
@@ -1,10 +1,54 @@
+import logging
+import platform
+import sys
 import uuid
 from dataclasses import dataclass
-from typing import Dict, Optional, Union
+from functools import lru_cache
+from pathlib import Path
+from typing import Dict, Optional, Tuple, Union
 
 from ggshield import __version__
 
 from .scan_mode import ScanMode
+
+
+logger = logging.getLogger(__name__)
+
+
+@lru_cache(None)
+def get_os_info() -> Tuple[str, str]:
+    if sys.platform.lower() == "linux":
+        return parse_os_release(Path("/etc/os-release"))
+    else:
+        return platform.system().lower(), platform.version()
+
+
+def parse_os_release(os_release_path: Path) -> Tuple[str, str]:
+    """
+    Extract and return Linux's OS-name and OS-Version
+    If extraction fails, we return ('linux', 'unknown')
+    """
+    error_tuple = "linux", "unknown"
+
+    try:
+        with open(os_release_path) as f:
+            lines = f.readlines()
+
+        # Build a dictionary from the os-release file contents
+        data_dict = {}
+        for line in lines:
+            if "=" in line:
+                key, value = line.split("=")
+                key, value = key.strip(), value.strip()
+                data_dict[key] = value.strip('"')
+
+        if "ID" not in data_dict:
+            return error_tuple
+
+        return data_dict["ID"], data_dict.get("VERSION_ID", "unknown")
+    except Exception as exc:
+        logger.warning(f"Failed to read Linux OS name and version: {exc}")
+        return error_tuple
 
 
 @dataclass
@@ -15,6 +59,8 @@ class ScanContext:
 
     def __post_init__(self) -> None:
         self.command_id = str(uuid.uuid4())
+        self.os_name, self.os_version = get_os_info()
+        self.python_version = platform.python_version()
 
     def get_http_headers(self) -> Dict[str, str]:
         """
@@ -26,8 +72,10 @@ class ScanContext:
             "Version": __version__,
             "Command-Path": self.command_path,
             "Command-Id": self.command_id,
+            "OS-Name": self.os_name,
+            "OS-Version": self.os_version,
+            "Python-Version": self.python_version,
         }
-
         if self.extra_headers:
             headers = {**headers, **self.extra_headers}
 
--- a/tests/unit/core/test_utils.py
+++ b/tests/unit/core/test_utils.py
@@ -1,5 +1,6 @@
 import os
-from typing import Dict, List, Optional
+import sys
+from typing import AnyStr, Dict, List, Optional, Tuple
 from unittest.mock import Mock, patch
 
 import click
@@ -21,6 +22,7 @@ from ggshield.core.utils import (
 )
 from ggshield.scan import Commit, File, Files, ScanContext, ScanMode, SecretScanner
 from ggshield.scan.repo import cd
+from ggshield.scan.scan_context import parse_os_release
 from tests.unit.conftest import (
     _PATCH_WITH_NONEWLINE_BEFORE_SECRET,
     _SECRET_RAW_FILE,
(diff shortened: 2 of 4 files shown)


ggshield/scan/scan_context.py
```diff
@@ -15,6 +17,9 @@ class ScanContext:
 
     def __post_init__(self) -> None:
         self.command_id = str(uuid.uuid4())
+        self.os = platform.system()
+        self.os_version = platform.version()
```
agateau-gg, line 21: `platform.version()` is not really useful on Linux. You can get more useful information for the `os_name` and `os_version` fields by parsing `/etc/os-release` and reading the ID and VERSION_ID fields.

I suggest you add a helper function at the top of the file to do something along those lines:

```python
def get_os_info() -> (str, str):
    if linux:
        return parse_os_release()
    else:
        # Use lower() to match the linux output
        return platform.system().lower(), platform.version()

def parse_os_release() -> (str, str):
    # ...
```

Then use it like this: `self.os_name, self.os_version = get_os_info()`

agateau-gg (changes requested):

ggshield/scan/scan_context.py
alaabenfatma: Thank you for this review @agateau-gg 

I did not know about the `/etc/os-release` trick before, thanks :pray: 

Let me know what you think about the new changes :smile_cat:

ggshield/scan/scan_context.py
```diff
@@ -1,12 +1,35 @@
...
+
+def parse_os_release() -> Tuple[str, str]:
+    """Return both Linux OS-name and OS-Version"""
+    os_release_path = pathlib.Path("/etc/os-release")
+    with open(os_release_path) as f:
+        lines = f.readlines()
+
+    # Build a dictionary from the metadata
+    data_dict = {
+        key.strip(): value.strip() for key, value in (line.split("=") for line in lines)
```
agateau-gg, line 28: There is an issue here: `value` may be surrounded with double-quotes. These quotes must be stripped:

```
In [1]: from ggshield.scan.scan_context import parse_os_release

In [2]: parse_os_release()
Out[2]: ('ubuntu', '"22.04"')
```

The function must also be hardened: if it fails, no scans happen :/ Right now it is going to fail if:

- the file does not exist
- the file cannot be read
- the file contains a line without a `=` ([os-release man page](https://www.linux.org/docs/man5/os-release.html) says line starting with `#` are comments)
- one of ID or VERSION_ID keys are not present. The man page says VERSION_ID is optional. It is the case on Arch linux because it's a rolling-release distribution:
  ```
  $ docker run --rm archlinux:latest cat /etc/os-release
  NAME="Arch Linux"
  PRETTY_NAME="Arch Linux"
  ID=arch
  BUILD_ID=rolling
  ANSI_COLOR="38;2;23;147;209"
  HOME_URL="https://archlinux.org/"
  DOCUMENTATION_URL="https://wiki.archlinux.org/"
  SUPPORT_URL="https://bbs.archlinux.org/"
  BUG_REPORT_URL="https://bugs.archlinux.org/"
  LOGO=archlinux-logo
  ```

If the file cannot be read, we should fallback to `linux` for the name and `unknown` for the version.

If VERSION_ID is not set it should also be `unknown`.


I suggest you add a test just for parse_os_release(). To make it easier to test, you can modify it to take the path of the file to parse instead, this way you can pass it any file, using the `tmp_path` Pytest fixture. Then the code would look like this:

```
def get_os_info() -> Tuple[str, str]:
    if sys.platform.lower() == "linux":
        return parse_os_release(Path("/etc/os-release"))
    else:
        return platform.system().lower(), platform.version()


def parse_os_release(os_release_path: Path) -> Tuple[str, str]:
    ...
```

agateau-gg (changes requested):

ggshield/scan/scan_context.py
agateau-gg: One more thing: you can add `@lru_cache(None)` (from functools module) to `get_os_info()` so that the results are cached.

ggshield/scan/scan_context.py
```diff
@@ -1,12 +1,56 @@
...
+            key, value = key.strip(), value.strip()
+            data_dict[key] = value.strip('"')
+
+    if "ID" not in data_dict:
+        return error_tuple
+
+    if "VERSION_ID" not in data_dict:
+        data_dict["VERSION_ID"] = "unknown"
+
+    return data_dict["ID"], data_dict["VERSION_ID"]
```
agateau-gg, line 51:
```suggestion
    return data_dict["ID"], data_dict.get("VERSION_ID", "unknown")
```

ggshield/scan/scan_context.py
alaabenfatma: Thanks @agateau-gg. I learned many new things from this PR :pray: 
By the way, I am skipping the new tests when the platform != Linux (it will not be relevant otherwise). 

This is the decorator:
```python
@pytest.mark.skipif(
    sys.platform.lower() != "linux", reason="This test is only relevant on Linux."
)
```

Let me know what you think about the state of this PR :smile_cat:

ggshield/scan/scan_context.py
agateau-gg: While catching individual exceptions is often the right thing to do, in this case I'd rather have a large try..except block. Something like this:

```python
try:
   # All the code, without exception handling
except Exception as exc:
   logger.warning("Failed to read Linux OS name and version: %s", exc)
   return error_tuple
```

This is because we are in a similar situation as `ggshield/core/check_updates.py`: this is code which is useful to us, but should never stop the user if it fails.

tests/unit/core/test_utils.py
```diff
@@ -324,3 +327,29 @@ def test_load_dot_env_loads_git_root_env(
...
+    ],
+)
+def test_parse_os_release(
+    tmp_path,
+    file_contents: AnyStr,
+    file_permissions: int,
+    expected_tuple: Tuple[str, str],
+):
+    file = tmp_path / "os-release"
+    file.touch()
```
agateau-gg, line 351: This is not necessary

tests/unit/core/test_utils.py
```diff
@@ -324,3 +327,29 @@ def test_load_dot_env_loads_git_root_env(
...
+
+    file.write_text(file_contents)
+    file.chmod(file_permissions)
+    assert parse_os_release(Path(file)) == expected_tuple
```
agateau-gg, line 355:
```suggestion
    assert parse_os_release(file) == expected_tuple
```

file is already a Path.

agateau-gg (changes requested):
Getting there!

ggshield/scan/scan_context.py
alaabenfatma: At first, I thought it'd be better to catch the exceptions individually so that we can handle them gracefully one day _maybe_. But yes I agree, I'll bundle it all in one large try/catch.

agateau-gg (approved):

PR #444, opened 2022-12-20
pre-receive: do not scan commits overwritten by push --force

### LINKED ISSUE #437: Pre-receive hook does not behave as expected in the event of a push force
## Environment

- ggshield version: `>= 1.13.6` (to have the latest pre-receive scan implementation version)
- Operating system (Linux, macOS, Windows): Any
- Operating system version: Any
- Python version: Any

## Describe the bug

When a user reworks a local branch (typically after rebasing the target branch) and makes a push force, the pre-receive hook will list all commits on the remote branch + all commits on the local branch being pushed force. This is not the expected behavior as the commits on the remote branch are already known to the remote, and hence should have already been handled by the pre-receive hook.  

### Steps to reproduce:

1. Chose a git repository
2. Make two separate clones : one will be called L (local) for convenience, and the other one R (remote).
3. Mark R as being a remote in L : `git remote add my_remote PATH_TO_R/.git`
4. Add a ggshield pre-receive hook in R : Create a `PATH_TO_R/.git/hooks/pre-receive` file containing :  
```
#!/bin/sh
set -e
ggshield --verbose --debug secret scan pre-receive
```
5. Create a branch `pre-receive-test` in L, add a commit (it has a sha : `sha-first-commit`), and push to R. GGShield should run and scan only the commit you added.
6. Now, go in R, and add a new commit on the main branch, we'll call its sha `sha-commit-main-branch`. Go in L, pull changed from main, and rebase `pre-receive-test` on main. You now have one commit on the `pre-receive-test` branch, the same as before, but its sha changed because of the rebase. Let's call it `sha-first-commit-after-rebase`.
7. In L, push force the changes to R : `git push -f my_remote`
8. GGshield pre-receive hook will do the following (in the logs) : `git rev-list --reverse sha-first-commit...sha-first-commit-after-rebase --max-count 51`
9. This results in scanning three commits (`sha-first-commit`, `sha-commit-main-branch`, `sha-first-commit-after-rebase`) instead of scanning only one : `sha-first-commit-after-rebase`.

### Actual result:

Commits from both the remote branch and the local branch being pushed force are scanned

### Expected result:

Only the commits from the branch being pushed are scanned.

agateau-gg:
## Description

When a branch is force-pushed, the pre-receive hook currently rescans the overwritten commits. This waste time and can report issues on code which is going away since it is being overwritten.

The pre-receive hook currently scans the commit range it receives via stdin (see [pre-receive hook doc][doc]), unless the code is pushing a new branch (because in this case the range start ref is the all-zeroes sha).

This PR fixes the bug by ignoring the range start ref received via stdin and always determining the start of the pushed code using the code currently used for the "new branch" case.

[doc]: https://git-scm.com/docs/githooks#pre-receive

## What has been done

- Some improvements to our `Repository` helper class: it learned how to `add()` and `push()`.
- Rewrote the pre-receive tests to use real repositories instead of mocking all git calls. This was required by the fix breaking most pre-receive tests, because the tests were too reliant on which git commands were run.
- Removed some no-longer-useful tests.
- Actually implement the fix.

## Issue

Fixes #437

### DIFF
--- a/ggshield/cmd/secret/scan/prereceive.py
+++ b/ggshield/cmd/secret/scan/prereceive.py
@@ -172,15 +172,7 @@ def prereceive_cmd(
         f"{before}...{after}", max_count=config.max_commits_for_hook + 1
     )
 
-    if not commit_list:
-        click.echo(
-            "Unable to get commit range.\n"
-            f"  before: {before}\n"
-            f"  after: {after}\n"
-            "Skipping pre-receive hook\n",
-            err=True,
-        )
-        return 0
+    assert commit_list, "Commit list should not be empty at this point"
 
     if len(commit_list) > config.max_commits_for_hook:
         click.echo(
--- a/ggshield/cmd/secret/scan/prereceive.py
+++ b/ggshield/cmd/secret/scan/prereceive.py
@@ -1,4 +1,5 @@
 import _thread as thread
+import logging
 import os
 import sys
 import threading
@@ -21,6 +22,8 @@ from ggshield.output.text.message import remediation_message
 from ggshield.scan import ScanContext, ScanMode
 from ggshield.scan.repo import scan_commit_range
 
+
+logger = logging.getLogger(__name__)
 
 REMEDIATION_MESSAGE = """  A pre-receive hook set server side prevented you from pushing secrets.
   Since the secret was detected during the push BUT after the commit, you need to:
@@ -97,9 +100,10 @@ def find_branch_start(commit: str) -> Optional[str]:
     return None
 
 
-def parse_stdin() -> Tuple[str, str]:
+def parse_stdin() -> Optional[Tuple[str, str]]:
     """
-    Parse stdin and return the first and last commit to scan
+    Parse stdin and return the first and last commit to scan,
+    or None if there is nothing to do
     """
     prereceive_input = sys.stdin.read().strip()
     if not prereceive_input:
@@ -108,16 +112,21 @@ def parse_stdin() -> Tuple[str, str]:
     # TODO There can be more than one line here, for example when pushing multiple
     # branches. We should support this.
     line = prereceive_input.splitlines()[0]
-    old_commit, new_commit, _ = line.split(maxsplit=2)
+    logger.debug("stdin: %s", line)
+    _old_commit, new_commit, _ = line.split(maxsplit=2)
 
-    if old_commit == EMPTY_SHA:
-        # Pushing to a new branch
-        start_commit = find_branch_start(new_commit)
-        if start_commit is None:
-            # branch does not contain any new commit
-            old_commit = new_commit
-        else:
-            old_commit = f"{start_commit}~1"
+    if new_commit == EMPTY_SHA:
+        # Deletion event, nothing to do
+        return None
+
+    # ignore _old_commit because in case of a force-push, it is going to be overwritten
+    # and should not be scanned (see #437)
+    start_commit = find_branch_start(new_commit)
+    if start_commit is None:
+        # branch does not contain any new commit
+        old_commit = new_commit
+    else:
+        old_commit = f"{start_commit}~1"
 
     return (old_commit, new_commit)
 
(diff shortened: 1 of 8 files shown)


tests/unit/cmd/scan/test_prereceive.py
```diff
@@ -94,50 +105,17 @@ def test_stdin_input_secret(
...
-    @patch("ggshield.cmd.secret.scan.prereceive.get_list_commit_SHA")
     @patch("ggshield.cmd.secret.scan.prereceive.scan_commit_range")
     def test_stdin_breakglass_2ndoption(
         self,
         scan_commit_range_mock: Mock,
-        get_list_mock: Mock,
         cli_fs_runner: CliRunner,
     ):
         """
         GIVEN 20 commits through stdin input but breakglass active
```
pierrelalanne, line 115: GIVEN commits through stding...

tests/unit/cmd/scan/test_prereceive.py
```diff
@@ -94,50 +105,17 @@ def test_stdin_input_secret(
...
         WHEN the command is run
         THEN it should return 0
```
pierrelalanne, line 117: THEN it should return 0 without scanning any commits

ggshield/cmd/secret/scan/prereceive.py
```diff
@@ -108,16 +112,21 @@ def parse_stdin() -> Tuple[str, str]:
...
-
-    if old_commit == EMPTY_SHA:
-        # Pushing to a new branch
-        start_commit = find_branch_start(new_commit)
-        if start_commit is None:
-            # branch does not contain any new commit
-            old_commit = new_commit
-        else:
-            old_commit = f"{start_commit}~1"
+    logger.debug("stdin: %s", line)
```
pierrelalanne, line 115: 👍

ggshield/cmd/secret/scan/prereceive.py
```diff
@@ -155,19 +164,21 @@ def prereceive_cmd(
...
             err=True,
         )
         return 0
 
-    if after == EMPTY_SHA:
-        click.echo("Deletion event or nothing to scan.", err=True)
-        return 0
-
     assert before != EMPTY_SHA
+    assert after != EMPTY_SHA
```
pierrelalanne, line 181: We already handled the `if before_after is None`, do we need this ?

pierrelalanne (commented):
Looks good to me !

ggshield/cmd/secret/scan/prereceive.py
agateau-gg: Since I removed the `if after == EMPTY_SHA:`, I feel more confident that case is indeed handled by the `if before_after is None` with this asset in place.

pierrelalanne: Okay, that's not a blocker for me of course. 
But if we take it the other way around : 
- if this assert fails, it means after == EMPTY_SHA
- This means in turn that before_after is something like (_, EMPTY_SHA)
- This means that parse_stdin() returned such a tuple
- Looking at the code for parse_stdin, this means that `new_commit` was EMPTY_SHA
- But if new_commit is EMPTY_SHA, parse_stdin returns None.
That's why the assert looks useless to me.

pierrelalanne (approved):

ggshield/cmd/secret/scan/prereceive.py
agateau-gg: Your analysis is correct, but the goal of this assert is not to get triggered by the current code. It has two jobs:

1. Document the assumption made while writing this code (useful for future code readers)
2. Trigger if we do something wrong in the future, like modifying `parse_stdin()` and making it return `EMPTY_SHA`, invalidating our assumption

It is a second line of defense. An assert triggering indicates a programmer error, it must never trigger for normal program behavior (including user errors).

pierrelalanne: Super clear, thanks for the explanation.

PR #449, opened 2022-12-29
Handle GitHub Rate Limits During Update Checks

rgajason:
GGShield checks for updates by connecting to GitHub anonymously.  The anonymous API quota is relatively small.  For organizations where hundreds of users all connect via a common egress IP, this can result in significant error messages (collectively across users) / confusion during GGShield operations.  For example:

```
% git commit -m 'My Awesome Commit' awesome.py
Failed to check: {"message":"API rate limit exceeded for 1.2.3.4. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

This PR address this issue by checking for quota exceeded condition and rescheduling the update check for after the quota has been reset.

Additionally, it adds headers to the update check request that are recommended by GitHub:

User Agent: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#user-agent-required
GH API Version: https://docs.github.com/en/rest/overview/api-versions?apiVersion=2022-11-28
Media type (accept): https://docs.github.com/en/rest/overview/media-types?apiVersion=2022-11-28

### DIFF
--- a/ggshield/core/check_updates.py
+++ b/ggshield/core/check_updates.py
@@ -63,6 +63,11 @@ def check_for_updates() -> Optional[str]:
     try:
         resp = requests.get(
             "https://api.github.com/repos/GitGuardian/GGShield/releases/latest",
+            headers={
+                "Accept": "application/vnd.github+json",
+                "User-Agent": f"GGShield {__version__}",
+                "X-Github-Api-Version": "2022-11-28",
+            },
             timeout=CHECK_TIMEOUT,
         )
     except Exception as e:
--- a/ggshield/core/check_updates.py
+++ b/ggshield/core/check_updates.py
@@ -75,6 +75,25 @@ def check_for_updates() -> Optional[str]:
         return None
 
     if resp.status_code != 200:
+        # Handle GitHub rate limit responses gracefully
+        # https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting
+        if int(resp.headers.get("X-RateLimit-Remaining", -1)) == 0:
+            logger.debug("GitHub rate limit exceeded - rescheduling update check")
+
+            # Reset the next update check based on when the GH API quota resets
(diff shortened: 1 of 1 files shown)


ggshield/core/check_updates.py
```diff
@@ -63,13 +63,37 @@ def check_for_updates() -> Optional[str]:
...
         return None
 
     if resp.status_code != 200:
+        # Handle GitHub rate limit responses gracefully
+        # https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting
+        if int(resp.headers.get("X-RateLimit-Remaining", -1)) == 0:
+            logger.debug("GitHub rate limit exceeded - rescheduling update check")
+
+            # Reset the next update check based on when the GH API quota resets
+            check_at = int(resp.headers.get("X-RateLimit-Reset", -1.0)) - 24 * 60 * 60
```
agateau-gg, line 84: nitpick: since we convert to int, we can make the default value an int too.
```suggestion
            check_at = int(resp.headers.get("X-RateLimit-Reset", -1)) - 24 * 60 * 60
```

ggshield/core/check_updates.py
```diff
@@ -63,13 +63,37 @@ def check_for_updates() -> Optional[str]:
...
+            if check_at < 0:
+                # Somehow we've hit the rate limit and the reset header is missing
+                # This can only happen if GH changes their responses
+                logger.error("Failed rescheduling update check")
```
agateau-gg, line 88: I think we should `return None` here to avoid writing `check_at: -1` in the cache file.

agateau-gg (changes requested):
Two minor remarks, but this looks great already. Thanks a lot for taking care of the bug and for adding the missing headers to the request.

rgajason:
@agateau-gg both changes implemented as requested

agateau-gg (approved):
Thanks, merging the changes in.

PR #453, opened 2023-01-06
Improve logs in several places, make `--check-for-updates` a common option

pierrelalanne:


agateau-gg (approved):
Looks great! +1 for the cleanly separated commits!

Issue #304, opened 2022-07-20
Add tests for GitHub actions


agateau-gg:
## Introduction

Because of some CI changes, our GitHub "secret" action is no longer continuously tested. The "iac" action is also not continuously tested.

This is a problem, and it has already bit us in the past: we had to release v1.13.5 because a hidden breaking change was not spotted early enough.

### `scanning` job

The `scanning` job is supposed to test the GitHub action from the main repository. Unfortunately we silently broke the job when we introduced the "iac" action: because we moved the root `action.yml` to `actions/secret/action.yml` (see 192fc32caa00dcfb20fbaf3525f66180a830befa), the job no longer finds it and falls back to executing the root Docker default command, which [displays ggshield usage message](https://github.com/GitGuardian/ggshield/actions/runs/3830050206/jobs/6518201093):

```
Scan commit with ggshield

Run ./
Building docker image
[...]

Issue #212, opened 2022-04-27
Print name of what is being scanned when scanning with --verbose


agateau-gg:
(this is a follow-up of #211)

When scanning with `--verbose` it would be useful, when possible, to provide more details on the progress of the scanning process:
- `scan path`, `scan archive` and `scan pypi` should print the filenames
- `scan repo` and `scan commit-range` should print the commit ids
- `scan docker` should print the scanned layers

PR #454, opened 2023-01-10
Fix "scanning" job

agateau-gg:
Use the correct path to the secret action.

This is not a complete fix for #304, but it at least bring back some tests.


pierrelalanne:
The log part is not surprising as we are using the latest image published : `FROM gitguardian/ggshield`. It should disappear after the next release. WDYT ?  
I am more puzzled by the two commits scanned. Any idea ?

agateau-gg:
> The log part is not surprising as we are using the latest image published : `FROM gitguardian/ggshield`. It should disappear after the next release. WDYT ?

Ah right, you are correct :+1:

> I am more puzzled by the two commits scanned. Any idea ?

I think that's because of the way the CI works. If I am not mistaken, GitHub creates a "test" branch from the target branch and merges the source branch in it. The CI then runs on this test branch. This means that if the source branch contains commit A, then the test branch is going to contain commit A + the merge commit.

This will be easier to read when #212 is done ;)

PR #460, opened 2023-01-26
Make the `ggshield scan` deprecation message more alarming

alaabenfatma:
# Introduction
Currently, we show a warning message when the deprecated command `ggshield scan` is used. The message we show is in yellow.
We would like to print it in red and inform the user that this command will be completely removed starting from version 1.15.0

# Method
We introduced a new boolean argument to the `display_warning` function called `critical`, which is set to False by default. If `¢ritical` is True, we show the message in red, and yellow otherwise.
### Why not create a new style?
We use the `error` style to display the message in red. We did not create another style `critical_warning` because we'll be duplicating the same style properties (red color) as the `error` style.

### DIFF
--- a/ggshield/cmd/scan/__init__.py
+++ b/ggshield/cmd/scan/__init__.py
@@ -17,7 +17,7 @@ from ggshield.cmd.secret.scan import (
     repo_cmd,
     scan_group_impl,
 )
-from ggshield.core.text_utils import display_warning
+from ggshield.core.text_utils import display_error
 
 
 @click.group(
@@ -60,7 +60,9 @@ def deprecated_scan_group(
     """
     Deprecated: use `ggshield secret scan (...)` instead.
     """
-    display_warning(
-        "Warning: Using `ggshield scan (...)` is deprecated. Use `ggshield secret scan (...)` instead.",
+    # We use `display_error` to print this warning message in red color
+    display_error(
+        "Warning: The `ggshield scan (...)` commands are deprecated and will be removed in version 1.15.0. "
+        "Use `ggshield secret scan (...)` instead."
     )
     return scan_group_impl(ctx)


ggshield/cmd/scan/__init__.py
```diff
@@ -61,6 +61,8 @@ def deprecated_scan_group(
     Deprecated: use `ggshield secret scan (...)` instead.
     """
     display_warning(
-        "Warning: Using `ggshield scan (...)` is deprecated. Use `ggshield secret scan (...)` instead.",
+        "Warning: The command `ggshield scan (...)` is deprecated and will be removed in version 1.15.0. "
```
agateau-gg, line 64:
```suggestion
        "Warning: The `ggshield scan (...)` commands are deprecated and will be removed in version 1.15.0. "
```

ggshield/core/text_utils.py
```diff
@@ -123,8 +123,9 @@ def display_error(msg: str) -> None:
     click.echo(format_text(msg, STYLE["error"]), err=True)
 
 
-def display_warning(msg: str) -> None:
-    click.echo(format_text(msg, STYLE["warning"]), err=True)
+def display_warning(msg: str, critical: bool = False) -> None:
```
agateau-gg, line 126: I am not fond of this because I am pretty sure the API change is going to stay there after we remove the deprecation warning. I think it's simpler to switch to `display_error()` instead (with a comment explaining why we use it there), even if it's not an error.

agateau-gg (changes requested):

ggshield/core/text_utils.py
alaabenfatma: I was hesitant of using `display_error()` because we were displaying a warning and not an error. But I agree, it's enough for the task and we can add a comment to explain why we use it.

Thanks for the review :pray:

agateau-gg (approved):
Looks good, thanks!

PR #462, opened 2023-01-27
Test our GitHub actions

agateau-gg:
## Description

This MR does the following:

- Introduces two GitHub actions: `actions-unstable/iac` and `actions-unstable/secret`. These actions behave like their `actions/` counterparts except they use `unstable` version of our Docker image and can update GGShield to use a version from any git reference.
- Make our CI uses these two actions.
- Fix the "git dubious ownership in repository" errors which appeared following the latest security fix in Git


actions-unstable/README.md
```diff
@@ -0,0 +1,6 @@
+This directory contains GitHub actions used internally by our CI to test the "real" GitHub actions. They differ from the real GitHub actions by:
```
pierrelalanne, line 1: Not sure that the "real" is that explicit ? What about "to test the after-merge GitHub actions" ? or "the GitHub action corresponding to the current state of the repo" ?

actions-unstable/iac/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM gitguardian/ggshield
```
pierrelalanne, line 1: Aren't we missing an `:unstable` here ?

actions-unstable/secret/action.yml
```diff
@@ -0,0 +1,26 @@
+name: 'ggshield scan'
```
pierrelalanne, line 1: name: 'ggshield secret scan'

actions-unstable/secret/action.yml
```diff
@@ -0,0 +1,26 @@
...
+description: 'Scan commits for hardcoded secrets and security policy breaks.'
```
pierrelalanne, line 2: description: 'Scan commits for hardcoded secrets.'

actions-unstable/secret/action.yml
```diff
@@ -0,0 +1,26 @@
...
+author: GitGuardian <support@gitguardian.com>
+
+branding:
+  icon: 'shield'
+  color: 'blue'
+
+inputs:
+  args:
+    description: |
+      Arguments to be passed to ggshield secret scan
```
pierrelalanne, line 12: We do not mention the possible scan mode ? Is this intentional ?

scripts/README.md
```diff
@@ -6,6 +6,14 @@ This directory contains scripts to help with ggshield development.
 
 Build .pyz, .deb and .rpm packages.
 
+## action-entrypoint-generator/action-entrypoint-generator
+
+We have two GitHub actions: one to run `secret scan` and another to run `iac scan`. They are defined in `actions/secret` and `actions/iac`. These two actions use the common Dockerfile from the root of the repository, which provides a way to run GGShield inside a Docker container.
+
+To test that changes made inside pull requests do not cause problems with our GitHub actions, we have a separate set of actions: `actions-unstable/secret` and `actions-unstable/iac`. The difference between these actions and those defined in `actions`, is that if the `GITGUARDIAN_GGSHIELD_REF` environment variable is set, then they install GGShield from this ref, so that we test the code from the pull request and not the code currently in the `main` branch.
```
pierrelalanne, line 13: We do not mention the base image being `unstable` instead of `latest` ?

.github/workflows/main.yml
```diff
@@ -152,29 +152,48 @@ jobs:
...
-    name: Test GitHub secret action
+  test_github_secret_scan_action:
+    name: Test GitHub actions/secret
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
       - name: Scan commits
```
pierrelalanne, line 163: We could be more explicit with "Scan commits for hardcoded secrets"

.github/workflows/main.yml
```diff
@@ -152,29 +152,48 @@ jobs:
...
+
+  test_github_iac_scan_action:
+    name: Test GitHub actions/iac
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Scan commits
```
pierrelalanne, line 182: ... And "Scan commits for IaC vulns"

.github/workflows/main.yml
```diff
@@ -152,29 +152,48 @@ jobs:
...
+        uses: ./actions/iac
```
pierrelalanne, line 183: Should be `./actions-unstable/iac` here

scripts/action-entrypoint-generator/action-entrypoint-generator
```diff
@@ -0,0 +1,86 @@
...
+Generate the entrypoint.sh files for our GitHub actions.
+"""
+import argparse
+import sys
+from pathlib import Path
+
+
+TEMPLATE = """#!/usr/bin/env bash
+set -euo pipefail
+# AUTOGENERATED FILE, DO NOT EDIT!
```
pierrelalanne, line 12: This line is not in-sync with the entrypoints committed ?

scripts/action-entrypoint-generator/action-entrypoint-generator
```diff
@@ -0,0 +1,86 @@
...
+# This file has been generated by the `action-entrypoint-generator` script
+# defined in `scripts/action-entrypoint-generator`. To make changes to this
+# file, modify the script and rerun it.
+
+# GitHub overrides $HOME. Set it back to the home directory of our `app` user,
+# otherwise the call to `git config` fails.
+export HOME=/home/app
+
+# Mark the current directory as safe. If we don't do this, git commands fail
```
pierrelalanne, line 21: Same remark here if I am not mistaken.

scripts/action-entrypoint-generator/action-entrypoint-generator
```diff
@@ -0,0 +1,86 @@
...
+
+@UPDATE_GGSHIELD@
+args=("$@")
+@COMMAND@
+"""
+
+# This script contains the code used by the unstable actions to update ggshield
+UPDATE_GGSHIELD_CODE = (Path(__file__).parent / "update-ggshield.sh").read_text()
+
+COMMANDS_FOR_DOMAIN = {
```
pierrelalanne, line 33: Should we use the term `vertical` instead ? 🤔

scripts/action-entrypoint-generator/action-entrypoint-generator
```diff
@@ -0,0 +1,86 @@
...
+        base_output_dir / ("actions" if stable else "actions-unstable") / domain
+    )
+    output_dir.mkdir(exist_ok=True, parents=True)
+    output_file = output_dir / "entrypoint.sh"
+    print(f"Creating {output_file}")
+    output_file.write_text(content)
+
+
+def main():
+    assert (DEFAULT_OUTPUT_DIR / "actions").exists()
```
pierrelalanne, line 60: Do we need this as the `generate` function will `mkdir` the final location ?  
If we still need this, shouldn't we also check the `actions-unstable` location ?

.github/workflows/main.yml
```diff
@@ -152,29 +152,48 @@ jobs:
             dist
             packages
 
-  scanning:
-    name: Test GitHub secret action
+  test_github_secret_scan_action:
+    name: Test GitHub actions/secret
```
pierrelalanne, line 156:
```suggestion
    name: Test GitHub actions-unstable/secret
```

.github/workflows/main.yml
```diff
@@ -152,29 +152,48 @@ jobs:
...
           GITHUB_PUSH_BEFORE_SHA: ${{ github.event.before }}
           GITHUB_PUSH_BASE_SHA: ${{ github.event.base }}
           GITHUB_PULL_BASE_SHA: ${{ github.event.pull_request.base.sha }}
           GITHUB_DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
           GITGUARDIAN_API_KEY: ${{ secrets.GITGUARDIAN_API_KEY }}
           GITGUARDIAN_API_URL: ${{ secrets.GITGUARDIAN_API_URL }}
+          GITGUARDIAN_GGSHIELD_REF: ${{ github.ref }}
+
+  test_github_iac_scan_action:
+    name: Test GitHub actions/iac
```
pierrelalanne, line 175:
```suggestion
    name: Test GitHub actions-unstable/iac
```

actions-unstable/iac/Dockerfile
agateau-gg: Oups, good catch.

actions-unstable/secret/action.yml
agateau-gg: I just copied the actions/secret/action.yml here. I agree those needs to be updated, but this can be done in a separate issue.

scripts/README.md
agateau-gg: Good point, fixed.

.github/workflows/main.yml
agateau-gg: :+1:

scripts/action-entrypoint-generator/action-entrypoint-generator
agateau-gg: Indeed, I updated the comments here but did not regenerate the scripts. Fixed now.

scripts/action-entrypoint-generator/action-entrypoint-generator
agateau-gg: Sounds good, changed to it.

scripts/action-entrypoint-generator/action-entrypoint-generator
agateau-gg: This is just a sanity check to ensure I got the correct number of `.parent` calls when defining `DEFAULT_OUTPUT_DIR`. It will trigger if we ever move the script around.

.github/workflows/main.yml
agateau-gg: Fixed

agateau-gg: I am not sure about this: the goal is really to test the actions in "actions/*", but due to the way our actions are defined, we can't test them directly, we have to use the "actions-unstable/*" actions instead. One of the reasons for generating the endpoint.sh files is to ensure "actions/$X" and "actions-unstable/$X" remains in sync (maybe we should also generate the Dockerfile and action.yml? I refrained from doing so, but did silly mistakes because of that 🤔)

pierrelalanne: I'd say we start with this version that is already great. We'll see later if it is painful to maintain actions in sync ?    
About the wording, we could just say "Test GitHub action for secret" and "Test GitHub action for iac" ?

pierrelalanne (approved):

Issue #168, opened 2022-01-27
CI does not run on Windows or macOS


agateau-gg:
There is an issue with our CI setup, which make it run all jobs on Linux:

![image](https://user-images.githubusercontent.com/91945295/151414892-a13061ec-1835-497a-858b-330eb5c29412.png)

![image](https://user-images.githubusercontent.com/91945295/151414942-6d5cf749-a044-42ed-978a-f2200060c8a7.png)

## TODO

There is quite some work to do here, so it's better to do it in separate steps.

- [x] Make CI run on Windows and macOS (#164)
- [x] Fix test_text_output (#180)
- [x] Fix test_scan_docker (#204)
- [x] #461

PR #463, opened 2023-01-27
feat(test_path): #168 unskip failing tests on Windows

pierrelalanne:
The --exclude option was not working as expected on Windows because of failing regexes when excluding filepaths.  
We use `pathlib.PurePosixPath` to convert Windows path to posix path and apply regex on them.  

We also add an extra test to test the exclusion of a whole directory from the scan.

### DIFF
--- a/ggshield/core/filter.py
+++ b/ggshield/core/filter.py
@@ -3,6 +3,7 @@ import math
 import operator
 import re
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


agateau-gg (approved):
Looks good!

Makes me think we should make ggshield use Path instead of str where it makes sense. Going to file an issue for this.

PR #472, opened 2023-02-03
Make all README.md badges clickable

agateau-gg:
- Point Stars badge to stargazers
- Point Build badge to Actions tab"

Without this, clicking on those badges opens the image of the badge, which is useless.


alaabenfatma (approved):
Looks good to me!

Issue #456, opened 2023-01-11
Scan does not fail in github action when missing credentials


gg-mmill:
## Environment
GGShield github action from  https://github.com/GitGuardian/ggcanary/blob/main/.github/workflows/main.yml

## Describe the bug

When running without credentials, the job does not fail, and commits appear to be scanned (at least in the UI)

Steps to reproduce:

1. Use given github action in a repository without any credential configured
2. Commit on the repository

#### Actual result:

The commit appear to be scanned in the action:

```

Run GitGuardian/gg-shield/actions/secret@main
[...]

PR #473, opened 2023-02-06
feat(auth): Only execute scans when a valid API key exists.

alaabenfatma:
### Related issue
https://github.com/GitGuardian/ggshield/issues/456

### Solution
We propose checking that a healthy connection with API can be established.
If we fail to do it successfully, then we conclude we cannot interact with the API for the time being. Thus, we abort.

### Implementation
We use the `health_check` function from the `GGClient` class to retrieve the health of the API connection we tried to establish. We do it in two different locations:
1. In the `SecretScanner` class in order to check for the validity of the key within the class constructor
    - This can be toggled off by setting the argument `check_api_health` to `False` (defaults to `True`)
2. Within the `scan_commit_range()` function because it creates different scanner instances: One per scanned commit.

#### Previous behavior (bad :no_entry: )
```bash
❯ ggshield secret scan path ./README.md
status_code=401 detail=Invalid API key.
Scanning Path... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1 files scanned out of 1 0:00:00

ERROR: Invalid API key..
```

#### New behavior (better :heavy_check_mark: )
```bash
❯ ggshield secret scan path ./README.md


ERROR: Invalid API key..
```

### DIFF
--- a/ggshield/core/errors.py
+++ b/ggshield/core/errors.py
@@ -88,6 +88,15 @@ class MissingTokenError(AuthError):
         super().__init__(instance, f"No token is saved for this instance: '{instance}'")
 
 
+class APIKeyCheckError(AuthError):
+    """
+    Raised when checking the API key fails
+    """
+
+    def __init__(self, instance: str, message: str):
+        super().__init__(instance, message)
+
+
 def format_validation_error(exc: ValidationError) -> str:
     """
     Take a Marshmallow ValidationError and turn it into a more user-friendly message
@@ -121,7 +130,7 @@ def handle_exception(exc: Exception, verbose: bool) -> int:
     )
     click.echo()
 
-    display_error(f"ERROR: {exc}.")
+    display_error(f"ERROR: {exc}")
     if isinstance(exc, UnicodeEncodeError) and platform.system() == "Windows":
         display_error(
             "\n"
--- a/ggshield/scan/scanner.py
+++ b/ggshield/scan/scanner.py
@@ -12,6 +12,7 @@ from pygitguardian.iac_models import IaCScanResult
 from pygitguardian.models import Detail, ScanResult
 
 from ggshield.core.cache import Cache
+from ggshield.core.client import check_client_api_key
 from ggshield.core.errors import UnexpectedError
 from ggshield.core.filter import (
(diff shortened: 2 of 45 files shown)


ggshield/scan/scanner.py
```diff
@@ -175,7 +175,13 @@ def __init__(
         ignored_matches: Optional[Iterable[IgnoredMatch]] = None,
         ignored_detectors: Optional[Set[str]] = None,
         ignore_known_secrets: Optional[bool] = None,
+        check_api_health: Optional[bool] = True,
```
agateau-gg, line 178: nitpick: the type for this can be `bool`: it does not need to be optional since it has a default value (making `ignore_known_secrets` optional was a mistake I think)

ggshield/scan/scanner.py
```diff
@@ -175,7 +175,13 @@ def __init__(
...
     ):
+        if check_api_health:
+            health_response = client.health_check()
+            if not health_response.success:
+                raise click.ClickException(health_response.detail)
```
agateau-gg, line 183: Two things here:

1. You can turn this code into a static method like `check_can_scan()`. This way you can reuse it in `repo.scan_commit_range()`
2. The exception should be AuthError when health_response.status_code is 401 and UnexpectedError otherwise.

ggshield/scan/scanner.py
agateau-gg: Also, you should set `check_api_health` to False when SecretScanner is created in repo.py.

agateau-gg (changes requested):

ggshield/core/errors.py
```diff
@@ -88,6 +88,15 @@ def __init__(self, instance: str):
         super().__init__(instance, f"No token is saved for this instance: '{instance}'")
 
 
+class APIKeyCheckError(AuthError):
```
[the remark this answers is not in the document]
agateau-gg: Oh yes, this has been bothering me for a while. Going to fix it.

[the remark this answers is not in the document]
agateau-gg: Should be good now.

pierrelalanne (approved):

PR #475, opened 2023-02-08
chore(wording): improve description of --check-for-updates option

pierrelalanne:


ggshield/cmd/common_options.py
```diff
@@ -114,7 +114,7 @@ def debug_callback(
     "--check-for-updates/--no-check-for-updates",
     is_flag=True,
     default=None,
-    help="Check for ggshield updates.",
+    help="Make a daily check for ggshield updates before executing commands.",
```
agateau-gg, line 117: I suggest:

```suggestion
    help="After executing commands, check if a new version of ggshield is available.",
```

I changed "before" to "after" since the check is now done just before exiting.

agateau-gg (changes requested):

ggshield/cmd/common_options.py
pierrelalanne: Oops, I forgot about that. Nice catch !

agateau-gg (approved):

Issue #480, opened 2023-02-13
Config Error Messages Replace "-" With "_"


rgajason:
## Environment

- ggshield version: 1.14.3
- Operating system (Linux, macOS, Windows): macOS
- Operating system version: Ventura 13.1
- Python version: 3.10.10

## Describe the bug

The "Unrecognized key in config" error message displays an underscore instead of hyphen when an invalid configuration is present.

Steps to reproduce:

Create an invalid configuration file such as (note `ignored-matches` misspelled):

**`.gitguardian.yml`**
```yaml
version: 2
secret:
  ignore-matches:
    - name: some-secret-i-want-to-ignore
      match: 4f2f11652a0<snip>5ead42463
```

Execute a `git commit` that includes the secret you wish to ignore (with GGShield configured as a pre-commit hook).

Actual result:

Error message:

> Unrecognized key in config: ignore_matches

Note the underscore when the configuration has a hyphen.

Expected result:

Error message:

> Unrecognized key in config: ignore-matches

(hyphen instead of underscore)

pierrelalanne:
Thank you for raising this issue @rgajason, we will look into this.

agateau-gg:
Sure, you're welcome to fix this! Do you have any idea in mind? I am not sure what would be the best way to fix it 🤔

karo-fox:
I'm thinking of modyfing `filter_fields` method in `FilteredConfig` ([here's its code](https://github.com/GitGuardian/ggshield/blob/b7b9a8e42163d40635dc5601e8825e10866145e0/ggshield/core/types.py)) so it will display better warning. Maybe we can change the message to something like: `Unrecognized key in config: {invalid key}, expected {valid key}`

agateau-gg:
I think that's not what this issue is about 😕.

For historical reasons, ggshield accepts config keys written using both `-` or `_`, so `ignored-matches` and `ignored_matches` are the same. Internally this works by replacing all `-` with `_` (this is done by [core.config.load_yaml_dict()](https://github.com/GitGuardian/ggshield/blob/main/ggshield/core/config/utils.py#L26)).

The problem of this feature is: if a user adds a `does-not-exist` (dash-separated) key to their config, then ggshield is going to report that `does_not_exist` (underscore-separated) is not a valid key, which is confusing.

This problem is potentially complicated to fix because at the point we find a non-existing key, we do not have the original key name anymore, so we can't report it.

karo-fox:
I came up with possible solution:
[...]

PR #483, opened 2023-02-22
Config error messages #480

karo-fox:
Now FilteredConfig replaces "-" with "_".

`replace_in_keys()` was removed from `load_yaml_dict()` and `save_yaml_dict()` and is explicitly called from `prepare_auth_config_dict_for_parse()`, `UserConfig.save()`, `UserConfig._update_from_file()` (when the config's version is 1).


ggshield/core/types.py
```diff
@@ -18,8 +18,8 @@ def filter_fields(cls, data: Dict, **kwargs: Any) -> Dict:
         field_names = {field_.name for field_ in fields(cls)}
         filtered_fields = {}
         for key, item in data.items():
-            if key in field_names:
-                filtered_fields[key] = item
+            if key in field_names or key.replace("-", "_") in field_names:
+                filtered_fields[key.replace("-", "_")] = item
```
agateau-gg, line 22: I would change this to:

```suggestion
            filtered_key = key.replace("-", "_")
            if filtered_key in field_names:
                filtered_fields[filtered_key] = item
```

It's a bit simpler and avoid performing the `replace()` call twice.

ggshield/core/config/auth_config.py
```diff
@@ -69,6 +70,7 @@ def prepare_auth_config_dict_for_parse(data: Dict[str, Any]) -> Dict[str, Any]:
 
     Stop if there is not exactly one account.
     """
+    replace_in_keys(data, "-", "_")
```
agateau-gg, line 73: Can you add a line in the docstring saying that for compatibility issues we turn `-` chars in keys into `_`?

agateau-gg (changes requested):
Minor remarks, looks good!

ggshield/core/config/auth_config.py
karo-fox: Done!

ggshield/core/types.py
karo-fox: Done!

agateau-gg (approved):
Looks good now, thanks a lot!

We have a release planned tomorrow morning, but I prefer not including this PR that late in the cycle. I will merge it tomorrow afternoon. Hope you don't mind.

karo-fox:
Sure, no problem

agateau-gg:
I just merged #485, which should fix the CI failures for PR from forks. Can you rebase your branch and push-force it? I would like to ensure it fixes the problem.

PR #484, opened 2023-02-22
Fix fails on pull-requests from forks (PR from a fork)

### LINKED ISSUE #374: CI fails for pull requests from forks
## Description

When a pull request is opened from a fork, the CI fails because it cannot access the `GITGUARDIAN_API_KEY` secret required to run functional tests. This is caused by a security feature on GitHub actions: pull requests across forks cannot access secrets, neither secrets from the source repository nor from the target repository.

## Definition of done

- [ ] Investigate the possible workarounds. A limited investigation shows that the [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event instead of the `pull_request` one would give access to the secrets because in this case the scripts run in the context of the target repository. Running untrusted code from this event is however a security issue, so `pull_request_target` cannot be used as a direct replacement for `pull_request`: the submitted code needs to be manually reviewed before triggering the run.
- [ ] If a viable solution has been found, implement the changes. If not, disable functional tests for pull-requests coming from forks. Do not disable them based on whether `GITGUARDIAN_API_KEY` is set or not: we want the tests to fail for a CI run on pushes and on pull-requests from branches within the repository.

agateau-gg:
These tests fail because they need the GITGUARDIAN_API_KEY variable, but it's not available for pull-requests coming from forks.

Fixes #374

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,6 +113,10 @@ jobs:
           fail_ci_if_error: false
 
       - name: Run functional tests
+        # For security reasons, secrets are not available when running tests for a pull request from a fork. This causes
+        # all functional tests to fail because GITGUARDIAN_API_KEY is not defined. To avoid this, skip this step if we
+        # were triggered by a pull request from a fork.
+        if: ${{ !github.event.pull_request.head.repo.fork }}
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


agateau-gg:
Merged via #485.

PR #485, opened 2023-02-22
Fix fails on pull-requests from forks

### LINKED ISSUE #374: CI fails for pull requests from forks

agateau-gg:
Do not run steps which require the GITGUARDIAN_API_KEY variable. These steps always fail because the variable is not available for pull-requests coming from forks.

Fixes #374

## Testing

This PR merges a branch from the GitGuardian/ggshield repository, so all steps should run.

PR #484 merges the same branch, but from my fork of the repository, and this time not all steps run. 
Update: looks like GitHub is confused and now shows the same results as for this PR. You can look at this [previous run](https://github.com/GitGuardian/ggshield/actions/runs/4243739433) to see skipped jobs (the GitHub action test jobs) and steps (the functional tests inside the build jobs). There is also this [run of the performance benchmark](https://github.com/GitGuardian/ggshield/actions/runs/4243739426) which is skipped too.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,11 @@
 name: Application Main Branch
 
+# About steps requiring the GITGUARDIAN_API_KEY:
+#
+# For security reasons, secrets are not available when a workflow is triggered by a pull request from a fork. This
+# causes all steps requiring the GITGUARDIAN_API_KEY to fail. To avoid this, we skip those steps when we are triggered
+# by a pull request from a fork.
+
 on:
   pull_request:
   push:
@@ -113,6 +119,8 @@ jobs:
           fail_ci_if_error: false
 
       - name: Run functional tests
+        # See note about steps requiring the GITGUARDIAN_API at the top of this file
+        if: ${{ !github.event.pull_request.head.repo.fork }}
         shell: bash
         run: |
           make functest GITGUARDIAN_API_KEY=${{ secrets.GITGUARDIAN_API_KEY }} GITGUARDIAN_API_URL=${{ secrets.GITGUARDIAN_API_URL }}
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -113,6 +113,10 @@ jobs:
           fail_ci_if_error: false
 
       - name: Run functional tests
+        # For security reasons, secrets are not available when running tests for a pull request from a fork. This causes
```
pierrelalanne, line 116: Should we duplicate the comment somewhere else ?

pierrelalanne (approved):
Looks good to me !  
Interestingly enough, GitHub seems to group the checks from both PRs [here](https://github.com/GitGuardian/ggshield/pull/485/checks) and we can see that it behaves as expected.

PR #488, opened 2023-02-23
fix: add :latest tag to image name before save

### LINKED ISSUE #468: `ggshield secret scan docker` scans all versions of images if `:latest` is omitted
## Environment

- ggshield version: 1.14.3
- Operating system (Linux, macOS, Windows): -
- Operating system version: -
- Python version: -

## Describe the bug

Steps to reproduce:

When scanning a Docker image using `ggshield secret scan docker` without adding the `:latest` tag, ggshield downloads **all** versions of the image.

1. Run command `ggshield secret scan docker gitguardian/ggshield:latest`
2. Notice it scanned 186 files and took 6s
3. Run command `ggshield secret scan docker gitguardian/ggshield`

Actual result:

GGShield scanned 188 files and took 35s.

Expected result:

GGShield should only have scanned 186 files in 6s.

The difference in number of files is not large here, but it could be on other images. The difference in execution time is massive.

## Explanation

Under the hood, GGShield runs `docker save <image_name> -o archive.tar`. Unlike other docker commands, `docker save` does not automatically append `:latest` if it's not present (see [`docker save` doc](https://docs.docker.com/engine/reference/commandline/save/))

## TODO

- [ ] Before running `docker save`, check if the image name includes a tag. If it's not the case, append `:latest` to it.

karo-fox:
Checks if image name is tagged, and adds :latest if it is not

Fixes #468

### DIFF
--- a/ggshield/scan/docker.py
+++ b/ggshield/scan/docker.py
@@ -41,6 +41,11 @@ FILEPATH_BANLIST_PATTERNS = {
 
 LAYER_TO_SCAN_PATTERN = re.compile(r"\b(copy|add)\b", re.IGNORECASE)
 
+IMAGE_NAME_PATTERN = (
+    r"^(?P<repository>[\w.\-_]+((?::\d+|)(?=/[a-z0-9._-]+/[a-z0-9._-]+))|)?(?:/|)"
+    r"(?P<image>[a-z0-9.\-_]+(?:/[a-z0-9.\-_]+|))(:(?P<tag>[\w.\-_]{1,127})?|)$"
+)
+
 
 class InvalidDockerArchiveException(Exception):
     pass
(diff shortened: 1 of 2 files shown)


ggshield/scan/docker.py
```diff
@@ -41,6 +41,11 @@
 
 LAYER_TO_SCAN_PATTERN = re.compile(r"\b(copy|add)\b", re.IGNORECASE)
 
+IMAGE_NAME_PATTERN = (
+    r"^(?P<repository>[\w.\-_]+((?::\d+|)(?=/[a-z0-9._-]+/[a-z0-9._-]+))|)?(?:/|)"
+    r"(?P<image>[a-z0-9.\-_]+(?:/[a-z0-9.\-_]+|))(:(?P<tag>[\w.\-_]{1,127})?|)$"
+)
```
agateau-gg, line 47: This looks a bit complicated, do you think it would be enough to only check if the end of the name looks like a tag? I think a regular expression like this would work `:[a-zA-Z0-9_][-.a-zA-Z0-9_]{0,127}$`.

I made a few changes compared to your expression, based on the [docker tag doc](https://docs.docker.com/engine/reference/commandline/tag/#description):
- Allow 128 chars, but do not allow `.` and `-` on first char
- Use `a-zA-Z0-9` instead of `\w` because `\w` matches `_` and matches more than `a-zA-Z` unless `re.ASCII` is set (it matches `é` for example)

Also, `IMAGE_NAME_PATTERN` should be the result of `re.compile(<the_pattern>)` for consistency with the other `*PATTERN` constants.

agateau-gg (changes requested):
Minor remark, looks good overall :+1:

ggshield/scan/docker.py
karo-fox: Done!

agateau-gg (approved):

agateau-gg:
Looks good now 👍🏻!

PR #489, opened 2023-02-24
fix: allow iac scan command to run with no argument

gg-mmill:
Allow `ggshield iac scan` to run with no provided argument (taking current directory as default)


agateau-gg (approved):
Looks good!

PR #493, opened 2023-03-13
Improve error message when a user has no API key set

pierrelalanne:
When a user has no API key set, we add instructions about how to authenticate to GitGuardian's dashboard and get one as part of the error message.  

We also reference our documentation and mention how to deal with the login if the user have an on-prem install.

### DIFF
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -20,7 +20,18 @@ def create_client_from_config(config: Config) -> GGClient:
             # This can happen when the user first tries the app and has not gone through
             # the authentication procedure yet. In this case, replace the error message
             # complaining about an unknown instance with a more user-friendly one.
-            raise APIKeyCheckError(e.instance, "GitGuardian API key is needed.")
+            raise APIKeyCheckError(
+                e.instance,
+                """A GitGuardian API key is needed to use ggshield.
+To get one, authenticate to your dashboard by running:
+
+    ggshield auth login
+
+If you are using an on-prem version of GitGuardian, \
+use the --instance option to point to it.
+Read the following documentation for more information: \
+https://docs.gitguardian.com/ggshield-docs/reference/auth/login""",
+            )
         else:
             raise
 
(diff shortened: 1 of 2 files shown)


ggshield/core/client.py
```diff
@@ -20,7 +20,16 @@ def create_client_from_config(config: Config) -> GGClient:
...
             # complaining about an unknown instance with a more user-friendly one.
-            raise APIKeyCheckError(e.instance, "GitGuardian API key is needed.")
+            raise APIKeyCheckError(
+                e.instance,
+                "A GitGuardian API key is needed to use ggshield.\n"
+                "To get one, authenticate to your dashboard by running :\n\n"
+                "\t`ggshield auth login`\n\n"
+                "If you are using an on-prem version of GitGuardian, "
+                "use the --instance option to point to it.\n"
+                "Read the following documentation for more information : "
```
agateau-gg, line 30:
```suggestion
                "To get one, authenticate to your dashboard by running:\n\n"
                "\t`ggshield auth login`\n\n"
                "If you are using an on-prem version of GitGuardian, "
                "use the --instance option to point to it.\n"
                "Read the following documentation for more information: "
```

No spaces before `:` in English :)

Also, maybe you can use a triple-quote string for the message?

agateau-gg (changes requested):

ggshield/core/client.py
pierrelalanne: Fixed 👌   
About the triple-quote, I'm not sure it is more readable... we would have something like that :

```
            raise APIKeyCheckError(
                e.instance,
                """A GitGuardian API key is needed to use ggshield.
To get one, authenticate to your dashboard by running:

    `ggshield auth login`

If you are using an on-prem version of GitGuardian, \
use the --instance option to point to it.
Read the following documentation for more information: \
https://docs.gitguardian.com/ggshield-docs/reference/auth/login
""",
            )
```

Let me know if you still want to go with the triple quotes, I'll do the changes.

agateau-gg: I think it's easier to imagine the end result: for example I didn't catch that there were back-ticks around `ggshield auth login`, which I would remove: the surrounding blank lines and the indentation are enough to understand it's a command, and not having the back-ticks makes it easier to copy-and-paste.

pierrelalanne: Makes sense.  
Here is how it looks : 
image

agateau-gg (approved):

PR #494, opened 2023-03-14
Fix reading the time we last checked for updates

### LINKED ISSUE #480: Config Error Messages Replace "-" With "_"

agateau-gg:
Do not fail with a strange error message.
    
Reading the date of the last check from `update_check.yaml` failed because the name of the key is `check-at`, not `check_at`. This used to work because `load_yaml_dict()` transparently turned `-` into `_`, but it does not do this anymore now that we fixed #480.
    
- Fix the key name.
- Move loading and saving to separate functions to make the code easier to test.
- Add more tests.

### DIFF
--- a/tests/unit/core/test_check_updates.py
+++ b/tests/unit/core/test_check_updates.py
@@ -7,7 +7,11 @@ from _pytest.monkeypatch import MonkeyPatch
 from pyfakefs.fake_filesystem import FakeFilesystem
 
 import ggshield.core
-from ggshield.core.check_updates import CACHE_FILE, check_for_updates
+from ggshield.core.check_updates import (
+    CACHE_FILE,
+    check_for_updates,
+    load_last_check_time,
+)
 
 
 @patch("requests.get")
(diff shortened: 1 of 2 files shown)


tests/unit/core/test_check_updates.py
```diff
@@ -103,3 +107,36 @@ def test_check_for_updates_does_nothing_if_cache_cant_be_saved(
...
+    """
+    GIVEN invalid content for the cache file
+    WHEN load_last_check_time() is called
+    THEN it does not raise any exception
+    AND it returns None
+    """
+    Path(CACHE_FILE).parent.mkdir(parents=True, exist_ok=False)
+    Path(CACHE_FILE).write_text(content)
+
+    assert load_last_check_time() is None
```
pierrelalanne, line 142: Should we use the [caplog fixture](https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture) to verify that the warnings were correctly logged ?

pierrelalanne (commented):
Minor remark, otherwise looks good to me

pierrelalanne (approved):

PR #497, opened 2023-03-21
Clean dependencies listed in the Pipfile

pierrelalanne:
We used to list dependencies of ggshield both in the Pipfile and in the setup.py.   

This can lead to discrepancies between the two files, and hence packages being not correctly shipped with gg-shield.  

To prevent this in the future, this PR gets rid of gg-shield dependencies in the Pipfile. Pipenv will install the package along with its dependencies anyway.

**Extra note:** We also had to pin pytest to avoid conflicting dependencies.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,19 +4,8 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [packages]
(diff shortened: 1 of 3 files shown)


agateau-gg (approved):
Looks good :+1:

PR #505, opened 2023-04-04
feat: reformat outputs to always show if the secret is known

alina-tuholukova-gg:


setup.py
```diff
@@ -39,7 +39,7 @@ def get_version() -> str:
         "marshmallow>=3.18.0,<3.19.0",
         "marshmallow-dataclass>=8.5.8,<8.6.0",
         "oauthlib>=3.2.1,<3.3.0",
-        "pygitguardian>=1.5.0,<1.6.0",
+        "pygitguardian @ git+https://github.com/GitGuardian/py-gitguardian.git@alina/include-incident-url",  # TODO fix
```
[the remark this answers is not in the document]
agateau-gg: The benchmark broke because it installs dependencies using `pipenv install --system` (see .github/workflows/perfbench.yml, in the "Install dependencies" step). I think you can fix this by replacing `--system` with `--skip-lock`.

changelog.d/20230404_144426_alina.tuholukova_add_known_secrets_url.md
```diff
@@ -0,0 +1,42 @@
...
+### Added
+
+- A bullet item for the Added category.
+
+-->
+
+### Changed
+
+- The output of the scan command includes now the information if the secret is known by user's GitGuardian
+  dashboard. If yes, the output also includes the incident url.
```
agateau-gg, line 23: I think this can be simplified to:

```suggestion
- `ggshield secret scan` output now includes a link to the incident if the secret is already known on the user's GitGuardian dashboard.
```

ggshield/scan/scanner.py
```diff
@@ -228,7 +230,7 @@ def _scan_chunk(
             self.client.multi_content_scan,
             documents,
             self.headers,
-            ignore_known_secrets=self.ignore_known_secrets,
+            ignore_known_secrets=True,
```
agateau-gg, line 233: If I am not mistaken, on the server side this argument is now ignored, is this correct? If so, I would rather remove it from here than pass an ignored value.

agateau-gg (changes requested):

ggshield/scan/scanner.py
alina-tuholukova-gg: it's not ignored. Finally we want the information about incident url to be returned only for ggshield, so using this argument will say gim to include this info

agateau-gg (changes requested):
Looks as good as it can be given the decided server behavior.

Just a minor remark on tests.

agateau-gg (approved):

PR #503, opened 2023-03-31
Refactor file reading

agateau-gg:
## Description

This PR paves the way to supporting server-side config and reading from various sources. It does so by refactoring the File class and introducing an abstract base class for scannable content, called Scannable.

Scannable has the following API:

- `url` property: Act as a unique identifier for the Scannable. May use custom protocols if required
- `path` property: "path" part of the API, returned as a `pathlib.Path`. Useful for local files
- `is_longer_than(self, size: int) -> bool`: returns true if the length of the *decoded* content is greater than `size`. Used by SecretScanner to decide if afile should be scanned or not
- `content` property: Returns the *decoded* content

File now inherits from Scannable and implements its API.

This PR also introduces StringScannable, which inherits from Scannable too. It is used for scannable content already loaded in memory. Individual patches of a commit are now represented using StringScannable.

## Performance

Interestingly, the performance bench results are very different between my machine and the CI.

CI results:

| command                                       | repository |         prod |      current |   delta |
| --------------------------------------------- | ---------- | -----------: | -----------: | ------: |
| iac scan --exit-zero .                        | Pillow     |  1.58s ±0.10 |  1.54s ±0.04 | -0.04 ≈ |
| iac scan --exit-zero .                        | compose    |  1.45s ±0.02 |  1.42s ±0.13 | -0.03 ≈ |
| iac scan --exit-zero .                        | leaky-repo |  0.97s ±0.04 |  1.01s ±0.07 | +0.04 ≈ |
| iac scan --exit-zero .                        | sqlite     |  0.73s ±0.01 |  0.74s ±0.01 | +0.01 ≈ |
| iac scan --exit-zero .                        | vue        |  1.10s ±0.01 |  1.08s ±0.03 | -0.02 ≈ |
| secret scan --exit-zero commit-range HEAD~9.. | Pillow     |  1.90s ±0.11 |  1.94s ±0.03 | +0.04 ≈ |
| secret scan --exit-zero commit-range HEAD~9.. | compose    |  1.91s ±0.03 |  1.97s ±0.02 | +0.06 ≈ |
| secret scan --exit-zero commit-range HEAD~9.. | leaky-repo |  1.77s ±0.06 |  1.84s ±0.04 | +0.07 ≈ |
| secret scan --exit-zero commit-range HEAD~9.. | sqlite     |  0.90s ±0.01 |  0.91s ±0.00 | +0.01 ≈ |
[...]

PR #506, opened 2023-04-06
Refactor file reading, Docker part

agateau-gg:
## Description

This PR is a follow-up for #503. It reworks the way we read the content of Docker images to avoid enforcing document size limits manually.

It also adds calls to `ggshield secret scan docker` to our performance benchmark.

After this change, `DOCUMENT_SIZE_THRESHOLD_BYTES` is only used by `SecretScanner`, except for tests. This puts us in a better position to make this value customizable for on-prem instances.

## Details

The work is done by introducing a `DockerContentScannable` class, which knows how to read content inside a Docker image archive (which is a tar archive containing other tar archives, one per layer)

It also introduces a `DockerFiles` class, which inherits from `Files`, and keeps a reference to the overall opened tar archive, so that it stays open.

Some of the code from `File` has been moved to helper static methods in `Scannable` so that `DockerContntScannable` can use them.

### DIFF
--- a/scripts/perfbench/run_cmd.py
+++ b/scripts/perfbench/run_cmd.py
@@ -7,7 +7,7 @@ import typing
 from dataclasses import asdict
 from pathlib import Path
 from shutil import which
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import click
 from perfbench_utils import (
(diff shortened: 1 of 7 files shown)


scripts/perfbench/run_cmd.py
```diff
@@ -127,6 +129,55 @@ def run_one_command(
     )
 
 
+def run_docker_command(
```
pierrelalanne, line 132: It looks like we duplicate a lot of code here 🤔

pierrelalanne (approved):
One small nit. Otherwise LGTM.

pierrelalanne (approved):

PR #509, opened 2023-04-17
Cache Docker successful scans

agateau-gg:
## Description

This PR makes `ggshield secret scan docker` remember the IDs of layers which did not contain policy breaks so that it does not scan them again.

This is implemented by storing the IDs of clean layers in a cache file on disk. The file name is based on the secrets engine version so that when a new version of the secrets engine is deployed the cache is invalidated.

## Important commits

1st commit rework scanning a bit to avoid loading the content of a layer too early: we don't want to load the content if the layer is known to be clean.

2nd commit implements the cache itself and make use of it in the scan implementation.

3rd commit factorizes some test data used by docker tests. It was actually not needed after all, but the code is cleaner this way so I kept it.

4th commit adds a functional test. Unfortunately it does not pass on Windows in our CI. I tried to reproduce locally but could not, so I skipped it.

7th commit modifies our benchmark to ensure we always measure the performance of `secret scan docker` with an empty cache.

### DIFF
--- a/ggshield/scan/docker.py
+++ b/ggshield/scan/docker.py
@@ -125,15 +125,14 @@ class LayerInfo:
 class DockerImage:
     def __init__(self, tar_file: tarfile.TarFile):
         self.tar_file = tar_file
-        self.manifest, self.config, self.config_scannable = _get_config(self.tar_file)
+        manifest, config, self.config_scannable = _get_config(self.tar_file)
+        self.layer_infos = list(
+            filter(_should_scan_layer, _get_layer_infos(manifest, config))
+        )
 
-    def get_layers(self) -> Iterable[Tuple[LayerInfo, Files]]:
-        for layer_info in filter(
-            _should_scan_layer, _get_layer_infos(self.manifest, self.config)
-        ):
-            scannables = list(_get_layer_files(self.tar_file, layer_info))
-            if scannables:
-                yield (layer_info, Files(scannables))
+    def get_layer(self, layer_info: LayerInfo) -> Files:
+        scannables = list(_get_layer_files(self.tar_file, layer_info))
+        return Files(scannables)
 
 
 def _get_config(archive: tarfile.TarFile) -> Tuple[Dict, Dict, Scannable]:
@@ -331,13 +330,17 @@ def docker_scan_archive(
                 scanner_ui=ui,
             )
 
-        for info, files in docker_image.get_layers():
+        for info in docker_image.layer_infos:
+            layer = docker_image.get_layer(info)
+            file_count = len(layer.files)
+            if file_count == 0:
+                continue
             print()
             display_heading(f"Scanning layer {info.get_id()}")
-            with RichSecretScannerUI(len(files.files)) as ui:
+            with RichSecretScannerUI(file_count) as ui:
                 results.extend(
                     scanner.scan(
-                        files.files,
+                        layer.files,
                         scanner_ui=ui,
                     )
                 )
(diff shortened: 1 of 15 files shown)


ggshield/scan/docker.py
```diff
@@ -336,13 +348,15 @@ def docker_scan_archive(
...
-            with RichSecretScannerUI(file_count) as ui:
-                results.extend(
-                    scanner.scan(
-                        layer.files,
-                        scanner_ui=ui,
-                    )
-                )
+            layer_id = info.get_id()
+            if layer_id in layer_id_cache:
+                display_heading(f"Skipping layer {layer_id}: already scanned")
```
pierrelalanne, line 353: It may be clearer to add that the layer did not contain any secret.

pierrelalanne: Also, this can become a bit verbose, the caching logic is somehow an implementation detail... shouldn't we not print this message by default ?

ggshield/scan/docker.py
```diff
@@ -336,13 +348,15 @@ def docker_scan_archive(
...
+            else:
+                display_heading(f"Scanning layer {info.get_id()}")
+                with RichSecretScannerUI(file_count) as ui:
+                    layer_results = scanner.scan(layer.files, scanner_ui=ui)
+                if not layer_results.has_policy_breaks:
```
pierrelalanne, line 358: A slightly tricky question. We could decide to cache layers that have no secret policy breaks only, instead of no policy breaks at all. This means caching more layers.  
On the other hand, ScanResult does not provide such a property, and it means implementing it by iterating over policy breaks to see if the ScanResult has secrets policy breaks. What do you think ?

pierrelalanne (commented):
The PR might not be 100% ready, but I had some time to look at it.

ggshield/scan/docker.py
agateau-gg: I think it is better to be explicit here because otherwise the user is going to be surprised by the very different duration of the scan for cached layers.

ggshield/scan/docker.py
agateau-gg: Assuming we did that: if we ever re-enable again checks for other policy breaks (which I hope we do at some point), then we would have to revisit this because we would potentially silently skip those policy breaks.

That's why I think it's safer to be strict on the conditions required to bypass the scan.

pierrelalanne: Alright, makes sense.

tests/unit/scan/test_scan_docker.py
```diff
@@ -103,15 +103,24 @@ def test_docker_archive(self, image_path: Path):
...
+            infos_and_layers = list(
+                filter(
+                    lambda info_and_layer: info_and_layer[1].files,
+                    ((x, image.get_layer(x)) for x in image.layer_infos),
+                )
+            )
+
+            layer_ids = [x.get_id() for x, _ in infos_and_layers]
+            expected_layer_ids = list(expected_files_for_layers.keys())
+            assert layer_ids == expected_layer_ids
```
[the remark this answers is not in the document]
pierrelalanne: I don't think this was modified.

pierrelalanne (approved):
One tiny nit remaining. Otherwise LGTM.

tests/unit/scan/test_scan_docker.py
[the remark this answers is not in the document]
agateau-gg: This is changed in another commit: https://github.com/GitGuardian/ggshield/pull/509/commits/309b04b0138c7d2b44d844b8ecac68ecdf68951f#diff-f7a5abb273effcb014889e97d2dcf8e089ecb65783b7103a2b38ca121894cb4cR103

PR #515, opened 2023-04-27
Split IaC and Secret output handler classes

agateau-gg:
## Description

IaC has its own output handler classes, but they inherit from the abstract OutputHandler base class, which is shared with secret classes. It turns out the outputs are quite different and do not benefit from being shared. The goal of this PR is to split the output handler classes.

## What has been done

This PR introduces the IaCOutputHandler base class and make iac output handler classes inherit from it.

The code is then split so that we go from this tree:

```
ggshield/output
├── gitlab_webui
│   └── gitlab_webui_output_handler.py
├── json
│   ├── iac_json_output_handler.py
│   ├── json_output_handler.py
│   └── schemas.py
├── output_handler.py
└── text
    ├── iac_text_output_handler.py
    ├── message.py
    ├── text_output_handler.py
    └── utils.py
```

To two separate trees. One for iac:

```
ggshield/iac/output
├── iac_json_output_handler.py
├── iac_output_handler.py
├── iac_text_output_handler.py
└── schemas.py
```

And one for secrets:

```
ggshield/secret/output
├── messages.py
├── schemas.py
├── secret_gitlab_webui_output_handler.py
├── secret_json_output_handler.py
├── secret_output_handler.py
└── secret_text_output_handler.py
```

The `ggshield/output` directory no longer exists.

## Next

This is the first step in splitting iac and secret. Next step is splitting ScanCollection.


pierrelalanne (commented):
Very clear.  
One naive question, do we plan to have somehow a common `OutputHandler` at some point to share some interfaces between IaC, secrets and others ?

pierrelalanne (approved):

agateau-gg:
> Very clear. One naive question, do we plan to have somehow a common `OutputHandler` at some point to share some interfaces between IaC, secrets and others ?

I don't think it would be useful because a `secret scan` command is never going to use an IaC output handler. It's more useful to have a shared toolbox of functions to display source code snippets with annotations, which multiple vertical commands can use. This is what we have in `core.text_utils`. This module could use some cleanup, but that will be for another time :)

PR #516, opened 2023-04-27
Add honeytoken create command

iXioN:
Add a new `ggshield honeytoken create` commande to create honeytokens for workspace manager

### DIFF
--- a/changelog.d/20230427_182700_antonin.lacombe_create_honeytoken.md
+++ b/changelog.d/20230427_182700_antonin.lacombe_create_honeytoken.md
@@ -1,0 +1,4 @@
+### Added
+
+- Add a new `ggshield honeytoken create` command to let you create honeytokens if enabled in your workspace.
+  Learn more about Honeytokens in https://www.gitguardian.com/honeytoken
--- a/ggshield/cmd/honeytoken/__init__.py
+++ b/ggshield/cmd/honeytoken/__init__.py
@@ -1,0 +1,12 @@
+from typing import Any
+
+import click
+
+from ggshield.cmd.common_options import add_common_options
+from ggshield.cmd.honeytoken.create import create_cmd
+
+
+@click.group(commands={"create": create_cmd})
+@add_common_options()
+def honeytoken_group(**kwargs: Any) -> None:
+    """Commands to work with honeytokens."""
--- a/ggshield/cmd/honeytoken/create.py
+++ b/ggshield/cmd/honeytoken/create.py
@@ -1,0 +1,105 @@
+import random
+import string
+from pathlib import Path
+from typing import Any, Dict, Optional
+
+import click
+from pygitguardian import GGClient
+from pygitguardian.models import Detail, HoneytokenResponse
+
+from ggshield.cmd.common_options import add_common_options
+from ggshield.core.client import create_client_from_config
+from ggshield.core.errors import UnexpectedError
+
+
+def _generate_random_honeytoken_name() -> str:
+    """Generate a honeytoken name based on a random string of eight alphanumeric characters"""
+    letters_and_digits = string.ascii_letters + string.digits
+    random_str = "".join(random.choice(letters_and_digits) for i in range(8))
+    return f"ggshield-{random_str}"
+
+
+def _dict_to_string(data: Dict, space: bool = False) -> str:
+    """Returns a string with 'key=value' for each key-value pair in the dictionary."""
+    space_char = " " if space else ""
+    return "\n".join([f"{k}{space_char}={space_char}{v}" for k, v in data.items()])
+
+
+@click.command()
+@click.option(
+    "--name",
+    "name",
+    required=False,
+    type=str,
+    help="give a name to your honeytoken. Default ggshield-xxxxxxxx",
+)
+@click.option(
+    "--type",
+    "type_",
+    required=True,
+    type=click.Choice(("AWS",)),
+)
+@click.option(
+    "--description", "description", required=False, type=str, help="250 characters max"
+)
+@click.option(
+    "-o",
+    "--output",
+    "output_file",
+    type=click.Path(
+        path_type=Path, file_okay=True, dir_okay=False, readable=True, writable=True
+    ),
+    required=False,
+)
+@add_common_options()
+@click.pass_context
+def create_cmd(
+    ctx: click.Context,
+    name: Optional[str],
+    type_: str,
+    description: Optional[str],
+    output_file: Optional[Path],
+    **kwargs: Any,
+) -> int:
+    """
+    Create a honeytoken
+    """
+    # if name is not given, generate a random one
+    if not name:
+        name = _generate_random_honeytoken_name()
+    client: GGClient = create_client_from_config(ctx.obj["config"])
+    response = client.create_honeytoken(name, type_, description)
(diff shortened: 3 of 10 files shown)


changelog.d/20230427_182700_antonin.lacombe_create_honeytoken.md
```diff
@@ -0,0 +1,3 @@
+### Added
+
+- Add a new `ggshield honeytoken create` allowing you to create honeytokens if enables in your workspace
```
agateau-gg, line 3:
```suggestion
- Add a new `ggshield honeytoken create` command to let you create honeytokens if enabled in your workspace.
```

Do we have a link to the feature we could add here?

ggshield/cmd/honeytoken/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+import click
+
+from ggshield.cmd.common_options import add_common_options
+from ggshield.cmd.honeytoken.create import create_cmd
+
+
+@click.group(commands={"create": create_cmd})
+@add_common_options()
+def honeytoken_group(**kwargs: Any) -> None:
+    """Commands to work with honeytoken."""
```
agateau-gg, line 12:
```suggestion
    """Commands to work with honeytokens."""
```

ggshield/core/config/config.py
```diff
@@ -130,12 +135,24 @@ def api_url(self) -> str:
     @property
     def dashboard_url(self) -> str:
         """
-        The dashboard URL to use to use the dashboard
+        The dashboard URL to use the dashboard
```
agateau-gg, line 138: Good catch 😅

tests/unit/cmd/honeytoken/test_honeytoken.py
```diff
@@ -0,0 +1,62 @@
...
+        [
+            "honeytoken",
+            "create",
+        ],
+    )
+    assert result.exit_code == ExitCode.USAGE_ERROR
+
+
+@my_vcr.use_cassette("test_honeytoken_creat_ok_no_name")
+def test_honeytoken_creat_ok_no_name(cli_fs_runner: CliRunner) -> None:
```
agateau-gg, line 27: Can you change "creat" to "create" here and below?

tests/unit/cmd/honeytoken/test_honeytoken.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+
+    result = cli_fs_runner.invoke(
+        cli,
+        [
+            "honeytoken",
+            "create",
+        ],
+    )
+    assert result.exit_code == ExitCode.USAGE_ERROR
```
agateau-gg, line 23: We have 2 helper functions for these tests: `assert_invoke_ok()` and `assert_invoke_exited_with()`. The benefit of using them is they print the full stdout and stderr if the result is not what is expected. Can you make the tests use them instead of  `assert`?

ggshield/core/config/config.py
```diff
@@ -130,12 +135,24 @@ def api_url(self) -> str:
...
         # TODO change when instance_name can be a name instead of just a URL
         return self.instance_name
 
+    @property
+    def workspace_dashboard_url(self) -> str:
+        """
+        The dashboard URL to use the dashboard with workspace information
+        It's the dashboard URL from the configured instance
+        """
+        account = self.auth_config.get_instance(self._cmdline_instance_name).account
```
agateau-gg, line 150: I think this should use `self.instance_name` instead of `self._cmdline_instance_name`, and then you don't need the change in `authconfig.get_instance()`. But don't spend time on this if we end up not adding this property as we discussed.

ggshield/cmd/honeytoken/create.py
```diff
@@ -0,0 +1,97 @@
...
+
+    if isinstance(response, Detail):
+        raise UnexpectedError(response.detail)
+    honeytoken: HoneytokenResponse = response
+
+    token_to_display = {}
+    for k, v in honeytoken.token.items():
+        token_to_display[f"{honeytoken.type_}_{k}".lower()] = v
+
+    if sys.stdout.isatty():
```
agateau-gg, line 82: No other code in ggshield has a different output depending on whether stdout is a terminal or not. I would prefer if we added a `--json` option to produce a machine-friendly output instead.

Given the code here, the implementation should just be:

```python
click.echo(json.dumps(token_to_display, indent=2))
```

We even have a `json_output_option_decorator` decorator ready to use for this :). It's in the badly-named `core.utils` module.

agateau-gg (changes requested):

ggshield/cmd/honeytoken/create.py
iXioN: This is re requirement from the product, it will be used to generate honeytokens and commit them directly from command line

ggshield/cmd/honeytoken/create.py
```diff
@@ -0,0 +1,97 @@
...
+
+from ggshield.cmd.common_options import add_common_options
+from ggshield.core.client import create_client_from_config
+from ggshield.core.config import Config
+from ggshield.core.errors import UnexpectedError
+from ggshield.core.utils import urljoin
+
+
+def _generate_random_honeytoken_name() -> str:
+    """Generate a honeytoken name base on random string of six alphanumeric characters"""
```
gg-mmill, line 18:
```suggestion
    """Generate a honeytoken name based on a random string of eight alphanumeric characters"""
```

ggshield/cmd/honeytoken/create.py
```diff
@@ -0,0 +1,97 @@
...
+    letters_and_digits = string.ascii_letters + string.digits
+    random_str = "".join(random.choice(letters_and_digits) for i in range(8))
+    return f"ggshield-{random_str}"
```
gg-mmill, line 21: Should it be ggtoken or something like that ?

iXioN: the goal is to know that this honeytokens have been created via ggshield

ggshield/cmd/honeytoken/create.py
```diff
@@ -0,0 +1,97 @@
...
+    if not isinstance(response, (Detail, HoneytokenResponse)):
+        raise UnexpectedError("Unexpected honeytoken response")
+
+    if isinstance(response, Detail):
+        raise UnexpectedError(response.detail)
+    honeytoken: HoneytokenResponse = response
+
+    token_to_display = {}
+    for k, v in honeytoken.token.items():
+        token_to_display[f"{honeytoken.type_}_{k}".lower()] = v
```
agateau-gg, line 80: We do not use single-letter variables, except in list or dict comprehensions. But you can rewrite it as one :)

```suggestion
    token_to_display = {f"{honeytoken.type_}_{k}".lower(): v for k, v in  honeytoken.token.items()}
```

agateau-gg (changes requested):

ggshield/cmd/honeytoken/create.py
agateau-gg: I am "unresolving" this because the issue I raised has not been addressed. Can you look into it?

iXioN: I just implemented it, if you can review it it would be great

agateau-gg (approved):
Looks good now 👍🏻.

PR #523, opened 2023-05-10
fix(cli): # Fix the common scan option --ignore-known-secrets

Walz:
## Description

The flag `--ignore-known-secrets` does not work because the path of the corresponding value in the config is now `config.secret.ignore_known_secrets` since b37d1a1b52a4ed591d55fccc986c85cf9f545521


agateau-gg (approved):
Good catch, thanks!

PR #528, opened 2023-05-15
Make it possible to run the release script from any release branch

agateau-gg:
Our `scripts/release` branch refused to run unless it was on the main branch (or `--dev-mode` was set). This PR makes it accept running on other release branches.

### DIFF
--- a/scripts/release
+++ b/scripts/release
@@ -23,7 +23,9 @@ ACTION_PATHS = [ROOT_DIR / "actions" / x / "action.yml" for x in ["iac", "secret
 CASSETTES_DIR = ROOT_DIR / "tests" / "unit" / "cassettes"
 
 # The branch this script must be run from, except in dev mode.
-RELEASE_BRANCH = "main"
+RELEASE_BRANCHES = {"^main$", r"^\d+\.\d+\.x$"}
+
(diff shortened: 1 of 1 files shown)


scripts/release
```diff
@@ -23,7 +23,9 @@ ACTION_PATHS = [ROOT_DIR / "actions" / x / "action.yml" for x in ["iac", "secret
 CASSETTES_DIR = ROOT_DIR / "tests" / "unit" / "cassettes"
 
 # The branch this script must be run from, except in dev mode.
-RELEASE_BRANCH = "main"
+RELEASE_BRANCHES = {"main", r"\d+\.\d+\.x"}
```
pierrelalanne, line 26: One nit: We rely on the way we use the pattern here (`fullmatch` in our case). It might be slightly more explicit to have `"^main$"` as a pattern, and use `re.match` instead of fullmatch.

pierrelalanne (approved):
One small nit, otherwise all good for me.

scripts/release
agateau-gg: You mean `re.search()` right? because `re.match()` would only require `"main$"` 😅 (Python re module is a bit messy if you ask me...)

Going to change it.

PR #529, opened 2023-05-16
fix(pre-receive): #417 Execute pre-receive within a sub-process to accurately enforce the timeout

### LINKED ISSUE #417: pre-receive timeout is not precise enough
## Environment

- ggshield version: 1.13.6
- Operating system (Linux, macOS, Windows): Linux
- Operating system version: Ubuntu 20.04
- Python version: 3.8

## Describe the bug

`ggshield secret scan pre-receive` is supposed to stop after a predefined timeout, which defaults to 4.5s by default. However it can sometimes take more than twice this time to stop.

Steps to reproduce:

1. Setup a local bare repository with a pre-receive hook calling ggshield
2. Clone it
3. Make your internet connection slow enough to hit the timeout (For example use [wondershaper](https://github.com/magnific0/wondershaper) to limit the bandwidth of your network interface to 2KBps)
4. Measure the timeout

Actual result:

It took 10s for the hook to stop on my machine.

Expected result:

The difference between the defined timeout and the actual delay should be less than 1s.

## Possible explanation

I suspect this is because the timeout is implemented using a thread, and Python GIL gets in the way. If this is the case, then that bug can be solved by running the scan in a different process.

Walz:
Fixes #417 

I first tried the solution proposed in the issue, but it did not work. Threads can not be killed and Python will always wait for all of its threads to end before exiting. If an API call is stuck, nothing can be done from the same process.

I've implemented a more brutal solution: move the logic within a sub-process which can be killed when it takes too long. To test it, I've added a functional test with a fake server which won't response in the middle of a request.

In the unit tests, to not rewrite the existing tests, the `multiprocessing.Process` is mock to run in the main-process and `multiprocessing.Pipe` to not use a unix socket.

### DIFF
--- a/ggshield/cmd/secret/scan/prereceive.py
+++ b/ggshield/cmd/secret/scan/prereceive.py
@@ -1,24 +1,26 @@
-import _thread as thread
 import logging
+import multiprocessing
 import os
+import re
 import sys
-import threading
-from types import TracebackType
-from typing import Any, List, Optional, Tuple, Type
+from multiprocessing.connection import Connection
+from typing import Any, List, Optional, Set, Tuple
 
 import click
+from pygitguardian import GGClient
 
 from ggshield.cmd.secret.scan.secret_scan_common_options import (
     add_secret_scan_common_options,
     create_output_handler,
 )
 from ggshield.core.cache import ReadOnlyCache
+from ggshield.core.config import Config
 from ggshield.core.errors import UnexpectedError, handle_exception
 from ggshield.core.git_shell import get_list_commit_SHA, git
 from ggshield.core.text_utils import display_error
 from ggshield.core.utils import EMPTY_SHA, PRERECEIVE_TIMEOUT
 from ggshield.scan import ScanContext, ScanMode
-from ggshield.secret.output import SecretGitLabWebUIOutputHandler
+from ggshield.secret.output import SecretGitLabWebUIOutputHandler, SecretOutputHandler
 from ggshield.secret.output.messages import remediation_message
 from ggshield.secret.repo import scan_commit_range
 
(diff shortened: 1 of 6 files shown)


ggshield/cmd/secret/scan/prereceive.py
```diff
@@ -195,35 +205,28 @@ def prereceive_cmd(
...
+        display_error("\nPre-receive hook took too long")
+        process.kill()
         return 0
-    except Exception as error:
-        return handle_exception(error, config.verbose)
+
+    if process.exitcode == 0:
+        return_code: int = receiver.recv()
+        return return_code
+    return process.exitcode
```
agateau-gg, line 232: If I understand the multiprocess.Process doc correctly, we should be able to pass the exit code via the child process exit code: make the child finish with `sys.exit(return_code)` and forward it here.

This would simplify the code (and the tests) by removing the need for the Pipe. Have you tried this?

tests/functional/conftest.py
```diff
@@ -13,3 +18,49 @@
...
+            self.send_header("content-type", "application/json")
+            self.send_header("Content-Length", str(len(content)))
+            self.end_headers()
+            self.wfile.write(content)
+        else:
+            self.send_response(418)
+
+    def do_POST(self):
+        if "multiscan" in self.path:
+            content = b'{"detail":"Sorry, I overslept!"}'
```
agateau-gg, line 40: 🤣

tests/functional/secret/test_scan_prereceive.py
```diff
@@ -100,3 +103,44 @@ def test_scan_prereceive_push_force(tmp_path: Path) -> None:
...
+
+    # Prepare to mock process.run to access the stderr of the push
+    stderr = b""
+    subprocess_run = subprocess.run
+
+    def run_and_copy_stderr(*args, **kwargs):
+        process = subprocess_run(*args, **kwargs)
+        nonlocal stderr
+        stderr = process.stderr
+        return process
```
agateau-gg, line 135: That's some nifty code :) but have you tried using pytest [capsys](https://docs.pytest.org/en/7.3.x/reference/reference.html#pytest.capture.capsys) to capture stderr? It would be less intrusive: as much as possible, functional tests should run the app without modifying it.

agateau-gg (changes requested):
A few minor remarks, but this looks really great.

+1 for the slow server!

tests/functional/secret/test_scan_prereceive.py
Walz: I have edited `git_shell` to `log.debug` the stderr of the command. It simplifies the test, because we can use `caplog`. It will also make debugging easier and won't affect users unless they use `--debug`. WDYT ?

agateau-gg: Good idea 👍🏻.

agateau-gg (approved):
Looks good, thanks!

PR #535, opened 2023-05-19
fix: use GITGUARDIAN_DOMAINS in api_to_dashboard

ppetit-gitguardian:
As discussed earlier, this method is not often used but I think it should have the same logic as `dashboard_to_api_url()`.


agateau-gg (approved):
Looks good, and this is consistent with `dashboard_to_api_url()`.

PR #543, opened 2023-05-26
fix(json_output_handler): Compute the ignore sha before censoring the matches

Walz:
## Description

The field `ignore_sha` in the JSON output differs from the one outputted by the text output when `--show-secrets` is not used, because it is computed after censoring the matches.

The PR fixes the bug and add both unit and functional tests to prevent regression.

### DIFF
--- a/tests/unit/secret/output/test_json_output.py
+++ b/tests/unit/secret/output/test_json_output.py
@@ -7,6 +7,7 @@ import pytest
 from pytest_voluptuous import Partial, S
 from voluptuous import Optional, Required, validators
 
+from ggshield.core.filter import leak_dictionary_by_ignore_sha
 from ggshield.core.utils import Filemode
 from ggshield.scan import Commit, ScanContext, ScanMode, StringScannable
 from ggshield.secret import Result, Results, SecretScanCollection, SecretScanner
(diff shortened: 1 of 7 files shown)


tests/unit/secret/output/test_json_output.py
```diff
@@ -147,6 +148,15 @@ def test_json_output(client, cache, name, input_patch, expected_exit_code):
                 json_flat_results
             )
 
+        # all ignore sha should be in the output
+        assert all(
+            ignore_sha in json_flat_results
+            for result in results.results
+            for ignore_sha in leak_dictionary_by_ignore_sha(
+                result.scan.policy_breaks
+            ).keys()
```
agateau-gg, line 157: nit: you can remove `.keys()` here: the loop iterate on dictionary keys by default.

tests/unit/secret/output/test_text_output.py
```diff
@@ -135,6 +136,14 @@ def test_leak_message(result_input, snapshot, show_secrets, verbose):
 
     snapshot.assert_match(output)
 
+    # all ignore sha should be in the output
+    assert all(
+        ignore_sha in output
+        for ignore_sha in leak_dictionary_by_ignore_sha(
+            result_input.scan.policy_breaks
+        ).keys()
```
agateau-gg, line 144: same here

agateau-gg (changes requested):
Looks good! Just 2 minor nitpicks in the tests.

agateau-gg (approved):

PR #547, opened 2023-05-29
Fix GitLab web UI output handler

pierrelalanne:
## Context  
We recently experienced a bug where one of our users could not correctly visualize the results of a `scan pre-receive` in the GitLab UI. Especially, an error message was still displayed when pushing known secrets with the `ignore-known-secrets` option. Indeed, this option had not been ported to the GitLab web UI handler.  
While reproducing the bug, we realized that a regression now simply made it impossible to visualize the `GL-HOOK-ERR` message in the UI.   

## What we did
This PR does mainly two things:  
1. We adapt GitLab Web UI output handler to behave as expected when used with the `ignore-known-secrets` option  
To do so, we add an `ignore_known_secrets` property to the `SecretGitLabWebUIOutputHandler` object. We also modify `_process_scan_impl` to make sure that only relevant secrets incidents are raised by the GitLab web UI.  

2. We adapt GitLab Web UI output handler to fix the issue of unreadable error messages in the UI.
According to the [documentation](https://docs.gitlab.com/ee/administration/server_hooks.html#custom-error-messages), the `GL-HOOK-ERR` message needs to be output to `stderr` or `stdout`. We force this behavior by re-implementing the `process_scan` method.

## Testing the feature
In all the following tests, we have beed using a self hosted version of GitLab. A `pre-receive` hook was deployed globally. 
We mostly used two invalid secrets: 
- One is known: `apikey=key-bb3dcff54bdae26f4f4a4d3b93d148da`
- One is new    : `apikey=key-bb3dcff54bdae26f4f4a4d3b93d148db`

### Regular pre-receive hook
1. With no secret
Commit is accepted as expected ✅ 
image
image

2. With a known secret
Commit is rejected as expected, no matter if the secret is already known ✅ 
image

### pre-receive hook with ignore-known-secrets flag
1. With no secret
Commit is accepted as expected ✅ 
image
image

2. With a known secret
Commit is accepted as expected, known secret is ignored ✅ 
image
image

3. With a new secret
Commit is rejected as expected ✅ 
image

Note that we also tested the Web IDE:
image

## Other considerations  
I don't understand why we had empty messages. This should have been fixed in [this PR](https://github.com/GitGuardian/ggshield/pull/172), but when checking out the corresponding commit (00826608e) and testing, I get an unexpected error or an empty message.

image

### DIFF
--- a/ggshield/secret/output/secret_gitlab_webui_output_handler.py
+++ b/ggshield/secret/output/secret_gitlab_webui_output_handler.py
@@ -30,25 +30,46 @@ class SecretGitLabWebUIOutputHandler(SecretOutputHandler):
     See https://docs.gitlab.com/ee/administration/server_hooks.html#custom-error-messages
     """
 
-    def __init__(self, show_secrets: bool = False) -> None:
-        super().__init__(show_secrets=show_secrets, verbose=False)
+    def __init__(
+        self, show_secrets: bool = False, ignore_known_secrets: bool = False
+    ) -> None:
+        super().__init__(
+            show_secrets=show_secrets,
+            verbose=False,
+            ignore_known_secrets=ignore_known_secrets,
+        )
 
-    def _process_scan_impl(self, scan_collection: SecretScanCollection) -> str:
-        results = list(scan_collection.get_all_results())
-        if not results:
+    def _process_scan_impl(self, scan: SecretScanCollection) -> str:
+        results = list(scan.get_all_results())
+        # If no secrets or no new secrets were found
+        if not results or (self.ignore_known_secrets and not scan.has_new_secrets):
             return ""
 
-        policy_breaks = []
+        policy_breaks_to_report = []
         for result in results:
-            policy_breaks += result.scan.policy_breaks
+            if not self.ignore_known_secrets:
+                # Populate GL-HOOK-ERR with all policy breaks found
+                policy_breaks_to_report += result.scan.policy_breaks
+            else:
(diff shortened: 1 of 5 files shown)


ggshield/secret/output/secret_gitlab_webui_output_handler.py
```diff
@@ -30,25 +30,46 @@ class SecretGitLabWebUIOutputHandler(SecretOutputHandler):
...
+                        policy_breaks_to_report.append(policy_break)
 
         # Use a set to ensure we do not report duplicate incidents.
         # (can happen when the secret is present in both the old and the new version of
         # the document)
-        formatted_policy_breaks = {format_policy_break(x) for x in policy_breaks}
+        formatted_policy_breaks = {
+            format_policy_break(x) for x in policy_breaks_to_report
+        }
+        break_count = len(formatted_policy_breaks)
```
agateau-gg, line 80: This line is duplicated

agateau-gg (changes requested):
Minor remarks, looks good!

ggshield/secret/output/secret_gitlab_webui_output_handler.py
agateau-gg: Never mind, it's not 😅. Don't know what's wrong with my eyes.

pierrelalanne:
@Walz, are you confident to integrate this in the `1.16.0` ? If so, I let your merge 🙂

PR #554, opened 2023-06-05
Improve consistency of --help messages

agateau-gg:
Click default help messages are sentences: they start with an upper-case letter and end with a point. Make our help messages follow the same format for consistency.


Walz (approved):
I was going to add a comment about the missing `.` in the help of `--json` but I remembered the other PR which factorize all `--json` option.

LGTM.

agateau-gg:
> I was going to add a comment about the missing `.` in the help of `--json` but I remembered the other PR which factorize all `--json` option.
> 
> LGTM.

Ah right, this commit was originally on top of the `--json` PR, but I rebased it :)

PR #558, opened 2023-06-08
Subcommands in IaC scan: allow commit scan

Paul-GitGuardian:


ggshield/cmd/iac/scan/scan.py
```diff
@@ -97,7 +78,14 @@ def scan_cmd(
...
     output_handler = output_handler_cls(verbose=config.user_config.verbose)
-    return output_handler.process_scan(scan)
+
+    return (
+        output_handler.process_scan(IaCPathScanCollection(scan.id, scan.result))
+        if since is None
+        else output_handler.process_diff_scan(
+            IaCDiffScanCollection(scan.id, scan.result)
+        )
+    )
```
Paul-GitGuardian, line 88: @cyrvan-b 
This looks ugly to me. What do you think? Here is the situation:
- There are two result classes, IaCPathScanCollection and IaCDiffScanCollection, for the two scan types. They both inherit an abstract class IaCScanCollection
- There are (currently) two output handlers, IaCTextOutputHandler & IaCJSONOutputHandler, both inheriting an abstract class IaCOutputHandler.

Both types of scan result can use any of the output handler classes.

Paul-GitGuardian: OK I think having two different process methods in the output handlers will be necessary anyway. Adding a `process` method to the ScanCollection classes might work. I'm still open to hearing your opinion on this.

Paul-GitGuardian: We decided not do this this, as it would cause circular dependencies.
Instead, one function will be created in scan.py for each scan result type.

ggshield/core/dirs.py
```diff
@@ -7,15 +8,15 @@
...
+def get_config_dir() -> Any:
     try:
         # See tests/conftest.py for details
         return str(os.environ["GG_CONFIG_DIR"])
     except KeyError:
         return user_config_dir(appname=APPNAME, appauthor=APPAUTHOR)
 
 
-def get_cache_dir() -> str:
+def get_cache_dir() -> Any:
```
agateau-gg, line 19: The changes in these files should not be necessary now that `os.environ` is turned into an `str`.

agateau-gg (commented):
Hey 👋🏻, I know this PR is a work in progress, but I figured it would be helpful to get some early feedback:

The `cmd.utils` package should be moved to something like `core.clickutils`. The sub-directories in `cmd` are supposed to contain commands. See [doc/dev/architecture.md](https://github.com/GitGuardian/ggshield/blob/main/doc/dev/architecture.md) for more details.

Have you considered using a single tar to store both before and after states? It would result in a smaller payload. To verify this I created two dirs with checkouts of ggshield, one on `main`, one on `v1.15.0`. I then created individual tarballs of the dirs, and a tarball of the two dirs. The single tarball is much smaller than the sum of the two individual tarballs:

|name|size|
|-|-|
|ggshield.tar.gz|3.8 MB|
|ggshield2.tar.gz|4.1 MB|
|ggshield-combined.tar.gz|4.4 MB|

This can be implemented by injecting the tar object in `tar_from_ref_and_filepaths()` instead of creating it inside the function. Bonus point if it's a simple Protocol, making it easier to test.

Finally, I hate to be that guy, but... this PR is getting too big. Consider splitting it, maybe using an integration branch if it's not in a releasable state yet.

Paul-GitGuardian:
Than you for your early feedback!
Regarding the single tar: this was considered by @gg-teolohrer, but the current solution was easier to implement in GIM (as tar hashes are used in the cache).
However, seing the large difference in the tarball sizes, we decided to go back to this solution.

ggshield/cmd/iac/scan/__init__.py
```diff
@@ -1,12 +1,79 @@
...
+    """
+    try:
+        max_commits = os.getenv("GITGUARDIAN_MAX_COMMITS_FOR_HOOK", None)
+        if max_commits is not None:
+            return int(max_commits)
+    except BaseException as e:
+        display_error(f"Unable to parse GITGUARDIAN_MAX_COMMITS_FOR_HOOK: {str(e)}")
+        return None
+
+    return None
```
GG-HH, line 79: While this can be useful for secrets (we scan each commit), this should be useless for IaC because we only scan two states whatever the number of commits is.

ggshield/core/clickutils/default_command_group.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any, List
+
+import click
+
+
+# Use https://github.com/click-contrib/click-default-group instead?
```
GG-HH, line 6: @agateau-gg I could use your help on this one.

I don't think it is a good idea to use the lib because it adds a dependency for a very small code snippet that hasn't changed for years.

However, I think it is a good idea to keep the reference for documentation and to find help if we have an issue with it.

ggshield/core/clickutils/option_group.py
```diff
@@ -0,0 +1,50 @@
+from typing import Any, List, Mapping
+
+import click
+
+
+# All options subscribed form a group
+# One of them must be present (and no more)
+
+
+class OptionGroup(click.Option):
```
GG-HH, line 10: Should we add tests for such code ?

ggshield/iac/iac_scan_models.py
```diff
@@ -32,24 +38,53 @@ class IaCDiffScanResult(Base):
...
-    )
-    result = IaCDiffScanResult(
-        scan.id,
-        scan.type,
-        scan.iac_engine_version,
-        entities_with_incidents=IaCDiffScanEntities([], [], []),
-    )
-    result.status_code = 200
-    return result
+class MockClient(GGClient):
```
GG-HH, line 41: @agateau-gg my idea here is to subclass the existing client until we are sure about the implementation and then to migrate into pygitguardian.
In this MR, the "real" routes to GIM are not implemented.

GG-HH (commented):
Sounds almost good to me. A few questions for our friend @agateau-gg :)

GG-HH (approved):

PR #562, opened 2023-06-12
docs(honeytoken): Honeytoken improve help textes and error 

iXioN:
This PR improve the `honeytoken create` command, 
The help text has been updated by the product team. 
The error displayed when a 403 is received has changed to help the user to setup honeytokens on their account.

### DIFF
--- a/tests/unit/cmd/honeytoken/test_honeytoken.py
+++ b/tests/unit/cmd/honeytoken/test_honeytoken.py
@@ -60,3 +60,28 @@ def test_honeytoken_create_ok(cli_fs_runner: CliRunner) -> None:
         ],
     )
     assert_invoke_ok(result)
+
+
+@my_vcr.use_cassette("test_honeytoken_create_error_403")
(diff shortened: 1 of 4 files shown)


tests/unit/cmd/honeytoken/test_honeytoken.py
```diff
@@ -60,3 +60,28 @@ def test_honeytoken_create_ok(cli_fs_runner: CliRunner) -> None:
...
+            "create",
+            "--description",
+            "description",
+            "--type",
+            "AWS",
+            "--name",
+            "test_honey_token_error_403",
+        ],
+    )
+    assert result.exit_code == ExitCode.UNEXPECTED_ERROR
```
agateau-gg, line 86: You can use `assert_invoke_exited_with()` instead of checking the exit code. The advantage of `assert_invoke_exited_with()` is that in case of failure, it automatically dumps the stderr and stdout of the command, making debugging easier.

agateau-gg (changes requested):

agateau-gg (approved):
Looks good, thanks!

PR #565, opened 2023-06-14
feat(hmsl): add HMSL commands

ppetit-gitguardian:
This PR adds a new `hmsl` subgroup with new commands to interact with HasMySecretLeaked.

### DIFF
--- a/ggshield/cmd/hmsl/decrypt.py
+++ b/ggshield/cmd/hmsl/decrypt.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any, Dict, Iterator, List, TextIO, cast
+from typing import Any, Dict, Iterable, Iterator, TextIO, cast
 
 import click
 
@@ -49,10 +49,10 @@ def decrypt_cmd(path: str, mapping_file: TextIO, json_output: bool, **_: Any) ->
     # Decrypt the secrets thanks to the hashes contained in mapping
     try:
         secrets = decrypt(input, mapping)
-        show_results(list(secrets), mapping, json_output)
+        # Display the secrets
+        show_results(secrets, mapping, json_output)
     except (json.JSONDecodeError, TypeError):
         raise ParseError("Invalid format in input file.")
-    # Display the secrets
 
     return 0
 
@@ -70,7 +70,7 @@ def load_mapping(mapping_file: TextIO) -> Dict[str, str]:
         if line == "":
             # Skip empty lines
             continue
-        hash, _, name = line.partition(":")
+        hash, name = line.split(":", maxsplit=1)
         mapping[hash] = name
     return mapping
 
@@ -96,11 +96,12 @@ def decrypt(input: TextIO, mapping: Dict[str, str]) -> Iterator[Secret]:
 
 
 def show_results(
-    secrets: List[Secret], names: Dict[str, str], json_output: bool
+    secrets: Iterable[Secret], names: Dict[str, str], json_output: bool
 ) -> None:
     """
     Display the secrets.
     """
+    secrets = list(secrets)
     display_info(f"Found {len(secrets)} leaked {pluralize('secret', len(secrets))}.")
     data = [
         {
@@ -112,7 +113,7 @@ def show_results(
         for secret in secrets
     ]
     if json_output:
-        click.echo(json.dumps(data, indent=4))
+        click.echo(json.dumps(data))
     else:
         for i, secret in enumerate(data):
             click.echo(TEMPLATE.format(number=i + 1, **secret))
--- a/ggshield/cmd/hmsl/decrypt.py
+++ b/ggshield/cmd/hmsl/decrypt.py
@@ -1,0 +1,122 @@
+import json
+from typing import Any, Dict, Iterator, List, TextIO, cast
+
+import click
+
+from ggshield.cmd.common_options import add_common_options, json_option
+from ggshield.cmd.hmsl.fingerprint import input_arg
+from ggshield.core.errors import ParseError
+from ggshield.core.text_utils import display_info, pluralize
+from ggshield.hmsl import Match, Secret
+from ggshield.hmsl.crypto import make_hint
+
+
+# Types and constants
+TEMPLATE = """
+> Secret {number}
+Secret name: "{name}"
+Secret hash: "{hash}"
+Distinct locations: {count}
+First occurrence:
+    URL: "{url}"
+"""
+
+
+# Command
+
+
+@click.command()
+@add_common_options()
+@click.option(
+    "--mapping",
+    "-m",
+    "mapping_file",
+    type=click.File("r"),
+    default="mapping.txt",
+    show_default=True,
+    help="File containing the hashes and their names.",
+)
+@json_option
+@input_arg
+def decrypt_cmd(path: str, mapping_file: TextIO, json_output: bool, **_: Any) -> int:
+    """
+    Decrypt and show secrets information.
+    """
+    # Opens the file or stdin
+    input = cast(TextIO, click.open_file(path, "r"))
+    mapping: Dict[str, str] = load_mapping(mapping_file)
+
+    # Decrypt the secrets thanks to the hashes contained in mapping
+    try:
+        secrets = decrypt(input, mapping)
+        show_results(list(secrets), mapping, json_output)
+    except (json.JSONDecodeError, TypeError):
+        raise ParseError("Invalid format in input file.")
+    # Display the secrets
+
+    return 0
+
+
+# Helper methods
+
+
+def load_mapping(mapping_file: TextIO) -> Dict[str, str]:
+    """
+    Load the mapping from the file.
+    """
+    mapping: Dict[str, str] = {}
+    for line in mapping_file:
+        line = line.strip()
+        if line == "":
+            # Skip empty lines
+            continue
+        hash, _, name = line.partition(":")
+        mapping[hash] = name
+    return mapping
+
+
+def decrypt(input: TextIO, mapping: Dict[str, str]) -> Iterator[Secret]:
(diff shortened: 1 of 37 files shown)


ggshield/cmd/hmsl/decrypt.py
```diff
@@ -0,0 +1,122 @@
...
+        {
+            "name": names.get(secret.hash) or secret.hash,
+            "hash": secret.hash,
+            "count": secret.count,
+            "url": secret.url,
+        }
+        for secret in secrets
+    ]
+    if json_output:
+        click.echo(json.dumps(data, indent=4))
```
agateau-gg, line 115: We usually do not indent JSON output: it's intended for machines.

ggshield/cmd/hmsl/fingerprint.py
```diff
@@ -0,0 +1,174 @@
...
+)
+
+naming_strategy_option = click.option(
+    "--naming-strategy",
+    "-n",
+    type=click.Choice(list(NAMING_STRATEGIES.keys())),
+    default="key",
+    show_default=True,
+    help="""Strategy to generate the hints.
+            With "censored", only the firsts and last characters are displayed.
```
agateau-gg, line 66:
```suggestion
            With "censored", only the first and last characters are displayed.
```

ggshield/cmd/hmsl/fingerprint.py
```diff
@@ -0,0 +1,174 @@
...
+        secret = line.strip()
+        if secret == "":
+            # Skip empty lines
+            continue
+        if source == SourceType.FILE:
+            yield SecretWithKey(value=secret, key=None)
+
+
+def prepare(
+    secrets: List[SecretWithKey],
```
agateau-gg, line 137: unless your function needs random access, it's better to make arguments take `Iterable[T]` instead of `List[T]` because it makes it possible for callers to pass generators.

ggshield/cmd/hmsl/fingerprint.py
```diff
@@ -0,0 +1,174 @@
...
+# Command
+
+input_arg = click.argument(
+    "path",
+    type=click.Path(
+        exists=True,
+        dir_okay=False,
+        readable=True,
+        allow_dash=True,
+    ),
```
agateau-gg, line 56: You can use `path_type=Path` here this way the argument is a `Path` instance instead of an `str`. It's more expressive, and you get handy methods like `.open()` or `.read_text()`.

tests/unit/cmd/hmsl/test_check.py
```diff
@@ -0,0 +1,70 @@
...
+    THEN we should not find a match
+    """
+    non_matching_secret = str(uuid4())
+    with open("secrets.txt", "w") as f:
+        f.write(non_matching_secret + "\n")
+
+    result = cli_fs_runner.invoke(cli, ["hmsl", "check", "secrets.txt"])
+
+    assert "Found 0 leaked secrets" in result.output
+    assert_invoke_ok(result)
```
agateau-gg, line 32: It's better to call `assert_invoke_ok()` before the other asserts because in case of failure, it prints all of stdout and stderr.

tests/unit/cmd/hmsl/test_check.py
```diff
@@ -0,0 +1,70 @@
...
+@my_vcr.use_cassette
+def test_hmsl_check_random_secret(cli_fs_runner: CliRunner) -> None:
+    """
+    GIVEN a random
+    WHEN running the check command on it
+    THEN we should not find a match
+    """
+    non_matching_secret = str(uuid4())
+    with open("secrets.txt", "w") as f:
+        f.write(non_matching_secret + "\n")
```
agateau-gg, line 27: Please use the `tmp_path` fixture to create temporary files.

tests/unit/cmd/hmsl/test_check.py
```diff
@@ -0,0 +1,70 @@
...
+    # The cassettes where recorded before the service was operational in production
+    monkeypatch.setenv(
+        "GITGUARDIAN_HMSL_URL", "https://hasmysecretleaked.staging.gitguardian.tech"
+    )
+
+
+@my_vcr.use_cassette
+def test_hmsl_check_random_secret(cli_fs_runner: CliRunner) -> None:
+    """
+    GIVEN a random
```
agateau-gg, line 21:
```suggestion
    GIVEN a random secret
```

😉

tests/unit/hmsl/test_hmsl_utils.py
```diff
@@ -0,0 +1,60 @@
...
+def test_get_token(isolated_fs, gim: MagicMock):
+    """
+    GIVEN a logged in session to GIM
+    WHEN requesting a token
+    THEN we receive it
+    """
+    config = Config()
+    assert load_token_from_disk() is None
+    token = get_token(config)
+    assert gim.assert_called_once
```
agateau-gg, line 59:
```suggestion
    assert gim.assert_called_once()
```

tests/unit/hmsl/test_hmsl_utils.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.fixture
+def hmsl_no_env_vars(monkeypatch):
+    monkeypatch.delenv("GITGUARDIAN_SAAS_URL", raising=False)
+    monkeypatch.delenv("GITGUARDIAN_SAAS_API_KEY", raising=False)
+    monkeypatch.delenv("GITGUARDIAN_HMSL_URL", raising=False)
+    monkeypatch.delenv("GITGUARDIAN_HMSL_AUDIENCE", raising=False)
+
+
+@pytest.fixture
+def gim(monkeypatch):
```
agateau-gg, line 20: Took me a while to understand what this fixture does. Can you rename it to `create_jwt_mock`?

agateau-gg (changes requested):

ggshield/cmd/hmsl/decrypt.py
```diff
@@ -0,0 +1,122 @@
...
+    Decrypt and show secrets information.
+    """
+    # Opens the file or stdin
+    input = cast(TextIO, click.open_file(path, "r"))
+    mapping: Dict[str, str] = load_mapping(mapping_file)
+
+    # Decrypt the secrets thanks to the hashes contained in mapping
+    try:
+        secrets = decrypt(input, mapping)
+        show_results(list(secrets), mapping, json_output)
```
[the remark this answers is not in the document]
ppetit-gitguardian: show_results starts by printing the number of secrets found, so we need to exhaust the generator first. I guess we could also do it at the start of show_results().

ggshield/cmd/hmsl/fingerprint.py
ppetit-gitguardian: does it still works with stdin? (`allow_dash=True`)

ggshield/hmsl/client.py
```diff
@@ -0,0 +1,195 @@
...
+
+    @overload
+    def check(
+        self, hashes: Iterable[str], *, full_hashes: bool, decrypt: bool
+    ) -> Iterator[Match]:
+        ...
+
+    def check(  # type: ignore
+        self, hashes: Iterable[str], *, full_hashes: bool = False, decrypt: bool = True
+    ):
```
agateau-gg, line 106: I think the code would be nicer if `check()` did not try to do too much. From what I understand, the different use cases are the following:

- The `hmsl check` command calls `check(full_hashes=$arg, decrypt=True)`. It wants Secret instances.

- The `hmsl query` command calls `check(full_hashes=$arg, decrypt=False)`. It only needs dicts, and dumps them as JSON on stdout.


I propose splitting `HMSLClient.check()` in two functions, like this:

```python
class HMSLClient:
    def check(hashes, *, full_hashes: bool) -> Iterable[Secret]
    
    def query(hashes, *, full_hashes: bool) -> Iterable[Dict[str, Any]]
``` 

To avoid implementing the batch iteration twice, we can add a Python 3.7 friendly backport of the upcoming [itertools.batched()](https://docs.python.org/3.12/library/itertools.html#itertools.batched) to core.utils (would be useful in other places too):

```python
def batched(iterable: Iterable[T], batch_size: int) -> Iterable[Tuple[T, ...]]:
    it = iter(iterable)
    while True:
        batch = tuple(islice(it, batch_size))
        if batch:
            yield batch
        else:
            return
```

This makes it possible to implement `HMSLClient.check` like this:

```python
    def check():
        ...
        for batch in batched(hashes, batch_size):
            if full_hashes:
                yield from self.check_hashes(batch).secrets
            else:
                hints = {make_hint(hash): hash for hash in batch}
                for match in response.matches:
                    if match.hint in hints:
                        hash = hints[match.hint]
                        yield match.decrypt(hash)
```                

And `HMSLClient.query` like this:

```python
    def query():
        ...
        for batch in batched(hashes, batch_size):
            if full_hashes:
                yield from map(asdict, self.check_hashes(batch).secrets)
            else:
                yield from map(asdict, response.matches)
```

ggshield/hmsl/utils.py
```diff
@@ -0,0 +1,104 @@
...
+
+    try:
+        client = create_client(
+            api_url=config.saas_api_url,
+            api_key=config.saas_api_key,
+            allow_self_signed=config.allow_self_signed,
+        )
+        audience = config.hmsl_audience
+    except (MissingTokenError, AuthExpiredError):
+        logger.debug("No API key found, using HMSL anonymously")
```
agateau-gg, line 50: nit
```suggestion
        logger.debug("No API key found, using HMSL anonymously.")
```

agateau-gg (changes requested):

agateau-gg (approved):
Looks good ✨! Merging it.

PR #568, opened 2023-06-16
feat(sca): add client with compute_sca_files method

gg-mmill:
Changes targetting `main-sca`

### DIFF
--- a/ggshield/sca/client.py
+++ b/ggshield/sca/client.py
@@ -1,0 +1,42 @@
+from dataclasses import dataclass
+from typing import Dict, List, Optional, Union, cast
+
+import marshmallow_dataclass
+from pygitguardian.client import GGClient, is_ok, load_detail
+from pygitguardian.models import Base, BaseSchema, Detail, FromDictMixin
+
+
+@dataclass
+class ComputeSCAFilesResult(Base, FromDictMixin):
+    sca_files: List[str]
+    potential_siblings: List[str]
+
+
+ComputeSCAFilesResult.SCHEMA = cast(
+    BaseSchema,
+    marshmallow_dataclass.class_schema(ComputeSCAFilesResult, base_schema=BaseSchema)(),
+)
+
+
+class SCAClient:
+    def __init__(self, client: GGClient):
+        self._client = client
+
+    def compute_sca_files(
+        self,
+        touched_files: List[str],
+        extra_headers: Optional[Dict[str, str]] = None,
+    ) -> Union[Detail, ComputeSCAFilesResult]:
+        response = self._client.post(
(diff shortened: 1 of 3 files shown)


ggshield/sca/client.py
```diff
@@ -0,0 +1,42 @@
...
+        extra_headers: Optional[Dict[str, str]] = None,
+    ) -> Union[Detail, ComputeSCAFilesResult]:
+        response = self._client.get(
+            endpoint="compute_sca_files",
+            params={"touched_files": touched_files},
+            extra_headers=extra_headers,
+        )
+        result: Union[Detail, ComputeSCAFilesResult]
+        if is_ok(response):
+            result = response.json()
```
agateau-gg, line 37: Should probably be
```suggestion
            result = ComputeSCAFilesResult.from_dict(response.json())
```

gg-mmill:
I had trouble using the staging for testing, so using localhost for now

tests/unit/sca/conftest.py
```diff
@@ -0,0 +1,15 @@
+from os.path import dirname, join, realpath
+
+import vcr
+
+
+my_vcr = vcr.VCR(
```
GG-HH, line 6: Why don't you use the existing configuration ?

tests/unit/sca/test_client.py
```diff
@@ -0,0 +1,16 @@
+from pygitguardian import GGClient
+
+from ggshield.sca.client import ComputeSCAFilesResult, SCAClient
+from tests.unit.sca.conftest import my_vcr
+
+
+class TestSCAClient:
+    @my_vcr.use_cassette
```
GG-HH, line 8: The cassette is not included in the commits.

GG-HH (changes requested):

ggshield/sca/client.py
```diff
@@ -0,0 +1,43 @@
...
+        self._client = client
+
+    def compute_sca_files(
+        self,
+        touched_files: List[str],
+        extra_headers: Optional[Dict[str, str]] = None,
+    ) -> Union[Detail, ComputeSCAFilesResult]:
+        response = self._client.get(
+            endpoint="sca/compute_sca_files/",
+            params={"touched_files": ",".join(touched_files)},
```
agateau-gg, line 32: A bit out of scope, but I feel this endpoint should be a POST with a `{ "touched_files": [...}}` payload instead of a GET joining all files in a single parameter.

gg-mmill:
I have addressed the changes, could you review again ?

agateau-gg (approved):
Looks good to me 👍🏻.

GG-HH (approved):

PR #571, opened 2023-06-19
feat(sca): add empty scan commands

gg-mmill:
Not adding test for commands for now, I don't think there really is a point.


GG-HH (approved):
Nit: we could add the "shortcuts" for pre-commit and pre-push, as in SCA.

PR #573, opened 2023-06-20
feat(sca): #PLC-761 add sca full scan output schema

xblanchot-gg:
Defined the same output schema as the one defined in backend.

I don't think we need any schema for the input, we should do the same as what is done for IaC. Do you agree @amascia-gg ?

### DIFF
--- a/ggshield/sca/sca_scan_models.py
+++ b/ggshield/sca/sca_scan_models.py
@@ -1,0 +1,80 @@
+from dataclasses import dataclass, field
+from datetime import datetime
+from typing import List, Optional, cast
+
+import marshmallow_dataclass
+from pygitguardian.models import Base, BaseSchema, FromDictMixin
+from typing_extensions import Literal
+
+
+@dataclass
+class SCAVulnerability(Base, FromDictMixin):
+    severity: str
+    summary: str
+    cve_ids: List[str] = field(default_factory=list)
+    created_at: Optional[datetime] = None
+    fixed_version: Optional[str] = None
+
+
+SCAVulnerability.SCHEMA = cast(
+    BaseSchema,
+    marshmallow_dataclass.class_schema(SCAVulnerability, base_schema=BaseSchema)(),
+)
+
+SCADependencyType = Literal["direct", "transitive"]
+
+
+@dataclass
+class SCAVulnerablePackageVersion(Base, FromDictMixin):
+    package_full_name: str
+    version: str
+    ecosystem: str
+    dependency_type: Optional[SCADependencyType] = None
+    vulns: List[SCAVulnerability] = field(default_factory=list)
+
+
+SCAVulnerablePackageVersion.SCHEMA = cast(
+    BaseSchema,
+    marshmallow_dataclass.class_schema(
+        SCAVulnerablePackageVersion, base_schema=BaseSchema
+    )(),
+)
+
+
+@dataclass
+class SCALocationVulnerability(Base, FromDictMixin):
+    location: str
+    package_vulns: List[SCAVulnerablePackageVersion] = field(default_factory=list)
+
+
+SCALocationVulnerability.SCHEMA = cast(
+    BaseSchema,
+    marshmallow_dataclass.class_schema(
+        SCALocationVulnerability, base_schema=BaseSchema
+    )(),
+)
+
+
+@dataclass
+class SCAScanAllOutput(Base, FromDictMixin):
+    scanned_files: List[str] = field(default_factory=list)
+    found_package_vulns: List[SCALocationVulnerability] = field(default_factory=list)
+
+
+SCAScanAllOutput.SCHEMA = cast(
+    BaseSchema,
+    marshmallow_dataclass.class_schema(SCAScanAllOutput, base_schema=BaseSchema)(),
+)
+
+
+@dataclass
+class SCAScanDiffOutput(Base, FromDictMixin):
+    scanned_files: List[str] = field(default_factory=list)
+    added_vulns: List[SCALocationVulnerability] = field(default_factory=list)
+    removed_vulns: List[SCALocationVulnerability] = field(default_factory=list)
+
+
+SCAScanDiffOutput.SCHEMA = cast(
+    BaseSchema,
+    marshmallow_dataclass.class_schema(SCAScanDiffOutput, base_schema=BaseSchema)(),
+)


ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+
+import marshmallow_dataclass
+from pygitguardian.models import Base, BaseSchema, FromDictMixin
+
+
+@dataclass
+class SCAVulnerability(Base, FromDictMixin):
+    severity: str
+    summary: str
+    cve_ids: List[str] = field(default_factory=list)
```
amascia, line 12:
```suggestion
    cve_ids: list[str] = field(default_factory=list)
```

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+    created_at: Optional[str] = None
+    fixed_version: Optional[str] = None
+
+
+SCAVulnerabilitySchema = cast(
```
amascia, line 17: why do you need this additional variable?

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+@dataclass
+class SCAVulnerablePackageVersion(Base, FromDictMixin):
+    full_name: str
+    version: str
+    ecosystem: str
+    dependency_type: str
+    vulns: List[SCAVulnerability] = field(default_factory=list)
+
+
+SCAVulnerablePackageVersionSchema = cast(
```
amascia, line 33: Same as above

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+import marshmallow_dataclass
+from pygitguardian.models import Base, BaseSchema, FromDictMixin
+
+
+@dataclass
+class SCAVulnerability(Base, FromDictMixin):
+    severity: str
+    summary: str
+    cve_ids: List[str] = field(default_factory=list)
+    created_at: Optional[str] = None
```
amascia, line 13:
```suggestion
    created_at: datetime | None
```

I think we can do this with ggshield's python version. Also this is a datetime object in backend

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+
+SCAVulnerability.SCHEMA = SCAVulnerabilitySchema()
+
+
+@dataclass
+class SCAVulnerablePackageVersion(Base, FromDictMixin):
+    full_name: str
+    version: str
+    ecosystem: str
+    dependency_type: str
```
amascia, line 29: This is different in backend

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+    marshmallow_dataclass.class_schema(SCAVulnerablePackageVersion, BaseSchema),
+)
+
+SCAVulnerablePackageVersion.SCHEMA = SCAVulnerablePackageVersionSchema()
+
+
+@dataclass
+class SCALocationVulnerability(Base, FromDictMixin):
+    filepath: str
+    package_vulns: List[SCAVulnerablePackageVersion] = field(default_factory=list)
```
amascia, line 44:
```suggestion
    package_vulns: List[SCAVulnerablePackageVersion] = []
```

Can't we just do this?

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+    Type[BaseSchema],
+    marshmallow_dataclass.class_schema(SCALocationVulnerability, BaseSchema),
+)
+
+SCALocationVulnerability.SCHEMA = SCALocationVulnerabilitySchema()
+
+
+@dataclass
+class SCAFullScanOutput(Base, FromDictMixin):
+    location_vulns: List[SCALocationVulnerability] = field(default_factory=list)
```
amascia, line 57: There also is `scanned_files` parameter here

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+SCAFullScanOutputSchema = cast(
+    Type[BaseSchema],
+    marshmallow_dataclass.class_schema(SCAFullScanOutput, BaseSchema),
+)
+
+SCAFullScanOutput.SCHEMA = SCAFullScanOutputSchema()
```
amascia, line 65: You could also add `SCAScanCommitOutputSchema`, see backend

ggshield/sca/output/schemas.py
GG-HH: Jumping in on that topic: for the typing, we must stick to the "old" way to ensure that ggshield is compatible with previous versions of Python (3.7+), so you cannot use: `|`, "dict", "list" etc...

What I don't understand is why `created_at` is a `str` instead of a `datetime` ? Is is cast later ?

ggshield/sca/output/schemas.py
agateau-gg: Yes, some code in pygg does declare a FooSchema variable, that is kept because removing it would break API compatibility, see ServerMetadata for a recent example which does what Achille suggests.

ggshield/sca/output/schemas.py
agateau-gg: No, that's not the same, see https://docs.python.org/3/library/dataclasses.html#default-factory-functions.

ggshield/sca/output/schemas.py
xblanchot-gg: Added !

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,78 @@
...
+    found_package_vulns: list[SCALocationVulnerability] = field(default_factory=list)
+
+
+SCAScanAllOutput.SCHEMA = cast(
+    BaseSchema,
+    marshmallow_dataclass.class_schema(SCAScanAllOutput, base_schema=BaseSchema)(),
+)
+
+
+class SCAScanDiffOutput(Base, FromDictMixin):
```
agateau-gg, line 69: You're missing a `@dataclass` here, I think

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,78 @@
...
+    created_at: Optional[datetime] = None
+    fixed_version: Optional[str] = None
+
+
+SCAVulnerability.SCHEMA = cast(
+    BaseSchema,
+    marshmallow_dataclass.class_schema(SCAVulnerability, base_schema=BaseSchema)(),
+)
+
+SCADependencyType = Literal["direct", "transitive"]
```
agateau-gg, line 23: Would be nice if we could use an enum here, but maybe that's annoying to deserialize? 🤔

ggshield/sca/output/schemas.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+@dataclass
+class SCAVulnerablePackageVersion(Base, FromDictMixin):
+    package_full_name: str
+    version: str
+    ecosystem: str
+    dependency_type: Optional[SCADependencyType] = None
```
agateau-gg, line 31: When can we have `dependency_type` set to None?

agateau-gg (changes requested):

ggshield/sca/output/schemas.py
xblanchot-gg: We parse dependency files in `mescal` to declare a dependency as `direct` or `transitive`, but sometimes we do not know. We then set it to `None`.

ggshield/sca/output/schemas.py
xblanchot-gg: Looks like we can make it work easily by doing `class SCADependencyType(str, Enum):` , [see here](https://stackoverflow.com/questions/24481852/serialising-an-enum-member-to-json), but serialization will directly transform it into a `str`.

ggshield/sca/output/schemas.py
xblanchot-gg: Thanks, done

ggshield/sca/output/schemas.py
agateau-gg: Based on your other answer, am I correct when I say these models are going to be used to parse Mescal responses?

If this is correct then what matters here is *de*-serialization. According to the stackoverflow answer you posted, using an enum would not help as one would need to manually turn the string into the enum (Unless Marshmallow can help here, but I doubt it). So `SCADependencyType` can stay the way it is, except... these models should not be in an `sca.output` package. If you look at `secret.output.schema`, it only contains models used to implement the json output. These would another case of models which start in ggshield and finish in py-gitguardian.

agateau-gg (changes requested):

ggshield/sca/output/schemas.py
agateau-gg: Yes, sounds good.

agateau-gg (approved):

PR #569, opened 2023-06-19
Iac scan text and json output

cyrvan-b:
This PR is supposed to implement text and json output for iac diff scan.

PR #575, opened 2023-06-21
fix(iac-scan): exit-zero option was ignored for both iac scans

Paul-GitGuardian:
How it works:
We connect the exit-zero option for iac  scans to the existing callback for secret scans.
The option is then handled in main.py

Depends on #569 for some tests to work

### DIFF
--- a/ggshield/cmd/common_options.py
+++ b/ggshield/cmd/common_options.py
@@ -140,6 +140,17 @@ _check_for_updates = click.option(
 )
 
 
+exit_zero_option = click.option(
+    "--exit-zero",
+    is_flag=True,
+    default=None,
+    envvar="GITGUARDIAN_EXIT_ZERO",
+    help="Always return a 0 (non-error) status code, even if incidents are found."
+    "The env var GITGUARDIAN_EXIT_ZERO can also be used to set this option.",
+    callback=create_config_callback("exit_zero"),
+)
+
+
 def add_common_options() -> Callable[[AnyFunction], AnyFunction]:
     def decorator(cmd: AnyFunction) -> AnyFunction:
         _verbose_option(cmd)
--- a/ggshield/cmd/iac/scan/iac_scan_common_options.py
+++ b/ggshield/cmd/iac/scan/iac_scan_common_options.py
@@ -14,7 +14,11 @@ from typing import Any, Callable, Sequence
 
 import click
 
-from ggshield.cmd.common_options import add_common_options, json_option
+from ggshield.cmd.common_options import (
+    add_common_options,
+    exit_zero_option,
+    json_option,
+)
 from ggshield.core.client import create_client_from_config
 from ggshield.core.config.config import Config
 from ggshield.core.filter import init_exclusion_regexes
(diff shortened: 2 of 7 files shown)


ggshield/cmd/iac/scan/iac_scan_common_options.py
```diff
@@ -26,7 +30,10 @@
 _exit_zero_option = click.option(
     "--exit-zero",
     is_flag=True,
-    help="Always return 0 (non-error) status code.",
+    default=None,
```
cyrvan-b, line 33: I think that the default value should be False rather than None.

Paul-GitGuardian: I agree. However, when setting the default to `False`, the option flag isn't working anymore.
In other parts of the code (in secrets), flags always have a default value of None.
It seems some functions do not work with a non-None default.

I suggest we keep the None default for now.

cyrvan-b (approved):

GG-HH (changes requested):

tests/unit/cmd/iac/test_scan_common.py
```diff
@@ -0,0 +1,39 @@
...
+
+from ggshield.cmd.main import cli
+from tests.repository import Repository
+from tests.unit.conftest import _IAC_SINGLE_VULNERABILITY, my_vcr
+
+
+COMMANDS = [["scan"], ["scan", "all"], ["scan", "diff", "--ref", "HEAD~1"]]
+
+
+@my_vcr.use_cassette("test_iac_scan_exit_zero")
```
GG-HH, line 15: Isn't there an issue with the cassettes here ?
Here it will use the same cassette for each parameter of the parametrize.
This means that two of your cassettes above are useless.

What you can do is use `my_vcr` as a context manager and add the name of the cassette in the parameters. (You should be able to use the same cassette when exit zero is given and when it is not.)

ggshield/cmd/common_options.py
```diff
@@ -140,13 +140,25 @@ def log_file_callback(
...
+
+
 def add_common_options() -> Callable[[AnyFunction], AnyFunction]:
     def decorator(cmd: AnyFunction) -> AnyFunction:
         _verbose_option(cmd)
         _debug_option(cmd)
         _log_file_option(cmd)
         _allow_self_signed_option(cmd)
         _check_for_updates(cmd)
+        _exit_zero_option(cmd)
```
GG-HH, line 161: This changes the behaviour because it will be added to all the commands. I am not sure that we want to do this.
We should declare the `_exit_zero_option` here but actually use it only where it was used before.

GG-HH (changes requested):

tests/unit/cmd/iac/test_scan_common.py
```diff
@@ -0,0 +1,54 @@
...
+from tests.repository import Repository
+from tests.unit.conftest import _IAC_SINGLE_VULNERABILITY, my_vcr
+
+
+SCAN_COMMAND = ["scan"]
+SCAN_ALL_COMMAND = ["scan", "all"]
+SCAN_DIFF_COMMAND = ["scan", "diff", "--ref", "HEAD~1"]
+
+
+@my_vcr.use_cassette("test_iac_scan_exit_zero")
```
GG-HH, line 17: you forgot to remove this line.

GG-HH (changes requested):

tests/unit/cmd/iac/test_scan_common.py
Paul-GitGuardian: sorry...

GG-HH (approved):

PR #577, opened 2023-06-21
tests(iac-scan): add functional tests for all iac scan commands

Paul-GitGuardian:
Depends on #576 and #569


GG-HH:
Much better than what the ticket specified! Thank you!

GG-HH (approved):

PR #578, opened 2023-06-21
feat(sca): #PLC-763 sca scan all command

xblanchot-gg:
This PR implements SCA scan all command logic.
- input: a directory
- filters sca related files with a call to SCA compute files API
- (NOT IMPLEMENTED) launches the SCA scan with a call to SCA public api (not merged in backend for now)
- returns the raw_results (output handling will be done later)

### DIFF
--- a/ggshield/cmd/sca/scan.py
+++ b/ggshield/cmd/sca/scan.py
@@ -1,9 +1,18 @@
+import re
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, List, Optional, Set
 
 import click
+from pygitguardian.client import _create_tar
 
-from ggshield.core.text_utils import display_warning
+from ggshield.core.client import create_client_from_config
+from ggshield.core.config import Config
+from ggshield.core.errors import APIKeyCheckError, UnexpectedError
+from ggshield.core.text_utils import display_error, display_info, display_warning
+from ggshield.sca.client import ComputeSCAFilesResult, SCAClient
+from ggshield.sca.file_selection import get_all_files_from_sca_paths
+from ggshield.sca.sca_scan_models import SCAScanAllOutput, SCAScanParameters
+from ggshield.scan import ScanContext, ScanMode
 
 
 def display_sca_beta_warning(func):
@@ -57,14 +66,120 @@ def scan_diff_cmd(
 
 
 @scan_group.command(name="all")
+@click.argument(
+    "directory",
+    type=click.Path(exists=True, readable=True, path_type=Path, file_okay=False),
+    required=False,
+)
 @click.pass_context
 @display_sca_beta_warning
 def scan_all_cmd(
     ctx: click.Context,
     directory: Optional[Path],
     **kwargs: Any,
-) -> int:
+) -> SCAScanAllOutput:
     """
     Scan a directory for SCA vulnerabilities.
     """
-    return 0
+    if directory is None:
+        directory = Path().resolve()
+
+    # Adds client and required parameters to the context
+    update_context(ctx)
+
+    result = sca_scan_all(ctx, directory)
+
+    # TODO, should be handled by an output_handler in the future
+    return result
+
+
+def update_context(ctx: click.Context) -> None:
+    """
+    Basic implementation, should be populated with required parameters later.
+    """
+    config: Config = ctx.obj["config"]
+    ctx.obj["client"] = create_client_from_config(config)
+
+    # Empty for now
+    ctx.obj["exclusion_regexes"] = set()
+
+
+def sca_scan_all(ctx: click.Context, directory: Path) -> SCAScanAllOutput:
+    """
+    Scan an entire directory for SCA Vulnerabilities.
+
+    - List SCA related files with a first call to SCA compute files API
+    - Create a tar archive with the required files contents
+    - Launches the scan with a call to SCA public API
+    """
+    client = SCAClient(ctx.obj["client"])
+
+    sca_filepaths = get_sca_scan_all_filepaths(
+        directory=directory,
+        exclusion_regexes=ctx.obj["exclusion_regexes"],
+        verbose=ctx.obj["config"].verbose,
+        client=client,
+    )
+
+    if len(sca_filepaths) == 0:
+        display_info("No file to scan.")
+        # Not an error, return an empty SCAScanAllOutput
+        return SCAScanAllOutput()
+
+    # empty for now
+    scan_parameters = SCAScanParameters()
+
+    tar = _create_tar(directory, sca_filepaths)
+
+    # Call to full scan API and get results
+    scan_result = client.sca_scan_directory(
+        tar,
+        scan_parameters,
+        ScanContext(
+            command_path=ctx.command_path,
+            scan_mode=ScanMode.SCA_DIRECTORY,
+        ).get_http_headers(),
+    )
+
+    if not isinstance(scan_result, SCAScanAllOutput):
+        if scan_result.status_code == 401:
+            raise APIKeyCheckError(client.base_uri, "Invalid API key.")
+        display_error("Error scanning.")
+        display_error(str(scan_result))
+        raise UnexpectedError("Unexpected error while performing SCA scan all.")
+
+    return scan_result
+
+
+def get_sca_scan_all_filepaths(
+    directory: Path,
+    exclusion_regexes: Set[re.Pattern],
+    verbose: bool,
+    client: SCAClient,
+) -> List[str]:
+    """
+    Retrieve SCA related files of a directory.
+    First get all filenames that are not in blacklisted directories, then calls SCA compute files
+    API to filter SCA related files.
+    """
+    all_filepaths = get_all_files_from_sca_paths(
+        path=directory,
+        exclusion_regexes=exclusion_regexes,
+        verbose=verbose,
+        # If the repository is a git repository, ignore untracked files
+        ignore_git=True,
+    )
+
+    # API Call to filter SCA files
+    response = client.compute_sca_files(touched_files=all_filepaths)
+
+    if not isinstance(response, ComputeSCAFilesResult):
+        if response.status_code == 401:
+            raise APIKeyCheckError(client.base_uri, "Invalid API key.")
+        display_error("Error while filtering SCA related files.")
+        display_error(str(response))
+        raise UnexpectedError("Unexpected error while filtering SCA related files.")
+
+    # Only sca_files field is useful in the case of a full_scan,
+    # all the potential files already exist in `all_filepaths`
+    return response.sca_files
--- a/ggshield/sca/client.py
+++ b/ggshield/sca/client.py
@@ -1,82 +1,24 @@
-from dataclasses import dataclass
-from datetime import datetime
-from typing import Dict, List, Optional, Union, cast
+from typing import Dict, List, Optional, Union
 
-import marshmallow_dataclass
+import requests
 from pygitguardian.client import GGClient, is_ok, load_detail
-from pygitguardian.models import Base, BaseSchema, Detail, FromDictMixin
+from pygitguardian.models import Detail
 
-
-@dataclass
-class ComputeSCAFilesResult(Base, FromDictMixin):
-    sca_files: List[str]
-    potential_siblings: List[str]
-
-
-ComputeSCAFilesResult.SCHEMA = cast(
-    BaseSchema,
-    marshmallow_dataclass.class_schema(ComputeSCAFilesResult, base_schema=BaseSchema)(),
-)
-
-
-@dataclass
-class ExposedVulnerability(Base, FromDictMixin):
-    cve_ids: List[str]
-    severity: str
-    created_at: Optional[datetime]
-    fixed_version: Optional[str]
-    summary: str
-
-
-ExposedVulnerability.SCHEMA = cast(
-    BaseSchema,
-    marshmallow_dataclass.class_schema(ExposedVulnerability, base_schema=BaseSchema)(),
-)
-
-
-@dataclass
-class PackageVulnerability(Base, FromDictMixin):
-    package_full_name: str
-    version: str
-    ecosystem: str
-    dependency_type: Optional[str]
-    vulns: List[ExposedVulnerability]
-
-
-PackageVulnerability.SCHEMA = cast(
-    BaseSchema,
-    marshmallow_dataclass.class_schema(PackageVulnerability, base_schema=BaseSchema)(),
-)
-
-
-@dataclass
-class LocationOutput(Base, FromDictMixin):
-    location: str
-    package_vulns: List[PackageVulnerability]
-
-
-LocationOutput.SCHEMA = cast(
-    BaseSchema,
-    marshmallow_dataclass.class_schema(LocationOutput, base_schema=BaseSchema)(),
-)
-
-
-@dataclass
-class SCAScanDiffResult(Base, FromDictMixin):
-    scanned_files: List[str]
-    added_vulns: List[LocationOutput]
-    removed_vulns: List[LocationOutput]
-
-
-SCAScanDiffResult.SCHEMA = cast(
-    BaseSchema,
-    marshmallow_dataclass.class_schema(SCAScanDiffResult, base_schema=BaseSchema)(),
+from ggshield.sca.sca_scan_models import (
+    ComputeSCAFilesResult,
+    SCAScanAllOutput,
+    SCAScanDiffOutput,
+    SCAScanParameters,
 )
 
 
 class SCAClient:
     def __init__(self, client: GGClient):
         self._client = client
+
+    @property
+    def base_uri(self):
+        return self._client.base_uri
 
     def compute_sca_files(
         self,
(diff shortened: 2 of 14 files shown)


xblanchot-gg:
@gg-mmill , @amascia-gg still a draft, but can you take a quick look on this when you have a moment ?

ggshield/sca/filter.py
```diff
@@ -0,0 +1,60 @@
...
+    "site-packages",
+    ".idea",
+    "node_modules",
+    ".mypy_cache",
+    ".pytest_cache",
+    ".hypothesis",
+)
+
+
+def get_all_files_from_sca_potential_paths(
```
gg-mmill, line 27: Why "potential" paths ?

Also the docstring is wrong - but I see this is coming from IaC

gg-mmill (commented):
Sounds good, seems like we'll share quite some many utils with iac, we should probably do something about this

ggshield/cmd/sca/scan.py
```diff
@@ -45,8 +52,84 @@ def scan_all_cmd(
...
     """
-    return 0
+    if directory is None:
+        directory = Path().resolve()
+
+    result = sca_scan_all(ctx, directory)
+    return result
+
+
+def sca_scan_all(ctx: click.Context, directory: Path):
```
tchaumeny, line 66: Maybe this should also be typed

ggshield/sca/filter.py
```diff
@@ -0,0 +1,60 @@
...
+from pathlib import Path
+from typing import List, Set
+
+from ggshield.scan import Scannable
+from ggshield.scan.file import get_files_from_paths
+
+
+# List of filepaths to ignore for SCA scans
+
+SCA_IGNORE_LIST = (
```
tchaumeny, line 11: Shouldn't this list be fetched from an API and cached locally?

ggshield/cmd/sca/scan.py
xblanchot-gg: Indeed, but the actual return type is defined in another MR.

ggshield/sca/filter.py
xblanchot-gg: I don't think so. This was part of the other idea to filter SCA files, however, this would imply an other API and a third API call to perform an SCA scan. I think it's too much. It's just a list of filepaths, it's easy to maintain, and missing some elements in this list in a given version of GGShield will just imply that the first API call is larger.

ggshield/sca/filter.py
xblanchot-gg: Thanks for the docstring. 
About `potential`, I wanted to say that many of those files will not be SCA files, but indeed just `get_all_files_from_sca_paths` is better.

ggshield/sca/filter.py
gg-mmill: > it's easy to maintain

I agree this is probably ok for this usecase, but any change is not so easy to deploy, since we don't control the version of ggshield used by our clients.

xblanchot-gg: Indeed, but I don't think missing some elements here for some users is sufficiently critical to maintain an API. However, from a user point of view, they should be able to add other directories here: this is possible with the `exclude_regex` parameter.

ggshield/cmd/sca/scan.py
```diff
@@ -40,13 +50,106 @@ def scan_diff_cmd(
...
+
+    - List SCA related files with a first call to SCA compute files API
+    - Create a tar archive with the required files contents
+    - Launches the scan with a call to SCA public API
+    """
+    client = SCAClient(ctx.obj["client"])
+
+    sca_filepaths = get_sca_scan_all_filepaths(
+        directory=directory,
+        exclusion_regexes=ctx.obj["exclusion_regexes"],
```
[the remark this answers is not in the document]
xblanchot-gg: Indeed, added an `update_context` that only creates the client and an empty `ignore_regexes`.

gg-mmill:
Thanks, let's wait on the route to be available !

ggshield/sca/file_selection.py
```diff
@@ -0,0 +1,58 @@
...
+        recursive=True,
+        yes=True,
+        verbose=verbose,
+        ignore_git=ignore_git,
+    ).apply_filter(is_not_excluded_from_sca)
+
+    return [str(x.relative_to(path)) for x in files.paths]
+
+
+def is_not_excluded_from_sca(scannable: Scannable) -> bool:
```
tchaumeny, line 53: Maybe `is_excluded_from_sca` which returns `any(part in SCA_IGNORE_LIST for part in scannable.path.parts)` would be more direct (and that would match the description in the docstring)

ggshield/cmd/sca/scan.py
```diff
@@ -0,0 +1,166 @@
...
+        verbose=ctx.obj["config"].verbose,
+        client=client,
+    )
+
+    # No file to scan
+    if sca_filepaths is None:
+        display_info("\nNo file to scan.")
```
tchaumeny, line 107: Why do we have to prefix with `\n`?

ggshield/cmd/sca/scan.py
```diff
@@ -0,0 +1,166 @@
...
+        exclusion_regexes=exclusion_regexes,
+        verbose=verbose,
+        # If the repository is a git repository, ignore untracked files
+        ignore_git=False,
+    )
+
+    # API Call to filter SCA files
+    response = client.compute_sca_files(touched_files=all_filepaths)
+
+    if not response.success or not isinstance(response, ComputeSCAFilesResult):
```
tchaumeny, line 157: In which case can we have `response.success == True` and `not isinstance(response, ComputeSCAFilesResult)`?

ggshield/sca/client.py
```diff
@@ -0,0 +1,96 @@
...
+        filenames: List[str],
+        scan_parameters: SCAScanParameters,
+        extra_headers: Optional[Dict[str, str]] = None,
+    ):
+        """
+        Generates tar archive associated with filenames and launches
+        SCA scan via SCA public API.
+        """
+
+        tar = _create_tar(directory, filenames)
```
tchaumeny, line 68: Why do we import something prefixed with a `_` (this generally means the function is purely internal, not for external use)? And why do we proceed differently from IaC (see `tar_from_ref_and_filepaths`)?

ggshield/cmd/sca/scan.py
```diff
@@ -0,0 +1,166 @@
...
+            scan_mode=ScanMode.SCA_DIRECTORY,
+        ).get_http_headers(),
+    )
+
+    # TODO add SCAScanResult when merged
+    if not scan_result.success or not isinstance(scan_result, DummyReturnModel):
+        if scan_result.status_code == 401:
+            raise APIKeyCheckError(client.base_uri, "Invalid API key.")
+        display_error("\nError scanning.")
+        display_error(str(scan_result))
```
tchaumeny, line 129: Here you don't need to prepend with `\n`?

ggshield/sca/client.py
tchaumeny, line 68: Also, I'm not sure the tar creation part should belong to the client.
IMO the client should only be responsible for the communication with backend part.

xblanchot-gg: `_create_tar` is defined in py-gitguardian `client.py`, and used by methods in this client. We plan to move our client `SCAClient` in `GGClient`. 
If I'm not mistaken, `tar_from_ref_and_filepaths` is only used for the diff scan in IAC, the full scan uses `_create_tar`.

ggshield/cmd/sca/scan.py
xblanchot-gg: It seems that `response.success is True` only is status_code is `200`, whereas we can have `isinstance(response, ComputeSCAFilesResult)` for every `ok` codes. So looks like one of the two conditions is included in the other, and we could avoid the test on `response.success`.

ggshield/sca/file_selection.py
xblanchot-gg: It would be, however we rely on the `apply_filter` method of `Scannable` which requires to return `True` if the file is accepted. I modified the docstring, it was indeed outdated.

ggshield/cmd/sca/scan.py
xblanchot-gg: I removed the `\n`, they came from some copy of iac logs.

xblanchot-gg:
@gg-mmill @tchaumeny Ready for review !

ggshield/cmd/sca/scan.py
tchaumeny: So why do you keep it (here and on line 129)?

ggshield/sca/file_selection.py
tchaumeny: Nit, but in this case I would find `return not any(... in SCA_IGNORE_LIST ...)` more direct (it factorizes the `not`)

ggshield/cmd/sca/scan.py
xblanchot-gg: I just wanted to be sure. Removed.

ggshield/sca/file_selection.py
xblanchot-gg: If you prefer ! Modified

gg-mmill (commented):
Almost good to me

ggshield/cmd/sca/scan.py
```diff
@@ -57,14 +51,119 @@ def scan_diff_cmd(
...
+
+    # API Call to filter SCA files
+    response = client.compute_sca_files(touched_files=all_filepaths)
+
+    if not isinstance(response, ComputeSCAFilesResult):
+        if response.status_code == 401:
+            raise APIKeyCheckError(client.base_uri, "Invalid API key.")
+        display_error("Error while filtering SCA related files.")
+        display_error(str(response))
+        return None
```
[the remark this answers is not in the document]
xblanchot-gg: Indeed. I raise an error if `compute_sca_files` fails, and return an empty `SCAScanAllOutput` in `sca_scan_all` in case of an empty result from `get_sca_scan_all_filepaths`. WDYT ?

ggshield/cmd/sca/scan.py
```diff
@@ -57,14 +51,123 @@ def scan_diff_cmd(
...
+
+    # API Call to filter SCA files
+    response = client.compute_sca_files(touched_files=all_filepaths)
+
+    if not isinstance(response, ComputeSCAFilesResult):
+        if response.status_code == 401:
+            raise APIKeyCheckError(client.base_uri, "Invalid API key.")
+        display_error("Error while filtering SCA related files.")
+        display_error(str(response))
+        raise client.ListScaFilesError
```
gg-mmill, line 169: Let's just use `ggshield.core.errors.UnexpectedError`

ggshield/cmd/sca/scan.py
```diff
@@ -57,14 +51,123 @@ def scan_diff_cmd(
...
+    client = SCAClient(ctx.obj["client"])
+
+    try:
+        sca_filepaths = get_sca_scan_all_filepaths(
+            directory=directory,
+            exclusion_regexes=ctx.obj["exclusion_regexes"],
+            verbose=ctx.obj["config"].verbose,
+            client=client,
+        )
+    except client.ListScaFilesError:
```
gg-mmill, line 108: We don't want to catch the error this way, because we'll need to know an error happened when exiting, in order to set the exit code.

xblanchot-gg: We could as we return None in case of errors, but yes indeed we can just let it raise and catch it when we compute the return code from the command function.

ggshield/cmd/sca/scan.py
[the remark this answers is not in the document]
gg-mmill: Sorry I had not seen your comment. 
> return an empty SCAScanAllOutput in sca_scan_all in case of an empty result from get_sca_scan_all_filepaths

I do not really see the point. If an error happens in `get_sca_scan_all_filepaths`, there is no point returning an SCAScanAllOutput - something went wrong, the only thing we can do is output meaningful error message and exit code.

gg-mmill:
Thanks, all good - except I have added `file_selection.py` as well :angel:

ggshield/cmd/sca/scan.py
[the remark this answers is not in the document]
xblanchot-gg: No, if we return an empty list, no error occured. If we return an empty list, there is just no file to scan, so in this case we do not return `None` anymore, we return an empty `SCAScanAllOutput` because it is basically the expected result in this case.

In case of an error, we raise an `UnexpectedError`, and we do not reach the line where we return an empty `SCAScanAllOutput`. (In the previous proposition in case of error I did return `None`, now I just raise).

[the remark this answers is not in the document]
xblanchot-gg: The misunderstanding maybe came from the term `empty result` instead of `empty list`.

ggshield/cmd/sca/scan.py
```diff
@@ -57,14 +51,119 @@ def scan_diff_cmd(
...
+    Basic implementation, should be populated with required parameters later.
+    """
+    config: Config = ctx.obj["config"]
+    ctx.obj["client"] = create_client_from_config(config)
+
+    # Empty for now
+    ctx.obj["exclusion_regexes"] = set()
+
+
+def sca_scan_all(ctx: click.Context, directory: Path) -> Optional[SCAScanAllOutput]:
```
gg-mmill, line 91:
```suggestion
def sca_scan_all(ctx: click.Context, directory: Path) -> SCAScanAllOutput:
```

ggshield/cmd/sca/scan.py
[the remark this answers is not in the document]
gg-mmill: Oh ok, yes I get it, but since we'll probably need a try/catch at the upper level, it's not worth handling errors in `scan_all` with an optional return type IMO. Let's wait, but we can probably handle the ReadTimeout there as well, instead of inside the client.

ggshield/cmd/sca/scan.py
xblanchot-gg: There is still a `return None` line 146, in case of error in the second API call, where the client returns a `Detail` instance.

tchaumeny (approved):

ggshield/cmd/sca/scan.py
```diff
@@ -57,14 +66,120 @@ def scan_diff_cmd(
...
+        if scan_result.status_code == 401:
+            raise APIKeyCheckError(client.base_uri, "Invalid API key.")
+        display_error("Error scanning.")
+        display_error(str(scan_result))
+        raise UnexpectedError("Unexpected error while performing SCA scan all.")
+
+    return scan_result
+
+
+def get_sca_scan_all_filepaths(
```
GG-HH, line 154: nit: should it be in `ggshield/sca/file_selection.py` ?

GG-HH (changes requested):

ggshield/cmd/sca/scan.py
xblanchot-gg: Ok for me, moved.

tests/unit/cassettes/test_sca_scan_directory_valid.yaml
```diff
@@ -0,0 +1,82 @@
...
+        Content-Type:
+          - multipart/form-data; boundary=3250a65d9a33a5c55e28d6525645c312
+        User-Agent:
+          - pygitguardian/1.8.0 (Linux;py3.10.6)
+      method: POST
+      uri: https://api.gitguardian.com/v1/sca/sca_scan_all/
+    response:
+      body:
+        string:
+          '{"scanned_files":["home/xavierblanchot/gg_tools/ggshield/tests/unit/sca/../data/Pipfile.lock"],"found_package_vulns":[{"location":"home/xavierblanchot/gg_tools/ggshield/tests/unit/data/Pipfile.lock","package_vulns":[{"package_full_name":"vyper","version":"0.2.10","ecosystem":"pypi","dependency_type":"transitive","vulns":[{"cve_ids":["CVE-2021-41121"],"severity":"high","created_at":"2021-10-12T15:59:29Z","fixed_version":"0.3.0","summary":"Memory
```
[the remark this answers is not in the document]
xblanchot-gg: The absolute path comes from the way I define the path to my `Pipfile.lock`, with `Path(__file__)` in my test file. This does not happen in real cases. I removed the `..` by adding a `.resolve()` in my path (see `tests/unit/sca/test_client.py`)

ggshield/sca/client.py
```diff
@@ -40,3 +52,45 @@ def compute_sca_files(
...
+                endpoint="sca/full_scan/",
+                files={
+                    "directory": tar,
+                },
+                data={
+                    "scan_parameters": SCAScanParameters.SCHEMA.dumps(scan_parameters),
+                },
+                extra_headers=extra_headers,
+            )
+        except requests.exceptions.ReadTimeout:
```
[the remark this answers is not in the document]
agateau-gg: Would be nice, but that might be tricky to do.

xblanchot-gg:
@agateau-gg Indeed, it's just temporary waiting for the backend side to go in production. The cassettes were created with staging.

GG-HH (approved):

PR #582, opened 2023-06-26
feat(sca): #PLC-768 add warning beta message for sca commands

xblanchot-gg:
Adds a decorator to print a warning message for SCA features being in beta.

### DIFF
--- a/ggshield/cmd/sca/scan.py
+++ b/ggshield/cmd/sca/scan.py
@@ -3,6 +3,22 @@ from typing import Any, Optional
 
 import click
 
+from ggshield.core.text_utils import display_warning
+
+
+def display_sca_beta_warning(func):
+    """
+    Displays warning about SCA commands being in beta.
+    """
+
+    def func_with_beta_warning(*args, **kwargs):
+        display_warning(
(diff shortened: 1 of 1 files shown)


ggshield/cmd/sca/scan.py
```diff
@@ -41,6 +58,7 @@ def scan_diff_cmd(
 
 @scan_group.command(name="all")
```
amascia, line 59: We need to add these lines so that we can use the `ggshield sca scan all` command

```
@click.argument(
    "directory",
    type=click.Path(exists=True, readable=True, path_type=Path, file_okay=False),
    required=False,
)
```

amascia-gg, line 59: We need to add these lines so that we can use the ggshield sca scan all command

```
@click.argument(
    "directory",
    type=click.Path(exists=True, readable=True, path_type=Path, file_okay=False),
    required=False,
)
```

xblanchot-gg: Yep I saw it, I add it in the `scan all command` MR.

amascia-gg (approved):

PR #583, opened 2023-06-26
chore: update dependencies

agateau-gg:
- Update py-gitguardian from git ref to 1.8.0
- Update cryptography from 40.0.2 to 41.0.0, to fix OpenSSL vulnerability (see https://github.com/GitGuardian/ggshield/security/dependabot/6)


GG-HH (approved):
Sorry for the delay, I missed the notification.

PR #588, opened 2023-06-26
Fix error when executing a diff scan in a subfolder of a git repository

Paul-GitGuardian:
Scanning a subfolder should work as long as it is inside a git repository. This is the (working) behaviour for the `iac scan all` command.
However, it fails with the `iac scan diff` command, and throws an ugly error message

Currently, the only fix in this PR is to throw a prettier message to the user, suggesting them to execute the scan at the git root.
However, we might prefer fixing the issue.

It comes from the `git show` command used to create the tars. This only works with a path relative to the git root (not relative to the cwd).

### DIFF
--- a/ggshield/core/git_shell.py
+++ b/ggshield/core/git_shell.py
@@ -232,7 +232,7 @@ def tar_from_ref_and_filepaths(
 
     with tarfile.open(fileobj=tar_stream, mode="w:gz") as tar:
         for path in filepaths:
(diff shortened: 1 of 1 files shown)


GG-HH:
Hello Paul, I agree that we should fix the issue completely: a user should be able to scan a specific directory inside a git repository without issues.
Thanks for noticing.

Paul-GitGuardian:
The fix was actually trivial

GG-HH (approved):

PR #591, opened 2023-06-27
fix(iac): fix filter on extensions / remove .py extension

gg-teolohrer:
Validation was weak on file extensions.

### DIFF
--- a/ggshield/iac/filter.py
+++ b/ggshield/iac/filter.py
@@ -12,7 +12,6 @@ IAC_EXTENSIONS = {
     ".yml",
     ".yaml",
     ".jinja",
-    ".py",
     ".py.schema",
     ".jinja.schema",
     ".tf",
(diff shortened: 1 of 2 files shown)


ggshield/iac/filter.py
```diff
@@ -51,6 +50,8 @@ def get_iac_files_from_paths(
 def is_file_iac_file(scannable: Scannable) -> bool:
     if any(ext in IAC_EXTENSIONS for ext in scannable.path.suffixes):
         return True
+    if any(scannable.path.name.endswith(iac_ext) for iac_ext in IAC_EXTENSIONS):
+        return True
```
Paul-GitGuardian, line 54: Are you keeping both on purpose?

If not:

```suggestion
def is_file_iac_file(scannable: Scannable) -> bool:
    if any(scannable.path.name.endswith(iac_ext) for iac_ext in IAC_EXTENSIONS):
        return True
```

Paul-GitGuardian (changes requested):

ggshield/iac/filter.py
gg-teolohrer: Yes, they're different.
Before, the `.py.schema` files returned True only because the `.py` extension existed.
Now:
If `.tf` is in any extension of the file (like `something.tf.other`) we return True.
If `.py.schema` is the ending pattern of the file, we return True as well.
In the end, I reckon we will overhaul that part since it does particularly efficient nor well-designed.
But for now, it handles the patterns correctly.

Paul-GitGuardian (approved):

PR #593, opened 2023-06-27
fix: no changed iac files leads to empty diff iac scan

gg-teolohrer:
This PR fixes a problematic behavior in IaC diff scans.

# Problem description

When no IaC files have been changed and a diff scan is requested, the tar archives are still built and sent for scanning even though no additional IaC incident will be found.

While it is true the scanning backend could have gotten new vulnerability detectors in the meantime, we generally do not want the scan to raise warnings when no IaC files have been touched.

# Solution

With an additional function in `git_shell.py`, we can detect the status of files relative to the given git reference.

If any file has been added/removed/modified AND is an IaC file, we perform the diff scan as usual. Else, we notify that the scan was skipped and return 0.

# Implementation details

- the `git_diff_file_status(...)` function has been added to `git_shell.py`. This function returns the associated `Filemode`s associated with each file.
- the `Filemode` enum has been moved from `utils.py` to `git_shell.py` to avoid circular imports. As `Filemode` describes git file modes, it feels coherent with the scope of `git_shell.py`. Consequently, `from ggshield.core.utils import Filemode` have been changed to `from ggshield.core.git_shell import Filemode`.
- IaC diff scanning has been updated.
- "Skip diff scan" output handlers have been added.
- tests were updated.

### DIFF
--- a/ggshield/cmd/iac/scan/diff.py
+++ b/ggshield/cmd/iac/scan/diff.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Any, Optional, Sequence
+from typing import Any, Optional, Sequence, Union
 
 import click
 from pygitguardian.iac_models import IaCScanParameters
@@ -21,8 +21,12 @@ from ggshield.core.config.config import Config
 from ggshield.core.git_shell import INDEX_REF, Filemode, get_diff_files_status
 from ggshield.core.text_utils import display_info, display_warning
 from ggshield.iac.collection.iac_diff_scan_collection import IaCDiffScanCollection
-from ggshield.iac.filter import is_file_path_iac_file_path
-from ggshield.iac.iac_scan_models import IaCDiffScanResult, create_client_from_config
+from ggshield.iac.filter import is_iac_file_path
+from ggshield.iac.iac_scan_models import (
+    IaCDiffScanResult,
+    IaCSkipDiffScanResult,
+    create_client_from_config,
+)
 from ggshield.scan import ScanContext, ScanMode
 
 
@@ -106,14 +110,16 @@ def scan_diff_cmd(
         staged = False
 
     result = iac_scan_diff(ctx, directory, ref, staged)
-    scan = IaCDiffScanCollection(id=str(directory), result=result)
     output_handler = create_output_handler(ctx)
+    if isinstance(result, IaCSkipDiffScanResult):
+        return output_handler.process_skip_diff_scan()
+    scan = IaCDiffScanCollection(id=str(directory), result=result)
     return output_handler.process_diff_scan(scan)
 
 
 def iac_scan_diff(
     ctx: click.Context, directory: Path, ref: str, include_staged: bool
-) -> Optional[IaCDiffScanResult]:
+) -> Optional[Union[IaCDiffScanResult, IaCSkipDiffScanResult]]:
     config = ctx.obj["config"]
     client = ctx.obj["client"]
     exclusion_regexes = ctx.obj["exclusion_regexes"]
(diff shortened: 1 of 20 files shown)


ggshield/core/git_shell.py
```diff
@@ -202,6 +216,55 @@ def get_staged_filepaths(wd: Optional[str] = None) -> List[Path]:
     return [Path(path_str) for path_str in filepaths]
 
 
+def get_diff_files_status(
+    wd: Optional[str], ref: str, staged: bool = False, similarity: int = 100
+) -> Dict[Path, Filemode]:
+    """
+    Fecthes the statuses of modified files since a given ref.
```
cyrvan-b, line 223:
```suggestion
    Fetches the statuses of modified files since a given ref.
```

ggshield/core/git_shell.py
```diff
@@ -202,6 +216,55 @@ def get_staged_filepaths(wd: Optional[str] = None) -> List[Path]:
...
+    """
+
+    def parse_name_status_patch(patch: str) -> Dict[Path, Filemode]:
+
+        m = {
+            "A": Filemode.NEW,
+            "D": Filemode.DELETE,
+            "M": Filemode.MODIFY,
+            "T": Filemode.MODIFY,
```
cyrvan-b, line 232: "M" and "T" are both modified is this wanted?

gg-teolohrer: Yes, T is a type change, like symbolic link to hard link. In practice, it does not appear that much.

GG-HH: Nit: add a link to the documentation.

ggshield/core/git_shell.py
```diff
@@ -22,6 +23,19 @@
 logger = logging.getLogger(__name__)
 
 
+class Filemode(Enum):
```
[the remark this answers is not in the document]
gg-teolohrer: `utils` is dependant on `git_shell`.
I wanted to avoid circular imports or local imports in the function.
Also, I think it makes sense to have git Filemodes in `git_shell`. Maybe I could make a `git_utils` file instead, WDYT ?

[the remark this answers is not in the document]
GG-HH: No, the actual is fine then

GG-HH (approved):

ggshield/cmd/iac/scan/diff.py
```diff
@@ -105,14 +110,16 @@ def scan_diff_cmd(
...
+    if isinstance(result, IaCSkipDiffScanResult):
+        return output_handler.process_skip_diff_scan()
+    scan = IaCDiffScanCollection(id=str(directory), result=result)
     return output_handler.process_diff_scan(scan)
 
 
 def iac_scan_diff(
     ctx: click.Context, directory: Path, ref: str, include_staged: bool
-) -> Optional[IaCDiffScanResult]:
+) -> Optional[Union[IaCDiffScanResult, IaCSkipDiffScanResult]]:
```
agateau-gg, line 122: nit: this can be written as:
```suggestion
) -> Union[IaCDiffScanResult, IaCSkipDiffScanResult, None]:
```

ggshield/core/git_shell.py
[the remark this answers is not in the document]
agateau-gg: +1 for this. The less code we have in core.utils, the better.

ggshield/core/git_shell.py
```diff
@@ -202,6 +216,62 @@ def get_staged_filepaths(wd: Optional[str] = None) -> List[Path]:
     return [Path(path_str) for path_str in filepaths]
 
 
+def get_diff_files_status(
+    wd: Optional[str], ref: str, staged: bool = False, similarity: int = 100
```
agateau-gg, line 220: Since `wd` is optional, it should be after non-optional arguments and have a default value of `None`.

agateau-gg (changes requested):

The commit history could be shorter :)

gg-teolohrer:
Reworked the history to hide fixups and typo commits.

agateau-gg (approved):

PR #594, opened 2023-06-28
chore(pre-commit): # update to latest GGShield

agateau-gg:


pierrelalanne (approved):

pierrelalanne:
I think we're just missing a skip changelog label to make CI pass

PR #598, opened 2023-06-29
Fix: iac scan diff should only return an exit code 1 if new vulnerabilities are found

Paul-GitGuardian:
Currently, having unchanged or deleted vulnerabilities would also trigger an exit code 1.
This makes it more difficult to use the command in CI, as only new vulnerabilities should throw an alert.

### DIFF
--- a/ggshield/iac/collection/iac_diff_scan_collection.py
+++ b/ggshield/iac/collection/iac_diff_scan_collection.py
@@ -4,10 +4,12 @@ from ggshield.iac.collection.iac_scan_collection import (
     CollectionType,
     IaCScanCollection,
 )
-from ggshield.iac.iac_scan_models import IaCDiffScanEntities, IaCDiffScanResult
+from ggshield.iac.iac_scan_models import IaCDiffScanResult
 
 
 class IaCDiffScanCollection(IaCScanCollection):
+    result: Optional[IaCDiffScanResult]
+
     def __init__(
         self,
         id: str,
@@ -18,13 +20,4 @@ class IaCDiffScanCollection(IaCScanCollection):
 
     @property
     def has_results(self) -> bool:
-        if self.result is None:
-            return False
-        if isinstance(self.result.entities_with_incidents, IaCDiffScanEntities):
-            return (
-                bool(self.result.entities_with_incidents.unchanged)
-                or bool(self.result.entities_with_incidents.new)
-                or bool(self.result.entities_with_incidents.deleted)
-            )
-        else:
-            return bool(self.result.entities_with_incidents)
+        return self.result is not None and bool(self.result.entities_with_incidents.new)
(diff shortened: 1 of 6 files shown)


ggshield/iac/collection/iac_diff_scan_collection.py
```diff
@@ -4,10 +4,12 @@
     CollectionType,
     IaCScanCollection,
 )
-from ggshield.iac.iac_scan_models import IaCDiffScanEntities, IaCDiffScanResult
+from ggshield.iac.iac_scan_models import IaCDiffScanResult
 
 
 class IaCDiffScanCollection(IaCScanCollection):
+    result: Optional[IaCDiffScanResult]
```
GG-HH, line 11: I know it is not part of the MR, but we don't need a init in those classes, it is only for the `type` attribute which could be a class attribute since you never change it.

tests/unit/iac/collection/test_iac_scan_collection.py
```diff
@@ -1,68 +1,136 @@
...
-    GIVEN an IaC path scan collection
-    WHEN accessing its type
-    THEN the type is CollectionType.PathScan
-    """
-    collection = IaCPathScanCollection("0", IaCScanResult("0", "path_scan", "", []))
-    assert collection.type == CollectionType.PathScan
+def _generate_empty_path_collection() -> IaCPathScanCollection:
+    return IaCPathScanCollection(
+        "3ac2985e-dcf9-49ff-92fb-943548268de8",
+        IaCScanResult("3ac2985e-dcf9-49ff-92fb-943548268de8", "path_scan", "1.8.0", []),
```
GG-HH, line 16: nit: use named arguments (in general, we consider it necessary/usefull when we have 3 or more arguments). Especially when they have the same type (strings), it can be very easy to make mistakes in the order.

tests/functional/iac/test_iac_scan_diff.py
```diff
@@ -52,10 +53,11 @@ def test_iac_scan_diff_new_vuln(tmp_path: Path) -> None:
     # WHEN scanning the diff between current and HEAD~1
     # (meaning reference should be one commit behind)
     args = ["diff", "--ref", "HEAD~1", str(tmp_path)]
+    # THEN exit code is 1 (new vuln detected)
     result = run_ggshield_iac_scan(*args, cwd=tmp_path, expected_code=1)
 
     print(result.stdout)
```
GG-HH, line 59: :O Sorry to spot that now but we shouldn't have prints.

GG-HH (approved):
Sorry, I keep finding issues with code that is not part of the MR...
The removal of the print may be done in another MR if you find it more relevant.

GG-HH (approved):

gg-teolohrer (approved):

PR #601, opened 2023-07-04
teolohrer/iac 94 ggshield scan all dont perform scan if not needed

gg-teolohrer:
# Problem description

When no IaC files are present (or all IaC files are ignored), the `ggshield iac scan all` method builds an empty tar before requesting a scan.

# Updates

- The `IACSkipDiffScanResult` singleton has been generalized to `IACSkipScanResult` and used for the `iac scan all` (in addition to the `iac scan diff`)
- The correct check has been added to the `iac/scan.py` file.
- Tests have been updated consequently
- Tests have been parametrized to remove duplicates between `iac scan all` and `iac scan`
- Minor typos have been corrected


tests/functional/iac/test_iac_scan_all.py
```diff
@@ -8,22 +8,68 @@
 from tests.repository import Repository
 
 
-def test_iac_scan_all(tmp_path: Path) -> None:
+# `iac scan` is set for deprecation, but should behave exactly as `iac scan all` in the meantime
+pytestmark = pytest.mark.parametrize("scan_prefix", [["all"], []])
```
GG-HH, line 12: It seems very hacky to me to use a list and then unfold it to handle the case where no command is given.

I think you should define a helper to build the args that will handle explicitely the case where there is no command.

GG-HH (changes requested):

GG-HH (approved):

PR #602, opened 2023-07-04
Perf: in iac diff scan, allow file filter to operate without opening the files

Paul-GitGuardian:
When executing a `iac scan diff`,
- all files in the repo are listed
- non-IaC files are filtered out of the list
- remaining files are added to a tar

Currently, files are filtered based on their path only, but their content is read and passed to the filter function anyway.

This PR removes the `acceptation_func` from the tar creation function, which now expects the filepaths list to be already filtered.
Also, caching is added to the file opening function, in case the iac filter requires to read them in the future.

Execution time for `iac scan diff --ref 253d331b8b44e87d11238f715b904828330ed84e` on the ggshield repo:
- with the old behaviour: ~7,5s
- with this new behaviour: ~5,8s

### DIFF
--- a/ggshield/cmd/iac/scan/iac_scan_utils.py
+++ b/ggshield/cmd/iac/scan/iac_scan_utils.py
@@ -1,6 +1,6 @@
 from pathlib import Path
 from re import Pattern
-from typing import Iterable, Set, Type
+from typing import Iterable, Optional, Set, Type
 
 import click
 from pygitguardian import GGClient
@@ -44,7 +44,7 @@ def handle_scan_error(client: GGClient, detail: Detail) -> None:
     display_error(str(detail))
 
 
-def get_iac_filepaths(directory: Path, ref: str) -> Iterable[Path]:
+def get_git_filepaths(directory: Path, ref: str) -> Iterable[Path]:
     return (
         get_staged_filepaths(str(directory))
         if ref == INDEX_REF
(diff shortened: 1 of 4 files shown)


ggshield/core/git_shell.py
```diff
@@ -302,13 +308,14 @@ def tar_from_ref_and_filepaths(
...
-            raw_file_content = git(["show", f"{ref}:./{path}"], cwd=wd)
 
-            if acceptation_func is not None and not (
-                acceptation_func(path, raw_file_content)
-            ):
+            def _readfile() -> str:
+                return read_git_file(ref, path, wd)
+
+            if acceptation_func is not None and not acceptation_func(path, _readfile):
                 continue
```
Paul-GitGuardian, line 316: Note: using `lru_cache` for this might be a waste.
Each filepath goes through the filter (1st read) then, if accepted, is immediately added to the tar (2nd read).
We go on to the next file and no longer need to open the previous one, so space is wasted in the cache).

However, this does not seem to impact performances.

ggshield/cmd/iac/scan/iac_scan_utils.py
```diff
@@ -64,7 +64,7 @@ def filter_iac_filepaths(
 def get_iac_tar(directory: Path, ref: str, exclusion_regexes: Set[Pattern]) -> bytes:
     filepaths = get_iac_filepaths(directory, ref)
 
-    def _accept_file(path: Path, content: str) -> bool:
+    def _accept_file(path: Path, open_file: Callable[[], str]) -> bool:
```
sylvain-baud-gg, line 67: open_file is not used in _accept_file.

Why not remove the second argument and only keep path ?
It seems the only place the fhis function is used is in tar_from_ref_and_filepath but this does not use the open_file function.

ggshield/core/git_shell.py
```diff
@@ -302,13 +308,14 @@ def tar_from_ref_and_filepaths(
...
 
+            raw_file_content = _readfile()
```
sylvain-baud-gg, line 318:
```suggestion
            raw_file_content = read_git_file(ref, path, wd)
```

If the `open_file` argument is removed from acceptation_func

PR #606, opened 2023-07-05
Perf (implementation 2): in iac diff scan, allow file filter to operate without opening the files

Paul-GitGuardian:
This attempts to solve the same issue as #602.
This time, we provide two distinct acceptation functions as arguments to the `tar_from_ref_and_filepaths` function:
- One filtering on filepath only (does not open the file)
- One filtering on file content. The file is opened once. The filepath is also provided as an argument, because cross-validation on path and content might be needed in the future.

Pros:
- When validating on path only, there is no longer an unused `file_opener` argument in the acceptation function
Cons:
- However, the `path` argument in the second acceptation function might be unused.

### DIFF
--- a/ggshield/core/git_shell.py
+++ b/ggshield/core/git_shell.py
@@ -280,7 +280,8 @@ def read_git_file(ref: str, path: Path, wd: Optional[str] = None) -> str:
 def tar_from_ref_and_filepaths(
     ref: str,
     filepaths: Iterable[Path],
-    acceptation_func: Optional[Callable[[Path, Callable[[], str]], bool]] = None,
+    accept_path: Optional[Callable[[Path], bool]] = None,
+    accept_content: Optional[Callable[[str, Path], bool]] = None,
     wd: Optional[str] = None,
 ) -> bytes:
     """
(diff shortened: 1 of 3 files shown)


ggshield/core/git_shell.py
```diff
@@ -308,15 +312,15 @@ def tar_from_ref_and_filepaths(
 
     with tarfile.open(fileobj=tar_stream, mode="w:gz") as tar:
         for path in filepaths:
+            if accept_path is not None and not accept_path(path):
+                continue
 
-            def _readfile() -> str:
-                return read_git_file(ref, path, wd)
+            content = read_git_file(ref, path, wd)
```
cyrvan-b, line 318: content should not be read if `accept_content` is `None`.

ggshield/core/git_shell.py
```diff
@@ -280,7 +280,8 @@ def read_git_file(ref: str, path: Path, wd: Optional[str] = None) -> str:
 def tar_from_ref_and_filepaths(
     ref: str,
     filepaths: Iterable[Path],
-    acceptation_func: Optional[Callable[[Path, Callable[[], str]], bool]] = None,
+    accept_path: Optional[Callable[[Path], bool]] = None,
+    accept_content: Optional[Callable[[str, Path], bool]] = None,
```
cyrvan-b, line 284: Are we using `accept_content` yet? Or is it added for future usage?

ggshield/core/git_shell.py
Paul-GitGuardian: We need to read the content regardless of the filters, to add it to the tar.

PR #607, opened 2023-07-05
Perf (implementation 3): in iac diff scan, allow file filter to operate without opening the files

Paul-GitGuardian:
This attempts to solve the same issue as https://github.com/GitGuardian/ggshield/pull/602.
This time, we pre-filter the filepaths list provided to the `tar_from_ref_and_filepaths` function.
The `acceptation_func` still exists, but must be used only for filters based on content.

Pros:
- One unique acceptation function. A filepath filtering function was not necessary, as we already have to provide the filepath list.
Cons:
- The path argument in the second acceptation function might be unused.

### DIFF
--- a/ggshield/core/git_shell.py
+++ b/ggshield/core/git_shell.py
@@ -280,7 +280,7 @@ def read_git_file(ref: str, path: Path, wd: Optional[str] = None) -> str:
 def tar_from_ref_and_filepaths(
     ref: str,
     filepaths: Iterable[Path],
-    acceptation_func: Optional[Callable[[Path, Callable[[], str]], bool]] = None,
+    acceptation_func: Optional[Callable[[Path, str], bool]] = None,
     wd: Optional[str] = None,
 ) -> bytes:
     """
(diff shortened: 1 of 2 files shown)


ggshield/core/git_shell.py
```diff
@@ -280,20 +280,21 @@ def read_git_file(ref: str, path: Path, wd: Optional[str] = None) -> str:
...
-    acceptation_func: Optional[Callable[[Path, Callable[[], str]], bool]] = None,
+    acceptation_func: Optional[Callable[[Path, str], bool]] = None,
     wd: Optional[str] = None,
 ) -> bytes:
     """
     Builds a gzipped archive from a given git reference, and selected filepaths.
     The filepaths are typically obtained via `get_filepaths_from_ref` or `get_staged_filepaths`
     before being filtered.
     The archive is returned as raw bytes.
+    For better performance, pre-filter the provided filepaths, or use the `acceptation_func`
```
cyrvan-b, line 291: Filtering based on content could also be done before using this function. IMO it should be a function which generate a tar from a list of Path.

PR #613, opened 2023-07-05
Drop Python 3.7

agateau-gg:
## Description

This PR bumps our minimum Python version to 3.8. The reason for this change are:

- Some of our dependencies are starting to drop Python 3.7 support.
- Some recently added tests had issues when running on Python 3.7 + Windows.


gg-mmill (approved):
Seems good,  I did not find any other reference to python 3.7

PR #614, opened 2023-07-06
Mmillet/ /fixes

gg-mmill:
- fix `compute_sca_files` call and cassettes
- remove fix for python 3.7, whose support was just dropped (https://github.com/GitGuardian/ggshield/pull/613)

### DIFF
--- a/tests/unit/sca/conftest.py
+++ b/tests/unit/sca/conftest.py
@@ -1,10 +1,8 @@
 import io
 import os
-import sys
 import tarfile
 import tempfile
 from pathlib import Path
-from typing import Union
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


tests/unit/sca/conftest.py
```diff
@@ -112,26 +110,6 @@ def clean_directory(path: Path):
             os.remove(filepath)
 
 
-def fix_py37_win_tempdir_permissions(dirpath: Union[str, Path]) -> None:
```
xblanchot-gg, line 115: Why removing this ?

xblanchot-gg (approved):

tests/unit/sca/conftest.py
gg-mmill: remove fix for python 3.7, whose support was just dropped (https://github.com/GitGuardian/ggshield/pull/613)

PR #618, opened 2023-07-07
`iac scan prepush` command

Paul-GitGuardian:
Command meant to be used in a pre-push git hook to scan iac files for vulnerabilities.

Some parts to focus on:
- moved some functions from `ggshield/cmd/secret/scan/prepush.py` to `ggshield/core/git_hooks/prepush.py` in order for them to be used both in secrets and iac (and maybe sca in the future)

### DIFF
--- a/ggshield/cmd/iac/scan/diff.py
+++ b/ggshield/cmd/iac/scan/diff.py
@@ -18,7 +18,12 @@ from ggshield.cmd.iac.scan.iac_scan_utils import (
     handle_scan_error,
 )
 from ggshield.core.filter import is_filepath_excluded
-from ggshield.core.git_shell import INDEX_REF, Filemode, get_diff_files_status
+from ggshield.core.git_shell import (
+    INDEX_REF,
+    Filemode,
+    get_diff_files_status,
+    get_empty_tar,
+)
 from ggshield.core.text_utils import display_info, display_warning
 from ggshield.iac.collection.iac_diff_scan_collection import IaCDiffScanCollection
 from ggshield.iac.filter import is_iac_file_path
@@ -67,7 +72,7 @@ def scan_diff_cmd(
 
 
 def iac_scan_diff(
-    ctx: click.Context, directory: Path, ref: str, include_staged: bool
+    ctx: click.Context, directory: Path, ref: Optional[str], include_staged: bool
 ) -> Union[IaCDiffScanResult, IaCSkipScanResult, None]:
     config = ctx.obj["config"]
     client = ctx.obj["client"]
@@ -75,11 +80,18 @@ def iac_scan_diff(
 
     verbose = config.user_config.verbose if config and config.user_config else False
     if verbose:
-        display_info(f"> Scanned files in reference {ref}")
-        filepaths = filter_iac_filepaths(directory, get_git_filepaths(directory, ref))
-        for filepath in filepaths:
-            display_info(f"- {click.format_filename(filepath)}")
-        display_info("")
+        if ref is None:
+            display_info(
+                "> No file to scan in reference. This might be a new repository."
+            )
+        else:
+            display_info(f"> Scanned files in reference {ref}")
+            filepaths = filter_iac_filepaths(
+                directory, get_git_filepaths(directory, ref)
+            )
+            for filepath in filepaths:
+                display_info(f"- {click.format_filename(filepath)}")
+            display_info("")
 
     current_ref = INDEX_REF if include_staged else "HEAD"
     if verbose:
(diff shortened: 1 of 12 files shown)


tests/functional/utils.py
```diff
@@ -64,3 +65,20 @@ def recreate_censored_content(content: str, matched_string: str) -> str:
...
+
+
+# Use this in iac tests until iac hooks are added to the install command
+# Meaning they are added to install.py
+def create_local_hook(
+    path: Path,
+    hook_type: str,
+) -> None:
+    """Create hook directory (if needed) and pre-commit/pre-push file."""
+    hook_dir_path = path / ".git/hooks"
```
gg-teolohrer, line 77:
```suggestion
    hook_dir_path = path / ".git" / "hooks"
```

tests/functional/utils.py
```diff
@@ -64,3 +65,20 @@ def recreate_censored_content(content: str, matched_string: str) -> str:
     """Applies ggshield censoring to any occurrence of `matched_string` inside
     `content`. Returns the censored string."""
     return content.replace(matched_string, recreate_censored_string(matched_string))
+
+
+# Use this in iac tests until iac hooks are added to the install command
+# Meaning they are added to install.py
+def create_local_hook(
+    path: Path,
```
gg-teolohrer, line 73: Directly specifying the hook dir path would allow to reuse this function for pre-receive also.
In the pre-receive context, the hook is executed in the `.git/` directory.

tests/functional/utils.py
```diff
@@ -64,3 +65,20 @@ def recreate_censored_content(content: str, matched_string: str) -> str:
...
+    hook_type: str,
+) -> None:
+    """Create hook directory (if needed) and pre-commit/pre-push file."""
+    hook_dir_path = path / ".git/hooks"
+    hook_dir_path.mkdir(parents=True, exist_ok=True)
+    hook_path = hook_dir_path / hook_type
+
+    with hook_path.open("w") as f:
+        f.write("#!/bin/sh\n")
```
gg-teolohrer, line 82:
```suggestion
        f.write("#!/usr/bin/env sh\n")
```
Not entirely sure this is the better way. It would avoid edge cases where there is no /bin/sh file

gg-teolohrer (changes requested):

ggshield/core/git_hooks/prepush.py
```diff
@@ -0,0 +1,110 @@
...
+
+from ggshield.core.git_shell import git, is_valid_git_commit_ref
+from ggshield.core.text_utils import display_warning
+from ggshield.core.utils import EMPTY_SHA
+
+
+logger = logging.getLogger(__name__)
+
+
+OUTDATED_HOOK_MESSAGE = """The installed pre-push hook did not pass its command-line arguments to ggshield. This can cause the hook to fail if the name of the remote you are pushing to is not "origin".
```
cyrvan-b, line 14: NIT: 🏗 There is a `ggshield/core/constants.py` file, should we use a similar file for this kind of constants within iac commands?

Paul-GitGuardian: IMO these are specific to prepush operations and would not benefit from being shared in a constants file. What do you think @Walz?

GG-HH (changes requested):

ggshield/cmd/iac/scan/diff.py
```diff
@@ -94,22 +106,32 @@ def iac_scan_diff(
...
+            and is_iac_file_path(file)
+        ]
+
+        if len(modified_iac_files) == 0:
+            return IaCSkipScanResult()
+
+    reference_tar = (
+        get_iac_tar(directory, ref, exclusion_regexes)
+        if ref is not None
+        else get_empty_tar()
```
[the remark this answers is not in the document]
Paul-GitGuardian: This case comes up on the first push to a repository.
I think it is more coherent to display all vulnerabilities as new in this case.
This should not take much more time than a scan all, as the ref tar is empty.

ggshield/core/git_hooks/prepush.py
```diff
@@ -0,0 +1,110 @@
...
+        ]
+    )
+    ancestors = output.splitlines()
+
+    if ancestors:
+        return ancestors[0]
+    return None
+
+
+def collect_commits_from_stdin(remote_name: str) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
Paul-GitGuardian: Sorry for that.
All these functions are the same as previously in cmd/secret/scan/prepush.py
`collect_commits_refs` contains code extracted (and untouched) from the secret scan prepush command

ggshield/cmd/iac/scan/diff.py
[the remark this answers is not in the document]
GG-HH: Yes but it would simplify the code in ggshield a lot and it is a very specific case that shouldn't happen very often (never).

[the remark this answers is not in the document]
Paul-GitGuardian: Ok, removed it

ggshield/core/git_hooks/prepush.py
Walz: I think If it's only used in this file, it can stay here.

Walz (approved):
LGTM

GG-HH (approved):

gg-teolohrer (approved):

PR #615, opened 2023-07-06
Refactor: Replace scan diff hook options by new commands

Paul-GitGuardian:
Currently, the `iac scan diff` command handles three options to simplify git hooks: `--pre-commit`, `--pre-push` and `--pre-receive`. The objective of the `main-iac-cli-hooks` integration branch is to replace them with new `iac scan` subcommands.

This PR will refactor elements of the code in order to:
- make the implementation of the subcommands (not in the scope of this PR) easier
- factorize code

The duplicate code in `precommit.py`, `prepush.py` and `prereceive.py` can be ignored for now.

PR #619, opened 2023-07-07
Pbes/ /refactor iac scan commands

Paul-GitGuardian:
Merging #615 on main directly

Currently, the `iac scan diff` command handles three options to simplify git hooks: `--pre-commit`, `--pre-push` and `--pre-receive`. The objective of the `main-iac-cli-hooks` integration branch is to replace them with new `iac scan` subcommands.

This PR will refactor elements of the code in order to:
- make the implementation of the subcommands (not in the scope of this PR) easier
- factorize code

The duplicate code in `precommit.py`, `prepush.py` and `prereceive.py` can be ignored for now.


agateau-gg (changes requested):
Looks good, but I think this is worth a changelog entry: it modifies user-visible commands.

Paul-GitGuardian:
You're right, changelog added. Should I rebase all three commits into one?

PR #625, opened 2023-07-10
feat(iac): #IAC-112 update example with iac command

carla-gitguardian:
Update `.gitguardian.example.yaml` with IaC command

### DIFF
--- a/.gitguardian.example.yml
+++ b/.gitguardian.example.yml
@@ -36,3 +36,18 @@ secret:
   # Detectors to ignore.
   ignored-detectors: # default: []
     - Generic Password
+
+iac:
+  # Exclude files and paths by globbing
+  ignored-paths:
+    - '**/README.md'
+    - 'doc/*'
+    - 'LICENSE'
+
+  # IaC vulnerabilities to ignore
+  ignored-policies:
+    - GG_IAC_0000
+    - GG_IAC_0005
+
+  # Minimum severity of the policies
+  minimum-severity: HIGH


.gitguardian.example.yml
```diff
@@ -36,3 +36,15 @@ secret:
...
+  # Exclude files and paths by globbing
+  ignored-paths:
+    - 'tests/'
+
+  # Ignore IaC vulnerability types by their id
+  ignored-policies:
+    - GG_IAC_0000
+
+  # Precise vulnerabilities minimum severity to scan for
+  minimum-severity: LOW
```
sylvain-baud-gg, line 50: Maybe use another value for minimum-severity.

As LOW is the lowest possible minimum severity, this parameters doesn't change anything.

.gitguardian.example.yml
```diff
@@ -36,3 +36,15 @@ secret:
   # Detectors to ignore.
   ignored-detectors: # default: []
     - Generic Password
+
+iac:
+  # Exclude files and paths by globbing
+  ignored-paths:
+    - 'tests/'
```
sylvain-baud-gg, line 43:
```suggestion
    - 'tests/*'
```

.gitguardian.example.yml
```diff
@@ -36,3 +36,15 @@ secret:
   # Detectors to ignore.
   ignored-detectors: # default: []
     - Generic Password
+
+iac:
+  # Exclude files and paths by globbing
+  ignored-paths:
```
sylvain-baud-gg, line 42: Better to have an example with at least two ignored paths to show it is possible

.gitguardian.example.yml
```diff
@@ -36,3 +36,15 @@ secret:
...
+    - 'tests/'
+
+  # Ignore IaC vulnerability types by their id
+  ignored-policies:
+    - GG_IAC_0000
```
sylvain-baud-gg, line 47: Better to have an example with at least two ignored policies to show it is possible

.gitguardian.example.yml
```diff
@@ -36,3 +36,15 @@ secret:
...
+
+  # Precise vulnerabilities minimum severity to scan for
```
sylvain-baud-gg, line 49:
```suggestion
  # Minimum severity of the policies
```

.gitguardian.example.yml
```diff
@@ -36,3 +36,15 @@ secret:
   # Detectors to ignore.
   ignored-detectors: # default: []
     - Generic Password
+
+iac:
+  # Exclude files and paths by globbing
+  ignored-paths:
+    - 'tests/'
+
+  # Ignore IaC vulnerability types by their id
```
sylvain-baud-gg, line 45:
```suggestion
  # IaC vulnerabilities to ignore.
```

sylvain-baud-gg (changes requested):

.gitguardian.example.yml
carla-gitguardian: Both work, only `test/` is recursive and `test/*` is not. Why do you think adding `*` is best?

sylvain-baud-gg: I think it's good to be consistent with the secret example.

.gitguardian.example.yml
carla-gitguardian: This value was suggested by the linear ticket. As the other ones, maybe it is intended to not change anything ; such as having only one policy with id `GG_IAC_0000`?

sylvain-baud-gg: See Henri's answer on the Linear issue

GG-HH (approved):

sylvain-baud-gg (approved):

PR #626, opened 2023-07-11
sort paths to enable iac scan all caching

sylvain-baud-gg:
When receiving tar files in GIM, we first hash the tarfile bytes content to generate a key used for caching.
The goal is to not scan the same tar file twice.

However, the order of bytes are important and if we do not generate the tarfile using always the same path order, the tarfile will have different bytes order (event if the uncompressed directory is the same)


gg-teolohrer (approved):
LGTM, good catch!

PR #629, opened 2023-07-11
teolohrer/iac 116 create ggshield iac scan pre receive command

gg-teolohrer:
# Feature: add `ggshield iac scan pre-receive` command

## Description

In the pre-receive case, the command is executed in the `.git/` directory, not in the working tree.
This has several impacts on the existing `iac scan diff` implementation.

## Changes

- change `git_shell` commands to handle the case where the current directory is a `.git/` and not a working tree
- change `iac scan diff` to include the possibility to force the current state to a given reference. In prereceive context, we cannot move the HEAD (due to quarantine) so we have to explicitly dump the given reference.
- update a message about the impossibility to read a file: since we do not have direct access to the files in pre-receive context, we only hint at the location, but display no error.
- add tests
- add a workaround for the `--all` option on pre-receive, as we do not have access to the files. We perform a diff scan with unspecified reference, which defaults to only scanning the current files. We then convert the diff scan result into a path scan result for proper display.

### DIFF
--- a/ggshield/cmd/iac/scan/ci.py
+++ b/ggshield/cmd/iac/scan/ci.py
@@ -26,7 +26,7 @@ def scan_ci_cmd(
     minimum_severity: str,
     ignore_policies: Sequence[str],
     ignore_paths: Sequence[str],
-    all: bool,
+    scan_all: bool,
     directory: Optional[Path] = None,
     **kwargs: Any,
 ) -> int:
@@ -39,15 +39,19 @@ def scan_ci_cmd(
     if directory is None:
         directory = Path().resolve()
     update_context(ctx, exit_zero, minimum_severity, ignore_policies, ignore_paths)
-    if all:
+    if scan_all:
         result = iac_scan_all(ctx, directory)
         return display_iac_scan_all_result(ctx, directory, result)
 
     config = ctx.obj["config"]
     commit_list, _ = collect_commit_range_from_ci_env(config.verbose)
-    reference, current = commit_list[0], commit_list[-1]
-    if current and len(commit_list) >= 2:
-        result = iac_scan_diff(ctx, directory, reference, current=current)
-    else:
-        result = iac_scan_diff(ctx, directory, reference, include_staged=True)
+    reference, current_ref = commit_list[0], commit_list[-1]
+
+    # If we failed to fetch a current reference, we set it to HEAD
+    if len(commit_list) < 2 or not current_ref:
+        current_ref = "HEAD"
+
+    result = iac_scan_diff(
+        ctx, directory, reference, current_ref=current_ref, include_staged=True
+    )
     return display_iac_scan_diff_result(ctx, directory, result)
--- a/ggshield/cmd/iac/scan/diff.py
+++ b/ggshield/cmd/iac/scan/diff.py
@@ -17,8 +17,14 @@ from ggshield.cmd.iac.scan.iac_scan_utils import (
     get_iac_tar,
     handle_scan_error,
 )
+from ggshield.core.file_utils import get_empty_tar
 from ggshield.core.filter import is_filepath_excluded
-from ggshield.core.git_shell import INDEX_REF, Filemode, get_diff_files_status
+from ggshield.core.git_shell import (
+    INDEX_REF,
+    Filemode,
+    get_diff_files_status,
+    get_filepaths_from_ref,
+)
 from ggshield.core.text_utils import display_info, display_warning
 from ggshield.iac.collection.iac_diff_scan_collection import IaCDiffScanCollection
 from ggshield.iac.filter import is_iac_file_path
@@ -57,9 +63,11 @@ def scan_diff_cmd(
     display_warning(
         "This feature is still in beta, its behavior may change in future versions."
     )
+
     if directory is None:
         directory = Path().resolve()
     update_context(ctx, exit_zero, minimum_severity, ignore_policies, ignore_paths)
+
     result = iac_scan_diff(ctx, directory, ref, staged)
     return display_iac_scan_diff_result(ctx, directory, result)
 
@@ -67,23 +75,45 @@ def scan_diff_cmd(
 def iac_scan_diff(
     ctx: click.Context,
     directory: Path,
-    ref: str,
-    include_staged: bool = False,
-    current: Optional[str] = None,
+    previous_ref: Optional[str],
+    include_staged: bool,
+    current_ref: Optional[str] = None,
 ) -> Union[IaCDiffScanResult, IaCSkipScanResult, None]:
+    """
+    Performs a diff scan for IaC vulnerabilities,
+    comparing two git reference. Vulnerabilities are flagged as new, removed or
+    remaining depending on whether they appear in the `current_ref` and `previous_ref`
+    git references.
+
+    :param ctx: click.Context with CLI arguments
+    :param directory: path to the location we want to scan.
+    :param previous_ref: git reference to the state of reference for the analysis
+    :param include_staged: bool whether or not we want to consider the staged files
+    only when the current reference is set to None.
+    :param current_ref: optional git reference to the current state, defaults to None.
+    When set to None, the current state is the indexed files currently on disk.
+    :return: IacDiffScanResult if the scan was performed; IaCSkipScanResult if the scan
+    was skipped (i.e. no IaC files were detected or changed between the two references)
+    """
     config = ctx.obj["config"]
     client = ctx.obj["client"]
     exclusion_regexes = ctx.obj["exclusion_regexes"]
 
     verbose = config.user_config.verbose if config and config.user_config else False
     if verbose:
-        display_info(f"> Scanned files in reference {ref}")
-        filepaths = filter_iac_filepaths(directory, get_git_filepaths(directory, ref))
-        for filepath in filepaths:
-            display_info(f"- {click.format_filename(filepath)}")
-        display_info("")
+        if previous_ref is None:
+            display_info(
+                "> No file to scan in reference. This might be a new repository."
+            )
+        else:
+            display_info(f"> Scanned files in reference {previous_ref}")
+            filepaths = filter_iac_filepaths(
+                directory, get_git_filepaths(directory, previous_ref)
+            )
+            for filepath in filepaths:
+                display_info(f"- {click.format_filename(filepath)}")
+            display_info("")
 
-    current_ref = current
     if current_ref is None:
         current_ref = INDEX_REF if include_staged else "HEAD"
     if verbose:
@@ -92,28 +122,43 @@ def iac_scan_diff(
         else:
             display_info("> Scanned files in current state")
         filepaths = filter_iac_filepaths(
-            directory, get_git_filepaths(directory, current_ref)
+            directory, get_git_filepaths(directory=directory, ref=current_ref)
         )
         for filepath in filepaths:
             display_info(f"- {click.format_filename(filepath)}")
 
+    modified_iac_files = []
+
     # Check if IaC files were created, deleted or modified
-    files_status = get_diff_files_status(
-        wd=str(directory), ref=ref, staged=include_staged, similarity=100
-    )
-    modified_modes = [Filemode.NEW, Filemode.DELETE, Filemode.MODIFY]
-    modified_iac_files = [
-        file
-        for file, mode in files_status.items()
-        if mode in modified_modes
-        and not is_filepath_excluded(str(file), exclusion_regexes)
-        and is_iac_file_path(file)
-    ]
+    if previous_ref is None:
+        # This means we are scanning all commits up to now.
+        filepaths = get_filepaths_from_ref(wd=str(directory), ref=current_ref)
+        modified_iac_files = list(filter(is_iac_file_path, filepaths))
+    else:
+        files_status = get_diff_files_status(
+            wd=str(directory),
+            current_ref=current_ref,
+            ref=previous_ref,
+            staged=include_staged,
+            similarity=100,
+        )
+        modified_modes = [Filemode.NEW, Filemode.DELETE, Filemode.MODIFY]
+        modified_iac_files = [
+            file
+            for file, mode in files_status.items()
+            if mode in modified_modes
+            and not is_filepath_excluded(str(file), exclusion_regexes)
+            and is_iac_file_path(file)
+        ]
 
     if len(modified_iac_files) == 0:
         return IaCSkipScanResult()
 
-    reference_tar = get_iac_tar(directory, ref, exclusion_regexes)
+    reference_tar = (
+        get_iac_tar(directory, previous_ref, exclusion_regexes)
+        if previous_ref is not None
+        else get_empty_tar()
+    )
     current_tar = get_iac_tar(directory, current_ref, exclusion_regexes)
 
     scan_parameters = IaCScanParameters(
--- a/ggshield/cmd/iac/scan/prereceive.py
+++ b/ggshield/cmd/iac/scan/prereceive.py
@@ -1,42 +1,103 @@
+import logging
 from pathlib import Path
-from typing import Any, Optional, Sequence
+from typing import Any, Sequence
 
 import click
+from pygitguardian.iac_models import IaCScanResult
 
-from ggshield.cmd.iac.scan.diff import display_iac_scan_diff_result, iac_scan_diff
+from ggshield.cmd.iac.scan.diff import iac_scan_diff
 from ggshield.cmd.iac.scan.iac_scan_common_options import (
     add_iac_scan_common_options,
     all_option,
-    directory_argument,
     update_context,
 )
-from ggshield.core.text_utils import display_warning
+from ggshield.cmd.iac.scan.iac_scan_utils import (
+    IaCSkipScanResult,
+    create_output_handler,
+)
+from ggshield.core.git_hooks.prereceive import get_breakglass_option, parse_stdin
+from ggshield.core.git_shell import check_git_ref, is_valid_git_commit_ref
+from ggshield.core.utils import EMPTY_TREE
+from ggshield.iac.collection.iac_diff_scan_collection import IaCDiffScanCollection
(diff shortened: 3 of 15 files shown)


ggshield/cmd/iac/scan/prereceive.py
```diff
@@ -0,0 +1,167 @@
...
+from ggshield.core.git_shell import git
+from ggshield.core.text_utils import display_error
+from ggshield.core.utils import EMPTY_SHA, EMPTY_TREE, PRERECEIVE_TIMEOUT
+from ggshield.iac.collection.iac_diff_scan_collection import IaCDiffScanCollection
+
+
+logger = logging.getLogger(__name__)
+
+REMEDIATION_MESSAGE = """  A pre-receive hook set server side prevented you from pushing secrets.
+  Since the secret was detected during the push BUT after the commit, you need to:
```
GG-HH, line 28: This will need to be changed.

GG-HH (changes requested):

ggshield/cmd/iac/scan/prereceive.py
```diff
@@ -0,0 +1,167 @@
...
+  1. rewrite the git history making sure to replace the secret with its reference (e.g. environment variable).
+  2. push again."""
+
+BYPASS_MESSAGE = """\n     git push -o breakglass"""
+
+
+def get_prereceive_timeout() -> float:
+    try:
+        return float(os.getenv("GITGUARDIAN_TIMEOUT", PRERECEIVE_TIMEOUT))
+    except BaseException as e:
```
tchaumeny, line 38: Any specific reason why you'd like to catch all `BaseException` and not just `Exception` (which is what we generally want, see [here](https://docs.python.org/3/tutorial/errors.html#:~:text=BaseException%20is%20the%20common%20base,that%20the%20program%20should%20terminate.))

ggshield/cmd/iac/scan/prereceive.py
```diff
@@ -0,0 +1,167 @@
...
+        display_error(f"Unable to parse GITGUARDIAN_TIMEOUT: {str(e)}")
+        return PRERECEIVE_TIMEOUT
+
+
+def get_breakglass_option() -> bool:
+    """Test all options passed to git for `breakglass`"""
+    raw_option_count = os.getenv("GIT_PUSH_OPTION_COUNT", None)
+    if raw_option_count is not None:
+        option_count = int(raw_option_count)
```
tchaumeny, line 47: Since you catch errors related to parsing `GITGUARDIAN_TIMEOUT`, shouldn't you do the same for `GIT_PUSH_OPTION_COUNT`? (in case it's not a correct `int`)

ggshield/cmd/iac/scan/prereceive.py
gg-teolohrer: This comes from the `secrets` side for `pre-receive`, and will be refactored / corrected.

ggshield/cmd/iac/scan/prereceive.py
gg-teolohrer: This comes from the secrets side for pre-receive, and will be refactored / corrected.

ggshield/cmd/iac/scan/diff.py
```diff
@@ -110,7 +104,11 @@ def scan_diff_cmd(
 
 
 def iac_scan_diff(
-    ctx: click.Context, directory: Path, ref: str, include_staged: bool
+    ctx: click.Context,
+    directory: Path,
+    ref: str,
+    include_staged: bool,
+    current_ref: Optional[str] = None,
```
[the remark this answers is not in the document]
gg-teolohrer: Let me know if the naming is better, with the corresponding docstring.

ggshield/cmd/iac/scan/prereceive.py
gg-teolohrer: LMK if that's ok now.

ggshield/cmd/iac/scan/diff.py
```diff
@@ -137,22 +128,36 @@ def iac_scan_diff(
             display_info(f"- {click.format_filename(filepath)}")
 
     # Check if IaC files were created, deleted or modified
-    files_status = get_diff_files_status(
-        wd=str(directory), ref=ref, staged=include_staged, similarity=100
+    if previous_ref is None:
+        # This means we are scanning all commits up to now.
+        # TODO: We might want to check that some changes exist
+        # to prevent from scanning a new and empty git repository
+        pass
```
sylvain-baud-gg, line 135: Here the comment says "_we are scanning all commits up to now_".
However we previously displayed "_No file to scan in reference. This might be a new repository_" for the `previous_ref is None` case.

What is the real use case of previous_ref is none ?

Also why pass here ? A scan diff with None as previous ref should be equivalent to a scan all on the current ref no ? WIth all found vulns are categorized as "new"

ggshield/cmd/iac/scan/scan.py
```diff
@@ -1,89 +1 @@
...
-            command_path=ctx.command_path,
-            scan_mode=ScanMode.IAC_DIRECTORY,
-        ).get_http_headers(),
-    )
-
-    if not scan.success or not isinstance(scan, IaCScanResult):
-        handle_scan_error(client, scan)
-        return None
-    return scan
+# test
```
sylvain-baud-gg, line 1: Why not deleting the file ?

ggshield/cmd/iac/scan/all.py
```diff
@@ -0,0 +1,95 @@
...
+    update_context(ctx, exit_zero, minimum_severity, ignore_policies, ignore_paths)
+
+    result = iac_scan_all(ctx, directory)
+    return display_iac_scan_all_result(ctx, directory, result)
+
+
+def iac_scan_all(
+    ctx: click.Context, directory: Path
+) -> Union[IaCScanResult, IaCSkipScanResult, None]:
+    paths = get_iac_files_from_path(
```
sylvain-baud-gg, line 50: `get_iac_files_from_path` generates a list of path using the `get_filepaths` function relying on a set.

This cause the path list to have inconsistent order and therefore the tar generated have a different byte order when called twice on the same directory. This cause the hash_key of GIM cache to be different when scanning twice the same directory so the cache is ignored.

If this is the case, a path.sort() might be necessary here.

tests/functional/iac/test_scan_prereceive.py
```diff
@@ -0,0 +1,54 @@
...
+
+    local_repo = Repository.clone(remote_repo.path, tmp_path / "local")
+
+    hook_path = remote_repo.path / "hooks" / "pre-receive"
+    hook_path.write_text(HOOK_CONTENT)
+    hook_path.chmod(0o700)
+    return local_repo
+
+
+def test_iac_scan_prereceive(repo_with_hook) -> None:
```
sylvain-baud-gg, line 29: No GIVEN docstring  ?

tests/functional/iac/test_scan_prereceive.py
```diff
@@ -0,0 +1,54 @@
...
+    with pytest.raises(CalledProcessError) as exc:
+        repo_with_hook.push()
+
+    # AND the error message contains the leaked secret
+    stderr = exc.value.stderr.decode()
+    assert "1 new incident" in stderr
+    assert "vuln.tf" in stderr
+
+
+def test_iac_scan_prereceive_branch_without_new_commits(repo_with_hook) -> None:
```
sylvain-baud-gg, line 47: No GIVEN docstring  ?

ggshield/cmd/iac/scan/diff.py
```diff
@@ -57,57 +63,103 @@ def scan_diff_cmd(
...
-    modified_modes = [Filemode.NEW, Filemode.DELETE, Filemode.MODIFY]
-    modified_iac_files = [
-        file
-        for file, mode in files_status.items()
-        if mode in modified_modes
-        and not is_filepath_excluded(str(file), exclusion_regexes)
-        and is_iac_file_path(file)
-    ]
+    if previous_ref is None:
+        # This means we are scanning all commits up to now.
```
sylvain-baud-gg, line 135: Here the comment says "we are scanning all commits up to now".
However we previously displayed "No file to scan in reference. This might be a new repository" for the previous_ref is None case.

What is the real use case of previous_ref is none ?

gg-teolohrer: `previous_ref` is None when the pre-receive command failed to find a root for a newly pushed branch.

Here, `None` is interpreted as the global root for the git tree. I wish we could rely on `EMPTY_TREE` or `EMPTY_SHA`, but it is dependent on whether people pushed an empty commit when creating the repo.

For pre-receive, setting `previous_ref` to `None` is not exactly equivalent to performing a `scan all` : in the current implementation, scan all requires a directory and file paths. In the pre-receive context, we only have access to git objects. `scan diff` handles that properly, hence the `previous_ref=None`.

sylvain-baud-gg: Ok, clear for me. Thanks for the explications

ggshield/cmd/iac/scan/diff.py
```diff
@@ -57,57 +63,103 @@ def scan_diff_cmd(
...
 ) -> Union[IaCDiffScanResult, IaCSkipScanResult, None]:
+    """
+    Performs a diff scan for IaC vulnerabilities,
+    comparing two git reference. Vulnerabilities are flagged as new, removed or
+    remaining depending on whether they appear in the `current_ref` and `previous_ref`
+    git references.
+
+    :param ctx: click.Context with CLI arguments
+    :param directory: path to the location we want to scan. Defaults to the current
+    directory.
```
GG-HH, line 90: There is no default for the directory here.

ggshield/cmd/iac/scan/prereceive.py
```diff
@@ -1,42 +1,118 @@
...
+    before, after = before_after
+    if before == after:
+        click.echo(
+            "Pushed branch does not contain any new commit.",
+            err=True,
+        )
+        return 0
+
+    assert before != EMPTY_SHA
+    assert after != EMPTY_SHA
```
GG-HH, line 69: Should we factorize this with secrets ?

tests/functional/iac/test_iac_scan_diff.py
```diff
@@ -180,7 +180,7 @@ def test_iac_scan_diff_only_tracked_iac_with_ignore(tmp_path: Path) -> None:
     assert "No IaC files changed" in result.stdout
 
 
-@pytest.mark.parametrize("staged", (True, False))
+@pytest.mark.parametrize("staged", (False,))
```
GG-HH, line 183: Why did you removed this test case ?

GG-HH (changes requested):
Few comments.
Since there are a lot of commits and changes in the git core, I'd like this to be reviewed by agateau.
Can you re-write the history to have meaningful commits (especially regarding refactoring and command changes). Thanks

ggshield/cmd/iac/scan/diff.py
gg-teolohrer: removed

ggshield/cmd/iac/scan/prereceive.py
gg-teolohrer: done

tests/functional/iac/test_iac_scan_diff.py
gg-teolohrer: my bad I restored it

GG-HH (approved):

ggshield/core/git_shell.py
```diff
@@ -77,7 +86,7 @@ def check_git_dir(wd: Optional[str] = None) -> None:
 
 
 def get_git_root(wd: Optional[str] = None) -> str:
-    return git(["rev-parse", "--show-toplevel"], cwd=wd)
+    return git(["rev-parse", "--git-dir"], cwd=wd)
```
agateau-gg, line 89: This new implementation returns a different path. `--show-toplevel` returns the directory containing the `.git` dir, whereas `--git-dir` returns the `.git` dir itself:

```
~/src/ggshield
$ git rev-parse --show-toplevel 
/home/agateau/src/ggshield

~/src/ggshield
$ git rev-parse --git-dir       
.git

~/src/ggshield
$ cd ggshield/core

~/src/ggshield/ggshield/core
$ git rev-parse --show-toplevel
/home/agateau/src/ggshield

~/src/ggshield/ggshield/core
$ git rev-parse --git-dir      
/home/agateau/src/ggshield/.git
```

I suspect this change breaks [.env loading from `utils.py`](https://github.com/GitGuardian/ggshield/blob/main/ggshield/core/utils.py#L218). But no tests broke, so either I am wrong, or there are some holes in the test coverage for .env loading 😟.

It also does not always return an absolute path, which could be a problem.

ggshield/core/git_shell.py
```diff
@@ -319,3 +346,10 @@ def tar_from_ref_and_filepaths(
             tar.addfile(tarinfo, fileobj=data)
 
     return tar_stream.getvalue()
+
+
+def get_empty_tar() -> bytes:
```
agateau-gg, line 351: `tar_from_ref_and_filepaths()` in `git_shell`, it makes less sense to me to have a `get_empty_tar()` function in this module.

I suggest either moving `get_empty_tar()` to `core.file_utils` or create a `core.tar_utils` module and move both `get_empty_tar()` and `tar_from_ref_and_filepaths()` to it.

ggshield/cmd/iac/scan/ci.py
```diff
@@ -39,15 +39,19 @@ def scan_ci_cmd(
...
     if current and len(commit_list) >= 2:
-        result = iac_scan_diff(ctx, directory, reference, current=current)
+        result = iac_scan_diff(
+            ctx, directory, reference, current_ref=current, include_staged=True
+        )
     else:
-        result = iac_scan_diff(ctx, directory, reference, include_staged=True)
+        result = iac_scan_diff(
+            ctx, directory, reference, current_ref="HEAD", include_staged=True
+        )
```
agateau-gg, line 56: Can be simplified to:

```python
reference, current_ref = commit_list[0], commit_list[-1]
if not current_ref or len(commit_list) < 2:
   current_ref = "HEAD"
result = iac_scan_diff(
   ctx, directory, reference, current_ref=current_ref, include_staged=True)
```

agateau-gg (changes requested):
+1 for factoring out code of `secret.scan.prereceive` to `core.git_hooks.prereceive` (though I would have preferred if there were only one commit moving the code instead of one duplicating and another one removing the just created duplication)

Another +1 for moving `PRERECEIVE_TIMEOUT` from `core.utils` :wink:.

On the other hand, this PR would have been easier to review if there were no changes made in one commit and reverted in the next one.

When a PR is too large to be reviewed as a whole it helps reviewers a lot if each commit is self-contained, making it possible to review the PR commit by commit. It's not only good for reviews though, it also helps our future selves next time we inspect the git history to understand a problem.

Do not hesitate to rebase and squash related changes so that the commits in your PR tell a story.

ggshield/core/git_shell.py
gg-teolohrer: Good catch. I used this workaround to have a generic solution even when we're not in a working tree, like with pre-receive. Latest version should restore previous behavior.

ggshield/core/git_shell.py
```diff
@@ -76,8 +85,20 @@ def check_git_dir(wd: Optional[str] = None) -> None:
         raise UsageError("Not a git directory.")
 
 
+@lru_cache(None)
```
agateau-gg, line 88: We should not cache the return of this function because if `wd` is None the function uses the current working directory, making it dependent of outside parameters. For example something like this:

```python
os.chdir("foo")
foo_root = get_git_root()
os.chdir("bar")
bar_root = get_git_root()  # error: get_git_root() returns the root for foo
```

This is admittedly unlikely to happen right now, and changing the current working directory in a running process is not a good idea but:
- we have been bitten by that in the past, better not add another time-bomb;
- part of our code does change the current working directory 😞.

This implementation also causes `git rev-parse --git-dir`  and `git rev-parse --show-toplevel` to be called multiple times, which can be slow on Windows (because Windows is slow at spawning processes). A way to avoid this would be to lru_cache the results of the calls. Something like this:

```python
@lru_cache
def _git_rev_parse(option: str, wd: str) -> Optional[str]:
    try:
        return git(["rev-parse", option], cwd=wd)
    except subprocess.CalledProcessError:
        return None
```

Then it can be used by the other functions:

```python
def is_git_dir(wd: str) -> bool:
    return _git_rev_parse("--git-dir", wd) is not None

def is_git_working_tree(wd: str) -> bool:
    return _git_rev_parse("--show-toplevel", wd) is not None

def get_git_root(wd: Optional[str] = None) -> str:
    if wd is None:
        wd = os.getcwd()
    check_git_dir(wd)
    top_level = _git_rev_parse("--show-toplevel", wd)
    if top_level is not None:
        return top_level
    return str(Path(_git_rev_parse("--git-dir", cwd=wd)).resolve())

```

tests/unit/core/test_file_utils.py
```diff
@@ -0,0 +1,21 @@
...
+def test_get_empty_tar():
+    # WHEN creating an empty tar
+    empty_tar_bytes = get_empty_tar()
+    tar_stream = BytesIO(empty_tar_bytes)
+
+    # THEN the file is considered as a .tar
+    version = sys.version_info
+    # `tarfile.is_tarfile` won't work until Python 3.9
+    if version.major > 3 or version.major == 3 and version.minor > 8:
+        assert tarfile.is_tarfile(tar_stream)
```
agateau-gg, line 17: Can be simplified to:
```suggestion
    # `tarfile.is_tarfile` won't work until Python 3.9
    if sys.version_info >= (3, 9):
        assert tarfile.is_tarfile(tar_stream)
```

agateau-gg (changes requested):
A few last minor remarks, looking good!

agateau-gg (approved):
Nice :+1:

PR #630, opened 2023-07-12
ignore-policy-gg-iac-0079-on-self-scan

sylvain-baud-gg:
In the iac scan tests, we need to test the CI integration. To do this we scan the HEAD of this repo.
However with the update of iac-api, new policies have been added.
Among them, the policy GG_IAC_0079 is detected in this repo.
Information on the policy :
https://docs.gitguardian.com/iac-scanning/policies/GG_IAC_0079
Present in these files :
Dockerfile
scripts/update-pipfile-lock/Dockerfile
tests/functional/data/docker-leaking-in-env/Dockerfile
tests/functional/data/docker-leaking-in-layer/Dockerfile 

For now, we chose to ignore it  to enable the CI to pass but another solution would be to correct this vulnerabilities.

### DIFF
--- a/.gitguardian.yml
+++ b/.gitguardian.yml
@@ -1,0 +1,9 @@
+# Required, otherwise ggshield considers the file to use the deprecated v1 format
+version: 2
+
+iac:
+  # IaC vulnerabilities to ignore
+  # We don't want to fix this vulnerability because many CI systems (including GitHub action and Azure pipelines)
+  # expect the user inside the container to be root.
+  ignored-policies:
+    - GG_IAC_0079


agateau-gg (changes requested):
A few remarks:

We don't want to fix this vulnerability because many CI systems (including GitHub action and Azure pipelines) expect the user inside the container to be root. I am not fond of that idea, but we have no control over what the CI systems expect.

We actually used to have a separated user, but we had to change that recently (see https://github.com/GitGuardian/ggshield/commit/93a53d100617a6b695587058073b7090aebbfc87).

Can you trim the file so it contains entries for config values which are not the default ones?

sylvain-baud-gg:
> A few remarks:
> 
> We don't want to fix this vulnerability because many CI systems (including GitHub action and Azure pipelines) expect the user inside the container to be root. I am not fond of that idea, but we have no control over what the CI systems expect.
> 
> We actually used to have a separated user, but we had to change that recently (see [93a53d1](https://github.com/GitGuardian/ggshield/commit/93a53d100617a6b695587058073b7090aebbfc87)).
> 
> Can you trim the file so it contains entries for config values which are not the default ones?

Thanks for the additional information.

I trimmed the file to keep only non-default value.
Also I added a part of your explication as a comment for it to be documented.

I used `git commit --amend` to keep a clean history.

agateau-gg (approved):
Looks good, +1 for adding a comment explaining why we ignore the policy!

Issue #561, opened 2023-06-09
ggshield and pygitguardian disagree on the way to check the maximum document size


agateau-gg:
## The problem

Sometimes a file is considered as small enough to be scanned by ggshield, but then it is rejected by py-gitguardian. This is because ggshield checks the character length of the **decoded** string whereas py-gitguardian checks the byte length of the **utf-8 encoded** string.

ggshield can't just use the byte length of the file it reads though: it may not be utf-8, and GitGuardian API expects utf-8 encoded content.

This diagram summarizes the current situation:

```
     content in bytes (any encoding)
             ↓
---------[ggshield]--------------------
             ↓
      detect encoding
      decode content
      decoded size too large? → exit
             ↓
       decoded content
             ↓ 
------[py-gitguardian]-----------------
             ↓
      encode to utf-8
      encoded size too large? → exit
             ↓
          { HTTP }
             ↓
------[GitGuardian API]----------------
    expects utf-8 content
             ↓
    check encoded size
```

This file: [561.txt](https://github.com/GitGuardian/ggshield/files/11980596/561.txt), consists of one `é` character followed by enough `e` characters so that the encoded size is too long by one byte. It can be used to reproduce the bug:

```
$ ggshield secret scan path 561.txt
Scanning Path... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 1 files scanned out of 1 0:00:00

Error: Scanning failed: file exceeds the maximum allowed size of 1048576B
```

### Do not check the document size on ggshield side

This would make ggshield load and decode **all** files in memory, including huge ones, only to have them rejected by py-gitguardian. That is wasteful.

### Change ggshield check to use the encoded size

This would mean each file would be encoded twice: once by ggshield, then again by py-gitguardian. ggshield cannot pass the encoded content to py-gitguardian, because py-gitguardian only accepts decoded content.

## Proposed solution

The proposed solution is to make ggshield do the utf-8 encoding (and check the size) and make py-gitguardian accept encoded content.

Details:

1. Add a static `encode_content(str) -> bytes` method to `pygitguardian.Document`. This would perform what the `post_load` methods of `DocumentSchema` do. Replace said `post_load` methods with a call to `Document.encode_content()`.

[...]

PR #631, opened 2023-07-12
Fix ggshield secret scan trying to scan too large documents

agateau-gg:
## Description

`Scannable.is_longer_than()` currently compares the size passed to it to the decoded size, but what we actually want to compare it to is the *UTF-8 encoded size*.

To implement this without reading the content multiple times, `Scannable` implementations now store the UTF-8 encoded size, and `Scannable` helper methods (`_is_file_longer_than()` and `_decode_bytes()`) return the UTF-8 encoded size in addition to their current return value.

When the content they work on is already UTF-8, the helpers try to avoid reading it until necessary.

This PR also integrates the latest py-gitguardian changes, which are also related to this issue.

Issue: #561

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "12b3ddffa6eb5dfb81bfe851b5df635e17a92a0873514b27bfaab9bfd5809f35"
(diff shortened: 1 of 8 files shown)


Walz (approved):
LGTM

PR #634, opened 2023-07-13
imp(hmsl): #EXPL-150 more cheerfull message when no leak has been found

phililippe:
This PR proposes improvements in messages displayed to users when no leak has been found by HMSL:
- more cheerful text message if no leak is found
- warning color in other cases
- count of leaks found to avoid empty json array as sole output if no leak has been found


GG-HH (approved):

GG-HH:
Thanks!

PR #637, opened 2023-07-17
IaC scans: display a link to the policy documentation

Paul-GitGuardian:
https://linear.app/gitguardian/issue/IAC-136/ggshield-add-link-to-the-policy-documentation

### DIFF
--- a/tests/unit/iac/output/test_iac_diff_text_output.py
+++ b/tests/unit/iac/output/test_iac_diff_text_output.py
@@ -11,6 +11,27 @@ from ggshield.iac.collection.iac_diff_scan_collection import IaCDiffScanCollecti
 from ggshield.iac.output.iac_text_output_handler import IaCTextOutputHandler
 
 
+POLICY_DOC_URL = "https://docs.gitguardian.com/iac-scanning/policies/GG_IAC_0021"
+
+
+def _generate_iac_file_result() -> IaCFileResult:
+    return IaCFileResult(
+        filename="FILENAME.tf",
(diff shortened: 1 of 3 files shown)


tests/unit/iac/output/test_iac_diff_text_output.py
```diff
@@ -213,6 +214,7 @@ def test_iac_scan_diff_vuln_verbose():
     assert_iac_diff_version_displayed(output)
     assert re.search(r"FILENAME\.tf.*1 new incident detected", output)
     assert_iac_diff_summary_displayed(output, new=1, unchanged=1, deleted=1)
+    assert_documentation_url_displayed(output, "DOCUMENT_URL")
```
GG-HH, line 217: Could we change the fixture and the test to have something that looks like a documentation URL ?

tests/unit/iac/output/test_iac_text_output.py
```diff
@@ -124,6 +125,14 @@ def assert_no_failures_displayed(result: Result):
     assert "Error scanning. Results may be incomplete." not in result.stdout
 
 
+def assert_documentation_url_displayed(result: Result):
+    assert re.search(
+        r"\((GG_IAC_\d{4})\).+https://docs.gitguardian.com/iac-scanning/policies/\1",
```
GG-HH, line 130: Nit: your regex is incorrect because some characters aren't escape such as the dots in the URL

GG-HH (changes requested):

PR #638, opened 2023-07-18
Improve help message for --ignore-path (iac) and --exclude (secret)

Paul-GitGuardian:
`--ignore-path` and `--exclude` expect a pattern rather than a file path.

Regex patterns do not work, as special characters are escaped.

Here is how the input value is transformed:
```
# Handle start/end of pattern
if pattern[-1] != "/":
    pattern += "$"
if pattern[0] == "/":
    pattern = "^" + pattern[1:]
else:
    pattern = "(^|/)" + pattern

# Replace * and ** sequences
pattern = re.sub(r"\\\*\\\*/", "([^/]+/)*", pattern)
pattern = re.sub(r"\\\*", "([^/]+)", pattern)
```

And here is how the filtering is done:
```
def is_filepath_excluded(filepath: str, exclusion_regexes: Set[re.Pattern]) -> bool:
    return any(r.search(str(PurePosixPath(Path(filepath)))) for r in exclusion_regexes)
```

### DIFF
--- a/ggshield/cmd/iac/scan/iac_scan_common_options.py
+++ b/ggshield/cmd/iac/scan/iac_scan_common_options.py
@@ -62,7 +62,9 @@ _ignore_path_option = click.option(
     default=None,
     type=click.Path(),
     multiple=True,
-    help="Do not scan the specified paths.",
+    help="""
+    Do not scan paths that match the specified glob-like patterns.
+    """,
 )
 
 all_option = click.option(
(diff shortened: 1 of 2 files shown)


Paul-GitGuardian:
@GG-HH Who in the secrets team should review this?

GG-HH (approved):

Walz (commented):
I think we can simply say ggshield accept [glob-like patterns](https://en.wikipedia.org/wiki/Glob_(programming)). Most people using a CLI should be familiar with it.

NIT: both are correct English but we use filename (without space) in the rest of the doc.

(We will need to update the [documentation](https://docs.gitguardian.com/ggshield-docs/reference/overview#options) when this change is release.)

Paul-GitGuardian:
I've made the change, thank you.
FYI the documentation change is here: https://gitlab.gitguardian.ovh/gg-code/public-docs/-/merge_requests/846

PR #645, opened 2023-07-24
feat(sca): add configuration from config file

xblanchot-gg:


ggshield/core/config/user_config.py
```diff
@@ -81,6 +83,25 @@ class IaCConfig(FilteredConfig):
...
+        )
+
+
+@marshmallow_dataclass.dataclass
+class SCAIgnoredVulnerability(FilteredConfig):
+    """
+    A model of an ignored vulnerability for SCA
+    """
+
+    ghsa_id: str = field(metadata={"validate": validate_ghsa_id})
```
tchaumeny, line 99: Couldn't you declare a regex directly (see https://github.com/lovasoa/marshmallow_dataclass#custom-newtype-declarations)?

ggshield/core/config/user_config.py
```diff
@@ -81,6 +83,25 @@ class IaCConfig(FilteredConfig):
...
+    path: str
```
tchaumeny, line 100: What does the `path` represent here? Maybe add explanations in the docstring of `SCAIgnoredVulnerability`?

ggshield/core/config/user_config.py
xblanchot-gg: I did not manage to make it work

PR #628, opened 2023-07-11
cbouchard/iac-120/iac_scan_ci_command

cyrvan-b:
Create the ci command and add tests.

Issue: [IAC-120](https://linear.app/gitguardian/issue/IAC-120/create-ggshield-iac-scan-ci-command).

PR #646, opened 2023-07-25
feat(iac): github action should now use the new CI command

Paul-GitGuardian:
Depends on #628 to be deployed for the stable action to work.

Updates the IaC Github Action to use the new `ggshield iac scan ci` command instead of `ggshield iac scan all`.

### DIFF
--- a/changelog.d/20230725_144801_paul.beslin.ext_update_github_action_iac_command.md
+++ b/changelog.d/20230725_144801_paul.beslin.ext_update_github_action_iac_command.md
@@ -1,0 +1,3 @@
+### Changed
+
+- The IaC Github Action now executes the new `ggshield iac scan ci` command. This means the action will fail only if the changes introduce a new vulnerability. To fail if any vulnerability is detected, use the `ggshield iac scan ci --all` command
--- a/docker/actions-iac-entrypoint.sh
+++ b/docker/actions-iac-entrypoint.sh
@@ -12,4 +12,4 @@ git config --global --add safe.directory "$PWD"
 
 
 args=("$@")
-ggshield iac scan all -v ${args[@]}
+ggshield iac scan ci -v ${args[@]}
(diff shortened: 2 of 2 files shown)


Paul-GitGuardian:
This changes the behavior of the action: it will now only block if new vulnerabilities are introduced.
Is it ok?

changelog.d/20230725_144801_paul.beslin.ext_update_github_action_iac_command.md
```diff
@@ -0,0 +1,3 @@
+### Changed
+
+- The IaC Github Action now executes the new `ggshield iac scan ci` command. This means the action will fail only if the changes introduce a new vulnerability.
```
gg-teolohrer, line 3:
```suggestion
- The IaC Github Action now executes the new `ggshield iac scan ci` command. This means the action will fail only if the changes introduce a new vulnerability. To fail if any vulnerability is detected, use the `ggshield iac scan ci --all` command
```

gg-teolohrer (changes requested):
The change in behavior seems OK to me, but we should describe the alternative in the changelog.

cyrvan-b:
LGTM with the change on changelogs.

gg-teolohrer (approved):

cyrvan-b (approved):

GG-HH:
@Paul-GitGuardian we should update the action files as well, the description will lack some options offered by the CI scan.

docker/actions-iac-entrypoint.sh
```diff
@@ -12,4 +12,4 @@ git config --global --add safe.directory "$PWD"
 
 
 args=("$@")
-ggshield iac scan all -v ${args[@]}
+ggshield iac scan ci -v ${args[@]}
```
GG-HH, line 15:
```suggestion
ggshield iac scan ci ${args[@]}
```
We shouldn't have verbose enabled by default

PR #652, opened 2023-07-26
feat(sca): Added pre-push command for sca

amascia-gg:
added pre-push command with unit and functional tests


tests/unit/cmd/sca/test_scan_prepush.py
```diff
@@ -0,0 +1,147 @@
...
+    scan_arg: Optional[List[str]],
+    cassette: str,
+    pipfile_lock_with_vuln,
+) -> None:
+    """
+    GIVEN a remote repository and a local clone
+    WHEN executing the prepush command with or without the '--all' option
+    THEN the scan output format is correct
+    """
+    remote_repo = Repository.create(tmp_path / "remote", bare=True)
```
gg-jonathangriffe, line 33: nit : I get why there is a remote_repo, but I'm not convinced this is necessary.

ggshield/cmd/sca/scan/prepush.py
```diff
@@ -0,0 +1,64 @@
...
+    minimum_severity: str,
+    ignore_paths: Sequence[str],
+    all: bool,
+    **kwargs: Any,
+) -> int:
+    """
+    Scan as pre-push for SCA vulnerabilities.
+    By default, it will return vulnerabilities added in the pushed commits.
+    """
+    directory = Path().resolve()
```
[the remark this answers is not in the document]
gg-jonathangriffe: But maybe it isn't necessary as it doesn't seem to be there in IaC

tests/unit/cmd/sca/test_scan_prepush.py
amascia-gg: I don't understand what you suggest to modify

tests/functional/sca/test_scas_scan_ci.py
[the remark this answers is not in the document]
amascia-gg: Yes, thanks!

ggshield/cmd/sca/scan/prepush.py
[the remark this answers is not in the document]
amascia-gg: No we don't need it as it is used as a hook so it will always be the root path of the repository

tests/unit/cmd/sca/test_scan_prepush.py
gg-jonathangriffe: In this test, since there is no push, having a local_repo is enough, and we don't need to clone it from a remote repo. This line is not really useful. 
I don't know if you should really remove it since it still makes the context of the test clearer.

amascia-gg: I would prefer to leave it as is for the context if it's ok with you

ggshield/cmd/sca/scan/prepush.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Adds client and required parameters to the context
+    update_context(ctx, exit_zero, minimum_severity, ignore_paths)
+
+    _, remote_commit = collect_commits_refs(prepush_args)
+    # Will happen if this is the first push on the branch
+    has_no_remote_commit = (
+        remote_commit is None or "~1" in remote_commit or remote_commit == EMPTY_SHA
```
tchaumeny, line 52: Why does `"~1" in remote_commit` mean it's the first push?

gg-jonathangriffe (approved):

ggshield/cmd/sca/scan/prepush.py
amascia-gg: The "~1" part is a shorthand for the parent commit in Git. If it appears in the remote_commit, it indicates that the current branch is pointing to a commit that has no remote counterpart.

PR #659, opened 2023-07-31
Add sca GitHub actions

amascia-gg:
added sca GitHub actions


agateau-gg (changes requested):
Two things:
- actions-sca-entrypoint.sh must be made executable
- can you modify `scripts/release` to add `sca` to the `ACTION_PATHS` list

agateau-gg (approved):

PR #660, opened 2023-08-01
feat(sca): add ignored_vulnerabilities parameters to sca commands

amascia-gg:
- added sca in configuration file .gitguardian.example.yaml
- added `ignored_vulnerabilities` pararms for sca scans.

### DIFF
--- a/ggshield/cmd/sca/scan/sca_scan_utils.py
+++ b/ggshield/cmd/sca/scan/sca_scan_utils.py
@@ -7,6 +7,7 @@ from pygitguardian.client import _create_tar
 
 from ggshield.cmd.common_options import use_json
 from ggshield.core.config.config import Config
+from ggshield.core.config.user_config import SCAConfig
 from ggshield.core.errors import APIKeyCheckError, UnexpectedError
 from ggshield.core.git_shell import INDEX_REF
 from ggshield.core.text_utils import display_error, display_info, display_warning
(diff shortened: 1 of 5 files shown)


ggshield/cmd/sca/scan/sca_scan_utils.py
```diff
@@ -67,7 +67,10 @@ def sca_scan_all(ctx: click.Context, directory: Path) -> SCAScanAllOutput:
         empty_output.status_code = sca_filter_status_code
         return empty_output
 
-    scan_parameters = SCAScanParameters(config.user_config.sca.minimum_severity)
+    scan_parameters = SCAScanParameters(
+        minimum_severity=config.user_config.sca.minimum_severity,
+        ignored_vulnerabilities=config.user_config.sca.ignored_vulnerabilities,
+    )
```
GG-HH, line 73: Nit: could be added in a specific function that will be called here and in `sca_sca_diff` to ensure we get the same parametrization.

tests/unit/core/config/test_user_config.py
```diff
@@ -261,6 +261,32 @@ def test_sca_config(self, local_config_path):
...
+            local_config_path,
+            {
+                "version": 2,
+                "sca": {
+                    "ignored_vulnerabilities": [
+                        {
+                            "identifier": "GHSA-aaaa-bbbb-cccc",
+                            "path": "Pipfile",
+                            "comment": "Not my prob",
+                            "until": "2023-05-01T00:00:00.000Z",
```
GG-HH, line 280: What happens if you have no "Z" in the config ? (ie: you have a naive datetime)

amascia-gg: Modify code to transform all datetimes' timezone to UTC

GG-HH (approved):

PR #665, opened 2023-08-02
feat(sca): add scan diff command

GG-HH:
Add scan diff command. This will allow users to perform a differential scan with custom reference.

### DIFF
--- a/tests/functional/sca/test_sca_scan_diff.py
+++ b/tests/functional/sca/test_sca_scan_diff.py
@@ -1,0 +1,50 @@
+from pathlib import Path
+
+from tests.functional.utils import run_ggshield_sca_scan
+from tests.repository import Repository
+
+
+def test_scan_diff(tmp_path: Path, pipfile_lock_with_vuln) -> None:
+    """
+    GIVEN a repository with a commit containing a vuln,
+    one clean commits on top of it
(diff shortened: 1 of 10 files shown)


tests/unit/cmd/sca/test_diff.py
amascia-gg, line 1: do we want to create a functional test as well?

tests/functional/sca/test_sca_scan_diff.py
```diff
@@ -0,0 +1,50 @@
+from pathlib import Path
+
+from tests.functional.utils import run_ggshield_sca_scan
+from tests.repository import Repository
+
+
+def test_scan_ci_diff(tmp_path: Path, monkeypatch, pipfile_lock_with_vuln) -> None:
```
amascia-gg, line 7:
```suggestion
def test_scan_diff(tmp_path: Path, monkeypatch, pipfile_lock_with_vuln) -> None:
```

amascia-gg: and it should work without `monkeypatch`

GG-HH: That is what happens when you copy paste a bit too fast... Sorry for that. I fixed it.

amascia-gg (approved):

PR #667, opened 2023-08-04
feat(sca): add sca pre-receive command

amascia-gg:
Added sca pre-receive command and functional tests

### DIFF
--- a/ggshield/cmd/sca/scan/prereceive.py
+++ b/ggshield/cmd/sca/scan/prereceive.py
@@ -1,0 +1,98 @@
+import logging
+from pathlib import Path
+from typing import Any, Sequence
+
+import click
+
+from ggshield.cmd.common_options import all_option
+from ggshield.cmd.sca.scan.sca_scan_utils import (
+    create_output_handler,
+    display_sca_beta_warning,
+    sca_scan_diff,
+)
+from ggshield.cmd.sca.scan.scan_common_options import (
+    add_sca_scan_common_options,
+    update_context,
+)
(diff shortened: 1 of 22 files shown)


ggshield/cmd/sca/scan/prereceive.py
```diff
@@ -0,0 +1,98 @@
...
+def scan_pre_receive_cmd(
+    ctx: click.Context,
+    scan_all: bool,
+    exit_zero: bool,
+    minimum_severity: str,
+    ignore_paths: Sequence[str],
+    **kwargs: Any,
+) -> int:
+    """
+    scan as a pre-receive git hook.
```
GG-HH, line 47:
```suggestion
    Scan as a pre-receive git hook.
```

tests/unit/cmd/sca/test_scan.py
```diff
@@ -115,8 +115,7 @@ def test_sca_scan_diff(client: GGClient, dummy_sca_repo: Repository):
         result = sca_scan_diff(
```
GG-HH, line 115: Not related to this line: could you add a unit test (ie: with a cassette) for the pre-receive scan ? IMO it will be useful for debugging in case we have an issue with the functional one (it will show a change in the API)

gg-jonathangriffe: I agree. I think one with a vuln and one without would be great.

gg-jonathangriffe (changes requested):

GG-HH (approved):

gg-jonathangriffe (approved):

PR #668, opened 2023-08-07
Fix: only display IaC files in scan all verbose mode

Paul-GitGuardian:
https://linear.app/gitguardian/issue/IAC-138/ggshield-change-verbose-logs-to-list-only-iac-files


gg-teolohrer (approved):
LGTM 👍

Issue #359, opened 2022-09-16
--instance option missing for some commands


nathaner:
**Is your feature request related to a problem? Please describe.**

When using `ggshield auth login` and `ggshield auth logout`, we can specify the instance we target using the `--instance` option. However, we can't specify this option in other commands that target a specific instance:
- `api_status`
- `quota`
- `secret`

**Describe the solution you'd like**

Add the option `--instance` to the commands above so we can target a specific instance.

**Describe alternatives you've considered**

For now, a bypass is to use the `GITGUARDIAN_INSTANCE` environment variable.

PR #670, opened 2023-08-07
Allow defining the default config instance

agateau-gg:
## Description

This PR makes it possible to set the default instance of GitGuardian to use with `ggshield config set instance <THE_INSTANCE_URL>`.

What the command does is setting the `instance` field of the user global configuration file (~/.gitguardian.yaml, ~/.gitguardian.yml or ~/.gitguardian). If no such file exists, it creates the ~/.gitguardian.yaml file.

The next PR completes this by allowing the default instance to be read with `ggshield config get instance`, unset with `ggshield config unset instance` and listed in `ggshield config list`.

## What has been done

The first 3 commits make some cleanups.

Then the next two rework the way we find the user config file to make it possible to implement and test `ggshield config set instance`.

The final commit really implements the feature.

## Related issue

This should make bug #359 a bit less annoying.

### DIFF
--- a/changelog.d/20230807_180315_aurelien.gateau_config_set_instance.md
+++ b/changelog.d/20230807_180315_aurelien.gateau_config_set_instance.md
@@ -1,0 +1,3 @@
+### Added
+
+- It is now possible to define the default instance using `ggshield config set instance <THE_INSTANCE_URL>`.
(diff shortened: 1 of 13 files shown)


Walz (approved):
LGTM.

PR #672, opened 2023-08-08
Merge main-sca into main

agateau-gg:
The code in main-sca has been reviewed as it was getting merged, so no need for manual reviews here, just wait until CI is 🟢.

PR #673, opened 2023-08-08
Update cryptography to fix security issue

agateau-gg:
This fixes https://github.com/advisories/GHSA-jm77-qphf-c4w8, and hopefully #672.


GG-HH (approved):
Thanks!

PR #675, opened 2023-08-08
chore(pre-commit): run ggshield-local on all files

GG-HH:
Remove filter on Python files

For an unknown reason, this filter was added. According to [the documentation](https://pre-commit.com/#config-types), it acts as a filter (logical AND) and will not trigger a scan for non python files.

To test the behavior before the change:
- modify the `.prettierignore`
- commit
- the `GitGuardian Shield` job is skipped


agateau-gg (approved):
Good catch, that's odd indeed.

PR #685, opened 2023-08-11
Unhide hmsl command group

agateau-gg:
This PR makes the `hmsl` command group visible, so that it can be discovered in the upcoming release.


GG-HH (approved):
Thanks!

PR #694, opened 2023-08-17
feat(iac): #IAC-211 add exception wrapper in commands

carla-gitguardian:
### Add a specific error raise for invalid git reference.
We used to call UsageError raising an exception which was not of type _ExitError. However, we are using the function handle_exception from this same file ggshield/core/errors.py, which tests if the exception raised is from type _ExitError else return an UNEXPECTED_ERROR. So with the old way and after adding the exception wrapper, we always ended up having this exit code whereas we wanted a USAGE_ERROR one.

### DIFF
--- a/ggshield/core/errors.py
+++ b/ggshield/core/errors.py
@@ -60,6 +60,15 @@ class ParseError(_ExitError):
         super().__init__(ExitCode.UNEXPECTED_ERROR, message)
 
 
+class InvalidGitRefError(_ExitError):
+    """
+    Raised when the git reference does not exist
+    """
+
+    def __init__(self, ref: str):
+        super().__init__(ExitCode.USAGE_ERROR, f"Not a git reference: {ref}.")
+
+
 class AuthError(_ExitError):
     """
     Base exception for Auth-related configuration error
--- a/ggshield/cmd/iac/scan/all.py
+++ b/ggshield/cmd/iac/scan/all.py
@@ -15,6 +15,7 @@ from ggshield.cmd.iac.scan.iac_scan_utils import (
     handle_scan_error,
 )
 from ggshield.core.config import Config
+from ggshield.core.errors import handle_exception
 from ggshield.core.text_utils import display_info
 from ggshield.iac.collection.iac_path_scan_collection import IaCPathScanCollection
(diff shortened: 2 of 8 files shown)


ggshield/core/errors.py
```diff
@@ -60,6 +60,15 @@ def __init__(self, message: str):
         super().__init__(ExitCode.UNEXPECTED_ERROR, message)
 
 
+class InvalidGitRefError(_ExitError):
+    """
+    Raised when the git reference does not exist
+    """
+
+    def __init__(self, ref: str):
+        super().__init__(ExitCode.USAGE_ERROR, f"Not a git reference: {ref}.")
```
Paul-GitGuardian, line 69: Is it correct for an Error class to expect a string argument that is not the error message? This is already the case for some other classes, for instance `UnknownInstanceError`.

WDYT @agateau-gg?

agateau-gg: I think it's OK because at the point the exception is raised we don't have any other information we could provide.

I don't understand what we gain from adding this exception class though: do you plan to catch it specifically in calling code, or is it going to be raised from other places? If not then the code could stay as it was.

agateau-gg (changes requested):

ggshield/core/errors.py
carla-gitguardian: We used to call `UsageError` raising an exception which was not of type `_ExitError`. However, we are using the function `handle_exception` from this same file `ggshield/core/errors.py`, which tests if the exception raised is from type `_ExitError` else return an `UNEXPECTED_ERROR`. So with the old way and after adding the exception wrapper, we always ended up having this exit code whereas we wanted a `USAGE_ERROR` one.
Do you think there is a better way to handle this situation?

agateau-gg: Oh right, good catch! No, I can't think of a better way to handle this :+1: (your explanation would make a great commit message, though :wink:).

agateau-gg (approved):

ggshield/core/errors.py
carla-gitguardian: Noted, I'll think about it next time. Thanks for your review :)

PR #696, opened 2023-08-17
HMSL Auth uses GITGUARDIAN_API_KEY if GITGUARDIAN_SAAS_API_KEY is not set

phililippe:
Fixes EXPL-201
When GITGUARDIAN_SAAS_API_KEY is not set and GITGUARDIAN_API_KEY is set, the auth will use the later.


GG-HH (approved):
Thank you!

PR #699, opened 2023-08-18
fix(ci): fetch gitlab target branch to fix sca and iac CIs

xblanchot-gg:
Gitlab CIs requires to first fetch target branch in order to get the HEAD commit sha, used in SCA and IaC CI jobs.


GG-HH (approved):
Nice!

agateau-gg (changes requested):
Looks good! But I think this is worth a changelog entry.

PR #702, opened 2023-08-21
Isolate functional tests from user Git config

fbochu:
Some users could have hooks defined in their Git template directory. The content of this directory is copied every time a repository is cloned or created. We have to isolate our functional tests to avoid the user Git config interferes with our tests.

### DIFF
--- a/scripts/run-functional-tests
+++ b/scripts/run-functional-tests
@@ -42,10 +42,10 @@ install_packages() {
 
 run_tests() {
     log_progress "Running tests"
-    pytest --disable-pytest-warnings -vvv tests/functional "${@}"
+    pytest --disable-pytest-warnings -vvv tests/functional "$@"
 }
 
 build_wheel
 create_venv
 install_packages
-run_tests "${@}"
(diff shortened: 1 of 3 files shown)


scripts/run-functional-tests
```diff
@@ -42,10 +42,16 @@ install_packages() {
 
 run_tests() {
     log_progress "Running tests"
-    pytest --disable-pytest-warnings -vvv tests/functional
+    pytest --disable-pytest-warnings -vvv tests/functional "${@}"
 }
 
+isolate() {
+    export GIT_CONFIG_GLOBAL=''
+    export GIT_CONFIG_SYSTEM=''
```
agateau-gg, line 50: I just found out we already hit that issue once in the past, but it was only fixed for the `unittest` Makefile target. I think we should unify this and add an autouse fixture to unset the variables in `tests/conftest.py` instead.

agateau-gg (changes requested):

agateau-gg (approved):
Looks good now, thanks.

Issue #521, opened 2023-05-05
Define boundaries between Python modules


agateau-gg:
## Goal

To help keeping the code-base manageable, we need clearer boundaries between the various Python modules inside GGShield.

The proposed architecture is the following:

```
 Command layer   | cmd.iac | cmd.secret | cmd.auth | cmd.config
---------------- | ---↓----|------↓-----|----↓-----|
 Vertical layer  |   iac   |   secret   |  auth    |
---------------- | ============================================
 Support layers  |                 scan | config
                 | -----------------↓-------↓------------------
                 |                    core
```

A module from the command layer can only import code from its matching vertical layer (`cmd.iac` can import from `iac` but not from `secret` or `auth`). Modules from the command and vertical layers can import from any support layer modules.

## Current problems and steps to fix them

- [ ] `core.config.user_config` imports `iac` to verify IaC policy syntax
    - [ ] Move IaC policy syntax code to `core.config.user_config`
- [ ] `core.oauth` imports `config`
    - [ ] Move `core.oauth` to `auth` (new module)
- [ ] `core.check_updates` imports `config.utils`
    - [ ] Move `core.config.utils.*_yaml_dict()` functions from `core.config.utils` to a new `core.yaml_utils` module
- [ ] `core.client` imports `config`
    - [ ] Move `core.client` to `core.config`

## Next steps

- [ ] Move `core.config` to `config`
- [ ] Document the final architecture in doc/dev/architecture.md
- [ ] Setup [import-linter](https://pypi.org/project/import-linter/) to enforce boundaries

PR #703, opened 2023-08-21
refactor: Move ggshield.iac to ggshield.verticals.iac

fbochu:
This is one of the step needed to regroup all verticals into one package and simplify the identification of the verticals layer as defined in #521.

refs: #521


agateau-gg (changes requested):
Can you move tests.unit.iac to tests.unit.verticals.iac too?

agateau-gg (approved):

PR #706, opened 2023-08-21
fix: file content could not be found

cyrvan-b:
In case of a scan launched on a directory within a repository, the related file content was not found.

### DIFF
--- a/tests/functional/iac/test_iac_scan_diff.py
+++ b/tests/functional/iac/test_iac_scan_diff.py
@@ -55,6 +55,43 @@ def test_iac_scan_diff_unchanged(tmp_path: Path) -> None:
     assert "0 new incidents detected" in result.stdout
 
 
+def test_iac_scan_diff_unchanged_inner_dir(tmp_path: Path) -> None:
+    # GIVEN a git repository
+    repo = Repository.create(tmp_path)
+    repo.create_commit()
+
+    # AND inner directory
+    inner_dir_path = tmp_path / "innerdir"
+    inner_dir_path.mkdir()
+
+    # AND a first commit with vulnerabilities
+    file1 = inner_dir_path / "file1.tf"
+    file1.write_text(_IAC_SINGLE_VULNERABILITY)
+    repo.add(file1)
+    repo.create_commit()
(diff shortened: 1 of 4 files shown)


tests/functional/iac/test_iac_scan_diff.py
```diff
@@ -87,6 +120,43 @@ def test_iac_scan_diff_new_vuln(tmp_path: Path) -> None:
...
+    result = run_ggshield_iac_scan(*args, cwd=inner_dir_path, expected_code=1)
+
+    # AND vulnerability of file1.tf shows as unchanged
+    # AND vulnerability of file2.tf shows as new
+    assert "0 incidents deleted" in result.stdout
+    assert "1 incident remaining" in result.stdout
+    assert "1 new incident detected" in result.stdout
+    # AND details about the new vulnerability are shown
+    assert "file2.tf" in result.stdout
+    assert "The incident was found between" not in result.stdout
```
GG-HH, line 157: Maybe we should test the presence of parts of the `_IAC_SINGLE_VULNERABILITY` content ? That would be more robust, in particular if we change the wording of the fallback message.

GG-HH: And same below

GG-HH (changes requested):

ggshield/iac/output/iac_text_output_handler.py
```diff
@@ -71,16 +71,27 @@ def group_incidents_by_filename(
 class IaCTextOutputHandler(IaCOutputHandler):
     nb_lines: ClassVar[int] = 3
 
+    def _get_source_path(self, scan: IaCPathScanCollection) -> Path:
```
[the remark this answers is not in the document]
cyrvan-b: After some tests it looks like their behavior is the same. Maybe calling `git rev-parse --show-toplevel` is less error prone as this is provided directly with git. I will propose a solution with that.

GG-HH:
Sounds good to me but there is a conflict with the main branch.

GG-HH (approved):

PR #707, opened 2023-08-21
feat(hmsl): add initial structure for check-secret-manager group

corenting:
- Add initial structure for the new `hmsl check-secret-manager`
  - Add initial structure for the first command, `hashicorp-vault` (corresponding code in the verticals module will be in a later PR). For now the command and the group are hidden due to the implementation not being done. The aim is to have an early feedback on the structure and to split the work in multiple and smaller PRs.
- Add note in doc to use an old version of pipenv when developing due to missing `--skip-lock` option in recent versions

I added the `skip-changelog` label due to the implementation not being done, I will add the changelog once the command is functional.

### DIFF
--- a/ggshield/cmd/hmsl/__init__.py
+++ b/ggshield/cmd/hmsl/__init__.py
@@ -5,6 +5,7 @@ import click
 from ggshield.cmd.common_options import add_common_options
 from ggshield.cmd.hmsl.api_status import status_cmd
 from ggshield.cmd.hmsl.check import check_cmd
+from ggshield.cmd.hmsl.check_secret_manager import check_secret_manager_group
 from ggshield.cmd.hmsl.decrypt import decrypt_cmd
 from ggshield.cmd.hmsl.fingerprint import fingerprint_cmd
 from ggshield.cmd.hmsl.query import query_cmd
@@ -14,6 +15,7 @@ from ggshield.cmd.hmsl.quota import quota_cmd
 @click.group(
     commands={
         "check": check_cmd,
+        "check-secret-manager": check_secret_manager_group,
         "decrypt": decrypt_cmd,
         "fingerprint": fingerprint_cmd,
         "query": query_cmd,
--- a/ggshield/cmd/hmsl/check_secret_manager/__init__.py
+++ b/ggshield/cmd/hmsl/check_secret_manager/__init__.py
@@ -1,0 +1,20 @@
+from typing import Any
+
+import click
+
+from ggshield.cmd.common_options import add_common_options
+from ggshield.cmd.hmsl.check_secret_manager.hashicorp_vault import (
+    check_hashicorp_vault_cmd,
+)
+
+
+@click.group(
+    commands={
+        "hashicorp-vault": check_hashicorp_vault_cmd,
+    },
+    hidden=True,
+)
+@add_common_options()
+def check_secret_manager_group(**kwargs: Any) -> None:
+    """Check if secrets from a secret manager have leaked."""
+    pass
--- a/ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
+++ b/ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
@@ -22,14 +22,15 @@ def _get_vault_token(use_vault_cli_token: bool) -> str:
                 f"could not get the token from Vault CLI: {str(exc)}"
             ) from exc
 
-    if "VAULT_TOKEN" in os.environ:
-        return os.environ["VAULT_TOKEN"]
+    env_token = os.getenv("VAULT_TOKEN")
+    if env_token is None:
+        raise click.UsageError(
+            "you need to specify the Vault token to use, either through the VAULT_TOKEN"
+            " environment variable or by using --use-cli-token to use the token "
+            "from the Vault CLI."
+        )
 
-    raise click.UsageError(
-        "you need to specify the Vault token to use, either through the VAULT_TOKEN"
-        " environment variable or by using --use-cli-token to use the token "
-        "from the Vault CLI."
-    )
+    return env_token
 
 
 @click.command(hidden=True)
--- a/ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
+++ b/ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
@@ -1,0 +1,97 @@
+import os
+from typing import Any, Optional
+
+import click
+
+from ggshield.cmd.common_options import add_common_options
+from ggshield.verticals.hmsl.secret_manager.hashicorp_vault import (
+    VaultCliTokenFetchingError,
+    get_vault_cli_token,
+)
+
+
+def _get_vault_token(use_vault_cli_token: bool) -> str:
+    """
+    Get the Vault token to use.
+    """
+    if use_vault_cli_token:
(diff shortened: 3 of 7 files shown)


ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
```diff
@@ -0,0 +1,77 @@
...
+    if use_vault_cli_token:
+        try:
+            return get_vault_cli_token()
+        except VaultCliTokenFetchingError as exc:
+            raise click.UsageError(
+                f"could not get the token from vault CLI: {str(exc)}"
+            ) from exc
+
+    if "VAULT_TOKEN" in os.environ:
+        return os.environ["VAULT_TOKEN"]
```
GG-HH, line 26: I think we usually use `getenv` in ggshield.

ggshield/verticals/hmsl/secret_manager/hashicorp_vault.py
```diff
@@ -0,0 +1,36 @@
...
+            capture_output=True,
+            text=True,
+            shell=True,
+            check=True,
+        )
+        json_content = json.loads(result.stdout)
+
+        return json_content["data"]["id"]
+    except subprocess.CalledProcessError as exc:
+        msg = "error when calling vault CLI."
```
GG-HH, line 30: Nit: should Vault be written with a capital V ? (For all messages)

corenting: We could indeed put Vault everywhere, it seems to be the naming convention used on the Hashicorp documentation even when referring to the CLI binary (like [here](https://developer.hashicorp.com/vault/docs/commands)).

agateau-gg (changes requested):
Thanks for opening the PR early, and for fixing the getting-started doc :+1:.

Just some minor remarks, it looks good overall. I am a bit worried about functional tests though. We don't have any for hmsl at the moment, but it's better to think about them earlier. How easy would it be to setup a kv store in a test and run `hmsl check-secret-manager` on it?

ggshield/cmd/hmsl/__init__.py
```diff
@@ -19,6 +20,7 @@
         "query": query_cmd,
         "quota": quota_cmd,
         "api-status": status_cmd,
+        "check-secret-manager": check_secret_manager_group,
```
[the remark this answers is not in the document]
agateau-gg: Also, it's quite long to type. Maybe we can have a shorter alias, like `check-sm`? (I wonder if Click supports aliases for groups, though).

ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
```diff
@@ -0,0 +1,77 @@
...
+    "--use-vault-cli-token",
+    "use_vault_cli_token",
+    is_flag=True,
+    show_default=True,
+    default=False,
+    help="Instead of using the VAULT_TOKEN environment variable, use the token used by the vault CLI.",
+)
+@click.argument(
+    "vault_path",
+    type=str,
```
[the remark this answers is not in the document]
corenting: It's a path on the remote vault mount, not a local filesystem

ggshield/cmd/hmsl/__init__.py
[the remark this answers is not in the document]
corenting: > Also, it's quite long to type. Maybe we can have a shorter alias, like check-sm? (I wonder if Click supports aliases for groups, though).

I looked into adding an alias but it's not supported by click directly. We could do something by using a custom class or something like [click-aliases](https://github.com/click-contrib/click-aliases), but I don't know if it's worth it. Wdyt?

corenting:
> Just some minor remarks, it looks good overall. I am a bit worried about functional tests though. We don't have any for hmsl at the moment, but it's better to think about them earlier. How easy would it be to setup a kv store in a test and run `hmsl check-secret-manager` on it?

There is a docker version of the server for dev purposes which could be used for this purpose if needed.

ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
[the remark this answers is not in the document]
agateau-gg: Oups, fair enough 😅

ggshield/cmd/hmsl/check_secret_manager/__init__.py
```diff
@@ -14,6 +14,40 @@
...
+)
+@click.option(
+    "--recursive",
+    "-r",
+    "recursive",
+    is_flag=True,
+    show_default=True,
+    default=False,
+    help="If the secret manager path is a directory and not a file, explore recursively.",
+)
```
agateau-gg, line 41: It's not a good idea to define the options at the group level, because Click has the annoying limitation of not propagating group options down to individual commands. This means that with the current code, you **must** pass the URL like this:

```
ggshield hmsl check-secret-manager --url https://example.com hashicorp-vault
```

If you define the URL later, like that:


```
ggshield hmsl check-secret-manager hashicorp-vault --url https://example.com
```

then it won't work :(

It's better to define the options at the command level, and factorize them in a common_options module later if necessary (see secret_scan_common_options.py for example)

agateau-gg (changes requested):

ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
```diff
@@ -0,0 +1,97 @@
...
+
+@click.command(hidden=True)
+@click.option(
+    "--use-cli-token",
+    "use_cli_token",
+    is_flag=True,
+    show_default=True,
+    default=False,
+    help="Instead of getting the token from the an environment variable, "
+    "get the token from the CLI tool.",
```
agateau-gg, line 43: nit: typo and repetition.
```suggestion
    help="Instead of getting the token from the environment variable, "
    "get it from the CLI tool.",
```

agateau-gg (changes requested):

ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+
+    # Get the Vault URL
+    if url is None and "VAULT_URL" not in os.environ:
+        raise click.UsageError(
+            "you need to specify the URL of your Vault, "
+            "either through --url or in the VAULT_URL environment variable"
+        )
+    if url is None and "VAULT_URL" in os.environ:
+        url = os.environ["VAULT_URL"]
```
agateau-gg, line 93: This can be made easier to follow like this:
```suggestion
    if url is None:
        url = os.getenv("VAULT_URL"):
        if url is None:
            raise click.UsageError(
                "you need to specify the URL of your Vault, "
                "either through --url or in the VAULT_URL environment variable"
            )
```

agateau-gg (changes requested):
Final nitpick!

ggshield/cmd/hmsl/check_secret_manager/hashicorp_vault.py
corenting: Fixed, alongside the typing of the `url` which was wrong

agateau-gg (approved):
Nice!

PR #713, opened 2023-08-22
docs(readme): update installation instructions

corenting:
- Changed the documentation installation method with pip to use `--user` to prevent
 installing system-wide. It's not recommended to install system wide and may not work anyway on OS where the system Python is marked as externally managed (see [PEP 668](https://peps.python.org/pep-0668/). It's the case for Debian 12 for example.
- Documented the fact that the tool can also be installed with [pipx](https://pypa.github.io/pipx/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Only metadata such as call time, request size and scan mode is stored from scans
 - [Installation](#installation)
   - [macOS - Using Homebrew](#macos---using-homebrew)
   - [Linux packages](#linux-packages)
-  - [Other Operating Systems - Using pip](#other-operating-systems---using-pip)
+  - [Other Operating Systems - Using pipx or pip](#other-operating-systems---using-pipx-or-pip)
     - [Installing](#installing)
     - [Updating](#updating)
 - [Initial setup](#initial-setup)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,7 +82,7 @@ The package should run on MacOS, Linux and Windows.
 To update `ggshield` you can add the option `-U/--upgrade` to the pip install command.
 
 ```shell
-$ pip install -U ggshield
+$ pip install --user -U ggshield
```
agateau-gg, line 85: If we explain how to install with `pipx` then we must also explain how to upgrade with it here.

changelog.d/20230822_160326_corentin.garcia_update_readme.md
```diff
@@ -0,0 +1,5 @@
+### Changed
```
agateau-gg, line 1: I don't think this needs to be in the changelog: the target audience for release notes are people who already have the app installed, they don't really care about how to do it :)

agateau-gg (changes requested):

README.md
```diff
@@ -59,12 +59,20 @@ Setup instructions:
 - [Deb packages](https://cloudsmith.io/~gitguardian/repos/ggshield/setup/#formats-deb)
 - [RPM packages](https://cloudsmith.io/~gitguardian/repos/ggshield/setup/#formats-rpm)
 
-## Other Operating Systems - Using pip
+## Other Operating Systems - Using pipx or pip
 
 ### Installing
 
+The recommended way to install `ggshield`` is to use [pipx](https://pypa.github.io/pipx/), which will install it an isolated environment:
```
agateau-gg, line 66:
```suggestion
The recommended way to install `ggshield` is to use [pipx](https://pypa.github.io/pipx/), which will install it an isolated environment:
```

README.md
```diff
@@ -59,12 +59,20 @@ Setup instructions:
...
+
+```shell
+$ pipx install ggshield
+```
+
+Alternatively, you can install with pip as an user package. This will not work if your Python installation is declared as externally managed (for example if using the system Python on some operating systems):
```
agateau-gg, line 72:
```suggestion
Alternatively, you can install with pip as a user package. This will not work if your Python installation is declared as externally managed (for example when using the system Python on operating systems like Debian 12):
```

agateau-gg (changes requested):
minor nitpicks

agateau-gg (approved):
Looks good, thanks!

I let you squash the fixups before I merge.

corenting:
> Looks good, thanks!
> 
> I let you squash the fixups before I merge.

Thanks for the review! I squashed the PR & rebased on the latest main.

PR #714, opened 2023-08-22
Create the auth "vertical"

agateau-gg:
## Intro

This PR introduces an `auth` "vertical". It's not a real vertical but it provides the same support as other `ggshield.verticals` packages to the `ggshield.cmd` packages.

## What has been done

I initially considered moving part of cmd.login and cmd.logout there but it would require more refactoring to make the code less tied to the CLI parsing and output. It is not worth the move without these refactoring IMHO and I do not want to spend time on it for now.

Instead I just moved OAuthClient and its related code there, and moved an helper date formatting function to core.textutils.

While I was at it, I simplified said helper date and made it output a more correctly formatted string.

## Ref

Part of #521.


fbochu (approved):
LGTM

PR #716, opened 2023-08-23
Enforce boundaries between modules

fbochu:
# Description

Enforce the boundaries between modules as described in #521, except for `ggshield.core` sub-modules that will be treated later.

Linked to #521

**Note to reviewers**: Review commit by commit is recommended as the first commits refactor some code to enforce boundaries.


scripts/check-import-linter-config.sh
```diff
@@ -0,0 +1,15 @@
...
+#: %c[31;1m =========== Error detected in the import linter config ===========%c[0m
+#: Please update and commit the configuration
+#: %c[30m
+#: $ ./scripts/generate-import-linter-config.py ./.importlinter
+#: $ git add .importlinter
+#: %c[0m
+
+
+diff --color -u .importlinter <(./scripts/generate-import-linter-config.py) \
+  || cat $0 | awk '$1 == "#:" { printf(substr($0, 3) "\n", 27, 27) }'
```
agateau-gg, line 15: This is quite a complicated way of printing the error message :). What about this?

```shell
MSG='\e[31;1m =========== Error detected in the import linter config ===========\e[0m
Please update and commit the configuration
\e[30m
$ ./scripts/generate-import-linter-config.py ./.importlinter
$ git add .importlinter
\e[0m
'

if ! diff --color -u .importlinter <(./scripts/generate-import-linter-config.py) ; then
    printf "$MSG"
    exit 1
fi
```

agateau-gg: Also, I would add this at the beginning so that the script can be called from any directory:

```bash
cd "$(dirname $0)/.."
```

.importlinter
```diff
@@ -0,0 +1,67 @@
+[importlinter]
```
agateau-gg, line 1: Can you make the script output a comment on the first line like this:

```
# this file has been generated by scripts/generate-import-linter-config.py, do not edit by hand!
```

This should help prevent wasting time for others :)

scripts/generate-import-linter-config.py
```diff
@@ -0,0 +1,201 @@
...
+                "ggshield.cmd.auth.** -> ggshield.verticals.hmsl.**",
+            ],
+            "unmatched_ignore_imports_alerting": "none",
+        },
+    ],
+}
+
+
+def get_submodules(*, name: str, prefixed: bool) -> list[str]:
+    """Retrieve the modules inclued in a package"""
```
agateau-gg, line 72:
```suggestion
    """Retrieve the modules included in a package"""
```

scripts/generate-import-linter-config.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def expand_glob(line: str) -> list[str]:
+    """
+    Expand the *glob*
+
+    supported globs:
+    - "xxx.*" => ["xxx.sub_module_a", "xxx.sub_module_b"]
+    - "xxx|*" => ["xxx.sub_module_a | xxx.sub_module_b"]
+    - "xxx" (no globs) => ["xxx"]
```
agateau-gg, line 90: The glob syntax makes `STATIC_CONFIG` hard to read because some `*` are expanded and others are not. I'd prefer if the globs expanded by the script used a different character, for example `%` (like in SQL).

agateau-gg (changes requested):
Minor remarks, looks great overall!

scripts/generate-import-linter-config.py
fbochu: This is a very good idea!

scripts/check-import-linter-config.sh
fbochu: Yes, your solution is much easier to read

scripts/generate-import-linter-config.py
```diff
@@ -0,0 +1,201 @@
...
+        "xxx.sub_module_b -> yyy.sub_module_b"
+      ]
+    """
+    if "{}" not in line:
+        return [line]
+
+    return [
+        line.replace("{}", module)
+        for module in get_submodules(
+            name=line[: line.index(".{}")],
```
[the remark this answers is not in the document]
fbochu: I don't think that this will increase the readability as the code is already affecting the ad-hoc value to a parameter named `name`, extracting this to a new variable named `name` and then affecting the `name` parameter with the `name` variable will not ease the reading of the code but will add one line.

But if you think is better for readability, I will do it.

scripts/generate-import-linter-config.py
fbochu: Whoops, thanks for the catch

.importlinter
fbochu: Very good idea! I will do it.

scripts/generate-import-linter-config.py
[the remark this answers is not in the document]
agateau-gg: I think it would help reducing the complexity of the list comprehension. multi-line list comprehensions are a code smell to me, but that's a personal opinion. I am fine with it.

scripts/generate-import-linter-config.py
```diff
@@ -0,0 +1,187 @@
...
+            "name": "verticals-cmd-transversals",
+            "type": "forbidden",
+            "source_modules": [
+                "ggshield.cmd.%",
+            ],
+            "forbidden_modules": [
+                "ggshield.verticals.%",
+            ],
+            "ignore_imports": [
+                "ggshield.cmd.main -> ggshield.verticals.**",
```
agateau-gg, line 63: I think `main` should not be allowed to import from `ggshield.verticals`. At least it does not need to right now.

This also made me realize we should move `main` out of `cmd`, because right now the script generates rules to allow `cmd.main` to import from `verticals.main`, which does not (and should not) exist. But we can do so in a separate PR.

Similarly, the script generates ignore rules to allow `cmd.utils` to import from `verticals.utils` which does not and should not exist either.

agateau-gg (changes requested):
Minor remarks, looks good

scripts/generate-import-linter-config.py
fbochu: By default, import rules are transitive, If `main` imports `cmd.hmsl` and `cmd.hmsl` imports `verticals.hmsl` then `main` import `verticals.hmsl`. We can be changed but only for the _forbidden_ contract, the _layers_ contract always allows indirect imports.

The script generates the config according to a simple set of rules, we can complexify it but is this worth the price?

agateau-gg (approved):

PR #717, opened 2023-08-24
chore(hmsl,cli): Use better help text and improve commands order

GG-HH:
- Clearer help text
- Order commands so that fingerprint is before query and decrypt


GG-HH:
@agateau-gg I don't know if we would like to generalize ordering of the commands in ggshield (to show the "most important" first). In that case, I could change the file in which we define the command ordering. WDYT ?

ggshield/cmd/hmsl/__init__.py
```diff
@@ -12,13 +12,25 @@
 from ggshield.cmd.hmsl.quota import quota_cmd
 
 
+class NaturalOrderGroup(click.Group):
+    """
+    Helper class to force the order of the commands in the help text.
+    Copied from https://github.com/pallets/click/issues/513#issuecomment-504158316
+    This will not work with python < 3.6.
```
corenting, line 19: nit:
```suggestion
    """
    Helper class to force the order of the commands in the help text.
    Copied from https://github.com/pallets/click/issues/513#issuecomment-504158316
```
as we don't support Python 3.6

ggshield/cmd/hmsl/fingerprint.py
```diff
@@ -47,7 +47,7 @@ def fingerprint_cmd(
     **_: Any,
 ) -> int:
     """
-    Collect secrets and prepare them to be queried.
+    Collect secrets and compute fingerprints.
```
corenting, line 50:
```suggestion
    Collect secrets and compute fingerprints to be used later by the `decrypt` command.
```

ggshield/cmd/hmsl/__init__.py
GG-HH: I know we don't support <py3.8 but I wanted to emphasis the reason why it works and warn anyone willing to re-use this code on other projects

ggshield/cmd/hmsl/fingerprint.py
GG-HH: The problem is that click will trim the sentence to 80 chars and replace with three dots any leftovers. Doing so will not improve a lot the message when doing `ggshield hmsl -h` and may make it worse.

corenting: Maybe we could separate in two parts? The initial string to keep for `ggshield hmsl -h` and a paragraph below for more help that will be displayed when doing `ggshield hmsl fingerprint -h`.

click does support increasing the max_content_width but I don't think it's something we want to do.

GG-HH: Indeed, I did so by using two lines :)

ggshield/cmd/hmsl/fingerprint.py
```diff
@@ -47,7 +47,9 @@ def fingerprint_cmd(
...
+
+    fingerprints are to be used later by the `decrypt` command.
```
agateau-gg, line 52: nit:
```suggestion
    Fingerprints are to be used later by the `decrypt` command.
```

ggshield/cmd/hmsl/__init__.py
```diff
@@ -12,13 +12,25 @@
 from ggshield.cmd.hmsl.quota import quota_cmd
 
 
+class NaturalOrderGroup(click.Group):
```
agateau-gg, line 15: This class would be a good candidate for the new top-level `ggshield.utils.click` package. `ggshield.utils` is where we want to store code which can easily be reused in other projects (no business-logic, no dependency on py-gitguardian...)

agateau-gg:
> @agateau-gg I don't know if we would like to generalize ordering of the commands in ggshield (to show the "most important" first). In that case, I could change the file in which we define the command ordering. WDYT ?

This sounds like a great idea :+1:.

agateau-gg (changes requested):

ggshield/cmd/hmsl/__init__.py
GG-HH: Indeed! I moved it!

agateau-gg (approved):
Nice!

PR #720, opened 2023-08-24
Create decorator for exception handling for secrets/iac/sca commands

amascia-gg:
Remove try/except blocks in commands from Secrets / IaC / SCA in favour of a decorator

### DIFF
--- a/ggshield/cmd/utils/common_decorators.py
+++ b/ggshield/cmd/utils/common_decorators.py
@@ -1,0 +1,32 @@
+from functools import wraps
+
+from ggshield.core.config import Config
+from ggshield.core.errors import handle_exception
+from ggshield.core.text_utils import display_warning
+
+
+def exception_wrapper(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as error:
+            config: Config = args[0].obj["config"]
+            return handle_exception(error, config.user_config.verbose)
+
+    return wrapper
+
+
+def display_beta_warning(func):
+    """
+    Displays warning about new verticals' commands being in beta.
+    """
+
+    @wraps(func)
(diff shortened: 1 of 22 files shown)


GG-HH (commented):
Ok to me, I am a bit confused by the `config: Config = args[0].obj["config"]` which seems a bit ugly to me.

GG-HH:
@agateau-gg WDYT ? Would you prefer if we remain with try/except blocks everywhere to be more explicit ?

ggshield/cmd/utils/common_decorators.py
```diff
@@ -0,0 +1,34 @@
...
+            return handle_exception(error, config.user_config.verbose)
+
+    return wrapper
+
+
+def display_beta_warning(func):
+    """
+    Displays warning about new verticals' commands being in beta.
+    """
+
```
agateau-gg, line 27: Should probably use `@wraps` here too, unless I am missing something.

agateau-gg (changes requested):
I like the idea of reducing indentation level, and the fact that the decorate means a bit less boilerplate ✨.

Would be even better to wrap the call to `cli.main()` in `ggshield.__main__.main()` inside a try..except block instead of having to decorate all functions, but I am not sure this is possible.

"Request changes" because of a minor remark in the beta decorator.

amascia-gg:
> I like the idea of reducing indentation level, and the fact that the decorate means a bit less boilerplate ✨.
> 
> Would be even better to wrap the call to `cli.main()` in `ggshield.__main__.main()` inside a try..except block instead of having to decorate all functions, but I am not sure this is possible.
> 
> "Request changes" because of a minor remark in the beta decorator.

The context does not exist at this stage just yet so it was not possible unfortunately.

agateau-gg (approved):
Looks good, thanks!

PR #721, opened 2023-08-24
Move git_shell from core to utils

agateau-gg:
## Context

We want to move git_shell from core to utils because it is *almost* independent of ggshield code, and other modules we want to move to utils do depend on it.

## What has been done

- Moved all tar-related code to a `core.tar_utils` module
- Introduced a `GitError` base exception class and replaced all click-based exceptions with exceptions inheriting from `GitError`. Updated `handle_exception()` accordingly. Interestingly, this highlighted 2 issues:
    1. a bug in test_check_git_ref_valid_git_path(), which was testing in the wrong directory.
    
    2. a bug in git command timeout handling. A timeout raised a click.Abort() exception, which was turned into an ExitCode.SUCCESS error code by errors.py.

## Ref

Issue #521


fbogg (approved):
LGTM

fbochu (approved):

PR #722, opened 2023-08-24
refactor: Move ggshield.core.file_utils to ggshield.utils.files

fbochu:
The functions inside the module are not ggshield specific and should be in the ggshield.utils module according to the architecture described in #521.

refs #521

### DIFF
--- a/ggshield/utils/files.py
+++ b/ggshield/utils/files.py
@@ -3,10 +3,15 @@ import re
 from pathlib import Path, PurePosixPath
 from typing import List, Set, Union
 
-import click
-
 from ggshield.utils._binary_extensions import BINARY_EXTENSIONS
 from ggshield.utils.git_shell import git_ls, is_git_dir
+
+
+class UnexpectedDirectoryError(ValueError):
+    """Raise when a directory is used where it is not excepted"""
+
+    def __init__(self, path: str):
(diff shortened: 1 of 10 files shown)


agateau-gg (commented):
To improve re-usability, it would be good to enforce the following for error handling in the `utils` package:
- do not use on Click exceptions
- do not refer to ggshield CLI options

A way to do this for `get_filepaths()` would be to make the function raise a ValueError and then have the caller (there is only one) turn it into a UsageError.

What do you think?

ggshield/utils/files.py
```diff
@@ -3,12 +3,17 @@
 from pathlib import Path, PurePosixPath
 from typing import List, Set, Union
 
-import click
-
 from ggshield.utils._binary_extensions import BINARY_EXTENSIONS
 from ggshield.utils.git_shell import git_ls, is_git_dir
 
 
+class UnexceptedDirectoryError(ValueError):
```
agateau-gg, line 10: typo: Unex**cep**tedDirectoryError -> Unex**pec**tedDirectoryError

agateau-gg (changes requested):
just a typo, looks good otherwise

agateau-gg (approved):

PR #725, opened 2023-08-24
Split core.utils

agateau-gg:
## What has been done

This PR blows the ugly core.utils to smithereens. The explosion radius took some other nearby functions with it.

- URL-related code has moved to `core.url_utils`
- The `Line` class and its related functions from `core.utils` and `core.text_utils` have moved to `core.lines`
- .env-related code has moved to `core.env_utils`
- Some date-related code from `core.utils` and `verticals.auth` moved to `utils.datetime`
- `utils.batched()` moved to `utils.itertools`
- Remaining tests in `test_utils` were not related to `core.utils` and were moved to more relevant test files, and an `utils.os` module was created

Bonus:
- move `cd()` from `verticals.secret.repo` to `utils.os`. This function is more used by test code than by `verticals.secret.repo`.

Best reviewed commit-by-commit.

Part of #521


ggshield/core/url_utils.py
```diff
@@ -0,0 +1,97 @@
+from urllib.parse import ParseResult, urlparse
+
+from click import UsageError
```
fbochu, line 3: We should avoid using `click` in the core layer, no? But maybe is for another commit.

ggshield/cmd/secret/scan/repo.py
```diff
@@ -13,11 +14,17 @@
 from ggshield.core.cache import Cache
 from ggshield.core.config import Config
 from ggshield.core.scan import ScanContext, ScanMode
-from ggshield.core.utils import REGEX_GIT_URL
 from ggshield.utils.git_shell import git
 from ggshield.verticals.secret.repo import scan_repo_path
 
 
+# Source: https://github.com/jonschlinkert/is-git-url MIT LICENSE
```
fbochu, line 21: Even if we only use this regex here, it is very generic and could be used in general code, maybe we should put it in `utils.git_shell` / `utils.git`, no?

:label: 1

ggshield/core/scan/commit.py
```diff
@@ -12,6 +11,11 @@
 _RX_HEADER_LINE_SEPARATOR = re.compile("[\n\0]:", re.MULTILINE)
 
 
+REGEX_HEADER_INFO = re.compile(
```
fbochu, line 14: See :label: 1

ggshield/core/lines.py
```diff
@@ -0,0 +1,160 @@
+import re
+from enum import Enum, auto
+from typing import Iterable, List, NamedTuple, Optional
+
+from ggshield.core.text_utils import LINE_DISPLAY, STYLE, format_line_count, format_text
+from ggshield.utils.git_shell import Filemode
+
+
+REGEX_PATCH_HEADER = re.compile(
```
fbochu, line 9: See :label: 1

fbochu (approved):
Overall LGTM, I had just left a few non-mandatory suggestions

ggshield/core/url_utils.py
agateau-gg: yes and yes :)

ggshield/cmd/secret/scan/repo.py
agateau-gg: I'd rather not because I am not fond of this regex: it prevents commands like `ggshield secret scan repo https://github.com/GitGuardian/ggshield`. One must use `ggshield secret scan repo https://github.com/GitGuardian/ggshield.git`.

ggshield/core/scan/commit.py
agateau-gg: For this one and the next one, I would prefer exposing a function which would take a string to parse and return a Python object instead of a raw regex. That would be higher-level. Maybe in another PR.

PR #728, opened 2023-08-26
migrate to setup.cfg

deronnax:
declarative config for python packages is the way go. Death to handwritten, custom-logic-riddled setup.py 🔥


setup.cfg
```diff
@@ -1,5 +1,65 @@
+[metadata]
+name = ggshield
+version = 1.18.1
```
agateau-gg, line 3: This duplicates the version number from `ggshield/__init__.py`. Is it possible to read the version from the file instead? If not then you modify `scripts/release` so that the `update_version()` function updates the version in this file too.

setup.cfg
```diff
@@ -1,5 +1,65 @@
...
+
+[options.entry_points]
+console_scripts = ggshield=ggshield.cmd.main:main
+
 [flake8]
 inline-quotes = double
 max-line-length = 120
 ignore = E203, W503
-exclude = **/snapshots/*.py, .venv, build
+exclude = **/snapshots/*.py, .venv, build
```
agateau-gg, line 65: nit: missing end-of-line

agateau-gg (changes requested):

deronnax:
@agateau-gg do you want a changelog entry for this one?

agateau-gg (approved):
Nice, thanks!

PR #731, opened 2023-08-28
GHA: run ci only on 'main' branch pushes

deronnax:
two avoid CI running twice, once on branch push and another time on pull requests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
   pull_request:
   push:
     branches:
-      - '*'
+      - 'main'
     tags-ignore:
       - '*'
     paths-ignore:


.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ on:
   pull_request:
   push:
     branches:
-      - '*'
+      - 'main'
```
agateau-gg, line 13: It should also be run on hotfix branches, so I suggest adding something like `- '[0-9]+.[0-9]+.x'`. Make sure you test this on a sandbox repo though, my suggestion might be wrong.

agateau-gg (changes requested):

.github/workflows/ci.yml
deronnax: hmmm hotfix branches still go through a pull request, am I wrong? If that's the case, we don't need to add them in `branches`

agateau-gg: Changes made in hotfix branches do go through a pull request, but the release itself is a direct push to the hotfix branch, just like the release of a new .0 version is a direct push to `main`.

agateau-gg:
Closed as stale.

PR #732, opened 2023-08-28
Fix the build

agateau-gg:
Use the new entrypoint. This got lost during the conversion from setup.py to setup.cfg.


GG-HH (approved):
Thanks!

PR #737, opened 2023-09-01
fix(github,ci): Github CI scans only last commit in case of push new …

xblanchot-gg:
…branch

### DIFF
--- a/ggshield/core/git_hooks/ci/previous_commit.py
+++ b/ggshield/core/git_hooks/ci/previous_commit.py
@@ -4,7 +4,12 @@ from typing import Optional
 import click
 
 from ggshield.core.errors import UnexpectedError
-from ggshield.utils.git_shell import EMPTY_SHA, get_last_commit_sha_of_branch, git
+from ggshield.utils.git_shell import (
+    EMPTY_SHA,
+    get_last_commit_sha_of_branch,
+    get_list_commit_SHA,
+    git,
+)
 
 from .supported_ci import SupportedCI
 
@@ -28,6 +33,8 @@ def get_previous_commit_from_ci_env(
 def github_previous_commit_sha(verbose: bool) -> Optional[str]:
     push_before_sha = github_push_previous_commit_sha()
     pull_req_base_sha = github_pull_request_previous_commit_sha()
+    head_sha = os.getenv("GITHUB_SHA", "HEAD")
+    event_name = os.getenv("GITHUB_EVENT_NAME")
 
     if verbose:
         click.echo(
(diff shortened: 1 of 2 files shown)


ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -43,9 +50,18 @@ def github_previous_commit_sha(verbose: bool) -> Optional[str]:
...
         return push_before_sha
 
+    if head_sha and event_name == "push":
+        if verbose:
+            click.echo("Could not find current branch previous commit.")
+            click.echo("Current branch may have been just pushed.")
+            click.echo("Only scan last commit.")
+        last_commits = get_list_commit_SHA(f"{head_sha}~1", max_count=1)
+        if len(last_commits) == 1:
+            return last_commits[0]
```
amascia-gg, line 63: if this condition is not met we are raising the `UnexpectedError`, is it what we want? just to check

ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -43,9 +50,18 @@ def github_previous_commit_sha(verbose: bool) -> Optional[str]:
     if pull_req_base_sha:
         return pull_req_base_sha
 
-    if push_before_sha:
+    if push_before_sha and push_before_sha != EMPTY_SHA:
         return push_before_sha
 
+    if head_sha and event_name == "push":
+        if verbose:
+            click.echo("Could not find current branch previous commit.")
```
amascia-gg, line 58:
```suggestion
            click.echo("Could not find previous commit for current branch.")
```

ggshield/core/git_hooks/ci/previous_commit.py
xblanchot-gg: Yes I took the behavior from secret, but maybe we prefer to scan ony last commit in any case where we fail to get what we want and log it ?

xblanchot-gg: However other cases should not happen: HEAD_SHA should always exist, and in case of a PR, we will have `pull_req_base_sha` defined. So as other cases seem unexpected, maybe raising in those cases is the right behavior.

ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -43,9 +50,20 @@ def github_previous_commit_sha(verbose: bool) -> Optional[str]:
...
         return pull_req_base_sha
 
-    if push_before_sha:
+    if push_before_sha and push_before_sha != EMPTY_SHA:
         return push_before_sha
 
+    if head_sha and event_name == "push":
+        if verbose:
+            click.echo(
+                "Could not find current branch previous commit for current branch."
```
amascia-gg, line 59:
```suggestion
                "Could not find previous commit for current branch."
```

ggshield/core/git_hooks/ci/previous_commit.py
amascia-gg: ok, fine by me!

amascia-gg (approved):

PR #738, opened 2023-09-04
chore(scan_mode): Add scan mode for sca diff scans

amascia-gg:
Add scan modes for SCA scan commands.


ggshield/cmd/sca/scan/ci.py
```diff
@@ -67,11 +69,15 @@ def scan_ci_cmd(
...
 
+        ci_mode = SupportedCI.from_ci_env()
+        scan_mode = f"{ScanMode.CI.value}/{ci_mode.value}"
         result = sca_scan_diff(
             ctx=ctx,
             directory=directory,
             previous_ref=previous_commit,
             current_ref=current_commit,
+            scan_mode=scan_mode,
+            ci_mode=ci_mode.name,
```
xblanchot-gg, line 80: `ci_mode` is already contained in `scan_mode`, this value is not necessary I think

tests/unit/cmd/sca/test_scan.py
```diff
@@ -338,36 +342,6 @@ def test_sca_text_handler_ordering(patch_scan_all, cli_fs_runner):
...
-Identifier: 
-CVE IDs: -
-
->>> : Incident 2 (SCA): toto@1.0.0
-Severity: Low
-Summary: 
-No fix is currently available.
-Identifier: 
-CVE IDs: -"""  # noqa W291
+        """> toto/Pipfile.lock: 2 incidents detected\n\n>>> : Incident 1 (SCA): bar@2.5.6\nSeverity: Critical\nSummary: \nNo fix is currently available.\nIdentifier: \nCVE IDs: -\n\n>>> : Incident 2 (SCA): foo@1.0.5\nSeverity: High\nSummary: \nNo fix is currently available.\nIdentifier: \nCVE IDs: -\n\n> Pipfile.lock: 2 incidents detected\n\n>>> : Incident 1 (SCA): titi@2.0.0\nSeverity: Medium\nSummary: \nNo fix is currently available.\nIdentifier: \nCVE IDs: -\n\n>>> : Incident 2 (SCA): toto@1.0.0\nSeverity: Low\nSummary: \nNo fix is currently available.\nIdentifier: \nCVE IDs: -"""  # noqa W291
```
xblanchot-gg, line 345: Why modifying this ? It was more readable in a multi-line string no ?

ggshield/cmd/sca/scan/ci.py
amascia-gg: I know but it's what is done in secrets so I took it, I thought the variable `Ci-Mode` might be used in gim backend but I need to check that

tests/unit/cmd/sca/test_scan.py
```diff
@@ -164,6 +166,7 @@ def test_sca_scan_diff_ignore_path(
             ctx=ctx,
             directory=dummy_sca_repo.path,
             previous_ref="branch_with_vuln",
+            scan_mode="",
```
xblanchot-gg, line 169: We may want to set valid scan_modes here in the tests

ggshield/cmd/sca/scan/ci.py
xblanchot-gg: Ok then !

tests/unit/cmd/sca/test_scan.py
amascia-gg: It's a nightmare to keep the old versions with unnecessary spaces, tried to revert but it completely broke my PR, re-starting again

PR #739, opened 2023-09-04
chore: add sca scan modes

amascia-gg:


ggshield/core/scan/scan_mode.py
```diff
@@ -15,4 +15,5 @@ class ScanMode(Enum):
     IAC_DIRECTORY = "directory"
     IAC_DIFF = "diff"
     DOCSET = "docset"
-    SCA_DIRECTORY = "sca_directory"
+    SCA_DIRECTORY = "directory"
+    SCA_DIFF = "diff"
```
xblanchot-gg, line 19: As we use the same `PRE_COMMIT`, `CI` etc as secret, could we also have just a generic `DIRECTORY` and generic `DIFF` that are shared with IaC ?

xblanchot-gg (approved):

ggshield/cmd/iac/scan/all.py
```diff
@@ -5,22 +5,17 @@
...
-)
-from ggshield.cmd.iac.scan.iac_scan_utils import (
-    IaCSkipScanResult,
-    create_output_handler,
-    handle_scan_error,
-)
+    add_iac_scan_common_options, update_context)
+from ggshield.cmd.iac.scan.iac_scan_utils import (IaCSkipScanResult,
+                                                  create_output_handler,
+                                                  handle_scan_error)
```
xblanchot-gg, line 11: Same here, you should run black on this file I think.

ggshield/cmd/iac/scan/diff.py
```diff
@@ -5,36 +5,26 @@
...
+from ggshield.cmd.iac.scan.iac_scan_utils import (IaCSkipScanResult,
+                                                  create_output_handler,
+                                                  filter_iac_filepaths,
+                                                  get_git_filepaths,
+                                                  get_iac_tar,
+                                                  handle_scan_error)
+from ggshield.cmd.utils.common_decorators import (display_beta_warning,
+                                                  exception_wrapper)
+from ggshield.cmd.utils.common_options import (directory_argument,
+                                               reference_option, staged_option)
```
xblanchot-gg, line 18: Same here, strange formatting

tests/unit/cmd/sca/test_scan.py
```diff
@@ -342,32 +340,32 @@ def test_sca_text_handler_ordering(patch_scan_all, cli_fs_runner):
 
 >>> : Incident 1 (SCA): bar@2.5.6
 Severity: Critical
-Summary: 
+Summary:
```
xblanchot-gg, line 343: If you remove the space, the test will fail. It should be ok with the space with the `noqa` at the end of the string

ggshield/cmd/iac/scan/all.py
amascia-gg: This is isort doing wonders :'(

PR #747, opened 2023-09-21
Send the scanned repository's origin url as an extra header

Paul-GitGuardian:
Compute the repository remote url. Pick one at random if several are found. This will allow a better connection between ggshield and GIM's database.
This will appear in the request headers: `'GGShield-Repository-URL': 'https://github.com/bridgecrewio/terragoat.git'`

All verticals will send this header. For IAC/SCA, the repository found is based on the provided path in the command.
For Secrets:

- No repository are sent for subcommands `archive`, `docker`, `docker-archive`, `docset`, `pypi`.
- Current working directory is sent for subcommands `ci`, `pre-commit`, `pre-push`, `pre-receive`, `range`. This is the directory that is checked by `check_git_dir`.
- For `path` subcommand, if the `paths` argument has exactly one element, it is sent. Otherwise cwd is sent.
- For `repo` subcommand, the `repository` argument or the local clone path is sent, depending on the type of the provided argument.

### DIFF
--- a/ggshield/core/scan/scan_context.py
+++ b/ggshield/core/scan/scan_context.py
@@ -1,9 +1,11 @@
 import platform
 import uuid
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Dict, Optional, Union
 
 from ggshield import __version__
+from ggshield.utils.git_shell import get_repository_url_from_path
 from ggshield.utils.os import get_os_info
 
 from .scan_mode import ScanMode
@@ -14,6 +16,7 @@ class ScanContext:
     scan_mode: Union[ScanMode, str]
     command_path: str
     extra_headers: Optional[Dict[str, str]] = None
+    target_path: Optional[Path] = None
 
     def __post_init__(self) -> None:
         self.command_id = str(uuid.uuid4())
@@ -34,6 +37,10 @@ class ScanContext:
             "OS-Version": self.os_version,
             "Python-Version": self.python_version,
         }
+        if self.target_path is not None:
+            repo_url = get_repository_url_from_path(self.target_path)
+            if repo_url is not None:
+                headers["Repository-URL"] = repo_url
         if self.extra_headers:
             headers = {**headers, **self.extra_headers}
 
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import re
 import subprocess
 from enum import Enum
 from functools import lru_cache
@@ -93,6 +94,27 @@ def _git_rev_parse_absolute(option: str, wd_absolute: Path) -> Optional[str]:
 
 def _git_rev_parse(option: str, wd: Path) -> Optional[str]:
     return _git_rev_parse_absolute(option=option, wd_absolute=wd.resolve())
+
+
+def simplify_git_url(url: str) -> str:
+    """
+    Removes elements from the git remote url.
+    - scheme
+    - credentials
+    - port
+    - extension
+    https://user:pass@mygitlab.corp.com:84/path/to/repo.git -> mygitlab.corp.com/toto/titi/tata
+    """
+    for (pattern, replace) in (
+        (r"https?://", ""),  # Scheme
+        (r".+@", ""),  # Credentials
+        (r":\d*/", "/"),  # Port
+        (r"\.git$", ""),  # Github/Gitlab/BitBucket extension (**.git)
+        (r"/_git/", "/"),  # Azure Devops extension (**/_git/**)
+        (":", "/"),  # Normalize ssh url to https format
+    ):
+        url = re.sub(pattern, replace, url)
+    return url
 
 
 def is_git_dir(wd: Union[str, Path]) -> bool:
@@ -263,6 +285,28 @@ def get_last_commit_sha_of_branch(branch_name: str) -> Optional[str]:
         return None
 
     return last_target_commit[0]
+
+
+def get_repository_url_from_path(wd: Path) -> Optional[str]:
+    """
+    Returns one of the repository remote urls. Returns None if no remote are found,
+    or the directory is not a repository.
+    """
+    remotes_raw: List[str] = []
+    try:
+        if not is_git_dir(wd):
+            return None
+        remotes_raw = git(["remote", "-v"], cwd=wd).splitlines()
+    except (subprocess.CalledProcessError, OSError):
+        return None
+
+    url: Optional[str] = None
+    for line in remotes_raw:
+        if match := re.search(r"^(.*)\t(.*) \(fetch\)$", line):
+            name, url = match.groups()
+            if name == "origin":
+                break
+    return simplify_git_url(url) if url else None
 
 
 def get_filepaths_from_ref(
--- a/tests/unit/core/scan/test_scan_context.py
+++ b/tests/unit/core/scan/test_scan_context.py
@@ -1,0 +1,52 @@
+from pathlib import Path
+from typing import Union
+
+from ggshield.core.scan.scan_context import ScanContext
+from ggshield.core.scan.scan_mode import ScanMode
+from tests.repository import Repository
+
+
+class TestScanContextRepositoryURL:
+    def _assert_repo_url_in_headers(
+        self, context: ScanContext, expected_url: Union[Path, str]
+    ):
+        assert context.get_http_headers().get("GGShield-Repository-URL") == str(
+            expected_url
+        )
+
+    def _assert_no_repo_url_in_headers(self, context: ScanContext):
+        assert context.get_http_headers().get("GGShield-Repository-URL") is None
+
+    def test_scan_context_no_repo(
+        self,
+        tmp_path: Path,
+    ):
+        """
+        GIVEN a directory which is not a git repo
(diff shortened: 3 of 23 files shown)


Paul-GitGuardian:
@sylvain-baud-gg IMO sending the entire URL is better, as GIM can then extract the required information from it (repo name, type of repo...)

sylvain-baud-gg (commented):
I only see a change on the scan diff . I think the iac_scan_all should also fetch the remote url and send it.

Some edge case to keep in mind : 
- Local repo only --> We want to send None
- Case of repo containing sub repos. --> We want to send the repo linked to the path provided.
- Launching the command inside a repoA to scan a repoB --> We want to send repoB

Paul-GitGuardian:
I suggest you review commits one by one

sylvain-baud-gg (approved):

Paul-GitGuardian:
@agateau-gg thank you for your review on my other PR, this one became our priority for the next days. Can you check it please? As said above, it is easier commit by commit imo

ggshield/core/scan/scan_context.py
```diff
@@ -34,6 +37,10 @@ def get_http_headers(self) -> Dict[str, str]:
             "OS-Version": self.os_version,
             "Python-Version": self.python_version,
         }
+        if self.target_path is not None:
+            repo_url = get_repository_url_from_path(self.target_path)
+            if repo_url is not None:
+                headers["Repository-URL"] = repo_url
```
agateau-gg, line 43: According to the PR description, the header should be `GGShield-Repository-URL`.

ggshield/utils/git_shell.py
```diff
@@ -95,6 +96,25 @@ def _git_rev_parse(option: str, wd: Path) -> Optional[str]:
     return _git_rev_parse_absolute(option=option, wd_absolute=wd.resolve())
 
 
+def _parse_from_git_url(url: str) -> str:
```
agateau-gg, line 99: This function is not really a parser. I would call it `simplify_git_url()` or `clean_git_url()`.

ggshield/utils/git_shell.py
```diff
@@ -95,6 +96,25 @@ def _git_rev_parse(option: str, wd: Path) -> Optional[str]:
...
+    https://user:pass@mygitlab.corp.com:84/path/to/repo.git -> mygitlab.corp.com/toto/titi/tata
+    """
+    stripped_scheme = re.sub(r"https?://", "", url)
+    stripped_credentials = re.sub(r".+@", "", stripped_scheme)
+    stripped_port = re.sub(r":\d*/", "/", stripped_credentials)
+    # Github/Gitlab/BitBucket
+    stripped_extension = re.sub(r"\.git$", "", stripped_port)
+    # Azure Devops
+    stripped_extension = re.sub(r"/_git/", "/", stripped_extension)
+    return stripped_extension
```
agateau-gg, line 115: reusing the `url` variable would be simpler here.

ggshield/utils/git_shell.py
GG-HH, line 99: I think we should have explicit test for this function with many cases.

ggshield/utils/git_shell.py
```diff
@@ -265,6 +285,28 @@ def get_last_commit_sha_of_branch(branch_name: str) -> Optional[str]:
...
+def get_repository_url_from_path(wd: Path) -> Optional[str]:
+    """
+    Returns one of the repository remote urls. Returns None if no remote are found,
+    or the directory is not a repository.
+    """
+    try:
+        if not is_git_dir(wd):
+            return None
+        remotes_raw = git(["remote", "-v"], cwd=wd).splitlines()
+        matches = (re.search(r"^.*\t(.*) \(.*\)$", remote) for remote in remotes_raw)
```
GG-HH, line 297: nit: performance may not be ideal in this regex. However, we shouldn't have too many/long items

ggshield/utils/git_shell.py
```diff
@@ -265,6 +285,28 @@ def get_last_commit_sha_of_branch(branch_name: str) -> Optional[str]:
...
+        return next(
+            (
+                _parse_from_git_url(match.group(1))
+                for match in matches
+                if match is not None and len(match.groups()) >= 1
+            ),
+            None,
+        )
+    except Exception:
+        return None
```
agateau-gg, line 307: A few remarks here:

I would prefer if the whole function were not wrapped in a `try..except`. Just wrap what can go wrong, and use more precise exception classes. This avoids hiding bugs.

The code is a bit convoluted, it could be simplified to this:

```python
    for line in git(["remote", "-v"]).splitlines():
        if match := re.search(r"^.*\t(.*) \(.*\)$", line):
            return _parse_from_git_url(match.group(1))
```

But neither the loop in the PR or that proposal prefer the `origin` remote, which is surprising. This version uses `origin` if available:

```python
    url: Optional[str] = None
    for line in git(["remote", "-v"]).splitlines():
        if match := re.search(r"^(.*)\t(.*) \(fetch\)$", line):
            name, url = match.groups()
            if name == "origin":
                break
    return _parse_from_git_url(url) if url else None
```

tests/unit/core/scan/test_scan_context.py
```diff
@@ -0,0 +1,100 @@
...
+                "https://user:password@github.com:84/GitGuardian/ggshield.git",
+                "github.com/GitGuardian/ggshield",
+            ),
+            (
+                "https://github.com/GitGuardian/ggshield.git",
+                "github.com/GitGuardian/ggshield",
+            ),
+            (
+                "git@github.com:GitGuardian/ggshield.git",
+                "github.com:GitGuardian/ggshield",
```
GG-HH, line 69: How difficult would it be to normalize it in order to return github.com**/**GitGuardian/ggshield ?

agateau-gg (changes requested):

tests/unit/core/scan/test_scan_context.py
```diff
@@ -0,0 +1,100 @@
...
+            ),
+            (
+                "https://gitlab.gitguardian.ovh/owner/project/repository.git",
+                "gitlab.gitguardian.ovh/owner/project/repository",
```
GG-HH, line 73: Nit: change the URL

ggshield/core/scan/scan_context.py
Paul-GitGuardian: "Repository-URL" is added to `headers`, which are all transformed by adding the `GGShield-` prefix, in the return below, so it ends up as `GGShield-Repository-URL` as expected

tests/unit/core/scan/test_scan_context.py
Paul-GitGuardian: I'm already assuming that repository names cannot contain `:` (this is correct for at least the 4 supported providers), so I'll simply replace occurrences with a `/`

ggshield/utils/git_shell.py
Paul-GitGuardian: Thank you for this level of detail.

As this change impacts not only iac, but also sca and secrets, we do not want to break existing features of other verticals in case something goes wrong, hence the big try/except. wdyt @GG-HH, can we remove it here?

Also I agree `origin` should have the priority. @GG-HH ?

ggshield/utils/git_shell.py
Paul-GitGuardian: I was using the variables names as explanations for the regexes.
Is this version better?

ggshield/utils/git_shell.py
GG-HH: I agree too for the priority of "origin".

For the try/except, I think we should keep it for the git command. For the matching, we should be confident enough about our code and be able to remove it.

ggshield/utils/git_shell.py
Paul-GitGuardian: I've moved the test urls from `test_scan_context` to the new test for this function, because I don't think testing all the urls twice is necessary. Do you agree?

ggshield/utils/git_shell.py
Paul-GitGuardian: I agree, but in most cases there will only be two remote lines here (one for fetch, one for push), and many characters seem to be allowed in remote names

ggshield/core/scan/scan_context.py
agateau-gg: Ah right, I forgot about this 👍🏻.

ggshield/utils/git_shell.py
agateau-gg: Looks good, thanks.

ggshield/utils/git_shell.py
```diff
@@ -265,6 +287,31 @@ def get_last_commit_sha_of_branch(branch_name: str) -> Optional[str]:
...
 
 
+def get_repository_url_from_path(wd: Path) -> Optional[str]:
+    """
+    Returns one of the repository remote urls. Returns None if no remote are found,
+    or the directory is not a repository.
+    """
+    remotes_raw: List[str] = []
+    try:
+        print("before remotes_raw")
```
GG-HH, line 297: Do not forget to remove those

GG-HH (approved):

agateau-gg (approved):

PR #748, opened 2023-09-21
chore(hmsl): Adapt message in case we find tons of matches

iXioN:
Add a new message when a secret has more than 100 locations
Screenshot 2023-09-21 at 16 32 56

### DIFF
--- a/ggshield/verticals/hmsl/output.py
+++ b/ggshield/verticals/hmsl/output.py
@@ -23,6 +23,8 @@ First occurrence:
     URL: "{url}"
 """
 
+TOO_MANY_SECRETS_THRESHOLD = 100
+
 
 def write_outputs(result: PreparedSecrets, prefix: str) -> None:
     """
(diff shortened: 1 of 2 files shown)


ggshield/verticals/hmsl/output.py
```diff
@@ -23,6 +23,18 @@
     URL: "{url}"
 """
 
+TOO_MANY_SECRET_THRESHOLD = 100
```
agateau-gg, line 26: nit: should be `TOO_MANY_SECRETS_THRESHOLD`. Same for `TOO_MANY_SECRET_TEMPLATE`.

ggshield/verticals/hmsl/output.py
```diff
@@ -23,6 +23,18 @@
...
+
+TOO_MANY_SECRET_TEMPLATE = """
+> Secret {number}
+Secret name: "{name}"
+Secret hash: "{hash}"
+Given the number of occurrences, your secret might be a template value
```
agateau-gg, line 32: nit: 
```suggestion
Given the number of occurrences, your secret might be a template value.
```

But I am actually wondering if the added line is not going to be lost in the output.

What do you think of always using TEMPLATE but with a `display_warning("Given the number of occurrences, your secret might be a template value.")` after it so that it stands out a bit more?

agateau-gg (changes requested):
Minor remarks.

Do you think we should mention this change in the changelog? If so, then you need to add a changelog entry (see https://github.com/GitGuardian/ggshield/blob/main/doc/dev/getting-started.md#changelog for details).

agateau-gg (approved):
Looks good, thanks!

PR #750, opened 2023-09-22
Add a way to override the default git timeout

agateau-gg:
## Description

This PR makes it possible to override the timeout applied to all git commands with the `GG_GIT_TIMEOUT` environment variable.

## What has been done

I got a bit annoyed by the multiple places where we read environment variables and turn them into int, float or boolean, sometimes in different ways (at least for boolean variables) so I went a bit overboard and added new functions to `utils.os` to read int, float or boolean values from the environment. This is the first commit.

the second commit uses the new function to read the timeout, and log it in debug mode.

### DIFF
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -7,11 +7,13 @@ from pathlib import Path
 from shutil import which
 from typing import Dict, List, Optional, Union
 
+from ggshield.utils.os import getenv_int
+
 
 EMPTY_SHA = "0000000000000000000000000000000000000000"
 EMPTY_TREE = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
 
-COMMAND_TIMEOUT = 45
+COMMAND_TIMEOUT = getenv_int("GG_GIT_TIMEOUT", 45)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


ggshield/utils/git_shell.py
```diff
@@ -7,11 +7,13 @@
...
 from typing import Dict, List, Optional, Union
 
+from ggshield.utils.os import getenv_int
+
 
 EMPTY_SHA = "0000000000000000000000000000000000000000"
 EMPTY_TREE = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
 
-COMMAND_TIMEOUT = 45
+COMMAND_TIMEOUT = getenv_int("GG_GIT_TIMEOUT", 45)
```
pierrelalanne, line 16: Should we keep a `DEFAULT_COMMAND_TIMEOUT = 45` to highlight the fact that this is a constant ?

pierrelalanne (approved):
Tiny remark, otherwise LGTM 👍

ggshield/utils/git_shell.py
agateau-gg: I don't think it's worth it, unless we use `DEFAULT_COMMAND_TIMEOUT` somewhere else.

PR #751, opened 2023-09-22
fix(config): ignore config regexes when needed

amascia-gg:
When we explicitly ask to scan a path that is in a configuration file, then we should ignore the config file and scan the path after displaying a warning.

### DIFF
--- a/ggshield/cmd/utils/common_options.py
+++ b/ggshield/cmd/utils/common_options.py
@@ -17,6 +17,8 @@ import click
 
 from ggshield.cmd.utils.debug_logs import setup_debug_logs
 from ggshield.core.config.user_config import UserConfig
+from ggshield.core.text_utils import display_warning
+from ggshield.utils.files import is_filepath_excluded
 
 
 AnyFunction = Callable[..., Any]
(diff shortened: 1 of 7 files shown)


ggshield/cmd/utils/common_options.py
```diff
@@ -36,6 +38,15 @@ def get_config_from_context(ctx: click.Context) -> UserConfig:
     return cast(UserConfig, ctx.obj["config"].user_config)
 
 
+def check_directory_in_ignored_path(ctx: click.Context, directory: Path) -> None:
+    """Check if the directory is in the ignored paths in the config."""
+    if is_filepath_excluded(next(directory.rglob("*")), ctx.obj["exclusion_regexes"]):
+        display_warning(
+            "Directory to scan is in the ignored paths in the config, ignoring config and proceeding with scan."  # noqa: E501
+        )
+        ctx.obj["exclusion_regexes"] = set()
```
xblanchot-gg, line 47: I'm not sure completely ignoring the config is the expected behaviour, we should still be able to ignore some files even if we ask to scan a directory named `tests`. In my opinion we should just check the `exclusion_regexes` relatively to the scanned directory. I think this should be done in `ggshield/utils/files.py:get_filepaths`. 

As this is a common function to all verticals, we should check that this does not break anything.

PR #758, opened 2023-09-28
chore: add headers to HMSL commands

corenting:
- Add User-Agent header to HMSLClient
- Add a `GGShield-HMSL-Command-Name`, similar to `GGShield-Command-Path` in the secret module but which takes an enum value of possible HMSL command

### DIFF
--- a/ggshield/verticals/hmsl/client.py
+++ b/ggshield/verticals/hmsl/client.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, Iterable, List, Optional, Union, cast
 
 import requests
 
+from ggshield import __version__
 from ggshield.utils.itertools import batched
 
 from .crypto import decrypt, make_hint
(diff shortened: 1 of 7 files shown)


ggshield/verticals/hmsl/client.py
```diff
@@ -152,11 +167,12 @@ def _query(self, endpoint: str, body: Dict[str, List[str]]) -> Any:
             headers = {"Authorization": f"Bearer {self.jwt}"}
         else:
             headers = {}
-        response = requests.post(
+        response = self.session.post(
             self.url + endpoint,
             json=body,
             headers=headers,
         )
+        print(f"Headers were: {response.request.headers}")
```
pierrelalanne, line 175: I don't think we want to keep this.

pierrelalanne (changes requested):

ggshield/verticals/hmsl/client.py
corenting: Ah yes I forgot to remove it, fixed

pierrelalanne (approved):

agateau-gg:
Would it be possible to use `ctx.command_path` instead of introducing an HMSL-specific enum?

tests/unit/verticals/hmsl/test_hmsl_client.py
```diff
@@ -34,10 +36,18 @@ def hashes():
     return [hash_string(letter) for letter in "abcdefghijklmnopqrstuvwxyz"]
 
 
-@pytest.fixture
-def hmsl_client():
+def get_hmsl_client(
+    hmsl_command_path: str | None = None,
```
agateau-gg, line 40: type must be `Optional[str]` because of python < 3.10 :/

agateau-gg (changes requested):
Looks good 👍🏻. Just a minor Python compatibility issue.

tests/unit/verticals/hmsl/test_hmsl_client.py
corenting: Fixed in [d4056c5](https://github.com/GitGuardian/ggshield/pull/758/commits/d4056c53e8f277e55f329440531be6a4a676fe1d)

corenting: There was also another case where I used `removeprefix` which was added in 3.9, I replaced it in [e3367d2](https://github.com/GitGuardian/ggshield/pull/758/commits/e3367d21f3a7ba172d964070719943e678185e1f)

agateau-gg (approved):
Looks good. I let you squash before merging.

corenting:
> Looks good. I let you squash before merging.

Thanks again for your review! I squashed & rebased the PR.

PR #760, opened 2023-10-04
feat(iac): add more scan modes based on context and --all flag

Paul-GitGuardian:
Adds more details in the scan mode passed as header (for IAC):
- what triggered the scan (manually, pre-commit, pre-push, pre-receive, ci)
- how it was performed (all, diff)


GG-HH:
Sounds good to me, except one thing: in other verticals, when scanning in CI, we have access to the type of CI (see supported_ci.py). I don't think that would be the case here.

Could we add some tests as well to ensure that the flag can be found in the headers of the query in the end ?

GG-HH (changes requested):
GH forces me to put a comment here. sorry.

Paul-GitGuardian:
> Sounds good to me, except one thing: in other verticals, when scanning in CI, we have access to the type of CI (see supported_ci.py). I don't think that would be the case here.
> 
> Could we add some tests as well to ensure that the flag can be found in the headers of the query in the end ?

Done for the CI mode.
I'm not sure how we can access request headers in tests, especially when the request is made in pygitguardian?

GG-HH (approved):

PR #761, opened 2023-10-05
feat(hmsl,vault): use the full path of the secret as hint in results

GG-HH:
We now return the full path of the secret when the naming strategy of the hint is the key


corenting (approved):
Look good to

PR #763, opened 2023-10-06
feat(cmd): modify exitcode when directory to scan is in ignore-paths

gg-jonathangriffe:


ggshield/cmd/iac/scan/all.py
```diff
@@ -57,6 +58,8 @@ def iac_scan_all(
 ) -> Union[IaCScanResult, IaCSkipScanResult, None]:
     config: Config = ctx.obj["config"]
 
+    assert_directory_not_ignored(directory, ctx.obj["exclusion_regexes"])
```
sylvain-baud-gg, line 61: As `ctx.obj["exclusion_regexes"]` is now used twice, we can define a variable : 
    `exclusion_regexes = ctx.obj["exclusion_regexes"]`

ggshield/cmd/utils/utils.py
```diff
@@ -0,0 +1,14 @@
...
+from click import UsageError
+
+from ggshield.utils.files import is_filepath_excluded
+
+
+def assert_directory_not_ignored(
+    directory: Union[str, Path], exclusion_regexes: Set[re.Pattern]
+):
+    if is_filepath_excluded(directory, exclusion_regexes):
+        raise UsageError("An ignored directory cannot be scanned.")
```
sylvain-baud-gg, line 14: Is it always a directory ?
Cannot it be a file sometime ?

ggshield/cmd/utils/utils.py
```diff
@@ -0,0 +1,14 @@
...
+
+from click import UsageError
+
+from ggshield.utils.files import is_filepath_excluded
+
+
+def assert_directory_not_ignored(
+    directory: Union[str, Path], exclusion_regexes: Set[re.Pattern]
+):
+    if is_filepath_excluded(directory, exclusion_regexes):
```
sylvain-baud-gg, line 13: If we scan a folder whose subfolders are all ignored, do we want to display a UsageError ? If yes, does this work with this code ?

Example with this architecture :
files_to_scan
   /tests
   /src

Our `paths_ignore` contains `tests/` and `src/,` and we scan `files_to_scan/`

sylvain-baud-gg (commented):
Thanks for the work on all verticals.

I think there is one edge case to check (See comment)

ggshield/cmd/utils/utils.py
gg-jonathangriffe: It can be a file in the secrets path scan. I will change it to "file or directory".

ggshield/cmd/utils/utils.py
gg-jonathangriffe: I think that according to the issue we do not want to display a UsageError.

ggshield/core/errors.py
```diff
@@ -146,7 +146,7 @@ def handle_exception(exc: Exception, verbose: bool) -> int:
     # Get exit code
     if isinstance(exc, _ExitError):
         exit_code = exc.exit_code
-    elif isinstance(exc, InvalidGitRefError):
+    elif isinstance(exc, InvalidGitRefError) or isinstance(exc, click.UsageError):
```
tchaumeny, line 149: `elif isinstance(exc, (InvalidGitRefError, click.UsageError)):`

tests/unit/cmd/sca/test_scan.py
```diff
@@ -376,3 +376,63 @@ def test_sca_text_handler_ordering(patch_scan_all, cli_fs_runner):
...
+            "scan",
+            "all",
+            "--ignore_path",
+            str(dummy_sca_repo.path),
+            str(dummy_sca_repo.path),
+        ],
+    )
+
+    assert_invoke_exited_with(result, ExitCode.USAGE_ERROR)
+    assert "An ignored file or directory cannot be scanned."
```
tchaumeny, line 406: 👀

tests/unit/cmd/sca/test_scan.py
```diff
@@ -376,3 +376,63 @@ def test_sca_text_handler_ordering(patch_scan_all, cli_fs_runner):
...
+            "--ref",
+            "branch_without_vuln",
+            "--ignore_path",
+            str(dummy_sca_repo.path),
+            str(dummy_sca_repo.path),
+        ],
+    )
+
+    assert_invoke_exited_with(result, ExitCode.USAGE_ERROR)
+    assert "AAn ignored file or directory cannot be scanned."
```
tchaumeny, line 437: 👀

tchaumeny: (typo + missing `in ...`)

tchaumeny (approved):

sylvain-baud-gg (approved):

ggshield/cmd/iac/scan/diff.py
```diff
@@ -22,11 +22,11 @@
     reference_option,
     staged_option,
 )
+from ggshield.cmd.utils.files import check_directory_not_ignored, is_filepath_excluded
```
agateau-gg, line 25: nitpick: importing `is_filepath_excluded()` like this works because `files.py` imports it, but it's an implementation detail. Would be better to keep the previous import.

agateau-gg (changes requested):
Just a minor remark about imports. Looks good overall 👍🏻.

PR #764, opened 2023-10-06
Add support for creating a token with extra scopes

agateau-gg:
This PR makes it possible to create a GitGuardian token with extra scopes in addition to the `scan` scope.

### DIFF
--- a/ggshield/verticals/auth/oauth.py
+++ b/ggshield/verticals/auth/oauth.py
@@ -423,32 +423,32 @@ class RequestHandlerWrapper:
                 """
                 callback_url: str = self_.path
                 parsed_url = urlparse.urlparse(callback_url)
-                if parsed_url.path == "/":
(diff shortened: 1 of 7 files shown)


ggshield/verticals/auth/oauth.py
```diff
@@ -423,32 +423,32 @@ def do_GET(
...
+                    # attach error message to the handler wrapper instance
+                    self.error_message = error.message
+                else:
+                    self_._end_request(
+                        301,
+                        urljoin(self.oauth_client.dashboard_url, "authenticated"),
+                    )
+
+                # indicate to the server to stop
+                self.complete = True
```
Walz: ~~This should be done above to keep the same behavior.~~  
It's fixed in a following commit

Walz (approved):
Tried to login with the scope `honeytokens:write` and it worked! I also tried with an invalid scope and it worked as expected.

LGTM.

PR #766, opened 2023-10-10
Xblanchot/sca 1066/clean sca models

xblanchot-gg:


xblanchot-gg:
@agateau-gg I modified the doc for make-pipfile-lock in the first commit, can you just check that I did not propose any mistake ?

agateau-gg (approved):
Looks good, thanks for taking care of the doc ✨.

ggshield/cmd/sca/scan/sca_scan_utils.py
```diff
@@ -14,27 +21,22 @@
...
-    SCAScanParameters,
-)
 
 
 def get_scan_params_from_config(sca_config: SCAConfig) -> SCAScanParameters:
     return SCAScanParameters(
         minimum_severity=sca_config.minimum_severity,
         ignored_vulnerabilities=[
-            ignored_vuln
+            SCAIgnoredVulnerability(
```
gg-jonathangriffe, line 37: I noticed that the code here transforms SCAIgnoredVulnerability (from ggshield) instances into SCAIgnoredVulnerability (from pygitguardian) instances. What is the purpose of having two separate classes ? If there is one shouldn't we avoid having the same name for both ?

gg-jonathangriffe: And why didn't we do this transformation before ?

xblanchot-gg: The one in py-gitguardian is more consise, and used as an intermediate type for `ScanParameters` that are sent to the backend. The one in GGShield is used to parse userconfig, and we did not wanted to add userConfig related information to a model in py-gitguardian. 
The one in GGShield could inherit from the first one if we want to. We perform the transformation here as it is exactly the purpose of the one in py-gitguardian, to be used in `SCAScanParameters`.

setup.cfg
```diff
@@ -43,7 +43,7 @@ install_requires =
     marshmallow>=3.18.0,<3.19.0
     marshmallow-dataclass>=8.5.8,<8.6.0
     oauthlib>=3.2.1,<3.3.0
-    pygitguardian>=1.10.0,<1.11.0
+    pygitguardian @ git+https://github.com/GitGuardian/py-gitguardian.git@cfa919cff68cc4d3ca40bf2bb8a6f24bc5fca786
```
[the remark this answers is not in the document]
xblanchot-gg: Indeed, I update the doc and i don't follow it !

ggshield/cmd/sca/scan/sca_scan_utils.py
xblanchot-gg: I can rename the one in the config `SCAConfigIgnoredVulnerability` if you find this confusing.

gg-jonathangriffe: It makes sense like this. I have no strong opinion on whether to rename it or not.

gg-jonathangriffe (approved):

PR #769, opened 2023-10-12
Draft: fix(circleci): fix circle_ci_range method

amascia-gg:
`CIRCLECI_RANGE` env var was always returning `...COMMIT` because the base_revision parameter of the CircleCI cannot be trusted. Therefore the scan was only done on the last commit of the push that could include 2+ commits.
This pull request proposes a fix to this.

This PR should be considered with the corresponding [PR](https://github.com/GitGuardian/gg-shield-orb/pull/9) in the ggshield orb.


amascia-gg:
closing since Issue comes from CircleCI and we cannot do anything about it for now

PR #771, opened 2023-10-12
chore(hmsl): handle missing urls

GG-HH:
This MR improve the model for Secret as well as the output to handle the cases where there is no url.

Please note that I choose to always keep a field "url" in the json output but it is nullable.


ggshield/verticals/hmsl/client.py
```diff
@@ -36,10 +36,14 @@ def decrypt(self, hash: str) -> Secret:
         key = bytes.fromhex(hash)
         payload = base64.b64decode(self.payload)
         decrypted = json.loads(decrypt(payload, key))
+        if decrypted.get("l"):
```
GG-HH, line 39: nit if "l" exists and is not a dict, the following code will fail

PR #774, opened 2023-10-13
Draft: feat(circle): add support for CircleCI for SCA/IaC ci scan

amascia-gg:
Add support for CircleCI SCA/IaC scan.
This will get the last commit before the first commit of a push event for example.


amascia-gg:
closing since Issue comes from CircleCI and we cannot do anything about it for now

PR #776, opened 2023-10-16
feat(jenkins): add support for Jenkins for SCA/IaC ci scan

gg-jonathangriffe:


ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -150,7 +150,43 @@ def gitlab_merge_request_previous_commit_sha(verbose: bool) -> Optional[str]:
...
+        click.echo(
+            f"GIT_PREVIOUS_COMMIT: {previous_commit}\n"
+            f"CHANGE_TARGET: {target_branch}\n",
+            err=True,
+        )
+
+    if target_branch:
+        return get_last_commit_sha_of_branch(f"origin/{target_branch}")
+
+    if previous_commit:
```
Blaxav, line 168: Can you add some logs also here ? In case of a job failure, we want to be able to understand in which case we did end up.

xblanchot-gg (approved):

PR #781, opened 2023-10-17
chore(wording): # Reword credits to quotas for clarity

pierrelalanne:
Avoid mixing credits and quota.  
Credits are one prefix call, on the landing page for instance.  
A quota is one quota as counted by the API.

### DIFF
--- a/ggshield/cmd/hmsl/quota.py
+++ b/ggshield/cmd/hmsl/quota.py
@@ -20,7 +20,7 @@ def quota_cmd(
     **kwargs: Any,
 ) -> int:
     """
-    Get the number of remaining credits for today.
+    Get the remaining quotas for today.
     """
 
     # Get our client
--- a/ggshield/cmd/quota.py
+++ b/ggshield/cmd/quota.py
@@ -17,7 +17,7 @@ from ggshield.core.errors import UnexpectedError
 @click.pass_context
 def quota_cmd(ctx: click.Context, **kwargs: Any) -> int:
     """
-    Show the remaining quota of API calls available for the entire workspace.
+    Show the remaining quotas of API calls available for the entire workspace.
     """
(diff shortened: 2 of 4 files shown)


ggshield/cmd/hmsl/quota.py
```diff
@@ -20,7 +20,7 @@ def quota_cmd(
     **kwargs: Any,
 ) -> int:
     """
-    Get the number of remaining credits for today.
+    Get the remaining quotas for today.
```
agateau-gg, line 23: I don't know if "credits" was the correct word, but I believe "quotas" is not the correct one here. A quota is a limit on something. The only reason it would be plural if is you have several different limits (a soft-limit and a hard-limit for example).

Maybe "API calls" is what we want here, just like in `ggshield quota`?

ggshield/cmd/quota.py
```diff
@@ -17,7 +17,7 @@
 @click.pass_context
 def quota_cmd(ctx: click.Context, **kwargs: Any) -> int:
     """
-    Show the remaining quota of API calls available for the entire workspace.
+    Show the remaining quotas of API calls available for the entire workspace.
```
agateau-gg, line 20:
```suggestion
    Show the remaining number of API calls available for the entire workspace.
```

agateau-gg (changes requested):

ggshield/cmd/hmsl/quota.py
GG-HH: It may be more complex but what do you guys think of "consumption units" ?

agateau-gg: Sounds a bit complex indeed. I thought it would make sense to sync this with the dashboard, but I can't see any mention to this there. Maybe I haven't been looking in the right place?

PR #784, opened 2023-10-18
Add comment and until date to iac config file

Paul-GitGuardian:
For both `ignored_policies` and `ignored_paths` in IAC config file, the expected data was changed.
Format before (still handled):
```
iac:
    ignored-policies:
        - "GG_IAC_0001"
```
New format handled:
```
iac:
    ignored-policies:
        - policy: "GG_IAC_0001"
              comment: "This policy isn't vulnerable for this repo."
              until: "2024-01-01T00:00:01Z"
```

Mixed formats are supported too.

Ignored policies and paths are only applied if no until date is provided, or if the date is in the future

### DIFF
--- a/ggshield/cmd/iac/scan/all.py
+++ b/ggshield/cmd/iac/scan/all.py
@@ -85,7 +85,7 @@ def iac_scan_all(
     client = ctx.obj["client"]
 
     scan_parameters = IaCScanParameters(
-        list(config.user_config.iac.ignored_policies),
+        list({ignored.policy for ignored in config.user_config.iac.ignored_policies}),
         config.user_config.iac.minimum_severity,
     )
     # If paths are not sorted, the tar bytes order will be different when calling the function twice
--- a/ggshield/cmd/iac/scan/diff.py
+++ b/ggshield/cmd/iac/scan/diff.py
@@ -174,7 +174,7 @@ def iac_scan_diff(
     current_tar = get_iac_tar(directory, current_ref, exclusion_regexes)
 
     scan_parameters = IaCScanParameters(
-        list(config.user_config.iac.ignored_policies),
+        list({ignored.policy for ignored in config.user_config.iac.ignored_policies}),
         config.user_config.iac.minimum_severity,
     )
 
--- a/ggshield/core/config/user_config.py
+++ b/ggshield/core/config/user_config.py
@@ -3,10 +3,10 @@ import re
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
+from typing import Any, Dict, List, Optional, Set, Tuple
 
 import marshmallow_dataclass
-from marshmallow import ValidationError, post_load
+from marshmallow import ValidationError, post_load, pre_load
 from pygitguardian.models import FromDictMixin
 
 from ggshield.core.config.utils import (
@@ -63,14 +63,86 @@ def validate_policy_id(policy_id: str) -> bool:
     return bool(POLICY_ID_PATTERN.fullmatch(policy_id))
 
 
-def validate_policy_ids(values: Iterable[str]) -> None:
-    invalid_excluded_policies = [
-        policy_id for policy_id in values if not validate_policy_id(policy_id)
-    ]
-    if len(invalid_excluded_policies) > 0:
-        raise ValidationError(
-            f"The policies {invalid_excluded_policies} do not match the pattern '{POLICY_ID_PATTERN.pattern}'"
+@marshmallow_dataclass.dataclass
+class ConfigIgnoredElement(FilteredConfig):
+    """
+    Base class for a config element which can be ignored temporarily.
+    Classes inheriting it must reject expired dates themselves.
+    """
+
+    comment: Optional[str]
+    until: Optional[datetime]
+
+    # Accept date yyyy-mm-dd instead of a full datetime
+    @pre_load
+    def parse_date(self, data, **kwargs):
+        if not (isinstance(data, dict)) or data.get("until") is None:
+            return data
+        try:
+            data["until"] = str(datetime.strptime(str(data["until"]), "%Y-%m-%d"))
+            return data
+        except ValueError:
+            return data
+
+    @post_load
+    def datetime_to_utc(self, data, **kwargs):
+        if data["until"] is not None:
+            data["until"] = data["until"].astimezone(timezone.utc)
+        return data
+
+
+def remove_expired_elements(
+    lst: List[ConfigIgnoredElement],
+) -> List[ConfigIgnoredElement]:
+    expired: List[ConfigIgnoredElement] = []
+    now = datetime.now(tz=timezone.utc)
+    for idx in range(len(lst) - 1, -1, -1):
+        ignored = lst[idx]
+        if ignored.until and ignored.until <= now:
+            lst.pop(idx)
+            expired.insert(0, ignored)
+
+    return expired
+
+
+def report_expired_elements(expired_lst: List[ConfigIgnoredElement]) -> None:
+    for element in expired_lst:
+        display_warning(
+            f"{element} has an expired 'until' "
+            f"date ({element.until}), please update your configuration file."
         )
+
+
+@marshmallow_dataclass.dataclass
+class IaCConfigIgnoredPath(ConfigIgnoredElement):
+    path: str
+
+    def __init__(self, path, comment=None, until=None, *args, **kwargs):
+        super().__init__(comment, until, *args, **kwargs)
+        self.path = path
+
+    def __str__(self):
+        return f"Path {self.path}"
+
+    @pre_load
+    def convert_paths(self, in_data, **kwargs):
+        return {"path": in_data} if isinstance(in_data, str) else in_data
+
+
+@marshmallow_dataclass.dataclass
+class IaCConfigIgnoredPolicy(ConfigIgnoredElement):
+    policy: str = field(metadata={"validate": validate_policy_id})
+
+    def __init__(self, policy, comment=None, until=None, *args, **kwargs):
+        super().__init__(comment, until, *args, **kwargs)
+        self.policy = policy
+
+    def __str__(self):
+        return f"Policy {self.policy}"
+
+    @pre_load
+    def convert_policies(self, in_data, **kwargs):
+        return {"policy": in_data} if isinstance(in_data, str) else in_data
 
 
 @marshmallow_dataclass.dataclass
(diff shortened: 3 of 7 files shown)


ggshield/core/config/user_config.py
```diff
@@ -80,12 +121,38 @@ class IaCConfig(FilteredConfig):
...
+        for ignored in data["ignored_policies"]:
+            if ignored.until is None or ignored.until > datetime.now(tz=timezone.utc):
+                valid_policies.append(ignored)
+            else:
+                display_warning(
+                    f"Policy {ignored.policy} has an expired 'until' "
+                    f"date ({ignored.until}), please update your configuration file."
+                )
+        data["ignored_policies"] = valid_policies
+        return data
```
Paul-GitGuardian, line 154: Do you think this can be factorized? The distinct warning messages bother me

ggshield/core/config/user_config.py
```diff
@@ -63,14 +63,55 @@ def validate_policy_id(policy_id: str) -> bool:
...
+            data["until"] = data["until"].astimezone(timezone.utc)
+        return data
+
+
+@marshmallow_dataclass.dataclass
+class IaCConfigIgnoredPath(ConfigIgnoredElement):
+    path: str
+
+    # If we set default values in the parent class,
+    # we would need to set them for the child class atributes too
```
sylvain-baud-gg, line 94:
```suggestion
    # we would need to set them for the child class attributes too
```

ggshield/core/config/user_config.py
```diff
@@ -63,14 +63,55 @@ def validate_policy_id(policy_id: str) -> bool:
...
+    @pre_load
+    def convert_paths(self, in_data, **kwargs):
+        return {"path": in_data} if isinstance(in_data, str) else in_data
+
+
+@marshmallow_dataclass.dataclass
+class IaCConfigIgnoredPolicy(ConfigIgnoredElement):
+    policy: str = field(metadata={"validate": validate_policy_id})
+
+    def __init__(self, policy, comment=None, until=None, *args, **kwargs):
```
sylvain-baud-gg, line 108: Could use the same comment here for default value

ggshield/core/config/user_config.py
```diff
@@ -80,12 +121,38 @@ class IaCConfig(FilteredConfig):
...
-    )
+    ignored_paths: List[IaCConfigIgnoredPath] = field(default_factory=list)
+    ignored_policies: List[IaCConfigIgnoredPolicy] = field(default_factory=list)
     minimum_severity: str = "LOW"
 
+    @post_load
+    def validate_ignored_paths(self, data, **kwargs):
+        valid_paths = []
+        for ignored in data["ignored_paths"]:
+            if ignored.until is None or ignored.until > datetime.now(tz=timezone.utc):
```
sylvain-baud-gg, line 132:
```suggestion
            if ignored.path in [i.path for i in valid.paths:
                display_warning("Path {ignored.path} is present twice in the config file}
                continue
            if ignored.until is None or ignored.until > datetime.now(tz=timezone.utc):
            
```

Something like this

ggshield/core/config/user_config.py
```diff
@@ -80,12 +121,38 @@ class IaCConfig(FilteredConfig):
...
+                    f"Path {ignored.path} has an expired 'until' "
+                    f"date ({ignored.until}), please update your configuration file."
+                )
+        data["ignored_paths"] = valid_paths
+        return data
+
+    @post_load
+    def validate_ignored_policies(self, data, **kwargs):
+        valid_policies = []
+        for ignored in data["ignored_policies"]:
```
sylvain-baud-gg, line 145: Same here, check duplicate value

ggshield/core/config/user_config.py
```diff
@@ -63,14 +63,55 @@ def validate_policy_id(policy_id: str) -> bool:
...
-def validate_policy_ids(values: Iterable[str]) -> None:
-    invalid_excluded_policies = [
-        policy_id for policy_id in values if not validate_policy_id(policy_id)
-    ]
-    if len(invalid_excluded_policies) > 0:
-        raise ValidationError(
-            f"The policies {invalid_excluded_policies} do not match the pattern '{POLICY_ID_PATTERN.pattern}'"
-        )
+@marshmallow_dataclass.dataclass
+class ConfigIgnoredElement(FilteredConfig):
```
sylvain-baud-gg, line 67: IMO, in this component, in case data is a dict, we should check the keys of the dict and display a warning if the dict has an unknown key

tests/unit/core/config/test_user_config.py
```diff
@@ -162,25 +173,159 @@ def test_load_ignored_matches_with_empty_names(self):
...
+            [
+                {
+                    "policy": "GG_IAC_0001",
+                    "comment": "mycomment",
+                    "until": "2050-01-01 00:00:00",
+                }
+            ],
+            ["GG_IAC_0001", {"policy": "GG_IAC_0002", "comment": "mycomment"}],
+        ),
+    )
```
sylvain-baud-gg, line 205: We have a test with `comment` but without `until`, we should also have a test with `until` but without `comment`

+ a test with duplicate entry would be great

ggshield/cmd/iac/scan/all.py
```diff
@@ -85,7 +85,7 @@ def iac_scan_all(
     client = ctx.obj["client"]
 
     scan_parameters = IaCScanParameters(
-        list(config.user_config.iac.ignored_policies),
+        list({ignored.policy for ignored in config.user_config.iac.ignored_policies}),
```
sylvain-baud-gg, line 88:
```suggestion
        [ignored.policy for ignored in config.user_config.iac.ignored_policies],
```
Unless it is possible to have duplicate value in  user_config.iac.ignored_policies.

ggshield/cmd/iac/scan/diff.py
```diff
@@ -174,7 +174,7 @@ def iac_scan_diff(
     current_tar = get_iac_tar(directory, current_ref, exclusion_regexes)
 
     scan_parameters = IaCScanParameters(
-        list(config.user_config.iac.ignored_policies),
+        list({ignored.policy for ignored in config.user_config.iac.ignored_policies}),
```
sylvain-baud-gg, line 177:
```suggestion
        [ignored.policy for ignored in config.user_config.iac.ignored_policies],
```
Unless it is possible to have duplicate value in  user_config.iac.ignored_policies.

sylvain-baud-gg (changes requested):
Good job, nice testing, some modification to be done.

ggshield/core/config/user_config.py
Paul-GitGuardian: We'll skip duplication checks for now.

ggshield/core/config/user_config.py
Paul-GitGuardian: Skipped for now, as other classes do not do this

ggshield/cmd/iac/scan/all.py
Paul-GitGuardian: Skipped as we still accept duplicate values

sylvain-baud-gg (approved):
LGTM.

ggshield/core/config/user_config.py
```diff
@@ -63,14 +64,81 @@ def validate_policy_id(policy_id: str) -> bool:
...
+    data[field] = valid_elements
+    return data
+
+
+@marshmallow_dataclass.dataclass
+class IaCConfigIgnoredPath(ConfigIgnoredElement):
+    path: str
+
+    # If we set default values in the parent class,
+    # we would need to set them for the child class attributes too
```
agateau-gg, line 113: I don't understand this comment. Which default values are we talking about here?

ggshield/core/config/user_config.py
```diff
@@ -63,14 +64,81 @@ def validate_policy_id(policy_id: str) -> bool:
...
-    if len(invalid_excluded_policies) > 0:
-        raise ValidationError(
-            f"The policies {invalid_excluded_policies} do not match the pattern '{POLICY_ID_PATTERN.pattern}'"
-        )
+@marshmallow_dataclass.dataclass
+class ConfigIgnoredElement(FilteredConfig):
+    comment: Optional[str]
+    until: Optional[datetime]
+
+    @abstractmethod
```
agateau-gg, line 72: `@abstractmethod` on a class which does not inherit from ABC does nothing. But inheriting from ABC here is probably not a good idea since this already inherits from a concrete class.

I would either remove the decorator, or remove the method entirely and re-implement `__str__()` in child classes instead (unless it does not work for some reason).

ggshield/core/config/user_config.py
```diff
@@ -63,14 +64,81 @@ def validate_policy_id(policy_id: str) -> bool:
...
+            return data
+
+    @post_load
+    def datetime_to_utc(self, data, **kwargs):
+        if data["until"] is not None:
+            data["until"] = data["until"].astimezone(timezone.utc)
+        return data
+
+
+def validate_ignored_elements_array(data: dict, field: str):
```
agateau-gg, line 94: This function does too much: it alter `data` in place, show warnings and return the modified `data`. I'd rather split it in two, a function to take expired elements from the list and return them, and another to report the expired elements. Something like this (untested):

```python
def remove_expired_elements(lst: List[ConfigIgnoredElement]) -> List[ConfigIgnoredElement]:
    expired: List[ConfigIgnoredElement] = []
    now = datetime.now(tz=timezone.utc)
    for idx in range(len(lst) - 1, -1, -1):
        ignored = lst[idx]
        if ignored.until and ignored.until <= now:
            lst.pop(idx)
            expired.insert(0, ignored)
    
    return expired
    

def report_expired_elements(expired_lst: List[ConfigIgnoredElement]) -> None:
    for element in expired_lst:
        display_warning(
            f"{element.short_repr()} has an expired 'until' "
            f"date ({element.until}), please update your configuration file."
        )
```

Then it can be used like that:

```
    def validate_ignored_vulns(self, data, **kwargs):
        expired_lst = remove_expired_elements(data["ignored_vulnerabilities"])
        report_expired_elements(expired_list)
        return data["ignored_vulnerabilities"]
```

tests/unit/core/config/test_user_config.py
```diff
@@ -15,6 +22,12 @@
 
 @pytest.mark.usefixtures("isolated_fs")
 class TestUserConfig:
```
agateau-gg, line 24: Have you tested the case where the `until` value is not a valid string? Do we get a user-friendly error message?

agateau-gg (changes requested):

tests/unit/core/config/test_user_config.py
Paul-GitGuardian: We get the same message from marshmallow as secret/sca:
```
Error: iac: 
  ignored_policies: 
    0: 
      until: ['Not a valid datetime.']
```

However when setting a number as the until date, the error message is messy, so I've added an explicit cast to str for the value to get the same message as above.

ggshield/core/config/user_config.py
```diff
@@ -63,14 +64,81 @@ def validate_policy_id(policy_id: str) -> bool:
...
+        try:
+            data["until"] = str(datetime.strptime(data["until"], "%Y-%m-%d"))
+            return data
+        except ValueError:
+            return data
+
+    @post_load
+    def datetime_to_utc(self, data, **kwargs):
+        if data["until"] is not None:
+            data["until"] = data["until"].astimezone(timezone.utc)
```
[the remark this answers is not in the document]
Paul-GitGuardian: It does not seem to work, the datetime is considered as 'offset naive' if `astimezone` is not called in post_load.

ggshield/core/config/user_config.py
Paul-GitGuardian: I tried to set default values for attributes in the parent class:
```
class ConfigIgnoredElement(FilteredConfig):
    comment: Optional[str] = None
    until: Optional[datetime] = None
```
It resulted in an error for child classes with required attributes. Seemingly the parent attributes are constructed first:
`TypeError: non-default argument 'path' follows default argument`

agateau-gg: I see. Indeed, this is to be expected. Not sure we need this comment.

tests/unit/core/config/test_user_config.py
agateau-gg: Looks good enough :+1:.

agateau-gg (approved):

PR #789, opened 2023-10-30
chore(sca): show error in case of HTTP 501

tchaumeny:
See https://linear.app/gitguardian/issue/SCA-1147/deactivate-correctly-in-on-prem

Note that the other API calls already display the associate error message.

### DIFF
--- a/ggshield/verticals/sca/file_selection.py
+++ b/ggshield/verticals/sca/file_selection.py
@@ -99,6 +99,8 @@ def sca_files_from_git_repo(
     if isinstance(sca_files_result, Detail):
         if sca_files_result.status_code == 401:
             raise APIKeyCheckError(client.base_uri, "Invalid API key.")
+        elif sca_files_result.status_code == 501:
+            raise UnexpectedError(sca_files_result.detail)
         raise UnexpectedError("Failed to select SCA files")
 
     sca_files = sca_files_result.sca_files


ggshield/verticals/sca/file_selection.py
```diff
@@ -99,6 +99,8 @@ def sca_files_from_git_repo(
     if isinstance(sca_files_result, Detail):
         if sca_files_result.status_code == 401:
             raise APIKeyCheckError(client.base_uri, "Invalid API key.")
+        elif sca_files_result.status_code == 501:
+            raise UnexpectedError(sca_files_result.detail)
```
xblanchot-gg, line 103: Should we define a specific error in this case ?

tchaumeny: I don't know, actually in other parts of the code (e.g. here https://github.com/GitGuardian/ggshield/blob/main/ggshield/cmd/sca/scan/sca_scan_utils.py#L240) we always display the content of any unexpected error. Maybe we should also do that here? (I didn't want to touch to existing code but maybe it's OK?)

xblanchot-gg: My point was just to not use `UnexpectedError`, from a semantic point of view. Maybe a `FeatureUnavailable` or something like that.

PR #791, opened 2023-10-31
feat(core): try to extract repository url from ci env variables

Paul-GitGuardian:
Following #747, we want CI commands to gather the current repository URL from the CI environment variables first, falling back on `git remote -v` if variables are not found.

### DIFF
--- a/tests/unit/core/scan/test_scan_context.py
+++ b/tests/unit/core/scan/test_scan_context.py
@@ -1,52 +1,123 @@
+import os
 from pathlib import Path
 from typing import Union
+from unittest import mock
+
+import pytest
 
 from ggshield.core.scan.scan_context import ScanContext
 from ggshield.core.scan.scan_mode import ScanMode
 from tests.repository import Repository
 
 
-class TestScanContextRepositoryURL:
-    def _assert_repo_url_in_headers(
-        self, context: ScanContext, expected_url: Union[Path, str]
-    ):
-        assert context.get_http_headers().get("GGShield-Repository-URL") == str(
-            expected_url
+DOMAIN = "https://github.com"
+SLUG = "from-env/repository"
+ENV_REPO_URL = f"{DOMAIN}/{SLUG}"
+EXPECTED_HEADER_ENV = "github.com/from-env/repository"
+
+REMOTE_URL = "https://github.com/from-remote/repository"
+FULL_REMOTE_URL = "https://user:password@github.com:84/from-remote/repository.git"
+EXPECTED_HEADER_REMOTE = "github.com/from-remote/repository"
+
+
+@pytest.fixture(scope="module")
+def fake_url_repo(tmp_path_factory: pytest.TempPathFactory) -> Repository:
+    repo = Repository.create(tmp_path_factory.mktemp("fake_url_repo"))
+    repo.git("remote", "add", "origin", REMOTE_URL)
+    return repo
+
+
+def _assert_repo_url_in_headers(context: ScanContext, expected_url: Union[Path, str]):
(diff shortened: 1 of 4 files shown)


tests/unit/core/scan/test_scan_context.py
```diff
@@ -1,12 +1,21 @@
...
 
 
 class TestScanContextRepositoryURL:
+    DOMAIN = "https://github.com"
+    SLUG = "expected/repository"
+    REPO_URL = f"{DOMAIN}/{SLUG}"
+    FULL_REPO_URL = "https://user:password@github.com:84/expected/repository.git"
+    FAKE_URL = "https://github.com/other/repository"
+    EXPECTED_HEADER = "github.com/expected/repository"
+
```
agateau-gg, line 18: nit: Those could be moved outside the class to avoid the need for the `self.` prefix.

Actually the class itself is not necessary: pytest is happy to use standalone functions. Not having a class saves indentation levels.

tests/unit/core/scan/test_scan_context.py
```diff
@@ -40,13 +49,71 @@ def test_scan_context_repository_url_parsed(self, tmp_path: Path):
...
 
         context = ScanContext(
             scan_mode=ScanMode.PATH,
             command_path="ggshield secret scan path",
             target_path=local_repo.path,
         )
-        self._assert_repo_url_in_headers(context, expected_url)
+        self._assert_repo_url_in_headers(context, self.EXPECTED_HEADER)
+
+    def test_ci_repo_found(self, tmp_path: Path) -> None:
```
agateau-gg, line 61: This test and `test_ci_no_env()` reimplement pytest parametrized tests. I assume this is to avoid recreating the repo for each iteration, but this can be avoided using module-level fixtures.

Can you rewrite them using parametrized tests? The code could look something like this (untested):


```python
@pytest.fixture(scope="module")
def fake_url_repo(tmp_path_factory: pytest.TempPathFactory) -> Repository:
    repo = Repository.create(tmp_path_factory.mktemp("fake_url_repo"))
    repo.git("remote", "add", "origin", FAKE_URL)
    return repo


@pytest.mark.parametrize(
    "env",
    (
        {"BUILD_BUILDID": "1", "BUILD_REPOSITORY_URI": REPO_URL},
        {"DRONE": "1", "DRONE_REPO_LINK": REPO_URL},
        {
            "GITHUB_ACTIONS": "1",
            "GITHUB_SERVER_URL": DOMAIN,
            "GITHUB_REPOSITORY": SLUG,
        },
        {"GITLAB_CI": "1", "CI_REPOSITORY_URL": REPO_URL},
        {"CIRCLECI": "1", "CIRCLE_REPOSITORY_URL": REPO_URL},
        {"BITBUCKET_COMMIT": "1", "BITBUCKET_GIT_HTTP_ORIGIN": REPO_URL},The advan
    )
)
def test_ci_repo_found(env, fake_url_repo) -> None:
    with mock.patch.dict(os.environ, env, clear=True):
        context = ScanContext(
            scan_mode=ScanMode.CI,
            command_path="ggshield secret scan path",
            target_path=local_repo.path,
        )

    _assert_repo_url_in_headers(context, EXPECTED_HEADER)
```

The advantage of this is you get more precise error messages in case of test failures and you have the ability to re-run only selected parameters.

agateau-gg (changes requested):
Minor remark on the tests. Looks good overall.

tests/unit/core/scan/test_scan_context.py
Paul-GitGuardian: The module scoped fixture is indeed exactly what I needed, thank you!

agateau-gg (approved):
Looks great now! ✨

PR #794, opened 2023-11-08
Implement pre-commit framework SKIP environment variable.

irgeek:
When `ggshield` is configured via the [pre-commit framework](https://pre-commit.com/) the user has the option to skip specific scans by setting the `SKIP` environment variable. When `ggshield` is configured directly as a `git` pre-commit hook, though, the user only has the option to disable _all_ pre-commit hooks via the `--no-verify` command-line option. Here we're "borrowing" the pre-commit framework's option so ggshield can be skipped the same way regardless of how it's configured.


agateau-gg (changes requested):
I like the idea but I would make a few changes to the implementation.

SKIP can apply to pre-push as well pre-commit, and it can apply to all verticals not only scan. For these reasons, I suggest the following:

1. add a `cmd.utils.hooks` module with a `check_user_requested_skip() -> bool` function in it, which would:
- read the env var
- print the info message if user requested skipping
- return True or False

2. Then make all precommit/prepush commands use this function.

irgeek:
@agateau-gg I've relocated my first implementation into a hooks module. Before I go scribbling in other files to add the check, can you just confirm this looks OK?

ggshield/cmd/utils/hooks.py
```diff
@@ -0,0 +1,32 @@
...
+from ggshield.core.text_utils import display_error
+
+
+def check_user_requested_skip() -> bool:
+    if get_skip_env_var():
+        return True
+    return False
+
+
+def get_skip_env_var() -> bool:
```
agateau-gg, line 12: I think you can merge these two, unless `check_user_requested_skip()` becomes responsible for showing the message.

ggshield/cmd/utils/hooks.py
```diff
@@ -0,0 +1,32 @@
...
+    framework](https://pre-commit.com/) the user has the option to skip
+    specific scans by setting the `SKIP` environment variable. When `ggshield`
+    is configured directly as a `git` pre-commit hook, though, the user only
+    has the option to disable _all_ pre-commit hooks via the `--no-verify`
+    command-line option. Here we're "borrowing" the pre-commit framework's
+    option so ggshield can be skipped the same way regardless of how it's
+    configured.
+    """
+    try:
+        skip_env_var = os.getenv("SKIP", None)
```
agateau-gg, line 24: nit: None is the default default value :)
```suggestion
        skip_env_var = os.getenv("SKIP")
```

ggshield/cmd/utils/hooks.py
```diff
@@ -0,0 +1,32 @@
...
+        if skip_env_var is not None:
+            skipped = set(p.strip().lower() for p in skip_env_var.split(","))
+            if "ggshield" in skipped:
+                return True
+    except BaseException as e:
+        display_error(f"Unable to parse SKIP environment variable: {str(e)}")
```
agateau-gg, line 30: I don't think we need to wrap the whole function like this: there's not really code which can fail there.

On the other hand, can you add a unit-test for this function?

ggshield/cmd/secret/scan/precommit.py
```diff
@@ -40,6 +41,11 @@ def precommit_cmd(
     Scan as a pre-commit hook all changes that have been staged in a git repository.
     """
     config: Config = ctx.obj["config"]
+
+    if check_user_requested_skip():
+        click.echo("Skipping GitGuardian pre-receive hook.", err=False)
```
agateau-gg, line 46: This is pre-commit here, not pre-receive ;). Actually if we make the message more generic (like "Skipping ggshield hook"), we can make `check_user_requested_skip()` print it and avoid duplicating the message in all hooks of all verticals.

agateau-gg (changes requested):
Minor remarks, this is going in the right direction 👍🏻.

tests/unit/cmd/utils/test_hooks.py
```diff
@@ -0,0 +1,23 @@
...
+def test_check_user_requested_skip_skipped(env_var_value):
+    with patch.dict("os.environ", {"SKIP": env_var_value}, clear=True):
+        assert check_user_requested_skip()
+
+
+@pytest.mark.parametrize("env_var_value", ["", "foo", "foo,bar"])
+def test_check_user_requested_skip_not_skipped(env_var_value):
+    with patch.dict("os.environ", {"SKIP": env_var_value}, clear=True):
+        assert not check_user_requested_skip()
+    pass
```
agateau-gg, line 18: This line can be removed

ggshield/cmd/utils/hooks.py
```diff
@@ -0,0 +1,30 @@
...
+    option so ggshield can be skipped the same way regardless of how it's
+    configured.
+    """
+    skip_env_var = os.getenv("SKIP")
+    if skip_env_var is not None:
+        skipped = set(p.strip().lower() for p in skip_env_var.split(","))
+        if "ggshield" in skipped:
+            return True
+
+    return False
```
agateau-gg, line 30: You can simplify this a bit using an early return and `any`:

```suggestion
    if skip_env_var is None:
        return False
        
    return any(p.strip().lower() == "ggshield" for p in skip_env_var.split(","))
```

agateau-gg (changes requested):
Looks good overall, thanks :+1:.

Just some minor remarks. It also needs a changelog entry (see https://github.com/GitGuardian/ggshield/blob/main/doc/dev/getting-started.md#changelog for details about the changelog entry)

agateau-gg:
As you pointed out, the lint check fails because the PR comes from a fork of the repository. Therefore I am merging it. Will look into that lint bug later.

StanislasGG:
Hello Aurelien, can you confirm that this will be included in next week's release ?

Thanks in advance

agateau-gg:
> Hello Aurelien, can you confirm that this will be included in next week's release ?
> 
> Thanks in advance

Yes, it will be there.

PR #796, opened 2023-11-09
chore: ensure no token is present before the test

GG-HH:
Please note that this is a bad fix because the isolated fs is not isolated and this should be fixed.

If I run this test alone with a hmsl_token in my file system, then it fails.

Please note that running the test will delete your current HMSL token if any.

### DIFF
--- a/ggshield/core/config/config.py
+++ b/ggshield/core/config/config.py
@@ -104,7 +104,7 @@ class Config:
     @property
     def api_url(self) -> str:
         """
-        The API URL to use to use the API
+        The API URL to use the API
         It's the API URL from the configured instance
         """
         # TODO change when instance_name can be a name instead of just a URL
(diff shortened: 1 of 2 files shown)


pierrelalanne:
Not sure to understand the context here @GG-HH ? Was this a failing test before the release process ?
Should we open a separate issue to find a more long term solution for the `isolated_fs` to have the expected behavior?

GG-HH:
Indeed, my description wasn't very explicit (and so is my proposed fix).
If you try to run this test alone while having a HMSL token in your cache, the test will fail because it doesn't use an isolated filesystem and the first assert will fail. I did not take time to investigate why the FS isn't isolated.

agateau-gg:
I think it would be interesting to understand the root cause of the problem here. Where is this cache stored?

agateau-gg:
Ah, I just had a look at the code and I think I know what's happening. It's because the path to the cache file is initialized globally as a Path. I had a similar issue with other parts of the code. A simple fix is to turn `TOKEN_PATH = ...` into a `get_token_path()` function.

GG-HH:
Thank you @agateau-gg , it indeed fixes the issue. I updated my MR.

I also fix a small typo in a docstring.

agateau-gg (approved):
Looks good, thanks!

PR #798, opened 2023-11-14
More refactoring of patch parser

agateau-gg:
This PR finishes removing the `Commit.files` property, which is currently responsible for loading whole commits in memory.

Changes in a Commit are no longer represented as StringScannable instances. Instead there is a dedicated class, CommitScannable. Using a dedicated class gives more control over the `path` and `url` properties.

Finally, I added some test to validate the JSON output because my changes in CommitInformation initially introduced new fields in the `extra_info` object. This was because all of CommitInformation was being turned into a dict and put in `extra_info`. Now we only store selected fields.


ggshield/core/scan/commit_information.py
```diff
@@ -0,0 +1,80 @@
...
+def _paths_from_name_only_line(text: str) -> List[Path]:
+    """
+    Parse a line generated by `git show` or `git diff` with the `--name-only -z`
+    options.
+
+    Filenames end with a \0, including the last one.
+
+    Return a list of paths.
+    """
+    return [Path(x) for x in text.rstrip("\0").split("\0")]
```
pierrelalanne, line 80: Naive question here, I have a `^@` separating filepaths when I launch the command on my machine, is this expected ? Could this vary depending on the git config ?

tests/functional/secret/test_scan_repo.py
```diff
@@ -1,56 +1,62 @@
...
-def test_scan_repo(tmp_path: Path) -> None:
-    # GIVEN a repository
-    repo = Repository.create(tmp_path)
+LEAK_CONTENT = f"password = {GG_VALID_TOKEN}"
+
+
+@pytest.fixture(scope="module")
+def leaky_repo(tmp_path_factory: pytest.TempPathFactory) -> Repository:
+    """
+    Create a repository, add a commit with a secret in it, then add 3 clean commits on
```
pierrelalanne, line 75: Only two clean commits ? The first one contains the secret.

ggshield/core/scan/commit_information.py
agateau-gg: I think the `^@` is the way your shell displays `\0`. You can see it more clearly using this small script I built to show invisible characters: [0cat](https://gist.github.com/agateau-gg/1bcafbfd85c34177d247cf475a0f8bd6).

tests/functional/secret/test_scan_repo.py
agateau-gg: Ah right, I forgot to call `create_commit()`. Fixing it.

ggshield/core/scan/commit_information.py
pierrelalanne: Ok, I just wanted to double check 👍

pierrelalanne (approved):

PR #800, opened 2023-11-15
Implement auto-ignore

Paul-GitGuardian:
The API now returns the status of each IaC incident on the GitGuardian app. We should use it to display only incidents that were not ignored on the app.

This applies to all IaC scan commands (all, diff, ci...) in all formats (text, JSON).

Depends on https://github.com/GitGuardian/py-gitguardian/pull/80

Commit 1 bumps the py-gitguardian version
Commits 2 and 3 contain some refactoring
Commit 4 is the main feature, described above
Commit 5 is a warning message in verbose mode when the source is not found by the API

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -185,45 +185,44 @@
         },
         "cryptography": {
             "hashes": [
-                "sha256:0c327cac00f082013c7c9fb6c46b7cc9fa3c288ca702c74773968173bda421bf",
-                "sha256:0d2a6a598847c46e3e321a7aef8af1436f11c27f1254933746304ff014664d84",
-                "sha256:227ec057cd32a41c6651701abc0328135e472ed450f47c2766f23267b792a88e",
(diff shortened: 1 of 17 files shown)


Paul-GitGuardian:
Hello @agateau-gg, the changes of this PR do not pass the tests yet as they require the typing changes in https://github.com/GitGuardian/py-gitguardian/pull/80. You can either review the code now or wait for the pygitguardian release. Thanks!

agateau-gg (changes requested):
CI is unhappy. To fix it:
- Follow instructions from https://github.com/GitGuardian/ggshield/blob/main/doc/dev/dependencies.md#using-an-unreleased-version-of-py-gitguardian
- Replace `type[IaCOutputHandler]` with `Type[IaCOutputHandler]`.

agateau-gg (approved):
Looks good now 👍🏻.

Paul-GitGuardian:
Thank you for your comments

PR #802, opened 2023-11-20
Print name of what is being scanned when scanning with --verbose

### LINKED ISSUE #212: Print name of what is being scanned when scanning with --verbose

Walz:
Closes #212 

When using `--verbose` additional logs are printed per element scan: `Scanned {name of the element}`:
- `scan path`, `scan archive` and `scan pypi` print the filenames
- `scan repo` and `scan commit-range` print the commit ids
- `scan docker` keep the same behavior and print the name of the layers

### DIFF
--- a/ggshield/verticals/secret/repo.py
+++ b/ggshield/verticals/secret/repo.py
@@ -50,6 +50,7 @@ def scan_repo_path(
                 matches_ignore=config.user_config.secret.ignored_matches,
                 scan_context=scan_context,
                 ignored_detectors=config.user_config.secret.ignored_detectors,
+                verbose=config.user_config.verbose,
             )
     except Exception as error:
         return handle_exception(error, config.user_config.verbose)
@@ -62,6 +63,7 @@ def scan_commits_content(
     matches_ignore: Iterable[IgnoredMatch],
     scan_context: ScanContext,
     progress_callback: Callable[..., None],
+    commit_scanned_callback: Callable[[Commit], None],
     ignored_detectors: Optional[Set[str]] = None,
 ) -> SecretScanCollection:  # pragma: no cover
     try:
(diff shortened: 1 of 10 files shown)


ggshield/verticals/secret/repo.py
```diff
@@ -62,14 +63,14 @@ def scan_commits_content(
     matches_ignore: Iterable[IgnoredMatch],
     scan_context: ScanContext,
     progress_callback: Callable[..., None],
+    commit_scanned_callbacks: Callable[[Commit], None],
```
agateau-gg, line 66: nit: should be `commit_scanned_callback` (no ending `s`) because it's a single callback (called multiple times).

ggshield/verticals/secret/rich_secret_scanner_ui.py
```diff
@@ -27,6 +32,9 @@ def __exit__(self, *args: Any) -> None:
         self.progress.__exit__(*args)
 
     def on_scanned(self, scannables: Sequence[Scannable]) -> None:
+        if self.verbose:
+            for scannable in scannables:
+                self.progress.console.print(f"Scanned [blue]{scannable.path}")
```
agateau-gg, line 37: For consistency, can you change the message string in `on_skipped()` to use `[blue]` for the path as well (or not use `[blue]` and let Rich do its magic, whichever you prefer). Otherwise it looks odd:

image

agateau-gg (changes requested):
Minor remarks, looks good overall!

There is a conflict to solve though, now that I merged the incremental parser patch.

ggshield/verticals/secret/rich_secret_scanner_ui.py
Walz: I've removed the `[blue]`.

agateau-gg (approved):
Looks good, thanks!

PR #805, opened 2023-11-28
feat(ignore): add output for no longer ignored incidents

carla-gitguardian:
Adapt ggshield's output when the grace period of an IAC incident ignored by a developer has been expired, for both all and diff scans.

### DIFF
--- a/ggshield/cmd/iac/scan/iac_scan_utils.py
+++ b/ggshield/cmd/iac/scan/iac_scan_utils.py
@@ -1,18 +1,22 @@
 from dataclasses import dataclass
 from pathlib import Path
 from re import Pattern
-from typing import Iterable, Optional, Set, Type
+from typing import Iterable, Optional, Set, Type, Union
 
 import click
 from pygitguardian import GGClient
+from pygitguardian.iac_models import IaCDiffScanResult, IaCScanResult
 from pygitguardian.models import Detail
 
 from ggshield.cmd.utils.context_obj import ContextObj
+from ggshield.core.config.user_config import UserConfig
 from ggshield.core.errors import APIKeyCheckError
+from ggshield.core.filter import init_exclusion_regexes
 from ggshield.core.tar_utils import INDEX_REF, tar_from_ref_and_filepaths
 from ggshield.core.text_utils import display_error
 from ggshield.utils.files import is_filepath_excluded
 from ggshield.utils.git_shell import get_filepaths_from_ref, get_staged_filepaths
+from ggshield.verticals.iac.collection.iac_scan_collection import IaCResult
 from ggshield.verticals.iac.filter import is_iac_file_path
 from ggshield.verticals.iac.output import (
     IaCJSONOutputHandler,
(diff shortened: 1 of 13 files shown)


ggshield/cmd/iac/scan/iac_scan_common_options.py
```diff
@@ -95,3 +98,63 @@ def update_context(
...
+) -> None:
+    """
+    GIVEN a list of vulnerabilities from a scan result
+    WHEN ignored policies and paths are configured but outdated
+    THEN augment the vulnerability with the date it was last ignored
+    """
+    if isinstance(result, IaCScanResult):
+        incidents_list = result.entities_with_incidents
+    elif isinstance(result, IaCDiffScanResult):
+        incidents_list = result.entities_with_incidents.new
```
sylvain-baud-gg, line 115: We want the text only for new incident ?

What about unchanged ?

ggshield/cmd/iac/scan/iac_scan_common_options.py
```diff
@@ -95,3 +98,63 @@ def update_context(
...
+            "until": outdated_ignored_path.until,
+        }
+        for outdated_ignored_path in outdated_ignored_paths
+    ]
+    outdated_ignored_policies_dict = {}
+    for outdated_ignored_policy in outdated_ignored_policies:
+        outdated_ignored_policies_dict[
+            outdated_ignored_policy.policy
+        ] = outdated_ignored_policy.until
+
```
sylvain-baud-gg, line 138:
```suggestion
    outdated_ignored_policies_dict = {outdated_policy.policy:outdated_policy.until for outdated_policy in outdated_ignored_policies}
```

ggshield/cmd/iac/scan/iac_scan_common_options.py
```diff
@@ -95,3 +98,63 @@ def update_context(
...
+                    file_ignored_until is None
+                    or file_ignored_until < outdated_ignored_path["until"]
+                ):
+                    file_ignored_until = outdated_ignored_path["until"]
+        for vulnerability in file_result.incidents:
+            # Check if policy was ignored
+            policy_ignored_until = outdated_ignored_policies_dict.get(
+                vulnerability.policy_id
+            )
+            until_dates = [policy_ignored_until, file_ignored_until]
```
sylvain-baud-gg, line 156:
```suggestion
            until_dates = (policy_ignored_until, file_ignored_until)
```

ggshield/cmd/iac/scan/iac_scan_utils.py
```diff
@@ -36,7 +36,9 @@ def create_output_handler(ctx: click.Context) -> IaCOutputHandler:
     else:
         output_handler_cls = IaCTextOutputHandler
     config: Config = ctx.obj["config"]
-    return output_handler_cls(verbose=config.user_config.verbose)
+    return output_handler_cls(
+        verbose=config.user_config.verbose, debug=config.user_config.debug
```
sylvain-baud-gg, line 40: Is it something we want to add debug here ?

sylvain-baud-gg (commented):
Can you please update tests ?

ggshield/cmd/iac/scan/iac_scan_common_options.py
carla-gitguardian: If I'm not mistaken, incidents details is only shown for new incidents and this is where the message about the date is shown. Or is there a verbose option that enables to show details for all incidents (new, unchanged and deleted)? Or do we want this augmentation for all incidents anyway, just so we keep them consistent?

ggshield/cmd/iac/scan/iac_scan_utils.py
carla-gitguardian: I'm not sure what this is for, I think it comes from Paul's MR which I rebased on.

carla-gitguardian: @Paul-GitGuardian ?

ggshield/cmd/iac/scan/iac_scan_common_options.py
sylvain-baud-gg: When verbose is on, we also display unchanged incidents.
We might want to validate with the product if we want to display the end of grace period in this case or not

ggshield/cmd/iac/scan/iac_scan_utils.py
Paul-GitGuardian: This was added in #800 
we pass the debug option to the output handler to optionally display some warnings based on it

ggshield/cmd/iac/scan/iac_scan_common_options.py
carla-gitguardian: Validated decision: we augment them all.

ggshield/verticals/iac/collection/filter_ignored.py
```diff
@@ -0,0 +1,25 @@
...
+
+def filter_unignored_incidents(
+    incidents: List[IaCVulnerability],
+) -> List[IaCVulnerability]:
+    return [
+        incident
+        for incident in incidents
+        if getattr(incident, "status", None) != "IGNORED"
+        and getattr(incident, "ignore_reason", None) != "ggshield"
+    ]
```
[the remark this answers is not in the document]
Paul-GitGuardian: It returns the incidents list without ignored incidents.
I'm adding a docstring

ggshield/core/config/user_config.py
```diff
@@ -169,17 +169,23 @@ class IaCConfig(FilteredConfig):
     ignored_paths: List[IaCConfigIgnoredPath] = field(default_factory=list)
     ignored_policies: List[IaCConfigIgnoredPolicy] = field(default_factory=list)
     minimum_severity: str = "LOW"
+    outdated_ignored_paths: List[IaCConfigIgnoredPath] = field(default_factory=list)
+    outdated_ignored_policies: List[IaCConfigIgnoredPolicy] = field(
```
agateau-gg, line 173: I am not fond of using a `*Config` class to store computed data. As the docstring of the class say, this class is supposed to store the content of the IAC config of a .gitguardian.yaml file. But the border is fuzzy here: the Config class itself does too much already IMO.

If we keep these fields, they must be marked as excluded in the UserConfig schema, like we did for `deprecation_messages` (another ugly hack). Otherwise they will be serialized if the user runs one of the commands that write to the .gitguardian.yaml file (such as `ggshield secret ignore --last-found`)

According to [Marshmallow doc](https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.Schema), you should be able to refer to these fields from the UserConfig schema using dot delimiters.

agateau-gg (changes requested):

ggshield/core/config/user_config.py
carla-gitguardian: Okay, let's keep it that way then. I did add a comment, isn't it what you meant?

agateau-gg (approved):

ggshield/core/config/user_config.py
carla-gitguardian: Alright, thanks for your review!

PR #809, opened 2023-12-01
fix: fix gitlab push pipelines for new branches

Paul-GitGuardian:
When pushing a new branch for the first time, the GGshield IAC and SCA CI cannot find the commit to use as ref.
As described [here](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html), `CI_COMMIT_BEFORE_SHA` is empty in that case.
Here we use git commands to find the starting commit of the new branch.

### DIFF
--- a/ggshield/core/git_hooks/ci/previous_commit.py
+++ b/ggshield/core/git_hooks/ci/previous_commit.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 from typing import Optional
 
 import click
@@ -8,6 +9,7 @@ from ggshield.utils.git_shell import (
     EMPTY_SHA,
     get_last_commit_sha_of_branch,
     get_list_commit_SHA,
+    get_new_branch_ci_commits,
     git,
 )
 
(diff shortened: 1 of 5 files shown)


GG-HH (approved):
Sounds good to me! 
I have one question still: Are we sure that GitLab will have cloned all the branches (at least the parent one) ?

Paul-GitGuardian:
Good catch, in fact it doesn't clone other branches. I've changed it to use remote branches instead.
Not sure if I can test the CI using this branch on a real repository.
At least I've tested the git commands in a Gitlab pipeline, they work.

GG-HH:
You can test it, just declare a job that will clone and install ggshield from your latest commit.

BTW, the CI is failing here

Paul-GitGuardian:
@GG-HH I've tested on a repository and made the necessary changes. Could you review it one last time please?

ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -112,6 +114,22 @@ def gitlab_previous_commit_sha(verbose: bool) -> Optional[str]:
...
 
+    # push_before_sha is also always EMPTY_SHA for the first commit of a new branch
+    current_branch = os.getenv("CI_COMMIT_BRANCH")
+    if current_branch:
+        new_commits = get_new_branch_ci_commits(current_branch, Path.cwd(), "origin")
+        if new_commits:
+            new_branch_before_sha = f"{new_commits[-1]}^1"
+            if verbose:
+                click.echo(
+                    f"new_branch_before_sha: {new_branch_before_sha}\n",
```
GG-HH, line 125: Should we have a more "user friendly" message ?

GG-HH (approved):
Nice!

PR #810, opened 2023-12-01
Test cleanups

agateau-gg:
Some minor test fixes. This was prompted by `pytest --help` taking a long time now that I have setup ggshield as a global git hook.


gg-mmill (approved):
Sounds good  !

PR #812, opened 2023-12-07
Draft: POC - Try Pluggy for HMSL modularity

Dedenne:
The goal of this branch is to assert the compatibility of the Pluggy format with what we want to do with HMSL plugins


agateau-gg:
Closed as stale.

PR #813, opened 2023-12-07
Remove mention of GITGUARDIAN_API_URL

Walz:
- Make use of GITGUARDIAN_INSTANCE in the unittest
- Remove mention of GITGUARDIAN_API_URL in documents


agateau-gg (approved):
Looks good, thanks for taking care of this.

PR #815, opened 2023-12-13
feat: consider paths in config file as relative to config path

gg-jonathangriffe:


ggshield/cmd/utils/files.py
```diff
@@ -7,6 +7,14 @@
...
 
 
-def check_directory_not_ignored(directory: Path, exclusion_regexes: Set[re.Pattern]):
-    if is_filepath_excluded(directory, exclusion_regexes):
-        raise UsageError("An ignored file or directory cannot be scanned.")
+IGNORED_ERROR_MESSAGE = "An ignored file or directory cannot be scanned."
+
+
+def check_directory_not_ignored(exclusion_regexes: Set[re.Pattern]):
+    if re.compile("(^|/)\\.$") in exclusion_regexes:
```
amascia-gg, line 14: Hi @xblanchot-gg, I'm taking over this MR and made some changes. However, I cannot figure out exactly what @gg-jonathangriffe was trying to do here, do you have an opinion?

Blaxav: I think the regex is false, and he wanted to check that path `/.` or `.` was not ignored. If it's the case, the reex should be `"(^|\/)\.$"`. However, I don't really understand why we would like to check that.

amascia-gg: yes that's what I thought but I don't see how that could match anything in `exclusion_regexes`, I'm thinking of reverting this so that `directory` is checked as before

PR #816, opened 2023-12-14
fix(iac,test): use tmp file in test

carla-gitguardian:
Fix a test creating a file outside the `tmp_path`


agateau-gg (approved):
Thanks! 👍🏻

PR #820, opened 2023-12-19
Fix IAC scan all on subdirectories

Paul-GitGuardian:
Currently, tars have two ways to be created in GGShield:

- For iac/sca scan diff, filepaths are listed using `git ls-tree --full-name`, which return filepaths relative to the git root
- For iac/sca scan all, filepaths are listed using `directory.rglob("*")` (where `directory` is the argument passed to the command), then returned relative to the directory using `path.relative_to(directory)`
- The difference comes from the necessity of listing files for a given git reference, for diff scans

However, the `all` behavior introduces two issues for IAC when calling the command with a Git subdirectory as argument:
1. File contents are not shown in the output, as the output handler looks for files relative to the git root, but is given filepaths relative to the subdir argument
2. Only the subdir is added to the tar sent to the API, preventing from linking existing incidents properly (as their paths are relative to the git root)
SCA is not impacted by these issues yet, but would be impacted by 2. when implemented.

This PR solves this issue by using the git root for all scans too as the root directory.
If the scanned directory isn't a Git repository, the current behavior is maintained.

### DIFF
--- a/ggshield/cmd/iac/scan/all.py
+++ b/ggshield/cmd/iac/scan/all.py
@@ -18,6 +18,7 @@ from ggshield.cmd.utils.common_decorators import exception_wrapper
 from ggshield.cmd.utils.common_options import directory_argument
 from ggshield.cmd.utils.context_obj import ContextObj
 from ggshield.cmd.utils.files import check_directory_not_ignored
+from ggshield.core.dirs import get_project_root_dir
 from ggshield.core.git_hooks.ci.supported_ci import SupportedCI
 from ggshield.core.scan import ScanContext, ScanMode
 from ggshield.core.text_utils import display_info
@@ -79,9 +80,12 @@ def iac_scan_all(
     if not paths:
         return IaCSkipScanResult()
 
+    root = get_project_root_dir(directory)
+    relative_paths = [str(x.resolve().relative_to(root)) for x in paths]
+
     if config.user_config.verbose:
         display_info("> Scanned files")
-        for filepath in paths:
+        for filepath in relative_paths:
             display_info(f"- {click.format_filename(filepath)}")
 
     client = ctx.obj["client"]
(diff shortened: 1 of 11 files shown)


Paul-GitGuardian:
For SCA, after the fix, the output will show filepaths relative to the git repository root, instead of relative to the given directory. Is that a problem?

ggshield/cmd/iac/scan/all.py
```diff
@@ -79,9 +81,12 @@ def iac_scan_all(
     if not paths:
         return IaCSkipScanResult()
 
+    root = get_project_root_dir(directory)
+    relative_paths = [str(x.resolve().relative_to(root)) for x in paths]
```
agateau-gg, line 85: Would be nice if `relative_paths` was a list of `Path` instead of a list of `str`. Can we remove the `str(...)` call?

ggshield/cmd/iac/scan/all.py
```diff
@@ -118,10 +123,11 @@ def display_iac_scan_all_result(
     directory: Path,
     result: Union[IaCScanResult, IaCSkipScanResult, None],
 ) -> int:
+    root = get_git_root(directory) if is_git_dir(directory) else directory.resolve()
```
agateau-gg, line 126: I think you can use your new `get_project_root_dir()` function here.

ggshield/cmd/sca/scan/sca_scan_utils.py
```diff
@@ -75,9 +76,14 @@ def sca_scan_all(
         empty_output.status_code = sca_filter_status_code
         return empty_output
 
+    root = get_project_root_dir(directory)
+    relative_paths = [
+        str((directory / x).resolve().relative_to(root)) for x in sca_filepaths
```
agateau-gg, line 81: Same remark here regarding `str(...)`.

agateau-gg (changes requested):
Minor remarks, looks good overall. +1 for the detailed PR description!

ggshield/cmd/iac/scan/all.py
Paul-GitGuardian: This would require changes in pygitguardian to accept Path lists (currently it expects string lists). Do you want me to open a PR?

ggshield/cmd/iac/scan/all.py
Paul-GitGuardian: Good catch thanks, in fact this is already done in the output handlers, so I've removed it here

agateau-gg (approved):

amascia-gg (approved):
Thanks for the fix, LGTM!

PR #823, opened 2024-01-03
 Fix ggshield creating too large payloads

fnareoh:
Aims at fixing the issue with scanning large documents where chunks can end up being larger than the maximum server payload, first reported in #555.

Depends on a change to py-gitguardian, [see PR here](https://github.com/GitGuardian/py-gitguardian/pull/85), to pass the server payload to the GGClient.

To facilitate the computation of chunk size an utf8_encoded_size property was added and calls a new method _read_content.
Some of the code is a bit clunky with `if self.content is None` statement even tough we just called a read method to make sure it wouldn't be None, but this way pyright doesn't complain.

I left a constant margin in the chunk size to encode the metadata and tested it with a scan of 10 000 files of 1Kb but it might be better to have it adapt depending on the number of files in the chunk ?


ggshield/core/scan/scannable.py
```diff
@@ -68,11 +70,27 @@ def is_longer_than(self, max_utf8_encoded_size: int) -> bool:
...
+
+    @property
     def content(self) -> str:
         """Return the decoded content of the scannable"""
-        raise NotImplementedError
+        if self._content is None:
+            self._read_content()
+        if self._content is not None:
+            return self._content
+        raise ValueError("content is None after reading")
```
agateau-gg, line 85: nit: I would write this like this, it feels more natural to me:

```suggestion
        if self._content is None:
            self._read_content()
            if self._content is None:
                raise ValueError("content is None after reading")
        return self._content
```

ggshield/core/scan/commit_utils.py
```diff
@@ -79,13 +83,14 @@ def path(self) -> Path:
...
-    def content(self) -> str:
-        return self._content
+        self._read_content()
+        if (
+            self._utf8_encoded_size is not None
+        ):  # it should not be, but just in case for typing
+            return self._utf8_encoded_size > max_utf8_encoded_size
+        raise ValueError(
+            f"For file {self.path}, _utf8_encoded_size is None after reading."
+        )
```
agateau-gg, line 93: As you said: "it should not be", so the check can be done using an assert at this point (it's a programmer error if it fails). It is enough to make pyright happy:

```suggestion
        self._read_content()
        assert self._utf8_encoded_size is not None
        return self._utf8_encoded_size > max_utf8_encoded_size
```

ggshield/core/scan/scannable.py
```diff
@@ -163,7 +181,10 @@ def __init__(self, url: str, content: str, filemode: Filemode = Filemode.FILE):
         self._url = url
         self._path: Optional[Path] = None
         self._content = content
-        self._utf8_encoded_size = None
+
+    def _read_content(self) -> None:
+        if self._content is not None and self._utf8_encoded_size is None:
```
agateau-gg, line 186: `self._content` cannot be None for StringScannable so we can simplify this a bit:

```suggestion
        if self._utf8_encoded_size is None:
```

ggshield/core/scan/scannable.py
```diff
@@ -181,10 +202,11 @@ def path(self) -> Path:
...
-    def content(self) -> str:
-        return self._content
+        self._read_content()
+        if (
+            self._utf8_encoded_size is not None
+        ):  # it should not be, but just in case for typing
+            return self._utf8_encoded_size > max_utf8_encoded_size
+        raise ValueError(
+            f"For file {self.path}, _utf8_encoded_size is None after reading."
+        )
```
agateau-gg, line 212: Same remark here regarding asserts

ggshield/verticals/secret/secret_scanner.py
```diff
@@ -150,6 +151,8 @@ def _start_scans(
         chunks_for_futures = {}
 
         chunk: List[Scannable] = []
+        max_payload_size = self.client.maximum_payload_size
+        utf8_encoded_chunk_size = _SIZE_METADATA_OVERHEAD
```
agateau-gg, line 155: Nit: if you set `max_payload_size` to `self.client.maximum_payload_size - _SIZE_METADATA_OVERHEAD`, then `utf8_encoded_chunk_size` can be 0. The nice thing about this is `utf8_encoded_chunk_size` is really the sum of all document chunk sizes.

agateau-gg (changes requested):
Made a few minor remarks, but this looks good! I like that it moves some duplicated code from Scannable concrete classes up to Scannable itself.

I think you need to rebase your py-gitguardian branch on top of the default branch to get the latest py-gitguardian changes because other parts of GGShield depend on it.

ggshield/verticals/secret/secret_scanner.py
fnareoh: very good point thanks !

agateau-gg (approved):
Looks good now!

PR #824, opened 2024-01-04
fix(iac): prepush scan all

carla-gitguardian:
This MR aims to fix 2 edge cases during IaC pre-push scans:
- Use scan diff (instead of all) when pushing a new branch. The ref commit should be the last one on the parent branch.
- Prevent staged files to be scanned during pre-push "all" scans. It should only occurs on the first push for each repo. This avoids to push a vulnerability that could have been fixed in the staged files only.

### DIFF
--- a/ggshield/core/scan/file.py
+++ b/ggshield/core/scan/file.py
@@ -61,6 +61,7 @@ def get_files_from_paths(
     display_scanned_files: bool,
     display_binary_files: bool,
     ignore_git: bool = False,
+    ignore_git_staged: bool = False,
 ) -> List[Scannable]:
     """
     Create a scan object from files content.
@@ -75,7 +76,11 @@ def get_files_from_paths(
     """
     try:
         filepaths = get_filepaths(
-            paths, exclusion_regexes, recursive, ignore_git=ignore_git
+            paths,
+            exclusion_regexes,
+            recursive,
+            ignore_git=ignore_git,
+            ignore_git_staged=ignore_git_staged,
         )
     except UnexpectedDirectoryError as error:
         raise click.UsageError(
--- a/ggshield/utils/files.py
+++ b/ggshield/utils/files.py
@@ -3,7 +3,7 @@ from pathlib import Path, PurePosixPath
 from typing import List, Set, Union
 
 from ggshield.utils._binary_extensions import BINARY_EXTENSIONS
-from ggshield.utils.git_shell import git_ls, is_git_dir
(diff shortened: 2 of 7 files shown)


Paul-GitGuardian (approved):

tests/functional/iac/test_iac_scan_prepush.py
```diff
@@ -97,3 +98,34 @@ def test_iac_scan_prepush_vuln_before_hook(
...
+    # AND a branch with a new commit
+    new_branch = "topic"
+    local_repo.create_branch(new_branch)
+    file = local_repo.path / "no_vuln.tf"
+    file.write_text(_IAC_NO_VULNERABILITIES)
+    local_repo.add("no_vuln.tf")
+    local_repo.create_commit()
+
+    # WHEN I try to push the new branch
+    # THEN the hook triggers a diff scan
```
agateau-gg, line 128: nit: I would move the `# THEN` comment *after* the `local_repo.git("push"...` line.

ggshield/utils/files.py
```diff
@@ -43,7 +44,16 @@ def get_filepaths(
...
+                if ignore_git_staged:
+                    _targets = {
+                        path / target
+                        for target in get_filepaths_from_ref(
+                            ref="HEAD",
+                            wd=path,
+                        )
+                    }
+                else:
+                    _targets = {path / target for target in git_ls(path)}
```
agateau-gg, line 56: nit: this can be factorized a bit. Something like this:

```python
paths = get_filepaths_from_ref("HEAD", wd=path) if ignore_git_staged else git_ls(path)
_targets = {path / x for x in paths}
```

agateau-gg (approved):
Some very minor remarks. Looks good otherwise.

ggshield/core/scan/file.py
```diff
@@ -61,6 +61,7 @@ def get_files_from_paths(
     display_scanned_files: bool,
     display_binary_files: bool,
     ignore_git: bool = False,
+    ignore_git_staged: bool = False,
```
sylvain-baud-gg, line 64: If I understand correctly, the `get_files_from_paths` function was previously always taking staged files.
If we introduce a boolean to toggle this, I think its default value should be what the function was previously doing. Especially since this function is used in 5 different files including Secret and SCA files. Are we sure we want to change the default behavior for them as well ?

carla-gitguardian: The default value does correspond to the previous behaviour.

PR #825, opened 2024-01-05
Fix IAC scan CI for new branches on Github

Paul-GitGuardian:
First commit refactors some IAC tests to use fixtures instead of global variables, and could be reviewed separately.

Other commits fix 2 bugs for IAC and SCA:
- CI scans failed when pushing a new branch, due to CI predefined variables being incorrect -> now the parent commit is calculated using git commands
- They failed on the first push to a repository for the same reason (and because there were no commits to compare HEAD to) -> a scan all is triggered instead

Previously, for these two cases, Gitlab CI failed, while Github CI fallback was to scan the last commit only.

CI is still misbehaving on force pushes, this will be fixed in another PR.

### DIFF
--- a/ggshield/core/git_hooks/ci/previous_commit.py
+++ b/ggshield/core/git_hooks/ci/previous_commit.py
@@ -12,6 +12,7 @@ from ggshield.utils.git_shell import (
     get_list_commit_SHA,
     get_new_branch_ci_commits,
     git,
+    is_valid_git_commit_ref,
 )
 
 from .supported_ci import SupportedCI
@@ -61,7 +62,7 @@ def github_previous_commit_sha(verbose: bool) -> Optional[str]:
         ref_type = os.getenv("GITHUB_REF_TYPE")
         ref_name = os.getenv("GITHUB_REF_NAME")
         if ref_type == "branch" and ref_name:
-            return get_new_branch_before_sha(ref_name, verbose)
+            return get_new_branch_parent_commit(ref_name, verbose)
 
         if verbose:
             click.echo("Could not find previous commit for current branch.")
@@ -124,7 +125,7 @@ def gitlab_previous_commit_sha(verbose: bool) -> Optional[str]:
     # push_before_sha is also always EMPTY_SHA for the first commit of a new branch
     current_branch = os.getenv("CI_COMMIT_BRANCH")
     if current_branch:
-        return get_new_branch_before_sha(current_branch, verbose)
+        return get_new_branch_parent_commit(current_branch, verbose)
 
     raise UnexpectedError(
         "Unable to get previous commit. Please submit an issue with the following info:\n"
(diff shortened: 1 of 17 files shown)


ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -282,6 +271,21 @@ def azure_pull_request_previous_commit_sha(verbose: bool) -> Optional[str]:
...
+        new_branch_before_sha = f"{new_commits[-1]}^1"
+        if verbose:
+            click.echo(
+                f"new_branch_before_sha: {new_branch_before_sha}\n",
+                err=True,
+            )
+        return new_branch_before_sha
+    else:
+        # New branch pushed with no commits
+        return "HEAD"
```
agateau-gg, line 322: If I am not mistaken, this only works if `branch_name` is the current branch. I think it would be more robust to return `branch_name` here.

And maybe use an early return:

```python
    new_commits = get_new_branch_ci_commits(branch_name, Path.cwd(), "origin")
    if not new_commits:
        return branch_name  
    new_branch_before_sha = f"{new_commits[-1]}^1"
    if verbose:
        click.echo(
            f"new_branch_before_sha: {new_branch_before_sha}\n",
            err=True,
        )
    return (
        new_branch_before_sha
        if is_valid_git_commit_ref(new_branch_before_sha)
        else None
    )
```

ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -282,6 +303,25 @@ def azure_pull_request_previous_commit_sha(verbose: bool) -> Optional[str]:
     return None
 
 
+def get_new_branch_before_sha(branch_name: str, verbose: bool) -> Optional[str]:
```
agateau-gg, line 306: This name is not very clear to me. Maybe get_new_branch_parent_commit()?

Also: can you add a docstring?

tests/conftest.py
```diff
@@ -202,6 +202,21 @@ def do_not_use_colors(monkeypatch):
...
+
+
+@pytest.fixture
+def iac_multiple_vulns() -> str:
+    return _IAC_MULTIPLE_VULNERABILITIES
+
+
+@pytest.fixture
+def iac_no_vulns() -> str:
+    return _IAC_NO_VULNERABILITIES
```
agateau-gg, line 217: It looks to me that using fixtures instead of constants makes tests look more complicated (more arguments) for no benefits, am I missing something?

agateau-gg (changes requested):

ggshield/core/git_hooks/ci/previous_commit.py
Paul-GitGuardian: You're right. I've made the change with two differences:
- the branch_name returned is prefixed with "refs/remotes/origin/", as it seems this is how other branches should be accessed in CI environments
- If this happens to not resolve to a valid ref, the function falls back to HEAD instead

tests/conftest.py
Paul-GitGuardian: This may be personal preference, but I find fixtures more elegant in tests. Should we keep the global variables, and simply remove the leading underscores from their names instead?

agateau-gg: Yes, I would prefer if we just removed the leading underscores. The advantage of fixtures is the ability to setup/teardown, and there is no need for this on constants.

amascia-gg (approved):
LGTM, thanks for the fix Paul!

PR #826, opened 2024-01-05
Finish migrating to ContextObj

agateau-gg:
## Description

This PR finishes the migration to ContextObj. It does two things:

- convert the remaining code to use ContextObj instead of dict access.
- remove dict-like support from ContextObj.

## Interesting trivia

Interestingly this pointed out harmless typing errors, for example:

```python
if not scan.success or not isinstance(scan, IaCScanResult):
    handle_scan_error(client, scan)
    return None
```

pyright complained because the 2nd argument of `handle_scan_error()` expects a `Detail` instance, but pyright thought it could be a `IaCScanResult`. From a typing perspective, it could happen if `scan.success` was False but `isinstance(scan, IaCScanResult)` was True. In practice it cannot happen because `scan` is always a `Detail` when `scan.success` is False.

I wondered why it did not complain before, and I realized it's because `client` was previously defined as `client = ctx.obj["client"]`, so pyright had no idea what type it was. And also had no idea what the return type of `client.iac_directory_scan()` was.

Now that `client` is defined as `client.ctx_obj.client`, pyright knows it's a `GGClient` instance and thus knows the return type of `client.iac_directory_scan()`.

### DIFF
--- a/ggshield/cmd/iac/scan/diff.py
+++ b/ggshield/cmd/iac/scan/diff.py
@@ -77,7 +77,8 @@ def scan_diff_cmd(
     result = iac_scan_diff(
         ctx, directory, ref, staged, scan_mode=ScanMode.DIRECTORY_DIFF
     )
-    augment_unignored_issues(ctx.obj["config"].user_config, result)
+    ctx_obj = ContextObj.get(ctx)
+    augment_unignored_issues(ctx_obj.config.user_config, result)
     return display_iac_scan_diff_result(ctx, directory, result)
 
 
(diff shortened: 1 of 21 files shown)


ggshield/cmd/iac/scan/diff.py
```diff
@@ -194,7 +196,7 @@ def iac_scan_diff(
         ).get_http_headers(),
     )
 
-    if not scan.success or not isinstance(scan, IaCDiffScanResult):
+    if not isinstance(scan, IaCDiffScanResult):
```
carla-gitguardian, line 199: I'm not sure to understand why you removed this condition here? Same in `all.py`.

carla-gitguardian (commented):
Otherwise LGTM. Interesting indeed to note the pyright's behaviour!

ggshield/cmd/iac/scan/diff.py
agateau-gg: That is required to fix pyright error: it thinks it is possible for both `scan.success` to be False and `scan` to be an instance of `IaCDiffScanResult`. If this were to happen, then the code in the `if` block would be wrong because it would pass an `IaCDiffScanResult` to `handle_scan_errors()`.

I can see how the change is not intuitive, but the alternative requires adding some `assert`, which is not super nice either:

```python
if not scan.success:
    assert isinstance(scan, Detail)
    handle_scan_error(client, scan)
    return None
assert isinstance(scan, IaCDiffScanResult)
return scan
```

The second `assert` is required because pyright does not know that if we did not enter the `if` block then `scan` is always an instance of `IaCDiffScanResult`.

carla-gitguardian (approved):

PR #831, opened 2024-01-10
Logging fixes

agateau-gg:
This small PR make a bunch of small logging fixes, mainly for `secret scan ci` but not only:

- Information reported when called with `--verbose` is now also logged, so using `--debug` without using `--verbose` does not end up with less information.
- When called with `--debug`, `secret scan ci` now also logs the list of commits to scan.
- Fix a case where we were using the root logger instead of the logger defined in the module


Paul-GitGuardian (approved):
Nice!
Maybe `log_vars` should be moved to an higher level? I think it could be used in `previous_commit.py` too

agateau-gg:
> Nice! Maybe `log_vars` should be moved to an higher level? I think it could be used in `previous_commit.py` too

Good idea, going to do this.

agateau-gg:
Closing this for now, I am reworking it.

PR #837, opened 2024-01-12
Handle IAC scans on CI after force pushing

carla-gitguardian:
This MR intends to fix the behaviour of CI jobs after a forced push, on both IAC and SCA scans.

It used to trigger an error, trying to access a commit from before the history was rewritten.

This situation will now trigger a scan on all commits instead.

### DIFF
--- a/ggshield/core/git_hooks/ci/previous_commit.py
+++ b/ggshield/core/git_hooks/ci/previous_commit.py
@@ -55,7 +55,7 @@ def github_previous_commit_sha(verbose: bool) -> Optional[str]:
         return pull_req_base_sha
 
     if push_before_sha and push_before_sha != EMPTY_SHA:
-        return push_before_sha
+        return get_commit_except_forced_push(push_before_sha, verbose)
 
     if head_sha and event_name == "push":
         # New branch pushed, try to get sha from git state
@@ -120,7 +120,7 @@ def gitlab_previous_commit_sha(verbose: bool) -> Optional[str]:
         return merge_req_base_sha
 
     if push_before_sha and push_before_sha != EMPTY_SHA:
-        return push_before_sha
+        return get_commit_except_forced_push(push_before_sha, verbose)
 
     # push_before_sha is also always EMPTY_SHA for the first commit of a new branch
     current_branch = os.getenv("CI_COMMIT_BRANCH")
@@ -182,8 +182,8 @@ def jenkins_previous_commit_sha(verbose: bool) -> Optional[str]:
     if target_branch:
         return get_last_commit_sha_of_branch(f"origin/{target_branch}")
 
-    if previous_commit:
-        return previous_commit
+    if previous_commit and previous_commit != EMPTY_SHA:
+        return get_commit_except_forced_push(previous_commit, verbose)
 
     if current_branch:
         try:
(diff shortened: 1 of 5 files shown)


Paul-GitGuardian (changes requested):
Also please add a test to `test_previous_commit` with the expected CI setup (previous_commit pointing to an invalid commit sha)

ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -183,6 +191,12 @@ def jenkins_previous_commit_sha(verbose: bool) -> Optional[str]:
         return get_last_commit_sha_of_branch(f"origin/{target_branch}")
 
     if previous_commit:
+        if previous_commit != EMPTY_SHA and not is_valid_git_commit_ref(
+            previous_commit
+        ):
+            if verbose:
+                click.echo("> This might be a forced push.")
+            return None
```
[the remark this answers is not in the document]
carla-gitguardian: Could it be EMPTY_SHA here in case of a new repo? Or would this case be handled somewhere else?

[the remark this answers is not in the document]
Paul-GitGuardian: I've not tested on Jenkins so I don't know, perhaps Jenkins never returns an EMPTY_SHA. I think you could set the line to `if push_before_sha and push_before_sha != EMPTY_SHA:` just in case, it wouldn't break anything

tests/unit/core/git_hooks/ci/test_previous_commit.py
```diff
@@ -121,3 +127,93 @@ def test_get_previous_commit_from_ci_env_new_repo(
...
+    """
+    remote_repository = Repository.create(tmp_path / "remote", bare=True)
+    repository = Repository.clone(remote_repository.path, tmp_path / "local")
+
+    repository.create_commit()
+    before_sha = repository.create_commit()
+
+    repository.git("commit", "--amend", "--allow-empty", "-m", "message")
+    repository.remove_unreachable_commits()
+
```
Paul-GitGuardian, line 190: nit: assert here (and in the other force push tests) that:
- `is_valid_git_commit_ref(before_sha)` is False (= we properly set an invalid SHA in CI env, which is what Github/Gitlab do)
- `repository.get_top_sha()` is different from `before_sha` (= the repository is ready for a force push)

Paul-GitGuardian (approved):
:+1:

ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -319,11 +319,20 @@ def get_new_branch_parent_commit(branch_name: str, verbose: bool) -> Optional[st
...
-    return (
-        new_branch_before_sha
-        if is_valid_git_commit_ref(new_branch_before_sha)
-        else None
-    )
+    if is_valid_git_commit_ref(new_branch_before_sha):
+        return new_branch_before_sha
+    else:
+        if verbose:
+            click.echo("> This might be a new repository.")
```
agateau-gg, line 326:
```suggestion
            click.echo("> This might be a new repository.", err=True)
```

ggshield/core/git_hooks/ci/previous_commit.py
```diff
@@ -319,11 +319,20 @@ def get_new_branch_parent_commit(branch_name: str, verbose: bool) -> Optional[st
...
+        return None
+
+
+def get_commit_except_forced_push(commit: str, verbose: bool) -> Optional[str]:
+    if not is_valid_git_commit_ref(commit):
+        if verbose:
+            click.echo("> This might be a forced push.")
```
agateau-gg, line 333:
```suggestion
            click.echo("> This might be a forced push.", err=True)
```

agateau-gg (changes requested):
Just minor nitpicks on output (progress output should use stderr instead of stdout)

agateau-gg (approved):

PR #839, opened 2024-01-17
Remove urlib3<2 pinning

agateau-gg:
The issue between urllib3 >= 2 and vcrpy has been fixed. Update vcrpy and remove the pinning.


xblanchot-gg:
Installed both with and with `--skip-lock`, works perfectly ! Thanks for the fix !

xblanchot-gg (approved):

PR #844, opened 2024-01-30
feat(sca): implement incidents output for autoignore feature

carla-gitguardian:
The autoignore feature modified SCA scan classes returned by py-gitguardian.
The added attributes should not be kept in the output schemas of ggshield.


changelog.d/20240130_093754_carla.martet.ext_implement_autoignored_incidents_output.md
```diff
@@ -0,0 +1,4 @@
+### Changed
+
+- Remove the following attributes from `SCAJSONVulnerablePackageVersion`: `url`, `status`, `ignored_until`, `ignore_reason`, `ignore_comment`.
+- Remove the `source_found` attribute from `SCAJSONScanAllOutput` and `SCAJSONScanDiffOutput`.
```
Paul-GitGuardian, line 4: I think changelogs should describe user-visible changes. Here there are none, as these attributes did not exist in the previous ggshield version

ggshield/verticals/sca/output/schemas.py
```diff
@@ -1,15 +1,41 @@
...
+from pygitguardian.sca_models import (
+    SCALocationVulnerability,
+    SCAScanAllOutput,
+    SCAScanDiffOutput,
+    SCAVulnerablePackageVersion,
+)
+
+
+@dataclass
+class SCAJSONVulnerablePackageVersion(SCAVulnerablePackageVersion):
```
Paul-GitGuardian, line 15: `SCAVulnerablePackageVersion` does not have these attributes, they must be removed from `SCAVulnerability`

Paul-GitGuardian (changes requested):
Can you test, with the updated pygitguardian version locally, that, when calling a sca command with `--json`:
- the request response contains the new attributes (`source_found` and all vulns attributes)
- the output does not contain them

Perhaps even write a test for this (check what exists for iac)

carla-gitguardian:
Hi @agateau-gg,
In this MR I wanted to exclude some attributes from the dataclasses returned by py-gitguardian. I did not find how to do it however, so I ended up redefining the classes in ggshield without the unwanted attributes to handle the output schemas. Would you think of a better way to do it?

agateau-gg:
> Hi @agateau-gg, In this MR I wanted to exclude some attributes from the dataclasses returned by py-gitguardian. I did not find how to do it however, so I ended up redefining the classes in ggshield without the unwanted attributes to handle the output schemas. Would you think of a better way to do it?

Just read the existing code and I have questions :)

In `SCAJsonOutputHandler._process_scan_*_impl()` methods: why do we deserialize the dict and then dump the object as JSON? It seems to me it would be simpler to run `json.dumps()` on the dict.

If there is an interest in serializing then deserializing, can you add a comment explaining it?

Assuming there is an interest in doing it, I think we can avoid basically duplicating all models by changing the code to use `schema.to_dict()` instead of `schema.dumps()` so that we have dictionaries and then remove the keys we don't want from said dictionaries.

agateau-gg:
Another point: something similar as already be done on the IAC side, have a look at ggshield/verticals/iac/output/schemas.py.

carla-gitguardian:
I'm not sure actually why it is serialized and then deserialized here, maybe @amascia-gg would have more clue?

As for IAC side, it had been done differently because schemas are exported from py-gitguardian for IaC whereas there are dataclasses for SCA. Maybe the change should directly occur in py-gitguardian to maintain better consistency? I don't know how much code would be impacted though.

agateau-gg:
> As for IAC side, it had been done differently because schemas are exported from py-gitguardian for IaC whereas there are dataclasses for SCA. Maybe the change should directly occur in py-gitguardian to maintain better consistency? I don't know how much code would be impacted though.

I am not sure I understand what difference you are referring to. Is it because in py-gitguardian IaC schemas are defined like this:

```python
IaCVulnerabilitySchema = cast(
    Type[BaseSchema], marshmallow_dataclass.class_schema(IaCVulnerability, BaseSchema)
)

IaCVulnerability.SCHEMA = IaCVulnerabilitySchema()
```

Whereas SCA schemas are defined like this?

```python
SCAIgnoredVulnerability.SCHEMA = cast(
    BaseSchema,
    marshmallow_dataclass.class_schema(
        SCAIgnoredVulnerability, base_schema=BaseSchema
    )(),
)
```

ie, if all that's missing is the `SCAIgnoredVulnerabilitySchema` class, then yes, let's define them in py-gitguardian.

carla-gitguardian:
I guess it is yes! I'll define them in py-gitguardian and use them the same way we did in IaC. I'll also look into this potentially unnecessary serialization. Thanks for your help!

Paul-GitGuardian:
Hi, some news:
- the load -> dump process on the dict apparently was necessary to parse the nested `created_at` datetime. Removing it resulted in an error (as `dump` expected a datetime, while the dict provided it as a string).
- however the `load` method fails because of the additional fields (such as `source_found`) that are not part of the schema, instead of ignoring them

There is no such issue for IAC because of the absence of datetime fields in the IAC schemas.

After spending **some** time trying to make it work, I decided to do as you suggested: discard the additional schemas and work with the dictionaries.

ggshield/verticals/sca/output/json_handler.py
```diff
@@ -1,22 +1,17 @@
...
 
 class SCAJsonOutputHandler(SCAOutputHandler):
     def _process_scan_all_impl(self, scan: SCAScanAllVulnerabilityCollection) -> str:
-        scan_dict = self.create_scan_all_dict(scan)
-        schema = SCAJSONScanAllOutputSchema
-        serialized_result = schema.load(scan_dict)
-        text = schema.dumps(serialized_result)
+        scan_dict = SCAJsonOutputHandler.create_scan_all_dict(scan)
+        text = json.dumps(scan_dict)
         return cast(str, text)
```
agateau-gg, line 15: nit: I think you can remove the `cast` here

agateau-gg (approved):
One minor nitpick. Not a blocker.

One thing to consider in the future is using JSON schemas to document and validate sca JSON outputs. I started doing so for secret outputs. It works like this:

- Schemas are defined in doc/schemas
- They are loaded as fixtures here: https://github.com/GitGuardian/ggshield/blob/main/tests/conftest.py#L312-L337
- They are checked by tests, for example here: https://github.com/GitGuardian/ggshield/blob/main/tests/functional/secret/test_scan_path.py#L89

Paul-GitGuardian:
> One minor nitpick. Not a blocker.
> 
> One thing to consider in the future is using JSON schemas to document and validate sca JSON outputs. I started doing so for secret outputs. It works like this:
> 
>     * Schemas are defined in doc/schemas
> 
>     * They are loaded as fixtures here: https://github.com/GitGuardian/ggshield/blob/main/tests/conftest.py#L312-L337
> 
>     * They are checked by tests, for example here: https://github.com/GitGuardian/ggshield/blob/main/tests/functional/secret/test_scan_path.py#L89

Ok that's great, I've made schemas for SCA, can you take a look?

I couldn't manage to import schema elements from other files, so I had to duplicate the `locations_vulnerabilities` block in both schemas. It looks like additional setup is required for that: https://python-jsonschema.readthedocs.io/en/latest/referencing/

tests/functional/sca/test_sca_scan_all.py
```diff
@@ -50,3 +55,21 @@ def test_scan_all_json_output(dummy_sca_repo: Repository) -> None:
...
+            ]
+            for key in [
+                "url",
+                "status",
+                "ignored_until",
+                "ignore_reason",
+                "ignore_comment",
+            ]
+        ]
+    )
```
agateau-gg, line 75: I think you can remove this code: validating the jsonschema takes care of it (but double-check just to be sure 😉)

Same below.

agateau-gg (changes requested):
Minor remark, hopefully you can remove some more code.

Good point about referencing. I must confess I don't know a lot about JSON schemas. I think we can live with the duplication for now, but this is something we should look into later if/when we use more schemas.

tests/functional/sca/test_sca_scan_all.py
Paul-GitGuardian: Yes it does throw on additional properties

agateau-gg (approved):
Looks good, but can you reword the last commit to not say "WIP" or squash it with the first one?

PR #845, opened 2024-01-30
chore(env): # Use py-gitguardian v1.13.0

pierrelalanne:
Change required before releasing ggshield 1.24.0

**Side effect:** pipenv also wanted to upgrade ipython to 8.20.0 which I prevented since this version is incompatible with Python 3.8.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,9 +43,7 @@ install_requires =
     marshmallow>=3.18.0,<3.19.0
     marshmallow-dataclass>=8.5.8,<8.6.0
     oauthlib>=3.2.1,<3.3.0
-    # TODO: replace this with a real version number as soon as a new version of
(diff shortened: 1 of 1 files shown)


agateau-gg (approved):

agateau-gg:
> Side effect: pipenv also wanted to upgrade ipython to 8.20.0 which I prevented since this version is incompatible with Python 3.8.

Interesting. Going to look into why we depend on ipython 🤔.

agateau-gg:
> > Side effect: pipenv also wanted to upgrade ipython to 8.20.0 which I prevented since this version is incompatible with Python 3.8.
> 
> Interesting. Going to look into why we depend on ipython 🤔.

Ah that's because of the dev dependency on ipdb.

PR #850, opened 2024-02-05
Hide vulnerabilities ignored in app

Paul-GitGuardian:
All vulnerabilities retrieved from the API are now filtered based on their statuses. Ignored vulnerabilities are no longer shown in the text/JSON outputs, for SCA scan all/diff commands.

The implementation is mostly similar to the one for IAC in #800.

### DIFF
--- a/tests/unit/verticals/sca/collection/test_filter_ignored.py
+++ b/tests/unit/verticals/sca/collection/test_filter_ignored.py
@@ -1,0 +1,129 @@
+from itertools import chain
+from typing import List
+
+import pytest
+from pygitguardian.sca_models import (
+    SCALocationVulnerability,
+    SCAVulnerability,
+    SCAVulnerablePackageVersion,
+)
+
+from ggshield.verticals.sca.collection.filter_ignored import (
+    filter_unignored_location_vulnerabilities,
+    filter_unignored_vulnerabilities,
+)
+
+
+BASE_VULN_INFO = {
+    "severity": "HIGH",
+    "summary": "Vulnerability summary",
+    "cve_ids": ["CVE-0000-0000"],
+    "created_at": None,
+}
+
+
+@pytest.fixture
+def ignored_vulns() -> List[SCAVulnerability]:
+    return [
+        SCAVulnerability.from_dict(
+            {
+                **BASE_VULN_INFO,
+                "status": "IGNORED",
+                **additional_info,
+            }
+        )
+        for additional_info in (
+            {"identifier": "id_0001"},
+            {"identifier": "id_0002", "ignore_reason": "ggshield"},
+            {
+                "identifier": "id_0003",
+                "url": "https://github.com/owner/repo/path",
+                "ignored_until": "2020-01-01T00:00:00",
+                "ignore_reason": "some reason",
+                "ignore_comment": "some comment",
+            },
+        )
+    ]
+
+
+@pytest.fixture
+def non_ignored_vulns() -> List[SCAVulnerability]:
+    return [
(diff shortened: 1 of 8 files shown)


tests/unit/verticals/sca/collection/test_filter_ignored.py
```diff
@@ -0,0 +1,129 @@
...
+    return [
+        SCAVulnerability.from_dict(
+            {
+                **BASE_VULN_INFO,
+                "status": "IGNORED",
+                **additional_info,
+            }
+        )
+        for additional_info in (
+            {"identifier": "id_0001"},
```
xblanchot-gg, line 36: identifier should be in `BASE_VULN_INFO`

tests/unit/verticals/sca/collection/test_filter_ignored.py
```diff
@@ -0,0 +1,129 @@
...
+):
+    """
+    GIVEN   a list of vulnerabilities
+    WHEN    calling filter_unignored_vulnerabilities
+    THEN    only non ignored vulnerabilities are returned
+    """
+    vulns = list(chain(*zip(ignored_vulns, non_ignored_vulns)))
+    expected_vulns = [vuln for vuln in vulns if vuln.status != "IGNORED"]
+    found_vulns = filter_unignored_vulnerabilities(vulns)
+    assert found_vulns == expected_vulns
```
xblanchot-gg, line 87: Are we sure the order of both lists will always be the same ? Otherwise, we may want to convert them to `set` in the assertion.

tests/unit/verticals/sca/collection/test_filter_ignored.py
```diff
@@ -0,0 +1,129 @@
...
+
+    found_files = filter_unignored_location_vulnerabilities(files)
+
+    assert len(found_files) == 2
+    assert found_files[0].package_vulns
+    assert [file.location for file in found_files] == ["location1", "location2"]
+    assert found_files[0].package_vulns[0].vulns == [
+        non_ignored_vulns[0],
+        non_ignored_vulns[1],
+    ]
```
xblanchot-gg, line 128: Same here, we might need to use `sets` to avoid this test being flaky, no ?

xblanchot-gg:
Thanks, I tested it using staging and it works perfectly ! I only have some small comments regarding the tests.

tests/unit/verticals/sca/collection/test_filter_ignored.py
```diff
@@ -0,0 +1,129 @@
...
+
+def test_filter_unignored_vulnerabilities(
+    ignored_vulns: List[SCAVulnerability], non_ignored_vulns: List[SCAVulnerability]
+):
+    """
+    GIVEN   a list of vulnerabilities
+    WHEN    calling filter_unignored_vulnerabilities
+    THEN    only non ignored vulnerabilities are returned
+    """
+    vulns = list(chain(*zip(ignored_vulns, non_ignored_vulns)))
```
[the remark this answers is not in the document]
Paul-GitGuardian: It is different, the chain zip will result in [ignored, non_ignored, ignored, ...], which I prefer as it adds more variance.

tests/unit/verticals/sca/collection/test_filter_ignored.py
Paul-GitGuardian: Due to the implementation, the list order will be conserved. Also it seems I can't cast to set as the objects are non-serializable.

xblanchot-gg (approved):

PR #851, opened 2024-02-06
chore(sca): ignore by default some more paths in sca

hcharaf:


hcharaf:
Hello @gg-mmill ! Do you mind reviewing this MR? (I can't technically assign you yet because I'm not yet a member of the repo)

gg-mmill:
Thanks for the MR, I don't think we should add new parameters for this though, having already the `exclusion_regexes`.

gg-mmill:
Also we may not need the `is_not_excluded_from_sca` function anymore

ggshield/verticals/sca/file_selection.py
```diff
@@ -32,6 +31,10 @@
     ".hypothesis",
 )
 
+SCA_EXCLUSION_REGEXES = {
+    re.compile(r".*" + pattern + ".*") for pattern in SCA_IGNORE_LIST
```
gg-mmill, line 35:
```suggestion
    re.compile(r".*/" + re.escape(pattern) + "/.*") for pattern in SCA_IGNORE_LIST
```
Some improvements over what I initially suggested
- add `/` to separate paths
- use `re.escape` , otherwise `.` in patterns get interpreted as regex patterns for example

Also, since `SCA_IGNORE_LIST` is not used anymore, you could just include the values in the comprehension, and get rid of the variable altogether.

gg-mmill (commented):
Seems good, can still be improved a bit !

gg-mmill (approved):
Good for me, thanks !

PR #852, opened 2024-02-26
chore(sca): ignore by default some more paths in sca

hcharaf:
Changed `SCA_EXCLUSION_REGEXES` pattern & added a test.

### DIFF
--- a/ggshield/verticals/sca/file_selection.py
+++ b/ggshield/verticals/sca/file_selection.py
@@ -16,7 +16,7 @@ from ggshield.utils.git_shell import get_filepaths_from_ref, get_staged_filepath
 
 # List of filepaths to ignore for SCA scans
 SCA_EXCLUSION_REGEXES = {
-    re.compile(r".*/" + re.escape(pattern) + "/.*")
+    re.compile(re.escape(pattern) + "/.*")
     for pattern in (
         "__pycache__",
         ".git",
--- a/tests/unit/verticals/sca/test_file_selection.py
+++ b/tests/unit/verticals/sca/test_file_selection.py
@@ -5,7 +5,9 @@ from typing import Set
 import pytest
 from pygitguardian import GGClient
 
+from ggshield.core.scan.file import get_files_from_paths
 from ggshield.verticals.sca.file_selection import (
+    SCA_EXCLUSION_REGEXES,
     get_all_files_from_sca_paths,
     sca_files_from_git_repo,
 )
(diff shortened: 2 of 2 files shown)


ggshield/verticals/sca/file_selection.py
```diff
@@ -16,9 +16,11 @@
 
 # List of filepaths to ignore for SCA scans
 SCA_EXCLUSION_REGEXES = {
-    re.compile(r".*/" + re.escape(pattern) + "/.*")
+    re.compile(re.escape(pattern) + "/.*")
     for pattern in (
         "__pycache__",
+        "front",
+        "docs",
```
GG-HH, line 23: I don't get why you add `front` and `docs` ? 
In some cases, those will be useful and it should be the user's choice to ignore those.

ggshield/verticals/sca/file_selection.py
```diff
@@ -16,9 +16,11 @@
 
 # List of filepaths to ignore for SCA scans
 SCA_EXCLUSION_REGEXES = {
-    re.compile(r".*/" + re.escape(pattern) + "/.*")
+    re.compile(re.escape(pattern) + "/.*")
```
GG-HH, line 19: Should we add tests ?

GG-HH (changes requested):

tests/unit/verticals/sca/test_file_selection.py
```diff
@@ -39,6 +46,32 @@ def test_get_all_files_from_sca_paths(tmp_path):
...
+            ["sca", "scan", "all", ".", "--verbose"],
+        )
+
+        pattern = r"^ignoring binary file: (.+)$"
+        matches = re.findall(pattern, result.output, re.MULTILINE)
+        matches_paths = [Path(match) for match in matches]
+
+        assert Path("file3.png") in matches_paths
+        assert Path("front/file1.png") in matches_paths
+        assert Path(".git/file2.png") not in matches_paths
```
GG-HH, line 72: Could you test more granularly ? By invoking `get_filepaths` or `is_filepath_excluded` for instance ?
Here the test rely on the pattern of a verbose message which is not very reliable (even if it test the global behavior we want to match)

GG-HH (changes requested):

tests/unit/verticals/sca/test_file_selection.py
```diff
@@ -39,6 +41,34 @@ def test_get_all_files_from_sca_paths(tmp_path):
...
 
+@pytest.mark.parametrize(
+    ("file_path", "expected"),
+    [("front/file1.png", True), (".git/file2.png", False), ("file3.png", True)],
+)
+def test_get_ignored_files(tmp_path, capsysbinary, file_path, expected):
+    """
+    GIVEN a directory
+    WHEN calling sca scan a directory
+    THEN certain ignored binary files should be omitted
```
GG-HH, line 52: This is not what we test here.

"THEN: excluded directory are not inspected"

GG-HH (approved):

PR #854, opened 2024-02-27
chore(deps): upgrade to last py-gitguardian release

fnareoh:


setup.cfg
```diff
@@ -39,15 +39,16 @@ install_requires =
...
     marshmallow>=3.18.0,<3.19.0
     marshmallow-dataclass>=8.5.8,<8.6.0
     oauthlib>=3.2.1,<3.3.0
-    pygitguardian @ git+https://github.com/GitGuardian/py-gitguardian.git@d3634704188b7f95036b8753b4dbc77217ebcf12
+    pygitguardian>=1.14.0,<1.15.0
     pyjwt>=2.6.0,<2.7.0
     python-dotenv>=0.21.0,<0.22.0
     pyyaml>=6.0.1,<6.1
     requests>=2.31.0,<2.32.0
+    urllib3
```
agateau-gg, line 51: Can you pin the urllib3 version, like the others are pinned? This ensures we don't get bad surprise when a not-100%-compatible new version is released.

PR #857, opened 2024-03-04
fix: Fix ignored paths in GGShield

gg-jonathangriffe:
This MR :
- Adds a fixture in auto use to clear the cache from functools.lru_cache between tests
- Makes ggshield get the local config from the root of the repo (if it can find git and is scanning a git repo)
- Corrects the check to return an error when scanning an ignored directory
- Fixes the test `test_retrieve_client_invalid_api_url`

### DIFF
--- a/ggshield/utils/files.py
+++ b/ggshield/utils/files.py
@@ -13,11 +13,17 @@ class UnexpectedDirectoryError(ValueError):
         self.path = path
 
 
-def is_filepath_excluded(
-    filepath: Union[str, Path], exclusion_regexes: Set[re.Pattern]
+def is_path_excluded(
+    path: Union[str, Path], exclusion_regexes: Set[re.Pattern]
 ) -> bool:
-    filepath = Path(filepath)
-    return any(r.search(str(PurePosixPath(filepath))) for r in exclusion_regexes)
+    path = Path(path)
+    if path.is_dir():
+        # The directory exclusion regexes have to end with a slash
+        # To check if path is excluded, we need to add a trailing slash
+        path_string = f"{PurePosixPath(path)}/"
+    else:
+        path_string = str(PurePosixPath(path))
+    return any(r.search(path_string) for r in exclusion_regexes)
 
 
 def get_filepaths(
@@ -54,7 +60,7 @@ def get_filepaths(
                 _targets = path.rglob(r"*")
 
             for file_path in _targets:
-                if not is_filepath_excluded(file_path, exclusion_regexes):
+                if not is_path_excluded(file_path, exclusion_regexes):
                     targets.add(file_path)
     return targets
 
--- a/tests/unit/cmd/iac/test_scan_all.py
+++ b/tests/unit/cmd/iac/test_scan_all.py
@@ -13,6 +13,7 @@ from pytest_mock import MockerFixture
 
 from ggshield.__main__ import cli
 from ggshield.core.errors import ExitCode
+from ggshield.utils.os import cd
 from tests.conftest import IAC_SINGLE_VULNERABILITY
 from tests.repository import Repository
 from tests.unit.conftest import assert_invoke_exited_with, my_vcr
@@ -244,16 +245,15 @@ def test_iac_scan_all_verbose(cli_fs_runner: CliRunner, cli_command) -> None:
 
 @patch("pygitguardian.client.GGClient.iac_directory_scan")
 def test_iac_scan_all_ignored_directory(
-    iac_directory_scan_mock: Mock, cli_fs_runner: CliRunner, cli_command
+    iac_directory_scan_mock: Mock, cli_fs_runner: CliRunner, cli_command, tmp_path: Path
 ) -> None:
     """
     GIVEN a directory which is ignored
     WHEN running the iac scan all command on this directory
     THEN an error is raised
     """
-    path = Path(".")
-    repo = Repository.create(path)
-    iac_file = path / "iac_file.tf"
+    repo = Repository.create(tmp_path)
+    iac_file = tmp_path / "iac_file.tf"
     iac_file.write_text(IAC_SINGLE_VULNERABILITY)
     repo.add(iac_file)
     repo.create_commit()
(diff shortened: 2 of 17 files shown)


tests/unit/core/scan/test_scan_context.py
```diff
@@ -114,7 +114,8 @@ def test_ci_no_env(env, fake_url_repo: Repository) -> None:
     WHEN there is no repository url in the environment
     THEN the remote url is sent by default
     """
-    with mock.patch.dict(os.environ, env, clear=True):
+    environ = {**os.environ.copy(), **env}
```
agateau-gg, line 117: If I am not mistaken, not reusing the existing environment is done on purpose: to ensure the `GITHUB_ACTIONS` environment variable is not set when running on CI. Why did you have to make this change?

ggshield/utils/files.py
```diff
@@ -13,11 +13,15 @@ def __init__(self, path: Path):
...
-def is_filepath_excluded(
-    filepath: Union[str, Path], exclusion_regexes: Set[re.Pattern]
+def is_path_excluded(
+    path: Union[str, Path], exclusion_regexes: Set[re.Pattern]
 ) -> bool:
-    filepath = Path(filepath)
-    return any(r.search(str(PurePosixPath(filepath))) for r in exclusion_regexes)
+    path = Path(path)
+    if path.is_dir():
+        path_string = f"{PurePosixPath(path)}/"
```
agateau-gg, line 21: Following our discussions, can you add a comment here to explain why we want the trailing slash?

tests/unit/cmd/iac/test_scan_all.py
```diff
@@ -244,15 +245,15 @@ def test_iac_scan_all_verbose(cli_fs_runner: CliRunner, cli_command) -> None:
...
+    iac_directory_scan_mock: Mock, cli_fs_runner: CliRunner, cli_command, tmp_path: Path
 ) -> None:
     """
     GIVEN a directory which is ignored
     WHEN running the iac scan all command on this directory
     THEN an error is raised
     """
-    path = Path(".")
-    repo = Repository.create(path)
+    path = Path(tmp_path)
```
agateau-gg, line 255: Since `tmp_path` is already a `Path`, I think you can just drop this line and use `tmp_path` below.

tests/unit/cmd/sca/test_scan.py
```diff
@@ -414,6 +414,54 @@ def test_scan_all_ignored_directory(
...
+    cli_fs_runner: CliRunner,
+    tmp_path: Path,
+    pipfile_lock_with_vuln: str,
+) -> None:
+    """
+    GIVEN a directory which is ignored in the config
+    WHEN running the sca scan all command on this directory
+    THEN an error is raised
+    """
+    path = Path(tmp_path)
```
agateau-gg, line 429: Same here, this line can be removed.

agateau-gg (changes requested):
Minor remarks, this looks good overall!

tests/unit/core/scan/test_scan_context.py
gg-jonathangriffe: This tests failed for windows because it couldn't find git. This test seemingly only succeeded before because it found the git path in the cache.

gg-jonathangriffe: I modified it to remove the GITHUB_ACTIONS env var

agateau-gg: Wow, side-effects everywhere!

Odd that GITHUB_ACTIONS is not set. According to the environment dump from my ci-sandbox repository (https://github.com/agateau-gg/ci-sandbox/actions/runs/7491395928/job/20392507364), it should be there.

There is an autouse fixture in tests/unit/cmd/scan/test_ci.py to remove it, but I would have expected it to not run outside of its module? 🤔

agateau-gg: I spent a bit of time looking at this and I think I understand what is happening: this test checks that when the CI env variables do not contain the necessary info, it falls back to getting the repository URL in some other way, and this way is independent from the CI type.

I think the `GITHUB_ACTIONS` env var is set but it does not change anything *visible*.

Here is a test I did:

I add this code inside the `with` block:

```python
        from ggshield.core.git_hooks.ci import SupportedCI
        print(f"SupportedCI={SupportedCI.from_ci_env()}")
```

and defined `GITHUB_ACTIONS` to `1` in my environment.

I then ran the test (using `-vs` to see the output) before and after your changes.

Before:

```
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env0] SupportedCI=SupportedCI.AZURE
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env1] SupportedCI=SupportedCI.DRONE
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env2] SupportedCI=SupportedCI.GITHUB
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env3] SupportedCI=SupportedCI.GITLAB
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env4] SupportedCI=SupportedCI.CIRCLECI
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env5] SupportedCI=SupportedCI.BITBUCKET
PASSED
```

After:

```
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env0] SupportedCI=SupportedCI.GITHUB
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env1] SupportedCI=SupportedCI.GITHUB
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env2] SupportedCI=SupportedCI.GITHUB
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env3] SupportedCI=SupportedCI.GITLAB
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env4] SupportedCI=SupportedCI.GITHUB
PASSED
tests/unit/core/scan/test_scan_context.py::test_ci_no_env[env5] SupportedCI=SupportedCI.GITHUB
PASSED
```

As you can see, the app does not think it is in the same environment. This looks wrong. Can you try by only copying the $PATH environment variable?

agateau-gg (approved):
Looks good, thanks!

Just need to squash the fixups.

PR #859, opened 2024-03-04
fix(tests): remove deprecated argument from pytest_report_header

gg-jonathangriffe:
This MR fixes `pytest_report_header`. The `startdir` argument was deprecated and removed in the new pytest release.


pierrelalanne (approved):

amascia-gg (approved):
Thanks for the fix!

PR #861, opened 2024-03-18
add ability to customize GITGUARDIAN PRERECEIVE MESSAGE

Renizmy:
Hello,

The main idea is to be able to customize the remediation message For example to set custom links to internal documentation . Maybe it could be better to set in the yaml configuration file, what do you think ?


agateau-gg:
Hi @Renizmy, thanks for this contribution!

This is an interesting idea, I am forwarding it to our product-owners.

agateau-gg:
Hi again!

We like your idea, but using an environment variable makes it difficult to set a multi-line message. We think it would be better to store this message in the configuration file. It could be added as `secret.prereceive-remediation-message`. One can then define the message in the global configuration file (`~/.gitguardian.yaml`) like this:

```yaml
version: 2
secret:
  prereceive-remediation-message: |
    A multiline
    remediation
    message
```

To do so you would need to add a `prereceive_remediation_message` field to the `SecretConfig` class, and then use it from the current code as `config.user_config.secret.prereceive_remediation_message`.

Do you feel like doing this?

Renizmy:
Indeed, the multi-line in an environnement variable is not a good idea .. Thank you for this information, here is the change

ggshield/core/config/user_config.py
```diff
@@ -45,6 +45,10 @@ class SecretConfig(FilteredConfig):
     ignored_matches: List[IgnoredMatch] = field(default_factory=list)
     ignored_paths: Set[str] = field(default_factory=set)
     ignore_known_secrets: bool = False
+    prereceive_remediation_message: str = """ A pre-receive hook set server side prevented you from pushing secrets.
+        Since the secret was detected during the push BUT after the commit, you need to:
+        1. rewrite the git history making sure to replace the secret with its reference (e.g. environment variable).
+        2. push again."""
```
agateau-gg, line 51: nitpick: I think it would be nicer to move the `REMEDIATION_MESSAGE` constant to this file and use it here, because right now this change changed the indentation level of the message (the 8 leading spaces are included in the output).

agateau-gg (changes requested):
Thanks for this new version. I made a request regarding indentation, but I think there is a bit more work to do here, because right now each line of the custom remediation message must start with a two-space indents for the whole error message to look consistent (see the starting spaces on each line of `REMEDIATION_MESSAGE`).

Ideally the `remediation_message()` function from verticals.secret.output.messages should be the one in charge of inserting these indentation spaces, so that callers do not have to hard-code them. Do you have time to look into this? If no, I can take care of it.

Renizmy:
Hi,
What do you think about it ?

ggshield/core/config/user_config.py
```diff
@@ -45,6 +45,10 @@ class SecretConfig(FilteredConfig):
     ignored_matches: List[IgnoredMatch] = field(default_factory=list)
     ignored_paths: Set[str] = field(default_factory=set)
     ignore_known_secrets: bool = False
+    prereceive_remediation_message: str = """A pre-receive hook set server side prevented you from pushing secrets.
+        Since the secret was detected during the push BUT after the commit, you need to:
+        1. rewrite the git history making sure to replace the secret with its reference (e.g. environment variable).
+        2. push again."""
```
agateau-gg, line 51: You still need to remove the 8 leading spaces on line 49 to 51, otherwise they appear in the output:

```suggestion
    prereceive_remediation_message: str = """A pre-receive hook set server side prevented you from pushing secrets.
Since the secret was detected during the push BUT after the commit, you need to:
1. rewrite the git history making sure to replace the secret with its reference (e.g. environment variable).
2. push again."""
```

agateau-gg (changes requested):

ggshield/verticals/secret/output/messages.py
```diff
@@ -1,6 +1,9 @@
 from ggshield.core.text_utils import STYLE, format_text
+from textwrap import TextWrapper
 
-
+def pad_text(text: str, padding: str ='  '):
+   return TextWrapper(initial_indent=padding,subsequent_indent=padding).indent(text)
```
agateau-gg, line 5: `TextWrapper` does not have an `indent()` method. You need to change your code to:
1. `import textwrap` instead of `from textwrap import TextWrapper`
2. replace this line with `textwrap.indent(text, "  ")`

agateau-gg (approved):
I took the liberty to fix the failing test and reformat the code a bit to please the lint step in the CI. We should be good now, thanks a lot for your contribution!

PR #862, opened 2024-03-20
Sign macOS binaries

agateau-gg:
This PR extends `scripts/build-standalone-exe` to produce signed macOS binaries.

## .pkg installers

Signing .zip archives is not supported so the PR switches to producing .pkg installers for macOS.

GGShield files are installed in `/opt/gitguardian/ggshield-$version` and a `ggshield` symlink is created in `/usr/local/bin` so that ggshield can be started without having to modify $PATH.

## How signing works

Signing and notarizing is done using [rcodesign](https://gregoryszorc.com/docs/apple-codesign/0.27.0/index.html), a cross-platform tool to sign and notarize macOS applications.

Signing a .pkg installer requires first signing all executables and libraries we ship, then creating the .pkg, then signing the .pkg itself.

`scripts/build-standalone-exe` does not sign binaries unless called with `--sign`. It then expect some environment variables to be set. They are documented in `doc/dev/standalone-executable.md`. They are defined as secrets in GitHub configuration, and then pulled as environment variables by the CI.

## Testing

The whole signing process can take a long time, so we don't want to do it for every pull request. For now I added a separate workflow file called `sign.yml`, which can be manually triggered from GitHub UI. The next step will be to integrate signing in the release process.

There's a chicken-and-egg problem though: this new workflow cannot be triggered until  `sign.yml` exists in the default branch, so to test the full process the workflow is currently triggered on pull requests too, but I will remove this before merging.

### DIFF
--- a/doc/dev/standalone-executables.md
+++ b/doc/dev/standalone-executables.md
@@ -1,0 +1,55 @@
+# Building standalone executables
+
+## Introduction
+
+`ggshield` is written in Python, and this sometimes makes deployment complicated.
+
+To solve those deployment issues, we provide standalone `ggshield` executables, that do not require a Python interpreter. This documentation explains how these executables are produced.
+
+The process of generating the executable is handled by the `scripts/build-standalone-exe` script. This script runs a series of "steps". It has a default list of steps, but you can tell it to run only specific steps using `scripts/build-standalone-exe step1 step2...`.
+
+All functions in the script starting with `step_` can be used as a step. This means you can get a list of all available steps with: `grep -o '^step_[a-z_]*' scripts/build-standalone-exe`.
+
+## Generating the executable
+
+We use [PyInstaller](https://pyinstaller.org) to generate `ggshield` standalone executable.
+
+## macOS-specific information
+
+For macOS, we produce a .pkg archive. The advantage of this file is that it can be installed by double-clicking on it or by using `sudo installer -pkg path/to/ggshield.pkg -target /`, and `ggshield` is immediately usable after install, without the need to alter `$PATH`.
+
+### Inside the pkg
+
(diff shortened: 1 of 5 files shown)


fnareoh (approved):
I read the doc and skim through the script, it looks good !

doc/dev/standalone-executables.md
```diff
@@ -0,0 +1,55 @@
...
+- `$MACOS_API_KEY_FILE`: Path to a JSON file holding the "App Store Connect API Key". This file is used by `rcodesign` for the notarization step. Follow [`rcodesign` documentation][rcodesign-api-key] to generate one.
+
+Attention: these 3 files should be treated as secrets (even if `$MACOS_P12_FILE` is protected by a password).
+
+[apple-signing-certificate]: https://help.apple.com/xcode/mac/current/#/dev154b28f09
+[rcodesign-api-key]: https://gregoryszorc.com/docs/apple-codesign/0.27.0/apple_codesign_getting_started.html#obtaining-an-app-store-connect-api-key
+
+#### Signing implementation details
+
+Although PyInstaller supports signing, it did not work at the time we tried it, so we use [rcodesign][] to do so.
```
pierrelalanne, line 49: Is there a missing link here ?
```suggestion
Although PyInstaller supports signing, it did not work at the time we tried it, so we use [rcodesign][] to do so.
```

agateau-gg: No: `[rcodesign][]` is a shortcut for `[rcodesign][rcodesign]`, and there is an `[rcodesign]: https://...` line below. See the doc rendered by GitHub: https://github.com/GitGuardian/ggshield/blob/d80be498a62b94e1c63759a17797be5badba94a1/doc/dev/standalone-executables.md.

PR #866, opened 2024-04-03
Do not crash if a commit has no author

agateau-gg:
## Context

There are some strange commits around there. [This one](https://github.com/awslabs/amazon-eks-ami/commit/f40da388.patch) for example, made ggshield crash because its Author field is mostly empty.

## What has been done

Make commit parser accept empty authors, added unit-test.

## Testing

- Unit-tests pass
- `ggshield secret scan repo https://github.com/awslabs/amazon-eks-ami.git` no longer fails


salome-voltz (approved):
Looks good, I just have a question: would it be possible for a commit to have no date, similarly to malformed commits with no author?

agateau-gg:
> Looks good, I just have a question: would it be possible for a commit to have no date, similarly to malformed commits with no author?

I hope not 😅 .

The way to produce a commit with no author is to actually set the git config value `user.name` to an empty string. The same thing can be done with `user.email` (which was already handled).

One can't do the same (AFAIK) for the date. It's possible to change the commit and author date with the `$GIT_COMMITTER_DATE` and `$GIT_AUTHOR_DATE` environment variables, so I just experimented messing with them: if they are empty git ignores them and uses the current date. If they contain invalid values it fails with a `fatal: invalid date format` message. We should be good 🤞🏻.

PR #870, opened 2024-04-05
Deprecate dash-separated config keys

agateau-gg:
## Context

This PR makes deprecate dash-separated config keys and suggest using underscore-separate keys instead. This means one must use `show_secrets` and not `show-secrets`.

Using dash syntax still works but ggshield prints a warning message when it finds them and no longer makes any effort to refer to a key in the syntax it was written.

## Output

Here is what the output looks like with a dash-separated key in the "global" config and 3 other dash-separated keys in the current directory (~/src/ggshield-alt in my example):

```
$ ggshield api-status
/home/agateau/.gitguardian.yaml: Config key show-secrets is deprecated, use show_secrets instead.
/home/agateau/src/ggshield-alt/.gitguardian.yaml: Config key ignored-matches is deprecated, use ignored_matches instead.
/home/agateau/src/ggshield-alt/.gitguardian.yaml: Config key ignored-paths is deprecated, use ignored_paths instead.
/home/agateau/src/ggshield-alt/.gitguardian.yaml: Config key ignored-policies is deprecated, use ignored_policies instead.
Status: healthy
App version: v2.62.1
Secrets engine version: 2.109.0
```

## Review

### DIFF
--- a/.gitguardian.example.yml
+++ b/.gitguardian.example.yml
@@ -4,7 +4,7 @@ version: 2
 # Set to true if the desired exit code for the CLI is always 0, otherwise the
 # exit code will be 1 if incidents are found.
 # The environment variable GITGUARDIAN_EXIT_ZERO=true can also be used toggle this behavior.
-exit-zero: false # default: false
+exit_zero: false # default: false
 
 verbose: false # default: false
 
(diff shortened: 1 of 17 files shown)


salome-voltz (approved):
Looks good :+1:

GG-HH:
Hello @agateau-gg , should we modify the .gitguardian.example.yaml too ?

agateau-gg:
> Hello @agateau-gg , should we modify the .gitguardian.example.yaml too ?

Good point! Just fixed it, and found a few other places to fix while checking.

fnareoh:
This MR looks good to me but I have a question with regard to migrating the configs and replacing `-` with `_` => If you have a config with the same variable but one with `-` and one with `_` [you will silently take the one with the dashes ](https://github.com/GitGuardian/ggshield/blob/3aa7e1a31806ddbfc45a59f714effbd8268eb281/ggshield/core/config/utils.py#L27), maybe you could issue a warning to the user in this (unlikely) case ?

PR #871, opened 2024-04-10
Release signed binaries

agateau-gg:
## Context

This PR extends our CI so that signed macOS binaries are built on each release.

## What has been done

Since testing CI code which only triggers on tags is not really possible, I used the following workarounds:

1. The existing sign.yml workflow has been turned into a more generic build_release_assets.yml workflow. This workflow is called by the tag.yml workflow. It can also be called manually.

2. I added a faketag.yml workflow. This is a copy of tag.yml except it triggers on pull requests and does not publish anything: each publish step is replaced by a step listing the downloaded artifacts which would be published. This allow verifies the changes to tag.yml as much as possible. You can see a run of it [here](https://github.com/GitGuardian/ggshield/actions/runs/8629377759).
  This faketag.yml will be removed before merging the PR.

## Review

First commit contains the real work. Second commit just adds faketag.yml and can mostly be ignored (apart from verifying changes made to tag.yml are also present in faketag.yml). I will revert that commit before merging.

### DIFF
--- a/.github/workflows/build_release_assets.yml
+++ b/.github/workflows/build_release_assets.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build_packages:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -43,9 +43,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-latest
+          - os: ubuntu-22.04
           - os: windows-2022
-          - os: macos-latest
+          - os: macos-13
             arch: x86_64
             sha256sum:
               python: 6378dfd22f58bb553ddb02be28304d739cd730c1f95c15c74955c923a1bc3d6a
--- a/.github/workflows/build_release_assets.yml
+++ b/.github/workflows/build_release_assets.yml
@@ -1,18 +1,49 @@
-name: Sign
+name: Build release assets
 
-# This is an on-demand workflow to test signing
+# This workflow is used by the tag workflow to build all release assets. It
+# can also be triggered manually.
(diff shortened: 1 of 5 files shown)


fnareoh (approved):
Looks good but I"m not sure I understand why you end up with [two macos binaries here](https://github.com/GitGuardian/ggshield/actions/runs/8629377759)  what is "macos-14" ?

agateau-gg:
> Looks good but I"m not sure I understand why you end up with [two macos binaries here](https://github.com/GitGuardian/ggshield/actions/runs/8629377759) what is "macos-14" ?

GitHub terminology is confusing:

- macos-14 is for ARM-based macOS, it uses macOS 14
- macos-latest is for Intel-based macOS

At least that was my understanding until I went back to https://github.com/actions/runner-images, and noticed macos-latest is now the same as macos-14! That changed on [April 2](https://github.com/actions/runner-images/commit/da5ebb36223a06890092a423a0466b28ea2de595.patch).

That's my punishment for not pinning runner versions I guess 😒. Going to pin this runner (and the Ubuntu one).

salome-voltz:
Looks good to me, I'm just not sure to understand how to trigger the Build release assets workflow manually ?

salome-voltz (approved):

agateau-gg:
> Looks good to me, I'm just not sure to understand how to trigger the Build release assets workflow manually ?

The place to trigger workflows is https://github.com/GitGuardian/ggshield/actions, but there is a trick: you can't trigger it right now because I renamed it, and the workflow must exist in the default branch to be able to trigger it 😞.

Right now you can try to trigger the [sign workflow](https://github.com/GitGuardian/ggshield/actions/workflows/sign.yml), the old name for build_release_assets, which is in default branch.

PR #877, opened 2024-04-16
fix(sca): fixing functional tests after remote data being modified

xblanchot-gg:
New vulnerability on packages used for SCA testing were declared, making the tests fail because of the test on the number of found vulnerabilities.
Modifying functional test so that they're not dependant anymore on the number of found vulnerabilities (assuming they will remain greater than 1).


agateau-gg (approved):
Thanks for the fix!

PR #884, opened 2024-04-25
feat(sca): allow ignoring vulnerabilities by CVE id

gg-jonathangriffe:
## Context

We want to allow ignoring vulnerabilities by CVE id in the config.

## What has been done

Ignore vulnerabilities identifier validation now additionally allows identifiers following the CVE id pattern.

## PR check list

### DIFF
--- a/ggshield/core/config/user_config.py
+++ b/ggshield/core/config/user_config.py
@@ -30,6 +30,7 @@ CURRENT_CONFIG_VERSION = 2
 _IGNORE_KNOWN_SECRETS_KEY = "ignore_known_secrets"
 
 GHSA_ID_PATTERN = re.compile("GHSA(-[a-zA-Z0-9]{4}){3}")
+CVE_ID_PATTERN = re.compile(r"CVE-\d{4}-\d{4,}")
 POLICY_ID_PATTERN = re.compile("GG_IAC_[0-9]{4}")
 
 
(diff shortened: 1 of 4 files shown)


ggshield/core/config/user_config.py
```diff
@@ -30,6 +30,7 @@
 _IGNORE_KNOWN_SECRETS_KEY = "ignore_known_secrets"
 
 GHSA_ID_PATTERN = re.compile("GHSA(-[a-zA-Z0-9]{4}){3}")
+CVE_ID_PATTERN = re.compile("CVE-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4,}")
```
amascia-gg, line 33: according to their [doc](https://cve.mitre.org/cve/identifiers/syntaxchange.html), the CVE pattern will only have digits so something like `re.compile(r"CVE-\d{4}-\d{4,}")` should do, doesn't it?

ggshield/core/config/user_config.py
```diff
@@ -203,10 +204,15 @@ def is_ghsa_valid(ghsa_id: str) -> bool:
     return bool(GHSA_ID_PATTERN.fullmatch(ghsa_id))
 
 
+def is_cve_id_valid(cve_id: str) -> bool:
+    return bool(CVE_ID_PATTERN.fullmatch(cve_id))
+
+
 def validate_vuln_identifier(value: str):
-    if not is_ghsa_valid(value):
+    if not is_ghsa_valid(value) and not is_cve_id_valid(value):
```
amascia-gg, line 212: nit

```suggestion
    if not (is_ghsa_valid(value) or is_cve_id_valid(value)):
```

I'm not a big fan of `not ... and not ...` but doesn't matter

tests/unit/cmd/sca/test_scan.py
```diff
@@ -625,6 +630,9 @@ def test_sca_scan_subdir_with_ignored_vuln(
             ],
         )
 
+        ignored_vulns = sca_scan_params_mock.call_args.kwargs["ignored_vulnerabilities"]
+        assert [i.identifier for i in ignored_vulns] == ["GHSA-rrm6-wvj7-cwh2"]
```
amascia-gg, line 634: don't you want to try adding CVE? I don't see how this change is related to the handling of CVEs

amascia-gg (commented):
thanks for the change!

tests/unit/cmd/sca/test_scan.py
gg-jonathangriffe: This change is not directly related to the handling of CVEs, I changed this to put it in a separate commit.
This test needed improvement since the cassette made it a false positive, as the Gitguardian API response will never have any incident. I added this to test we at least send the correct option to ignore vulnerabilities.

amascia-gg (approved):

PR #887, opened 2024-05-15
feat(secret): Add option --use-gitignore to command secret scan path

salome-voltz:
## Context

When running a scan at the root of a git checkout with `ggshield secret scan path`, files that are not part of the repository are scanned because the .gitignore is not honored. 
This creates irrelevant results. 
  
## What has been done

`secret scan path` now supports a `--use-gitignore` option to honor .gitignore files and related files (.git/info/exclude, $HOME/.config/git/ignore).

## Validation
- Clone a repo 
- Checkout a branch
- You can scan the current state of the checkout with `ggshield secret scan path -r --use-gitignore .`
- Modify some file
- You can scan the modified files (even unstaged) with the same command 

## PR check list

### DIFF
--- a/ggshield/utils/files.py
+++ b/ggshield/utils/files.py
@@ -1,9 +1,33 @@
 import re
+from enum import Enum, auto
 from pathlib import Path, PurePosixPath
 from typing import List, Set, Union
 
 from ggshield.utils._binary_extensions import BINARY_EXTENSIONS
-from ggshield.utils.git_shell import get_filepaths_from_ref, git_ls, is_git_dir
+from ggshield.utils.git_shell import (
+    get_filepaths_from_ref,
+    git_ls,
+    git_ls_unstaged,
+    is_git_dir,
+)
+
+
+class ListFilesMode(Enum):
+    """
+    Control `get_filepaths()` behavior:
+
+    - FILES_ONLY: list specified paths. Expect them to be plain files, raise an exception if one of them is not.
+    - ALL: list all specified paths. If one of the path is a directory, list all its paths recursively.
+    - ALL_BUT_GITIGNORED: like ALL, except those ignored by git (listed in .gitignore).
+    - GIT_COMMITTED_OR_STAGED: list all committed files and all staged files.
+    - GIT_COMMITTED: list only committed files.
+    """
+
+    FILES_ONLY = auto()
+    GIT_COMMITTED = auto()
+    GIT_COMMITTED_OR_STAGED = auto()
+    ALL_BUT_GITIGNORED = auto()
+    ALL = auto()
 
 
 class UnexpectedDirectoryError(ValueError):
@@ -29,15 +53,12 @@ def is_path_excluded(
 def get_filepaths(
     paths: List[Path],
     exclusion_regexes: Set[re.Pattern],
-    recursive: bool,
-    ignore_git: bool,
-    ignore_git_staged: bool = False,
+    list_files_mode: ListFilesMode,
 ) -> Set[Path]:
     """
     Retrieve the filepaths from the command.
 
     :param paths: List of file/dir paths from the command
-    :param recursive: Recursive option
     :param ignore_git: Ignore that the folder is a git repository
     :raise: UnexpectedDirectoryError if directory is given without --recursive option
     """
@@ -46,16 +67,19 @@ def get_filepaths(
         if path.is_file():
             targets.add(path)
         elif path.is_dir():
-            if not recursive:
+            if list_files_mode == ListFilesMode.FILES_ONLY:
                 raise UnexpectedDirectoryError(path)
 
-            if not ignore_git and is_git_dir(path):
+            _targets = set()
+            if list_files_mode != ListFilesMode.ALL and is_git_dir(path):
                 target_filepaths = (
                     get_filepaths_from_ref("HEAD", wd=path)
-                    if ignore_git_staged
+                    if list_files_mode == ListFilesMode.GIT_COMMITTED
                     else git_ls(path)
                 )
                 _targets = {path / x for x in target_filepaths}
+                if list_files_mode == ListFilesMode.ALL_BUT_GITIGNORED:
+                    _targets.update({path / x for x in git_ls_unstaged(path)})
             else:
                 _targets = path.rglob(r"*")
 
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -234,6 +234,32 @@ def git_ls(wd: Optional[Union[str, Path]] = None) -> List[str]:
     return git(cmd, timeout=600, cwd=wd).split("\n")
 
 
+def _get_submodules_paths(wd: Optional[Union[str, Path]] = None) -> List[str]:
+    cmd = ["submodule", "status"]
+    return [
+        submodule.split()[1]
+        for submodule in git(cmd, timeout=600, cwd=wd).splitlines()
+        if submodule
+    ]
+
+
+def git_ls_unstaged(wd: Optional[Union[str, Path]] = None) -> List[str]:
+    # git command to get list of unstaged files in repo
+    cmd = ["ls-files", "--others", "--exclude-standard"]
+    unstaged_files = git(cmd, timeout=600, cwd=wd).splitlines()
+
+    # --recurse-submodules is not compatible with --others so we
+    # need to iterate over submodules to get the list of unstaged files
+    for submodule in _get_submodules_paths(wd=wd):
+        unstaged_files.extend(
+            str(Path(submodule) / unstaged_file)
+            for unstaged_file in git_ls_unstaged(
+                wd=wd / Path(submodule) if wd else Path(submodule)
+            )
+        )
+    return unstaged_files
+
+
 def is_valid_git_commit_ref(ref: str, wd: Optional[Union[str, Path]] = None) -> bool:
     """
     Check if a reference is valid and can be resolved to a commit
--- a/tests/unit/cmd/scan/test_path.py
+++ b/tests/unit/cmd/scan/test_path.py
@@ -365,8 +365,73 @@ class TestScanDirectory:
         assert all(
             string in result.output
             for string in ["Do you want to continue", "not_committed"]
-        ), "not_committed files not should have been ignored"
+        ), "not_committed files should not have been ignored"
         assert result.exception is None
+
+    def test_scan_path_use_gitignore(
+        self, tmp_path: Path, cli_runner: CliRunner
+    ) -> None:
+        """
+        GIVEN a directory containing a gitignore
+        WHEN executing a scan with --use-gitignore option
+        THEN ggshield should not scan the ignored files
+        """
+
+        # We can't use cassettes for this test because we want to scan 2 files.
+        # Depending on the order of the files, the request and response will be different.
+
+        local_repo = Repository.create(tmp_path)
+
+        # files in repo
+        ignored_secret = local_repo.path / "ignored_file_secret"
+        found_secret = local_repo.path / "found_file_secret"
+        ignored_secret.write_text(VALID_SECRET_PATCH)
+        found_secret.write_text(VALID_SECRET_PATCH)
+
+        gitignore = local_repo.path / ".gitignore"
+        gitignore.write_text("ignored_file_secret")
+
+        # Submodule
+        submodule_path = tmp_path / "submodule_repo"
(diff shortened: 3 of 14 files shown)


tests/unit/cmd/scan/test_path.py
```diff
@@ -368,6 +368,26 @@ def test_scan_path_should_detect_non_git_files(self, cli_fs_runner):
...
+        GIVEN a directory containing a gitignore
+        WHEN executing a scan with --use-gitignore option
+        THEN ggshield should not scan the ignored files
+        """
+        os.makedirs("repo_git_ignore")
+        Path("repo_git_ignore/secret.js").write_text(VALID_SECRET_PATCH)
+        Path("repo_git_ignore/.gitignore").write_text("secret.js")
+        os.system("git init")
+        os.system("git add .")
+        os.system("git commit -m 'initial commit'")
```
agateau-gg, line 383: This is going to create a repository with what is in the current directory, so the commit is likely to be slow. Better use the `tmp_path` fixture to get a temporary directory to work in. Furthermore, instead of using `os.system()` to run git commands, you can use the `Repository` helper class.

You can find an example of this in `test_scan_context_repository()` in this file.

tests/unit/cmd/scan/test_path.py
```diff
@@ -368,6 +368,26 @@ def test_scan_path_should_detect_non_git_files(self, cli_fs_runner):
...
 
+    @my_vcr.use_cassette()
+    def test_scan_path_use_gitignore(self, cli_fs_runner: CliRunner) -> None:
+        """
+        GIVEN a directory containing a gitignore
+        WHEN executing a scan with --use-gitignore option
+        THEN ggshield should not scan the ignored files
+        """
+        os.makedirs("repo_git_ignore")
+        Path("repo_git_ignore/secret.js").write_text(VALID_SECRET_PATCH)
```
agateau-gg, line 379: It would be a better test if the repository also contained one secret to be found. This way the test proves that `--use-gitignore` does not cause all files to be ignored 😉.

agateau-gg (changes requested):

ggshield/utils/git_shell.py
```diff
@@ -234,6 +234,11 @@ def git_ls(wd: Optional[Union[str, Path]] = None) -> List[str]:
     return git(cmd, timeout=600, cwd=wd).split("\n")
 
 
+def git_ls_unstaged(wd: Optional[Union[str, Path]] = None) -> List[str]:
+    cmd = ["ls-files", "--other", "--exclude-standard"]
```
agateau-gg, line 238: As you said, this command does not work with submodules. This is a problem because users are going to believe we scan unknown files inside submodules, giving them a false sense of security. I am afraid we have to implement submodules recursion ourselves.

ggshield/utils/files.py
```diff
@@ -32,6 +37,7 @@ def get_filepaths(
     recursive: bool,
     ignore_git: bool,
     ignore_git_staged: bool = False,
+    include_git_unstaged: bool = False,
```
agateau-gg, line 40: We should avoid introducing a forth boolean argument: 3 was already too many.

I suggest we introduce a FileListMode enum to replace them. It would have the following possible values:

| Value                   | recursive | ignore_git | ignore_git_staged | include_git_unstaged |
| ----------------------- | --------- | ---------- | ----------------- | -------------------- |
| FILES_ONLY              |           |            |                   |                      |
| GIT_COMMITTED           | x         |            | x                 |                      |
| GIT_COMMITTED_OR_STAGED | x         |            |                   |                      |
| ALL_BUT_GITIGNORED      | x         |            |                   | x                    |
| ALL                     | x         | x          |                   |                      |

The values do not include all possible flag combinations, but I think the others do not make sense. What do you think?

`GIT_COMMITTED_OR_STAGED` is a bit long, shorter suggestions are welcome :)

agateau-gg (changes requested):

ggshield/utils/git_shell.py
salome-voltz: This has been a bit more tricky than expected, but it should work as expected now. I'm more worried about the robustness of the code now.

ggshield/utils/files.py
```diff
@@ -35,9 +44,7 @@ def get_filepaths(
     paths: List[Path],
     exclusion_regexes: Set[re.Pattern],
     recursive: bool,
```
agateau-gg, line 46: The enum replaces `recursive` too: `FILES_ONLY` is the equivalent of `recursive=False`.

ggshield/utils/files.py
```diff
@@ -55,17 +62,19 @@ def get_filepaths(
...
-                    else git_ls(path)
-                )
-                _targets = {path / x for x in target_filepaths}
-                if include_git_unstaged:
-                    _targets.update({path / x for x in git_ls_unstaged(path)})
-            else:
-                _targets = path.rglob(r"*")
+            _targets = set()
+            if list_files_mode != ListFilesMode.FILES_ONLY:
+                if is_git_dir(path) and not list_files_mode == ListFilesMode.ALL:
```
agateau-gg, line 67: I'd rewrite this as:

```
if list_files_mode != ListFilesMode.ALL and is_git_dir(path):
```

The reason for this is that it removes one place where git is required. When installing ggshield in an environment without git (for example a docker container), we should be able to run `ggshield secret scan path`.

ggshield/utils/files.py
```diff
@@ -11,6 +12,14 @@
 )
 
 
+class ListFilesMode(Enum):
```
agateau-gg, line 15: This enum needs a docstring, something like this:

```
Control `get_filepaths()` behavior:

- FILES_ONLY: list specified paths. Expect them to be plain files, raises an exception if one of them is not.
- ALL: list all specified paths. If one of them specified is a directory, list all its paths. 
- ALL_BUT_GITIGNORED: like ALL, except those ignored by git
- GIT_COMMITTED_OR_STAGED: list all committed files and all staged files
- GIT_COMMITTED: list only committed files
```

ggshield/utils/git_shell.py
```diff
@@ -234,9 +234,31 @@ def git_ls(wd: Optional[Union[str, Path]] = None) -> List[str]:
...
+    unstaged_files = git(cmd, timeout=600, cwd=wd).split("\n")
+
+    if wd is None:
+        wd = Path(".")
+    if _has_submodules(wd):
+        # --recurse-submodules is not compatible with --others so we need to
+        # get the list of unstaged files in each submodule
+        submodule_path = "s~^~$sm_path/~"
+        submodule_files = (
+            f"git ls-files -o --exclude-standard --full-name | sed {submodule_path}"
```
agateau-gg, line 255: This is not going to work on Windows, where `sed` is not available. Using the shell like this can also be dangerous (though it looks rather safe here).

I suggest parsing the output of `git submodule status` to get the path to all submodules and recalling `git_ls_unstaged` recursively on the new paths.

This can remove the need for `_has_submodules()`, which mostly does this. The code would be something like this:

```python
def git_ls_unstaged():
    # call ls-files
    for path in get_submodule_paths():
        # call git_ls_unstaged(path)
```

agateau-gg (changes requested):

ggshield/utils/git_shell.py
salome-voltz: I just had to handle the path of the files to ensure they will be found during the scanning process.

ggshield/utils/git_shell.py
```diff
@@ -234,6 +234,33 @@ def git_ls(wd: Optional[Union[str, Path]] = None) -> List[str]:
...
+def _get_submodules_paths(wd: Optional[Union[str, Path]] = None) -> List[str]:
+    find_path = "git rev-parse --show-toplevel"
+    cmd = ["submodule", "--quiet", "foreach", find_path]
+    submodules = [
+        submodule
+        for submodule in git(cmd, timeout=600, cwd=wd).split("\n")
+        if submodule
+    ]
+
+    return submodules
```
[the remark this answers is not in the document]
agateau-gg: Also, using `.splitlines()` instead of `split("\n")` removes the need for removing trailing empty lines:

```
>>> a="fff\nbbb\n"
>>> a.split("\n")
['fff', 'bbb', '']
>>> a.splitlines()
['fff', 'bbb']
```

agateau-gg (changes requested):

tests/unit/cmd/scan/test_path.py
```diff
@@ -365,9 +365,48 @@ def test_scan_path_should_detect_non_git_files(self, cli_fs_runner):
...
+        self, tmp_path: Path, cli_runner: CliRunner
+    ) -> None:
+        """
+        GIVEN a directory containing a gitignore
+        WHEN executing a scan with --use-gitignore option
+        THEN ggshield should not scan the ignored files
+        """
+
+        # We can't use cassettes for this test because we want to scan 2 files.
+        # Depending on the order of the files, the request and response will be different.
```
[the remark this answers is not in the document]
salome-voltz: The test doesn't hit the network, I removed the "--yes" option of the command. 
This was an intermediate solution ensuring that we plan to scan the correct number of files, which indirectly ensures that we ignore some files... This test is not ideal I agree but  I don't see the point of scanning the files if we create the response ourselves. 
Am I misunderstanding something?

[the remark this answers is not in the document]
agateau-gg: Oh right. I was about to say this looks fragile, but then I realized the test right above yours does exactly that... let's keep it like this then.

tests/unit/utils/test_git_shell.py
```diff
@@ -375,3 +376,31 @@ def test_tar_from_ref_and_filepaths(tmp_path):
...
+
+    # WHEN listing unstaged files
+    unstaged_files = git_ls_unstaged(wd=repo.path)
+
+    # THEN unstaged files in repo and submodule are returned
+    # as relative OR absolute paths
+    assert all(
+        any(item in unstaged_files for item in [str(path), path.name])
+        for path in [repo_file, submodule_file]
+    ), "Not all paths are found in unstaged_files"
```
agateau-gg, line 406: This got me trying `git_ls_unstaged()` manually and I think it would be more consistent if it had the same behavior as `git_ls()`: all returned paths should be relative to the directory passed as its argument.

I ask this because I am worried this could bite us back in the future.

agateau-gg (changes requested):

tests/unit/utils/test_git_shell.py
salome-voltz: There was a bit of path reconstruction to do, but I think we're good now!

tests/unit/utils/test_git_shell.py
```diff
@@ -375,3 +376,31 @@ def test_tar_from_ref_and_filepaths(tmp_path):
...
+
+    # WHEN listing unstaged files
+    unstaged_files = git_ls_unstaged(wd=repo.path)
+
+    # THEN unstaged files in repo and submodule are returned
+    # as relative OR absolute paths
+    assert all(
+        file.name in [Path(path).name for path in unstaged_files]
+        for file in [repo_file, submodule_file]
+    ), "Not all paths are found in unstaged_files"
```
agateau-gg, line 406: This only tests the file name, not the file path. But `git_ls_unstaged()` implementation is good now. I rewrote this part a bit and it still passes.

```suggestion
    # THEN unstaged files in repo and submodule are returned
    # as relative to repo.path
    expected_paths = {x.relative_to(repo.path) for x in (repo_file, submodule_file)}
    assert {Path(x) for x in unstaged_files} == expected_paths
```

tests/unit/cmd/scan/test_path.py
```diff
@@ -365,9 +365,48 @@ def test_scan_path_should_detect_non_git_files(self, cli_fs_runner):
...
+                [
+                    "secret",
+                    "scan",
+                    "path",
+                    "--recursive",
+                    "--use-gitignore",
+                    str(local_repo.path),
+                ],
+            )
+        print(result.output)
```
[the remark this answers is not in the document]
agateau-gg: That `print()` is tenacious 😉

agateau-gg (changes requested):

tests/unit/utils/test_git_shell.py
salome-voltz: `relative_to()` would have saved me some time, trying to work with `\` in path :laughing:

agateau-gg (approved):
Looks good!

PR #888, opened 2024-05-16
Minor refactoring of secret output code

agateau-gg:
## Context

While working on the SARIF output handler I came upon some code in the output which could be cleaner, so I made some minor refactoring.

## What has been done

All the changes are detailed in individual commit. Some are very benign. The most important one is the renaming of MatchIndices to MatchSpan and the associated changes in it.

## Validation

Tests still pass.

## PR check list


fnareoh (approved):
Thanks for the refacto! Just for my knowledge, is it standard practice to have Enum class be uppercase ?

agateau-gg:
> Thanks for the refacto! Just for my knowledge, is it standard practice to have Enum class be uppercase ?

There are no entries for them in PEP-8, but if you look at the official Enum doc (for example the [Enum HOWTO](https://docs.python.org/3/howto/enum.html)), values are all uppercase, and the howto says:

> Case of Enum Members
>
> Because Enums are used to represent constants, and to help avoid issues with name clashes between mixin-class methods/attributes and enum names, we strongly recommend using UPPER_CASE names for members, and will be using that style in our examples.

So I'd say yes.

PR #890, opened 2024-05-17
refactor: move make matches to results

fnareoh:
## Context

This MR starts refactoring the output of secret scanning result towards not having to save all the file content for each result.

## What has been done
This MR combines both version of `make_matches` from the output handlers files to a method `enrich_matches` in Result.

## Validation 

Test still pass !

## PR check list


ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -19,6 +22,41 @@ class Result(NamedTuple):
...
+        is_patch = self.filemode != Filemode.FILE
+        lines = get_lines_from_content(content, self.filemode)
+        for policy_break in self.scan.policy_breaks:
+            res = []
+            for match in policy_break.matches:
+                if match.index_start is None or match.index_end is None:
+                    res.append(match)
+                    continue
+                index_minified_start = match.index_start
+                index_minified_end = match.index_end
```
agateau-gg, line 36: What are those two? They don't exist in ExtendedMatch.

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -105,15 +104,16 @@ def process_result(
         # policy breaks and matches are modified in the functions leak_dictionary_by_ignore_sha and censor_content.
         # Previously process_result was executed only once, so it did not create any issue.
         # In the future we could rework those functions such that they do not change what is in the result.
-        policy_breaks = deepcopy(result.scan.policy_breaks)
+        result = deepcopy(_result)
```
agateau-gg, line 107: I think you can overwrite `result` with a deepcopy of itself. This way you don't need to rename `result` to `_result` in the arguments (which looks a bit odd, and could break if using named arguments)

ggshield/core/filter.py
```diff
@@ -211,3 +211,35 @@ def censor_content(content: str, policy_breaks: List[PolicyBreak]) -> str:
                 )
             )
     return content
+
+
+def censor_line_content(content: str, policy_breaks: List[PolicyBreak]) -> str:
```
agateau-gg, line 216: It looks like this code is not used. Can you remove it, or is something missing?

agateau-gg (changes requested):
Some remarks, but I like where this is going :)

ggshield/verticals/secret/secret_scan_collection.py
fnareoh: Ouf this is just a remain of another attempt a working refacto...

ggshield/verticals/secret/output/secret_text_output_handler.py
fnareoh: I changed the line anyway because in some cases it was not possible to make a deep copy of result but I'm not sure my current version truly works :confused:

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -19,6 +22,40 @@ class Result(NamedTuple):
...
+                        index_end=span.column_index_end,
+                        line_start=line_index_start,
+                        line_end=line_index_end,
+                        pre_line_start=line_start.pre_index,
+                        post_line_start=line_start.post_index,
+                        pre_line_end=line_end.pre_index,
+                        post_line_end=line_end.post_index,
+                    )
+                )
+            policy_break.matches = res
```
agateau-gg, line 57: If you introduce a `def enrich_match(Match, List[Lines]) -> Match:` function then you can simplify these loops to something like this:

```python
for policy_break in self.scan.policy_breaks:
   policy_break.matches = [enrich_match(lines, m) for m in policy_break.matches]
```

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -105,15 +104,16 @@ def process_result(
...
-        sha_dict = leak_dictionary_by_ignore_sha(policy_breaks)
+        sha_dict = leak_dictionary_by_ignore_sha(result.scan.policy_breaks)
 
         if self.show_secrets:
             content = result.content
         else:
-            content = censor_content(result.content, policy_breaks)
+            content = censor_content(result.content, result.scan.policy_breaks)
 
+        result.enrich_matches()  # important to keep this call after censor content
```
agateau-gg, line 116: Why is it important to keep this after `censor_content()`? It's going to be a problem to move this call earlier, no?

agateau-gg (changes requested):

ggshield/verticals/secret/output/secret_text_output_handler.py
fnareoh: for now `censor_content` is based on the ids with regard to the beginning of the file, [which enrich_matches overrides here](https://github.com/GitGuardian/ggshield/pull/890/files#diff-9353b47388d3c5cce7ee2ea9b8caa644d858c1167e4d6d65370a339e3326bba1R47) here, this is something that I want to change in the next refacto MR but for this one I wanted to start reasonably small...

agateau-gg: Ah OK, makes sense :+1:.

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -105,15 +104,16 @@ def process_result(
         # policy breaks and matches are modified in the functions leak_dictionary_by_ignore_sha and censor_content.
         # Previously process_result was executed only once, so it did not create any issue.
         # In the future we could rework those functions such that they do not change what is in the result.
-        policy_breaks = deepcopy(result.scan.policy_breaks)
+        result.scan.policy_breaks = deepcopy(result.scan.policy_breaks)
```
agateau-gg, line 107: I am not sure this call to deepcopy() is useful. The goal of the deepcopy() call in the main branch was to ensure we do not modify the policy breaks inside the `result` argument, but this version still modifies it.

In a previous iteration of this branch the code made a deepcopy of the whole `result` instance, which sounds like what we want. What was the issue with that?

agateau-gg (changes requested):

ggshield/verticals/secret/output/secret_text_output_handler.py
fnareoh: Making a deep copy of result breaks the functional test `tests/functional/secret/test_scan_docker.py::test_scan_docker_uses_cache` with the error: `Error: cannot pickle 'ExFileObject' object`

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -19,6 +22,22 @@ class Result(NamedTuple):
     file: Scannable  # filename that was scanned
     scan: ScanResult  # Result of content scan
 
+    def enrich_matches(self, lines: List[Line] = []) -> None:
```
agateau-gg, line 25: Using a mutable object as a default value is a bad practice: if the function ever modifies `lines`, the default will be modified for future calls.

It's safer to change lines to be `Optional[List[Line]] = None` and use `if not lines:` (which is more Pythonic than `if len(lines) == 0:` anyway).

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -105,15 +104,16 @@ def process_result(
...
+        sha_dict = leak_dictionary_by_ignore_sha(result.scan.policy_breaks)
 
         if self.show_secrets:
             content = result.content
         else:
-            content = censor_content(result.content, policy_breaks)
+            content = censor_content(result.content, result.scan.policy_breaks)
 
+        result.enrich_matches()  # important to keep this call after censor content
         lines = get_lines_from_content(content, result.filemode)
```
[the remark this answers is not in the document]
agateau-gg: Sounds good :+1:.

ggshield/verticals/secret/output/secret_text_output_handler.py
agateau-gg: I see. I think this can be fixed with this:

```python
result = Result(file=result.file, scan=deepcopy(result.scan))
```

This way we get a new result.scan instance which we can modify, but we reuse the existing result.file, which cannot be deepcopied.

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -27,12 +25,13 @@
 # MAX_SECRET_SIZE controls the max length of |-----| under a secret
 # avoids occupying a lot of space in a CI terminal.
 MAX_SECRET_SIZE = 80
+# The number of lines to display before and after a secret in the patch
+NB_CONTEXT_LINES = 3
```
agateau-gg, line 29: +1 for the simplification here :)

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -27,12 +25,13 @@
...
 
 
 class SecretTextOutputHandler(SecretOutputHandler):
-    nb_lines: ClassVar[int] = 3
-
-    def _process_scan_impl(self, scan: SecretScanCollection, top: bool = True) -> str:
+    def _process_scan_impl(self, scan: SecretScanCollection) -> str:
+        """Output  Secret Scan Collection in text format"""
```
agateau-gg, line 34:
```suggestion
        """Output Secret Scan Collection in text format"""
```

agateau-gg (changes requested):

PR #895, opened 2024-05-24
Agateau/fix extended match

agateau-gg:
## PR check list

### DIFF
--- a/ggshield/core/lines.py
+++ b/ggshield/core/lines.py
@@ -159,7 +159,6 @@ def get_lines_from_patch(content: str, filemode: Filemode) -> Iterable[Line]:
 
 def get_padding(lines: List[Line]) -> int:
     """Return the number of digit of the maximum line number."""
-    # value can be None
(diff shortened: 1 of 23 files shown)


ggshield/core/lines.py
```diff
@@ -26,13 +26,40 @@ class LineCategory(Enum):
...
+    +WORLD
+    ```
+
+    The values of pre_index and post_index are:
+
+    line          pre_index  post_index
+    " Hello"      56         56
+    "-world"      57         None
+    "+beautiful"  None       57
+    "+WORLD"      None       58
```
fnareoh, line 62: Actually here in [the lines content the first character is always removed when it is as patch](https://github.com/GitGuardian/ggshield/blob/f9c849ad90b45f8fbb96a241543faac17b4ff8ce/ggshield/core/lines.py#L106-L106)

agateau-gg:
Replaced with #905.

Issue #892, opened 2024-05-21
IndexError when GITGUARDIAN_INSTANCE=""


adrien-fichet:
## Environment

- ggshield version: 1.27.0
- Operating system (Linux, macOS, Windows): macOS
- Operating system version: 14.5
- Python version: 3.12

## Describe the bug

When running `ggshield` while the `GITGUARDIAN_INSTANCE` environment variable is set to an empty string, the following error is displayed:

```
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/ggshield/1.27.0/libexec/lib/python3.12/site-packages/ggshield/core/config/config.py", line 131, in api_key
    key = os.environ["GITGUARDIAN_API_KEY"]
          ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen os>", line 685, in __getitem__
KeyError: 'GITGUARDIAN_API_KEY'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/homebrew/bin/ggshield", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/homebrew/Cellar/ggshield/1.27.0/libexec/lib/python3.12/site-packages/ggshield/__main__.py", line 190, in main
    return cli.main(args, prog_name="ggshield", standalone_mode=not show_crash_log)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/ggshield/1.27.0/libexec/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/ggshield/1.27.0/libexec/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/ggshield/1.27.0/libexec/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #897, opened 2024-05-27
fix(auth_login): error handling with invalid instance URL

salome-voltz:
## Context
ggshield auth login flow errors were not handled when instance URL is invalid. 

## What has been done
Error thrown are handled, similarly to how error are handled with `GITGUARDIAN_INSTANCE='https://dashboard.gitguardian.com/abc' ggshield auth login --method=token` 

## Validation
Run the following commands:
- `GITGUARDIAN_INSTANCE='https://dashboard.gitguardian.com/abc' ggshield auth login`
-  `GITGUARDIAN_INSTANCE='https://dashboard.example.com' ggshield auth login`

## Note 
This also fixes the following issue: #892 
image

## PR check list

### DIFF
--- a/ggshield/cmd/auth/login.py
+++ b/ggshield/cmd/auth/login.py
@@ -2,6 +2,7 @@ import re
 from typing import Any, List, Optional, Tuple
 
 import click
+import requests
 
 from ggshield.cmd.utils.common_options import add_common_options
 from ggshield.cmd.utils.context_obj import ContextObj
@@ -189,7 +190,14 @@ def token_login(config: Config, instance: Optional[str]) -> None:
 
     # enforce using the token (and not use config default)
     client = create_client(api_key=token, api_url=config.api_url)
-    response = client.get(endpoint="token")
+    try:
+        response = client.get(endpoint="token")
+    except requests.exceptions.ConnectionError as e:
+        if "Failed to resolve" in str(e):
+            raise click.UsageError(f"Invalid instance: {instance}.")
+        else:
+            raise UnexpectedError(f"Failed to connect to {instance}.") from e
+
     if not response.ok:
         raise UnexpectedError("Authentication failed with token.")
 
--- a/ggshield/core/url_utils.py
+++ b/ggshield/core/url_utils.py
@@ -24,24 +24,33 @@ def clean_url(url: str, warn: bool = False) -> ParseResult:
     return parsed_url
 
 
+def validate_instance_url(url: str, warn: bool = False) -> ParseResult:
+    """
+    Validate a dashboard URL
+    """
+    parsed_url = clean_url(url, warn=warn)
+    if parsed_url.scheme != "https" and not (
+        parsed_url.netloc.startswith("localhost")
+        or parsed_url.netloc.startswith("127.0.0.1")
+    ):
+        raise UsageError(f"Invalid scheme for dashboard URL '{url}', expected HTTPS")
+    if any(parsed_url.netloc.endswith("." + domain) for domain in GITGUARDIAN_DOMAINS):
+        if parsed_url.path:
+            raise UsageError(
+                f"Invalid dashboard URL '{url}', got an unexpected path '{parsed_url.path}'"
+            )
+
+    return parsed_url
+
+
 def dashboard_to_api_url(dashboard_url: str, warn: bool = False) -> str:
     """
     Convert a dashboard URL to an API URL.
     handles the SaaS edge case where the host changes instead of the path
     """
-    parsed_url = clean_url(dashboard_url, warn=warn)
-    if parsed_url.scheme != "https" and not (
-        parsed_url.netloc.startswith("localhost")
-        or parsed_url.netloc.startswith("127.0.0.1")
-    ):
-        raise UsageError(
-            f"Invalid scheme for dashboard URL '{dashboard_url}', expected HTTPS"
-        )
+    parsed_url = validate_instance_url(dashboard_url, warn=warn)
+
     if any(parsed_url.netloc.endswith("." + domain) for domain in GITGUARDIAN_DOMAINS):
-        if parsed_url.path:
-            raise UsageError(
-                f"Invalid dashboard URL '{dashboard_url}', got an unexpected path '{parsed_url.path}'"
-            )
         parsed_url = parsed_url._replace(
             netloc=parsed_url.netloc.replace("dashboard", "api")
         )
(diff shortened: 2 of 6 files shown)


ggshield/core/url_utils.py
```diff
@@ -24,24 +24,32 @@ def clean_url(url: str, warn: bool = False) -> ParseResult:
...
+
+
+def dashboard_to_api_url(dashboard_url: str, warn: bool = False) -> str:
+    """
+    Convert a dashboard URL to an API URL.
+    handles the SaaS edge case where the host changes instead of the path
+    """
+    validate_instance_url(dashboard_url, warn=warn)
+
+    parsed_url = clean_url(dashboard_url, warn=warn)
```
fnareoh, line 51: You already do this in `validate_instance_url` is there a reason to make a second call ?

tests/unit/core/config/test_config.py
```diff
@@ -312,3 +313,19 @@ def test_updating_config_not_from_default_local_config_path(
...
+
+    @pytest.mark.parametrize(
+        "instance_url",
+        ["", "http://api.gitguardian.com/", "https://api.gitguardian.com/abc"],
+    )
+    def test_invalid_instance_url(self, monkeypatch, instance_url):
+        """
+        GIVEN an invalid instance url
+        WHEN loading the config
+        THEN it raises
```
fnareoh, line 325: Did you maybe forget to precise which error it raises ?

ggshield/cmd/auth/login.py
```diff
@@ -189,7 +190,11 @@ def token_login(config: Config, instance: Optional[str]) -> None:
 
     # enforce using the token (and not use config default)
     client = create_client(api_key=token, api_url=config.api_url)
-    response = client.get(endpoint="token")
+    try:
+        response = client.get(endpoint="token")
+    except requests.exceptions.ConnectionError:
```
fnareoh, line 195: Hum there could be causes for ConnectionError other than an Invalid instance, where you trying to catch here if the user set an invalid instance ?

fnareoh (changes requested):
Few comments and I have not tested yet!

ggshield/cmd/auth/login.py
salome-voltz: The problem is the following: it will crash when calling `'$GITGUARDIAN_INSTANCE=https://dashboard.example.com' ggshield auth login` since the instance URL is invalid. Currently, the whole stack trace is displayed in the console, which is not ideal. 
So I would like to catch the error when it happens. The problem is that we can know for sure in advance that the instance URL will be invalid, as it could be an on-prem instance. So I wanted to fix that by catching the error we get when we can't resolve the resource name. 
I agree that this might not be ideal, do you have another idea to handle the error a better way?

fnareoh: Hum, when I try on my current ggshield ` GITGUARDIAN_INSTANCE='https://dashboard.example.com' ggshield auth login` does not crash (it just gets stuck because the web page does not identify me), but in deed `GITGUARDIAN_INSTANCE='https://dashboard.example.com' ggshield auth login --method=token` crashes in a problematic way.
The full error is:
```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='dashboard.example.com', port=443): Max retries exceeded with url: /exposed/v1/token (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x725d5d2a4940>: Failed to resolve 'dashboard.example.com' ([Errno -2] Name or service not known)")"
```
I propose that you modify your solution to detect the  `"Failed to resolve {instance}"` in the error and in this case only raise the invalid instance ? What do you think ?

salome-voltz: I adjusted the error handling as you suggested, let me know if this works for you

ggshield/cmd/auth/login.py
```diff
@@ -189,7 +190,14 @@ def token_login(config: Config, instance: Optional[str]) -> None:
 
     # enforce using the token (and not use config default)
     client = create_client(api_key=token, api_url=config.api_url)
-    response = client.get(endpoint="token")
+    try:
+        response = client.get(endpoint="token")
+    except requests.exceptions.ConnectionError as e:
+        if "Failed to resolve" in str(e):
```
fnareoh, line 196: I think we could be even more precise by searching `Failed to resolve '{instance}'` ?

salome-voltz: not this easily, for example with `$GITGUARDIAN_INSTANCE='https://dashboard.example.com'`, the error would be `Failed to resolve 'dashboard.example.com'` so I would have to crop the instance variable to look for it... I'm afraid to introduce more unwanted behavior. 
Do you think I should still do it ?

ggshield/core/url_utils.py
```diff
@@ -24,24 +24,33 @@ def clean_url(url: str, warn: bool = False) -> ParseResult:
...
-    parsed_url = clean_url(dashboard_url, warn=warn)
+    parsed_url = clean_url(url, warn=warn)
     if parsed_url.scheme != "https" and not (
         parsed_url.netloc.startswith("localhost")
         or parsed_url.netloc.startswith("127.0.0.1")
     ):
-        raise UsageError(
-            f"Invalid scheme for dashboard URL '{dashboard_url}', expected HTTPS"
-        )
+        raise UsageError(f"Invalid scheme for dashboard URL '{url}', expected HTTPS")
```
fnareoh, line 36: What would happen if a self hosted client had its instance be something like `https://all-internal-tools.tech/gitguardian` ?
My understanding it that he could be blocked because of the new URL validation ?

ggshield/core/url_utils.py
```diff
@@ -24,24 +24,33 @@ def clean_url(url: str, warn: bool = False) -> ParseResult:
...
     if any(parsed_url.netloc.endswith("." + domain) for domain in GITGUARDIAN_DOMAINS):
         if parsed_url.path:
```
fnareoh, line 38: What would happen if a self hosted client had its instance be something like https://all-internal-tools.tech/gitguardian ?
My understanding it that he could be blocked because of the new URL validation ?

fnareoh: Never mind it is only if we are on gitguardian instance because of the previous condition !

fnareoh (approved):
Looks good to me !

agateau-gg (approved):
Looks good!

PR #899, opened 2024-05-28
Refactor: save lines in extended match and add censor method

fnareoh:
## Context

After having worked on factorizing `make_matches` the new goal is to start storing context lines in the Extended Match and have the output entirely depend those line and no longer the `result.file.content`. 
In particular this means a way to censor Extended Match based on the lines and the indices in MatchSpan and no longer relying on censor content.

## What has been done

This MR is divided into several commits and test should pass after each commit (best reviewed by commit?). 

Several important change have been made:
- the `Line` class has been changed from a `NamedTuple` to a `SimpleNamespace` to be able to modify the line.content for censoring.
- The display patch hunk indication is no longer present in the patch output as it is not really part of the file and we do not display the patch symbols '+/-/ ' . This means some test secrets are considered shorter and the censoring (proportional to the length of the secret) is shorter as well (so some test in output test text have been changed accordingly).
- The '+' have been removed of multi line secret in patches as they where not present in single ligne secret nor in the first and last lines of the multiline secret.
- Some test in test output text had falty Match index end creating falty output, it has been cleanned up. 

## Validation

Test pass and output is more coherent now ? I did my best to make it reviewable but it was quite tricky to work with...

## PR check list

### DIFF
--- a/changelog.d/20240604_110454_fnareoh_refacto_censor_lines.md
+++ b/changelog.d/20240604_110454_fnareoh_refacto_censor_lines.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Patch symbols at the start of lines are now always displayed, even for single line secrets.
--- a/ggshield/core/lines.py
+++ b/ggshield/core/lines.py
@@ -24,6 +24,16 @@ class LineCategory(Enum):
     DELETION = auto()
     EMPTY = auto()  # Used for empty lines and patch headers.
 
+    @property
+    def symbol(self) -> str:
+        if self == LineCategory.ADDITION:
+            return "+"
+        if self == LineCategory.DELETION:
+            return "-"
+        if self == LineCategory.EMPTY:
+            return ""
+        return " "
+
 
 class Line(SimpleNamespace):
     """
@@ -97,12 +107,6 @@ class Line(SimpleNamespace):
         # Patch
         pre_index = format_line_count(self.pre_index, padding)
         post_index = format_line_count(self.post_index, padding)
-
-        if self.category == LineCategory.ADDITION:
-            pre_index = " " * padding
-
-        elif self.category == LineCategory.DELETION:
-            post_index = " " * padding
 
         return PATCH_LINE_PREFIX.format(
             format_text(pre_index, line_count_style),
--- a/ggshield/core/lines.py
+++ b/ggshield/core/lines.py
@@ -1,6 +1,7 @@
 import re
 from enum import Enum, auto
-from typing import Iterable, List, NamedTuple, Optional
+from types import SimpleNamespace
+from typing import Iterable, List, Optional
 
 from ggshield.core.text_utils import STYLE, format_line_count, format_text
 from ggshield.utils.git_shell import Filemode
@@ -21,10 +22,10 @@ class LineCategory(Enum):
     ADDITION = auto()
     DATA = auto()
     DELETION = auto()
-    EMPTY = auto()
+    EMPTY = auto()  # Used for empty lines and patch headers.
 
 
-class Line(NamedTuple):
+class Line(SimpleNamespace):
     """
     Represent a line in a document.
 
@@ -66,6 +67,16 @@ class Line(NamedTuple):
     category: Optional[LineCategory] = None
     pre_index: Optional[int] = None
     post_index: Optional[int] = None
+
+    def __hash__(self):
+        return hash(
+            (
+                self.content,
+                self.category,
+                self.pre_index,
+                self.post_index,
+            )
+        )
 
     def build_line_count(self, padding: int, is_secret: bool = False) -> str:
         """Return the formatted line count."""
--- a/ggshield/verticals/secret/extended_match.py
+++ b/ggshield/verticals/secret/extended_match.py
@@ -3,8 +3,13 @@ from typing import Any, Dict, List, Optional
 from marshmallow import fields, post_dump
 from pygitguardian.models import Match, MatchSchema
 
+from ggshield.core.filter import censor_string
 from ggshield.core.lines import Line
 from ggshield.core.match_span import MatchSpan
+
+
+# The number of lines to display before and after a secret in the patch
+NB_CONTEXT_LINES = 3
 
 
 class ExtendedMatchSchema(MatchSchema):
@@ -40,6 +45,9 @@ class ExtendedMatch(Match):
     def __init__(
         self,
         span: MatchSpan,
+        lines_before_secret: List[Line],
+        lines_with_secret: List[Line],
+        lines_after_secret: List[Line],
         pre_line_start: Optional[int] = None,
         pre_line_end: Optional[int] = None,
         post_line_start: Optional[int] = None,
@@ -47,6 +55,9 @@ class ExtendedMatch(Match):
         **kwargs: Any,
     ):
         self.span = span
+        self.lines_before_secret = lines_before_secret
+        self.lines_with_secret = lines_with_secret
+        self.lines_after_secret = lines_after_secret
         self.pre_line_start = pre_line_start
         self.pre_line_end = pre_line_end
         self.post_line_start = post_line_start
@@ -68,6 +79,16 @@ class ExtendedMatch(Match):
         return cls(
             span=span,
             match=match.match,
+            lines_before_secret=lines[
+                max(
+                    span.line_index_start - NB_CONTEXT_LINES + 1, 0
+                ) : span.line_index_start
+            ],
+            lines_with_secret=lines[span.line_index_start : span.line_index_end + 1],
+            lines_after_secret=lines[
+                span.line_index_end
+                + 1 : min(span.line_index_end + NB_CONTEXT_LINES, len(lines))
+            ],
             match_type=match.match_type,
             index_start=span.column_index_start,
             index_end=span.column_index_end,
@@ -78,6 +99,32 @@ class ExtendedMatch(Match):
             pre_line_end=end_line.pre_index,
             post_line_end=end_line.post_index,
         )
+
+    def censor(self) -> None:
+        """
+        Censor the match and all the lines containing the secret.
+        Lines are modified in place, so the secret does not appear in the context of other secrets.
+        """
+        len_match = len(self.match)
+        self.match = censor_string(self.match)  # Censor the match
+        assert len(self.match) == len_match
+        match_split_lines = self.match.splitlines()
+        assert len(self.lines_with_secret) == len(match_split_lines)
+        # Censor the lines containing the secret by replacing the secret by its censored version.
+        for index_line, (line, match_split_line) in enumerate(
+            zip(self.lines_with_secret, match_split_lines)
+        ):
+            censor_start = self.span.column_index_start if index_line == 0 else 0
+            censor_end = (
+                self.span.column_index_end
+                if index_line == len(match_split_lines) - 1
+                else len(line.content)
+            )
+            line.content = (  # Modify the line to censor the match
+                line.content[:censor_start]
+                + match_split_line
+                + line.content[censor_end:]
+            )
 
     def __repr__(self) -> str:
         return ", ".join(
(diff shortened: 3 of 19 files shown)


ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -168,70 +169,29 @@ def leak_message_located(
     leak_msg = StringIO()
     max_width = shutil.get_terminal_size()[0] - offset if clip_long_lines else 0
 
-    lines_to_display = get_lines_to_display(flat_matches_dict, lines)
-
     old_line_number: Optional[int] = None
-    for line_number in sorted(lines_to_display):
-        multiline_end = None
-        line = lines[line_number]
+    for line, list_of_matches in flat_matches_dict.items():
```
agateau-gg, line 173: nit: I think `list_of_matches` can be renamed `matches`

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -240,46 +200,65 @@ def leak_message_located(
                         max_width,
                     )
                     leak_msg.write(formatted_line)
+                else:
+                    # The secret is on multiple lines
+                    detector_position = float("inf"), float("-inf")
+                    for index, line_of_secret in enumerate(match.lines_with_secret):
```
agateau-gg, line 205: similar here: `line_of_secret` → `line`

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -240,46 +200,65 @@ def leak_message_located(
...
+                flat_match_dict[line] = []
+            flat_match_dict[line].append(match)
+    # Sort the matches per line number
+    ordered_flat_match_dict: OrderedDict[Line, List[ExtendedMatch]] = OrderedDict(
+        sorted(
+            flat_match_dict.items(),
+            key=lambda item: item[0].pre_index or item[0].post_index or -1,
+        )
+    )
+    return ordered_flat_match_dict
```
agateau-gg: I think this can be a plain dict: `dict` preserves insertion order so when creating the dict like this, items will be correctly ordered.

Alternatively, the return value could be a `List[Tuple[Line, List[ExtendedMatch]]` since we don't need dict-like access outside of this function.

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -240,46 +200,65 @@ def leak_message_located(
...
     return leak_msg.getvalue()
 
 
 def flatten_policy_breaks_by_line(
     policy_breaks: List[PolicyBreak],
-) -> Dict[int, List[ExtendedMatch]]:
+) -> OrderedDict[Line, List[ExtendedMatch]]:
     """
     flatten_policy_breaks_by_line turns a list of policy breaks into a dictionary
     mapping a line number to a list of matches starting at that line.
```
agateau-gg, line 239: docstring needs to be updated: the returned dict keys are not line numbers anymore.

tests/unit/verticals/secret/output/snapshots/snap_test_text_output.py
```diff
@@ -258,15 +245,14 @@
...
-  6 | +****************+***********************************************
-  7 | +**********************+*****************************************
-  8 | +****+******Xme/ovcDeM1+3W/UmSHYUW4b3WYq4
+  2 | MIIBOgIBAAJBAIIRkYjxjE3KIZi******************************+******
+  3 | ****************************************************************
+  4 | ****************************************************************
+  5 | ***********+****************************************************
+  6 | ****************+***********************************************
+  7 | **********************+*****************************************
+  8 | ****+******Xme/ovcDeM1+3W/UmSHYUW4b3WYq4
```
agateau-gg: I think it's not a good idea to loose the "category" character: it is useful for users to know if the scanned commit added or removed the secret.

ggshield/verticals/secret/extended_match.py
```diff
@@ -69,16 +69,37 @@ def from_match(
...
+                [
+                    (
+                        str_line_match[1:]
+                        if index > 0 and line.category != LineCategory.DATA
+                        else str_line_match
+                    )
+                    for index, (str_line_match, line) in enumerate(
+                        zip(match.match.split("\n"), lines_with_secrets)
+                    )
+                ]
```
agateau-gg, line 98: You can remove the `[`, `]` here: it's simpler and faster (no need to build a list)

ggshield/verticals/secret/extended_match.py
```diff
@@ -69,16 +69,37 @@ def from_match(
...
         assert line_index_start is not None and line_index_end is not None
         line_index_start += int(is_patch) - 1  # convert to 0-based
         line_index_end += int(is_patch) - 1
+
+        lines_with_secrets = [
+            lines[line_index]
+            for line_index in range(span.line_index_start, span.line_index_end + 1)
+        ]
+        stripped_match = match.match
+        if is_patch and len(lines_with_secrets) > 1:
```
agateau-gg, line 86: nit: I find code like this more readable as:

```
if something:
    foo = 1
else:
    foo = 2
```

than:

```
foo = 2
if something:
    foo = 1
```

It avoids my brain going: "ok, foo is 2. Ah no, wait, actually it's 1".

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -167,6 +157,16 @@ def leak_message_located(
...
+    all_lines = []
+    for line, list_matches in flat_matches_dict.items():
+        all_lines.append(line)
+        for match in list_matches:
+            all_lines.extend(
+                match.lines_with_secret,
+            )
+    padding = get_padding(all_lines)
+    offset = get_offset(padding, is_patch)
+    print(padding, offset)
```
agateau-gg, line 169: :fire:

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -21,19 +21,31 @@ class Result(NamedTuple):
     # TODO: Rename `file` to `scannable`?
     file: Scannable  # filename that was scanned
     scan: ScanResult  # Result of content scan
+    is_on_patch: bool  # Is the scan on a patch
+
+    def __init__(self, file: Scannable, scan: ScanResult):
+        self.file = file
+        self.scan = scan
+        self.is_on_patch = file.filemode != Filemode.FILE
```
agateau-gg, line 29: This can be a property, then Result can remain a NamedTuple, which implements `__eq__()` for you.

agateau-gg (changes requested):
This is some impressive work!

The fact that the off-by-one errors in the test data went unnoticed for so long is a bit scary 😱. Great to see them fixed.

Other than some minor remarks, the only thing that worries me is the loss of the "category" character when showing patches.

Regarding the Windows CI failures: they may come from the fact that you split lines with `split("\n")` instead of `splitlines()`, but I could be wrong.

tests/unit/verticals/secret/test_censor_extended_matches.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def test_censor_match_for_plain_content_multiple_on_one_line():
+    """
+    GIVEN some content
+    AND a match in it
+    WHEN ExtendedMatch.from_match() is called
+    THEN it contains the expected values
+    """
+    print()
```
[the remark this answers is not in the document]
fnareoh: Oh no!!! I swear I had grep some diff (evidently not the right one) to make sure there where none but next time I'll double check the MR T.T

ggshield/verticals/secret/output/secret_text_output_handler.py
fnareoh: the problem here is that we already have a line define :)

agateau-gg: Oh, I missed that. Might be because that function is quite large (8 level of indentations at its peak). I won't object if you split it 😉, but not a blocker if you want to land this first.

fnareoh:
As discussed, I added back the display of patches hunk and unified the display of patch line "category".

The error on unit test on windows seems to have been due outdated cassettes with faulty indices I'm really surprised the other OS work (and it is a bit worrying).

agateau-gg (changes requested):
Just one last docstring comment and this can go in!

changelog.d/20240604_110454_fnareoh_refacto_censor_lines.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+The display of patch start of lines symbols has been made more uniform.
+It is always displayed not just in case of multiline secrets.
```
agateau-gg, line 4: I think this entry can be more concise, here is an attempt at shortening it (also, turn it into a list item, otherwise scriv is not going to be happy)

```suggestion
- Patch symbols at the start of lines are now always displayed, even for single line secrets.
```

agateau-gg (approved):
Made a small remark on the changelog entry, but approved nevertheless, so feel free to merge once this is fixed.

ggshield/verticals/secret/output/secret_json_output_handler.py
```diff
@@ -3,7 +3,7 @@
 from pygitguardian.client import VERSIONS
 from pygitguardian.models import PolicyBreak
 
-from ggshield.core.filter import censor_content, leak_dictionary_by_ignore_sha
+from ggshield.core.filter import leak_dictionary_by_ignore_sha
```
salome-voltz, line 6: Do we we still need function censor_content in ggshield/core/filter.py ? 
P-S: if you remove it, also remove the associate test ;)

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -150,136 +140,141 @@ def process_result(
...
     :param padding: The line padding
     :param offset: The offset due to the line display
+    :param is_patch: Whether the content is a patch
+    :param clip_long_lines: Whether to clip long lines
     """
     leak_msg = StringIO()
+    all_lines = []
+    for line, list_matches in flat_matches:
+        all_lines.append(line)
+        for match in list_matches:
```
salome-voltz, line 160: It's hard to understand why you add lines twice, maybe you can add a comment to explain it :)

ggshield/core/lines.py
```diff
@@ -23,8 +24,18 @@ class LineCategory(Enum):
     DELETION = auto()
     EMPTY = auto()
```
salome-voltz, line 25: Should we rename this Enum, maybe `PATCH_INFORMATION` (or any other better wordind)

salome-voltz:
Minor remarks for suggestions, but feel free to merge :smile:

ggshield/core/lines.py
fnareoh: Renaming enum can be a pain but I'll add a comment :)

ggshield/verticals/secret/extended_match.py
```diff
@@ -58,16 +69,46 @@ def from_match(
...
+        lines_with_secret = lines[span.line_index_start : span.line_index_end + 1]
+
+        match_split_lines = match.match.splitlines()
+        assert len(match_split_lines) == len(lines_with_secret)
+        if is_patch and len(lines_with_secret) > 1:
+            # The secret is multi-line and in patch, we can have the first character to remove
+            stripped_match = "\n".join(
+                (
+                    str_line_match[1:]
+                    if index > 0 and line.category != LineCategory.DATA
```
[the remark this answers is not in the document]
fnareoh: I thought about it more and we don't need to check because the range just intersect so if `len(str_line_match)==0` then it will just return the empty line which is ok !

ggshield/verticals/secret/output/secret_text_output_handler.py
fnareoh: I went the rout of not adding the line twice as it makes more sense anyway...

PR #902, opened 2024-05-31
Mmillet/sca 1670 refactor ggshield scan ci commands

gg-mmill:
## Context

Refactor `sca scan ci` and  `iac scan ci`:  we want to 
- compute the range of commits associated with the merge request, and not the push event
- only allow the command to run in merge requests pipelines (for now, a fallback behaviour is kept, so that the command will perform a full scan if that is not the case).

## What has been done

Refactor the commands

## Validation

Run the commands in merge request pipelines, with verbose mode.

## PR check list

### DIFF
--- a/ggshield/cmd/sca/scan/ci.py
+++ b/ggshield/cmd/sca/scan/ci.py
@@ -1,9 +1,7 @@
-import os
 from pathlib import Path
 from typing import Any, Optional, Sequence
 
 import click
-from click import UsageError
 
 from ggshield.cmd.sca.scan.sca_scan_utils import (
     create_output_handler,
@@ -15,12 +13,15 @@ from ggshield.cmd.sca.scan.scan_common_options import (
     update_context,
 )
 from ggshield.cmd.utils.common_decorators import exception_wrapper
-from ggshield.cmd.utils.common_options import all_option, directory_argument
+from ggshield.cmd.utils.common_options import directory_argument
 from ggshield.cmd.utils.context_obj import ContextObj
-from ggshield.core.errors import handle_exception
-from ggshield.core.git_hooks.ci import get_current_and_previous_state_from_ci_env
+from ggshield.core.git_hooks.ci.get_scan_ci_parameters import (
+    NotAMergeRequestError,
+    get_scan_ci_parameters,
+)
 from ggshield.core.git_hooks.ci.supported_ci import SupportedCI
 from ggshield.core.scan.scan_mode import ScanMode
+from ggshield.utils.git_shell import git
 from ggshield.verticals.sca.collection.collection import (
     SCAScanAllVulnerabilityCollection,
     SCAScanDiffVulnerabilityCollection,
@@ -31,7 +32,6 @@ from ggshield.verticals.sca.collection.collection import (
 @add_sca_scan_common_options()
 @click.pass_context
 @directory_argument
-@all_option
 @exception_wrapper
 def scan_ci_cmd(
     ctx: click.Context,
@@ -41,7 +41,6 @@ def scan_ci_cmd(
     ignore_fixable: bool,
     ignore_not_fixable: bool,
     directory: Optional[Path],
-    scan_all: bool,
     **kwargs: Any,
 ) -> int:
     """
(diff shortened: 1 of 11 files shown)


ggshield/utils/git_shell.py
```diff
@@ -445,3 +445,16 @@ def parse_name_status_patch(patch: str) -> Dict[Path, Filemode]:
...
     return git(["show", f"{ref}:{path.as_posix()}"], cwd=wd)
+
+
+def get_mr_commits(
+    target_branch: str, current_tip: str = "HEAD", wd: Optional[Union[str, Path]] = None
+) -> list[str]:
+    """
+    This function is used to list commits that belong to a merge request
+    Commits are returned in descending order (most recent commit first)
+    Note: I have no idea why each line needs to be stripped from ' characters
```
agateau-gg, line 456: I think that is because of the `'` in `"--format='%H'` 😉

ggshield/utils/git_shell.py
```diff
@@ -445,3 +445,16 @@ def parse_name_status_patch(patch: str) -> Dict[Path, Filemode]:
 def read_git_file(ref: str, path: Path, wd: Optional[Union[str, Path]] = None) -> str:
     # Use as_posix to handle git and Windows
     return git(["show", f"{ref}:{path.as_posix()}"], cwd=wd)
+
+
+def get_mr_commits(
```
agateau-gg, line 450: This function lists all commits not in the current branch, I think we can give it a name not linked to GitLab (MR is very GitLab-only). May `get_commits_not_in_branch()`?

agateau-gg (changes requested):

gg-mmill (commented):
Thanks for the review, made the requested changes. 
I'm now trying to debug the github actions, but this should work shortly

ggshield/cmd/sca/scan/ci.py
```diff
@@ -65,34 +63,41 @@ def scan_ci_cmd(
...
-            previous_ref=previous_commit,
+            previous_ref=reference_commit,
             current_ref=current_commit,
             scan_mode=scan_mode,
             ci_mode=ci_mode.name,
         )
         scan = SCAScanDiffVulnerabilityCollection(id=str(directory), result=result)
         return output_handler.process_scan_diff_result(scan)
-
-    except Exception as error:
```
gg-mmill, line 97: Note: the function is already decorated with `@exception_wrapper`, so we can remove the catch-all exception handler

amascia-gg (commented):
Thanks a lot for the refacto!

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
```diff
@@ -0,0 +1,106 @@
...
+            click.echo(
+                f"\tUsing first remote {remotes[0]}.",
+                err=True,
+            )
+        remote_prefix = f"{remotes[0]}/"
+
+    mr_tip = "HEAD"
+    if current_ci == SupportedCI.GITHUB:
+        # GitHub pipelines are all merge results
+        mr_tip = remote_prefix + os.environ["GITHUB_HEAD_REF"]
```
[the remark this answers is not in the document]
gg-mmill: It's always set when `GITHUB_BASE_REF` (`CI_TARGET_BRANCH_ASSOC` for github) is set, but indeed the check was not in the right place, this is now fixed !

Added some tests, which also uncovered another small bug.

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
```diff
@@ -0,0 +1,109 @@
...
+
+def get_scan_ci_parameters(
+    current_ci: SupportedCI,
+    wd: Optional[Union[str, Path]] = None,
+    verbose: bool = False,
+) -> Union[Tuple[str, str], None]:
+    """
+    Function used to gather current commit and reference commit, for the SCA/IaC scan
+    ci commands.
+    Return: None if the MR has no associated commits, otherwise a tuple (current_commit, reference commit)
```
agateau-gg, line 44: In which situation can there be an MR with no associated commits?

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
```diff
@@ -0,0 +1,109 @@
...
+        click.echo(
+            f"\tIdentified current ci as {current_ci.value}",
+            err=True,
+        )
+
+    if current_ci == SupportedCI.TRAVIS:
+        return travis_scan_ci_args()
+
+    # we will work with branch names, so we run a git fetch to ensure the names are locally available
+    git(["fetch"], cwd=wd)
```
agateau-gg, line 56: I understand it's handy to have the fetch here, but I think it would be better to have the caller do this, for two reasons:

- The function name does not look like a function which would alter the repository it is working on, this could lead to surprising behavior. Especially since right now it's not done for Travis.
- If we add more functions which require an up-to-date repository, and they each call `git fetch`, then we end up with multiple fetches where we only need one.

agateau-gg (changes requested):

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
gg-mmill: Well you can create a branch, push it, and then create a MR (no commit = "no now commit compared to target branch")

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
gg-mmill: It would really worsen the UX though - WDYT of fetching at the command level ? (so that we only do it once)

Also the cost of subsequent fetch is not negligible, but still much less than the first one.

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
```diff
@@ -0,0 +1,108 @@
...
+        if verbose:
+            click.echo(
+                f"\tUsing first remote {remotes[0]}.",
+                err=True,
+            )
+        remote_prefix = f"{remotes[0]}/"
+
+    target_branch_var = CI_TARGET_BRANCH_ASSOC.get(current_ci)
+    if not target_branch_var:
+        raise UnexpectedError(f"Using scan ci is not supported for {current_ci.value}.")
```
amascia-gg, line 75: nit: Should this be done before the `get_remotes` so that in the case of CircleCI we directly raise this error?

amascia-gg (approved):

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
agateau-gg: Fetching at the command level sounds good to me :+1:.

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
gg-mmill: I think it's ok, getting the remotes is a very light operation (does not involve network) - and also the case where there is no remote should not happen in practice, it was mostly used for tests (although that's not best practice :see_no_evil: )

gg-mmill: Note that we do have one test where a remote is set though

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
gg-mmill: Thanks, i  cleanup up the commits, should be ready then :)

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
agateau-gg: Ah right. Can you just add a comment explaining this?

agateau-gg (approved):

ggshield/core/git_hooks/ci/get_scan_ci_parameters.py
gg-mmill: Added

PR #904, opened 2024-06-04
Use allow-self-signed in auth login command

irgeek:
## Context

The `ggshield auth login` command isn't using the `--allow-self-hosted` flag.

## What has been done

Passing `allow_self_hosted` to the `create_client` function--as we're already doing for `auth logout`.

## Validation

Without the fix, trying run `ggshield auth login --allow-self-signed` against a GitGuardian instance using an untrusted certificate or being routed through a TLS bump proxy will result in an error:

```
requests.exceptions.SSLError: HTTPSConnectionPool(host='hostname.example.com', port=443): Max retries exceeded with url: /exposed/v1/token (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:1007)')))
```

With the fix the certificate error won't appear.

## PR check list

### DIFF
--- a/changelog.d/20240606_114213_james.sinclair_auth_login_allow_self_signed.md
+++ b/changelog.d/20240606_114213_james.sinclair_auth_login_allow_self_signed.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- The `ggshield auth login` command now respects the `--allow-self-signed` flag.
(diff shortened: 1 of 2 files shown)


agateau-gg (changes requested):
The change looks good, thanks!

I think this fix is worth a changelog entry. Can you add one (look at doc/dev/getting-started.md for details)?

agateau-gg:
Looks good, thanks!

PR #905, opened 2024-06-05
Garancegourdel/refacto output handlers to remove scanable from result

fnareoh:
## Context

Currently, for every `Result` we keep the entire file (more precisely `Scannable` object) in which the secret was found, and only at the very end used it to display the secrets and a bit of it's context. This can cause ggshield to run out of memory if a scan finds many secrets in large files.

## What has been done

In #890 we first worked on unifying the `make_matches` functions in between json and text output. Unfortunately, due to a lack of test on the json output, a breaking change was introduced and work MR was revert. This MR adds it back, with tests and a fix (thank you so much @agateau-gg )! 

Then the last 4 commits gradually replace the need for the full content when displaying and ultimatly removes it. In the process it removes a few incoherence's in the tests. Best reviewed by commits!

## Validation

Test should still pass.
When finding a lot of secrets in large files the memory consumption should augment much less.

## PR check list


fnareoh:
@agateau-gg  There is only the last commit you have not yet reviewed (it removes completly the Scannable from the Result).

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -12,24 +13,41 @@
...
     Return model for a scan which zips the information
     between the Scan result and its input file.
     """
 
-    # TODO: Rename `file` to `scannable`?
-    file: Scannable  # filename that was scanned
+    filename: str  # Name of the file/patch scanned
+    filemode: Filemode
+    filepath: Path
+    file_url: str
```
agateau-gg, line 25: Since not all scannables are files (cf the TODO comment), I propose the following name changes:

- filepath -> path
- file_url -> url

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -100,13 +99,8 @@ def process_result(
         """
         result_buf = StringIO()
 
-        # policy breaks and matches are modified in the functions leak_dictionary_by_ignore_sha and censor_content.
```
agateau-gg, line 105: `leak_dictionary_by_ignore_sha()` still modifies scan.policy_breaks, so I think we should continue creating a copy of result.

agateau-gg (changes requested):
Made some comments here, but we should treat this branch as an "integration" branch for your work on fixing the memory leaks. This makes review of new changes simpler than "just review the N last commits".

This means the changes to address my comments should be made in a new PR targeting this branch (like #906 did, after the rebase). When we are done, we merge this branch into `main`.

ggshield/verticals/secret/output/secret_text_output_handler.py
fnareoh: Hum if I understand well it mostly modifies the order of policy breaks and the order of the matches in the policy break ? It seems to be called several time redundantly including as part of the init of [SecretScanCollection](https://github.com/GitGuardian/ggshield/blob/5eeff39a8ab083f78db1a51720cbd22d4ec36c28/ggshield/verticals/secret/secret_scan_collection.py#L161-L161).

For coherence across execution, I think I would prefer always enforcing this order rather than copying the result but if you disagree I will trust your guts more than mine ! :smile:

fnareoh:
Sorry for the branches confusion and thank you for your patience! #908  addresses your fist comment.

ggshield/verticals/secret/output/secret_text_output_handler.py
agateau-gg: Oh, I missed that it was called some other places. This prompted me to make the experiment or removing all sorting from that method and... it seems to be fine! No tests broke.

Going to file a PR to just remove sorting, this way `leak_dictionary_by_ignore_sha()` no longer has any side-effect. Let's keep your code the way it is.

agateau-gg (approved):
Looks good!

PR #906, opened 2024-06-05
Simplify patch handling

agateau-gg:
## Context

While reviewing #905 I thought it was a bit sad that we had to add a field just for backward-compatibility, so I dived into what could be done to get rid of some of the patch special-cases cruft.

## What has been done

Line.content now always contains the full line, including the patch symbol.

This simplifies a bunch of things:
- MatchSpan.from_match() no longer cares about is_patch
- ExtendedMatch.from_match() can reuse match.match as is
- SecretTextOutputHandler no longer needs to reprint the patch symbol
- Line.category is gone, it's replaced by a `is_patch` boolean. This attribute is only used to decide whether to print patch or file line numbers.

If you look at the snapshot diff you will see two things.

First, patch output is more correct: the second `@` at the end of the `@@ ... @@` lines does not miss anymore.

For multiline secrets, the detector left "arm" (the `|___` symbol) starts one character earlier (see the snapshot diffs). This might be fixable but I don't think it's a problem.

## Validation

Tests still pass.

## PR check list


ggshield/core/lines.py
```diff
@@ -175,25 +151,22 @@ def get_lines_from_patch(content: str, filemode: Filemode) -> Iterable[Line]:
...
         elif line_type == "-":
             pre_index += 1
             line_pre_index = pre_index
-            line_content = line[1:]
-            category = LineCategory.DELETION
+            line_content = line
         elif line_type == "\\":
             # This type of line shouldn't contain any secret; no need to set indices
-            line_content = line[1:]
+            line_content = line
```
fnareoh, line 164: we can probably simplify the setting of line_content, it is almost always set to line ?

ggshield/core/match_span.py
```diff
@@ -23,39 +23,38 @@ class MatchSpan(NamedTuple):
     column_index_end: int
 
     @staticmethod
-    def from_match(match: Match, lines: List[Line], is_patch: bool) -> "MatchSpan":
+    def from_match(match: Match, lines: List[Line]) -> "MatchSpan":
```
fnareoh, line 26: :clap: :+1:  it really simplifies this function

fnareoh (changes requested):
Thanks a lot for this, it definitely simplifies the code...
You set this PR to your original fix branch, instead of `garancegourdel/refacto-output-handlers-to-remove-scanable-from-result` is that intentional ? I can take care of the switch if you want/remove my last commit so you can point to it.
You have functional test failing because they capture a new generic secret in the snapshot.

ggshield/core/lines.py
agateau-gg: Good point. Looking into it.

PR #910, opened 2024-06-10
Remove unnecessary sorting

agateau-gg:
## Context

`ggshield.core.filter.leak_dictionary_by_ignore_sha()` takes a list of policy breaks and returns a dict of sha => list[policy_breaks]. To do so it sorts the policy breaks, and the matches within the policy breaks. Since it sorts in-place, it modifies the input policy breaks, which is surprising and actually not necessary. Especially when the function is called in places where the caller only cares about the number of different shas.

This PR removes all sorting.

## What has been done

First commit renames the function to `group_policy_breaks_by_ignore_sha()`.

Second commit removes the sorting, and adjusts the tests so that they pass. This uncovered some holes in the tests, such as `test_json_output()` only checking indices of the first secret, and actually only the first match of said secret.

## Validation

All tests pass.

## PR check list


ggshield/core/filter.py
```diff
@@ -103,14 +102,8 @@ def group_policy_breaks_by_ignore_sha(
...
     """
-    policy_breaks.sort(
-        key=lambda x: min(
-            match.index_start if match.index_start else -1 for match in x.matches
-        )
-    )
-    sha_dict: Dict[str, List[PolicyBreak]] = OrderedDict()
+    sha_dict: Dict[str, List[PolicyBreak]] = {}
     for policy_break in policy_breaks:
-        policy_break.matches.sort(key=lambda x: x.index_start if x.index_start else -1)
```
fnareoh, line 121: I'm worried about some policy breaks changing sha's and breaking out client configs... If for some reason at some point we create a policy break with matches that are not in sorted by `index_start`  it will have a different sha then it used to because [get_ignore_sha does not sort by index_start](https://github.com/GitGuardian/ggshield/blob/7e29744322a86516b61ec52c881764d5e404b447/ggshield/core/filter.py#L91) ?

agateau-gg: I was worried about that too but after a closer inspection, I think we are fine: `get_ignore_sha()` sort matches by "match_type", which is actually the name of the match. Since match names are unique inside a secret (there can't be two `username` matches, or two `port` matches), sorting by match names always produce the same order, no matter what the initial order was.

`test_get_ignore_shas()` checks this: it shuffles matches if `duplicate` is true. The PR adds a parametrized entry with 4 matches to ensure we test this case.

fnareoh: Oh thank you, I did not look closely at that test and forgot that matches_type cannot be duplicated. Great ! :partying_face:

fnareoh (approved):

PR #912, opened 2024-06-14
Pyright update

agateau-gg:
## Context

This PR syncs pyright configuration to match py-gitguardian:

- update pyright to 1.1.367
- enable the same checks

## What has been done

Mostly simple fixes, but a few trickier changes are in their own commits. In particular IaCScanCollection becomes a generic class (see 2nd commmit).


fnareoh (changes requested):
Hum something went wrong in the pipeline ? :o

agateau-gg:
> Hum something went wrong in the pipeline ? :o

Yeah, Python 3.8 is getting too old to run latest pyright. Should be better now.

fnareoh (approved):
Interesting to see the changes surrounding dataclass vs SingleNameSpace I'll remember to avoid them in the future !

gg-mmill (approved):
Good to me !

PR #919, opened 2024-06-24
Remove urllib3<2 hack

agateau-gg:
Our CI included a workaround for a weird Windows-only failure with urllib3 and vcrpy. It seems it no longer happens so we can remove it.


fnareoh (approved):
Always nice to remove hacks !

Issue #869, opened 2024-04-05
Sarif Output Format Support


Jiri-Stary:
## Is your feature request related to a problem? Please describe.

I would like to get the output of ggshield in SARIF format, so it can be process afterwards.
My goal is to unify scanning experience with multiple scanning  tools, so i can show aggregated vulnerability information (number of issues and their severity)  on github workflow runs and / or on PRs.

## A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

Current json output is undocumented and does note define a schema

## Describe the solution you'd like

Provide option to define additional output format as SARIF instead of json.

mathieubellon:
Hi @Jiri-Stary 
FYI [GGShield 1.30](https://github.com/GitGuardian/ggshield/releases/tag/v1.30.1) is now supporting an XML output (SARIF format)

PR #920, opened 2024-06-24
Add support for SARIF output for secret scanning

agateau-gg:
## Context

This PR adds support for generating reports in [SARIF format](https://sarifweb.azurewebsites.net/) when scanning for secrets.

See #869.

## What has been done

The PR first introduces a `--format (text|json)` option, and make existing commands use it. `--format json` is a more verbose version of `--json`, but it makes it possible to add new formats without adding a new flag for each format.

Then it introduces a `sarif` value for `--format`. This value is only used by `secret scan` commands for now. Adding it to other verticals will be done separately.

### Representing multi-match secrets in SARIF output

There is a bit of an impedance mismatch between ggshield secret output and SARIF output. A SARIF result is supposed to have one location, but a secret often contains multiple matches, so each match needs to be a separate location. To reconcile this, the result location starts at the start of the first match and ends at the end of the last match. Then each match is added as a "related" location.

This can be seen in this screenshot of a GitHub import of a ggshield SARIF document:

image(1)

I am open to suggestions regarding better ways to represent multi-match secrets.

## Validation

The code can be tested in different ways.

The first thing is to generate some SARIF report. This can be done with a command like this:

```
ggshield secret scan path -ry --use-gitignore . --format sarif -o myreport.sarif
```

Then you can try to use the generated report in different places:

- [SARIF website validator](https://sarifweb.azurewebsites.net/Validation)
- VS Code provides at least two SARIF extensions
- GitHub can import SARIF documents and display the results in the "Security" page of the project (see [GitHub doc on this topic](https://docs.github.com/en/enterprise-cloud@latest/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github))

## PR check list

### DIFF
--- a/ggshield/cmd/utils/common_options.py
+++ b/ggshield/cmd/utils/common_options.py
@@ -18,6 +18,7 @@ import click
 
 from ggshield.cmd.utils.context_obj import ContextObj
 from ggshield.cmd.utils.debug_logs import setup_debug_logs
+from ggshield.cmd.utils.output_format import OutputFormat
 from ggshield.core.config.user_config import UserConfig
 
 
@@ -187,13 +188,22 @@ def add_common_options() -> Callable[[AnyFunction], AnyFunction]:
     return decorator
 
 
+def _set_json_output_format(
+    ctx: click.Context, param: click.Parameter, value: Optional[bool]
+) -> Optional[bool]:
+    if value:
+        ctx_obj = ContextObj.get(ctx)
+        ctx_obj.output_format = OutputFormat.JSON
+    return value
+
+
 json_option = click.option(
     "--json",
     "json_output",
     is_flag=True,
     default=None,
     help="Use JSON output.",
-    callback=create_ctx_callback("use_json"),
+    callback=_set_json_output_format,
 )
 
 
--- a/ggshield/cmd/utils/common_options.py
+++ b/ggshield/cmd/utils/common_options.py
@@ -202,7 +202,7 @@ json_option = click.option(
     "json_output",
     is_flag=True,
     default=None,
-    help="Use JSON output.",
+    help="Shorthand for `--format json`.",
     callback=_set_json_output_format,
 )
 
@@ -222,7 +222,7 @@ def _create_format_option(
     return click.option(
         "--format",
         type=click.Choice([x.value for x in formats]),
-        help="Output format.",
+        help="Format to use for the output.",
         callback=_set_output_format,
     )
 
--- a/ggshield/cmd/utils/common_options.py
+++ b/ggshield/cmd/utils/common_options.py
@@ -12,7 +12,7 @@ The `kwargs` argument is required because due to the way click works,
 """
 
 from pathlib import Path
-from typing import Any, Callable, Optional, TypeVar
+from typing import Any, Callable, List, Optional, TypeVar
 
 import click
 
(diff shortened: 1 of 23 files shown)


ggshield/cmd/utils/common_options.py
```diff
@@ -207,6 +207,23 @@ def _set_json_output_format(
...
+        ctx_obj.output_format = OutputFormat(value)
+    return value
+
+
+format_option = click.option(
+    "--format",
+    type=click.Choice([x.value for x in OutputFormat]),
+    help="Output format.",
+    callback=_set_output_format,
+)
```
fnareoh, line 237: I really like the addition of the `--format` option but we should note that it creates different behavior depending on the flags order:
* `--format sarif --json` => json output 
* `--json --format sarif` => sarif output
* `--format json --format sarif` => sarif output
I am concerned with confusion for users (some may think we could output to several format at the same time), do we want to clarify that in our help message ? Or even fail when  we are given several options ?

fnareoh: Also, maybe you could add a changelog for the addition of this option ?

tests/unit/verticals/secret/output/test_sarif_output.py
```diff
@@ -0,0 +1,291 @@
...
+            {
+                "id": int,
+                "physicalLocation": SARIF_PHYSICAL_LOCATION_DICT_SCHEMA,
+                "message": {"text": str},
+            }
+        ],
+        "partialFingerprints": {"secret/v1": str},
+        VOptional("hostedViewerUri"): str,
+    }
+)
```
fnareoh, line 92: I don't know if it is worth the trouble but on a result on `scan path` the [Sarif validator that you linked](https://sarifweb.azurewebsites.net/Validation) points out other rules that we don't follow and means that we could not be imported by Azure DevOps Advanced Security service and GitHub Advanced Security service.

fnareoh (commented):
Great addition :D , I tested and I have a few questions on options flags. 

For the multi match secret it does not seem ideal but it is probably the best we can do. Do we know how Github Advance security & others displays those with SARIF format ? 
I tested with the validator and pycharm extension it looks good to me !

ggshield/cmd/utils/common_options.py
agateau-gg: > I really like the addition of the --format option but we should note that it creates different behavior depending on the flags order

Good point! I don't want to make it too complex by failing if both are set, but I am going to improve the --help output to mention that the last one set wins.

> Also, maybe you could add a changelog for the addition of this option ?

Oh yes, definitely!

agateau-gg:
> For the multi match secret it does not seem ideal but it is probably the best we can do. Do we know how Github Advance security & others displays those with SARIF format ?

I have no idea :/

> I tested with the validator and pycharm extension it looks good to me !

Oh, I could not find a Pycharm extension! Need to look again.

ggshield/cmd/utils/common_options.py
```diff
@@ -187,13 +188,52 @@ def decorator(cmd: AnyFunction) -> AnyFunction:
...
     "json_output",
     is_flag=True,
     default=None,
     help="Use JSON output.",
-    callback=create_ctx_callback("use_json"),
+    callback=_set_json_output_format,
+)
+
+
+def _set_output_format(
```
salome-voltz, line 210: So, are we planning to deprecate `--json` over time, or will we keep the command ?

salome-voltz (approved):
It looks good ! I tested it and it was great :smile: 
I agree with Garance, output might be a bit weird for multimatch secrets but it's a detail

fnareoh (approved):

Issue #924, opened 2024-06-27
`ggshield scan docker` fails with missing path error when run from the Docker image


cdupuis:
## Environment

- ggshield version: `gitguardian/ggshield:latest`
- Operating system (Linux, macOS, Windows):
- Operating system version:
- Python version:

## Describe the bug

I expect the following command to not fail:

```
❯ docker run -it -v /var/run/docker.sock:/var/run/docker.sock -e GITGUARDIAN_API_KEY=${GITGUARDIAN_API_KEY} gitguardian/ggshield ggshield secret scan docker alpine
Saving docker image...
Error: [Errno 2] No such file or directory: 'docker'

Re-run the command with --verbose to get a stack trace.
```

fnareoh:
Hi, thank you for creating this issue.
 I believe this is due to the docker command not being available inside your docker. 
 The error [occurs when trying to run this command](https://github.com/GitGuardian/ggshield/blob/ef2a491833aa82a54f4706996a861ad6059c2766/ggshield/verticals/secret/docker.py#L299)  to download the docker image alpine, but we will improve the error by catching it  and printing something more explicit. :smile:

cdupuis:
> I believe this is due to the docker command not being available inside your docker.

This is your official Docker image, not my image. So shouldn't you add the Docker CLI to the ggshield image in that case to make this error disappear?

fnareoh:
Ah yes sorry I got confused, yes we will !

cdupuis:
I can confirm that running `ggshield` from an image built from the following `Dockerfile` works as expected:

```
FROM gitguardian/ggshield:latest

RUN \
    apt-get update \
    && apt-get install -y --no-install-recommends ca-certificates curl gnupg \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/* \
    && curl -fsSL https://get.docker.com | sh
```

fnareoh:
Thanks a lot :pray: , I doubled checked and opened a PR. We will just wait for the return of @agateau-gg  (our main maintainer, returning probably next Wednesday) to validate !

PR #926, opened 2024-06-28
fix(docker): add docker to the image so ggshield secret scan docker works

fnareoh:
## Context

Addresses #924, thanks a lot @cdupuis for raising the issue and providing a fix !

## What has been done

This MR adds docker to the image so the command `ggshield secret scan docker alpine` can download the image and works out of the box.

## Validation

 `ggshield secret scan docker alpine` works on the new image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Keep image in sync with scripts/update-pipfile-lock/Dockerfile
 FROM python:3.10-slim as build
 
-LABEL maintainer="GitGuardian SRE Team <support@gitguardian.com>"
+LABEL maintainer="GitGuardian <support@gitguardian.com>"
 
 ENV LC_ALL C.UTF-8
 ENV LANG C.UTF-8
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -15,9 +15,10 @@ WORKDIR /app
 RUN \
     apt-get update \
     && apt-get dist-upgrade -y --no-install-recommends \
-    && apt-get install -y --no-install-recommends git openssh-client \
+    && apt-get install -y --no-install-recommends git openssh-client ca-certificates curl gnupg  \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/*\
+    && curl -fsSL https://get.docker.com | sh
```
agateau-gg, line 21: I am not fond of blindly pulling from get.docker.com. Since we only use `docker pull` and `docker save`, would it be enough to get the distro version instead?

agateau-gg (changes requested):

fnareoh:
I'll go ahead an close this PR as it will probably be tackled differently.

PR #933, opened 2024-07-17
update certifi

corenting:
## Context

The `certifi` dependency was at an old version with a low CVE due to a root certificate removed in recent versions.

## What has been done

Dependency was updated with pipenv.

## Validation

N/A

## PR check list


GG-HH (approved):
Sounds good to me. How did you notice it ? Was it disruptive in your workflow ?
Thanks

PR #940, opened 2024-07-31
fix: make sure `secret scan archive` does not overwrite files

agateau-gg:
## Context

`shutil.unpack_archive()` can overwrite files outside the destination directory when unpacking a malicious tar archive. An attacker could use this to overwrite files on the user machine.

This can only happen with tar archives, but the behavior with zip is not optimal either: bad files are silently ignored.

## What has been done

With this change, the `secret scan archive` command now aborts if the archive contains bad files or bad symbolic links, regardless of whether the archive is a tar or a zip. For extra safety, the `secret scan pypi` command does the same.

## Validation

From within ggshield repository, run this command:

```
ggshield secret scan archive tests/unit/data/archives/bad.tar
```

With the main branch, you should end up with two files in `/tmp`: `/tmp/bad-absolute` and `/tmp/bad-relative`.

With this branch, the command should fail like this:

```
Error: Failed to unpack "/home/agateau/src/ggshield/tests/unit/data/archives/bad.tar" archive: Unsafe archive.
Bad paths:
../bad-relative
/tmp/bad-absolute
Bad links:
./subdir/bad-relative-symlink -> ../../bad-relative
./subdir/bad-absolute-symlink -> /tmp/bad-absolute
```

## PR check list


ggshield/utils/archive.py
```diff
@@ -0,0 +1,124 @@
...
+import stat
+import tarfile
+from pathlib import Path
+from typing import List, Optional, Tuple
+from zipfile import ZipFile, ZipInfo
+
+
+class UnsafeArchive(Exception):
+    """
+    Raised when an archive is considered unsafe
```
fnareoh, line 11: Might be worth being a bit more specific ? 
```suggestion
    Raised when an archive is considered unsafe: if it could cause to write/overwrite or link outside the target folder when unpacking.
```

ggshield/utils/archive.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+
+    def __init__(self):
+        self.bad_paths: List[Path] = []
+        self.bad_links: List[Tuple[Path, Path]] = []
+
+    def __str__(self) -> str:
+        lines: List[str] = ["Unsafe archive."]
```
fnareoh, line 19: I think it could be good to be a bit more explicit ?
```suggestion
        lines: List[str] = ["Unsafe archive, it could cause to write/overwrite or link outside the target folder when unpacking. To avoid vulnerabilities to malicious attack ggshield refuses to unpack it."]
```

ggshield/utils/archive.py
```diff
@@ -0,0 +1,124 @@
...
+        base.joinpath(path).resolve().relative_to(base)
+        return False
+    except ValueError:
+        return True
+
+
+def _is_bad_link(path: Path, target: Path, base: Path) -> bool:
+    # Links are interpreted relative to the directory containing the link
+    link_base = base.joinpath(Path(path).parent).resolve()
+    target = link_base.joinpath(target).resolve()
```
fnareoh, line 73: This function is a bit harder to understand, maybe you could add an example ?

tests/unit/utils/test_archive.py
```diff
@@ -0,0 +1,92 @@
...
+)
+from tests.unit.conftest import DATA_PATH
+
+
+ARCHIVES_PATH = DATA_PATH / "archives"
+BAD_ZIP_PATH = ARCHIVES_PATH / "bad.zip"
+BAD_TAR_PATH = ARCHIVES_PATH / "bad.tar"
+GOOD_ZIP_PATH = ARCHIVES_PATH / "good.zip"
+GOOD_WHL_PATH = ARCHIVES_PATH / "good.whl"
+GOOD_TAR_PATH = ARCHIVES_PATH / "good.tar"
```
fnareoh, line 21: Our documentation says `Supported archive formats are zip, tar, tar.gz,  tar.bz2 and tar.xz.` 
Could it be worth explicitly adding .whl to the doc ? and should we also test good and bad for ` tar.gz,  tar.bz2 and tar.xz` ?

fnareoh (approved):
Looks good to me, just minor comments (mostly wording). :) 
At first, I was worried about blocking users by just failling on those bad archive, but as the command scan archive is barely used anyway, the risk is low.

fnareoh (approved):

PR #942, opened 2024-08-06
Add a test secret as environment variable to hide it from the dashboard

GG-HH:
## Context

We cannot run the functional tests on an instance monitoring the repo because the GG_VALID_SECRET has been marked as ignored which means it is ignored when scanning (see documentation).

## What has been done
- Add the ability to override the value with the .env so that we can have a secret not known by the dashboard

BEWARE: I did not modify the environment variables yet in the ggshield repo.

## Validation

- functional tests should run with an account monitoring ggshield

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,6 +9,7 @@ GITGUARDIAN_INSTANCE=https://dashboard.gitguardian.com/
 # it unless the default value does not work for you
 #TEST_UNKNOWN_SECRET=
 # Set this to a single match secret unknown by your dashboard with a valid status
-# This is required to run functional test on an instance with a ggshield repo integrated
+# This is required to run functional tests on an account monitoring ggshield repository,
+# where the token in conftest.py has been ignored
 TEST_GG_VALID_TOKEN=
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -8,3 +8,7 @@ GITGUARDIAN_INSTANCE=https://dashboard.gitguardian.com/
 # TEST_KNOWN_SECRET, this one has a default value, so you do not have to set
 # it unless the default value does not work for you
 #TEST_UNKNOWN_SECRET=
+# Set this to a single match secret unknown by your dashboard with a valid status
+# This is required to run functional test on an instance with a ggshield repo integrated
```
agateau-gg, line 12:
```suggestion
# Set this to a single match secret unknown by your dashboard with a valid status
# This is required to run functional tests on an account monitoring ggshield repository,
# where the token in conftest.py has been ignored
```

agateau-gg (changes requested):
Minor suggestion on the explanation. Looks good otherwise.

agateau-gg (approved):

PR #943, opened 2024-08-06
ci: use codecov action v4

GG-HH:
## Context
Jobs are failing for an obscure reason : see for instance https://github.com/GitGuardian/ggshield/actions/runs/10262514554/job/28392351544

Let's try a simple fix which is also good for tech debt.

## What has been done

 I changed the value

## Validation

Can you see the coverage in codecov ?


agateau-gg (approved):
Thanks for taking care of this 👍🏻.

PR #946, opened 2024-08-07
feat(performance): reduce the number of files scanned after merge conflict

fnareoh:
## Context

The goal of this Merge Request is to improve the speed of the ggshield pre-commit scan when commiting after a merge conflict.

## What has been done

I created a script `gen_merge_commit.py` (temporarily added at the repository root) that creates a repo with two branches (master and feature_branch) a merge commit (with or without commit depending on the options) giving the following commit graph:
```
*   17d3a27 (HEAD -> feature_branch) Merge branch 'master' into feature_branch
|\  
| * db20e84 (master) Commit master n°2
| * 53a4cd0 Commit master n°1
| * a78f68d Commit master n°0
* | 9f5717c Commit feature branch
|/  
* e1db622 Initial commit
```

I then experimented and with it:
* Without conflict: our pre-commit does not seem to be called on merge
* With conflict: I implemented a proof of concept way to limit the number of files scan: when ggshield pre-commit is called 
       - if MERGE_HEAD does not exists, we assume it is a regular commit and we should scan everything (current behavior)
       - if MERGE_HEAD exists: we list the files about to be committed (git diff --staged --name-only), then compare the current shas of those files  (using the command `git ls-files --staged <list_files_diff>`) with the shas of HEAD and MERGE_HEAD (obtained via `git ls-tree  <list_files_diff>`), and if it differs from both, then the file has been modified and should be re-scanned... 

I'm opening this PR as a draft as it is clearly not ready (no proper automated testing) but the general idea should be reviewed...

### DIFF
--- a/changelog.d/20240819_162411_fnareoh_scrt_4759_optimize_files_to_scan_on_merge_commits.md
+++ b/changelog.d/20240819_162411_fnareoh_scrt_4759_optimize_files_to_scan_on_merge_commits.md
@@ -1,0 +1,5 @@
+### Added
+
+- The command `ggshield secret scan pre-commit` has a new flag `--merge-skip-unchanged`. It is off by default, if activated,
+  in the case of merge commit, it skips the scan of files that were not modified by merge. This is done for efficiency
+  but is less secure.
--- a/ggshield/core/scan/commit.py
+++ b/ggshield/core/scan/commit.py
@@ -13,6 +13,9 @@ from .commit_utils import (
     PATCH_PREFIX,
     STAGED_PREFIX,
     CommitScannable,
+    get_diff_files,
+    get_file_sha_in_ref,
+    get_file_sha_stage,
     parse_patch,
 )
 from .scannable import Scannable
@@ -72,6 +75,39 @@ class Commit:
                 yield from parse_patch(sha, patch, exclusion_regexes)
 
         return Commit(sha, parser, info)
+
+    @staticmethod
+    def from_merge(
+        exclusion_regexes: Optional[Set[Pattern[str]]] = None,
+        merge_branch: str = "MERGE_HEAD",
+        cwd: Optional[Path] = None,
+    ) -> "Commit":
+
+        diff_files = get_diff_files(cwd=cwd)
+
+        shas_in_merge_branch = dict(
+            get_file_sha_in_ref(merge_branch, diff_files, cwd=cwd)
+        )
+        shas_in_head = dict(get_file_sha_in_ref("HEAD", diff_files, cwd=cwd))
+
+        files_to_scan = set()
+        for path, sha in get_file_sha_stage(diff_files, cwd=cwd):
+            # The file is either new or different from both HEAD and MERGE_HEAD
+            if sha not in {shas_in_head.get(path), shas_in_merge_branch.get(path)}:
+                files_to_scan.add(path)
+
+        def parser_merge(commit: "Commit") -> Iterable[Scannable]:
+            patch = git(
+                ["diff", "--staged", *PATCH_COMMON_ARGS, *files_to_scan], cwd=cwd
+            )
+            yield from parse_patch(
+                STAGED_PREFIX,
+                DIFF_EMPTY_COMMIT_INFO_BLOCK + patch,
+                exclusion_regexes,
+            )
+
+        info = CommitInformation.from_staged(cwd)
+        return Commit(sha=None, patch_parser=parser_merge, info=info)
 
     @staticmethod
     def from_staged(
--- a/ggshield/core/scan/commit_utils.py
+++ b/ggshield/core/scan/commit_utils.py
@@ -1,10 +1,10 @@
 import re
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Iterable, List, Optional, Pattern, Set
+from typing import Iterable, List, Optional, Pattern, Set, Tuple
 
 from ggshield.utils.files import is_path_excluded
-from ggshield.utils.git_shell import Filemode
+from ggshield.utils.git_shell import Filemode, git
 
 from .scannable import Scannable
 
@@ -235,3 +235,37 @@ def parse_patch(
             else:
                 msg = f"Could not parse patch: {exc}"
             raise PatchParseError(msg)
+
+
+def get_file_sha_in_ref(
+    ref: str,
+    files: List[str],
+    cwd: Optional[Path] = None,
+) -> Iterable[Tuple[str, str]]:
+    """
+    Helper function to get the shas of files in the git reference.
+    """
+    output = git(["ls-tree", "-z", ref] + files, cwd=cwd)
+    for line in output.split("\0")[:-1]:
+        _, _, sha, path = line.split()
+        yield (path, sha)
+
+
+def get_file_sha_stage(
+    files: List[str], cwd: Optional[Path] = None
+) -> Iterable[Tuple[str, str]]:
+    """
+    Helper function to get the shas currently staged of files.
+    """
+    output = git(["ls-files", "--stage", "-z"] + files, cwd=cwd)
+    for line in output.split("\0")[:-1]:
+        _, sha, _, path = line.split()
+        yield (path, sha)
+
+
+def get_diff_files(cwd: Optional[Path] = None) -> List[str]:
+    """
+    Helper function to get the files modified and staged.
+    """
+    output = git(["diff", "--name-only", "--staged", "-z"], cwd=cwd)
+    return output.split("\0")[:-1]  # remove the trailing \0
--- a/tests/functional/secret/test_merge_commit.py
+++ b/tests/functional/secret/test_merge_commit.py
@@ -1,0 +1,78 @@
+from pathlib import Path
+from subprocess import CalledProcessError
+
+import pytest
+
+from tests.functional.utils_create_merge_repo import (
+    SecretLocation,
+    generate_repo_with_merge_commit,
+)
+
+
+@pytest.mark.parametrize(
+    "with_conflict",
+    [
+        True,
+        False,
+    ],
+)
+@pytest.mark.parametrize(
+    "secret_location",
+    [
+        SecretLocation.MASTER_BRANCH,
+        SecretLocation.FEATURE_BRANCH,
+        SecretLocation.NO_SECRET,
+    ],
+)
+@pytest.mark.parametrize(
+    "merge_skip_unchanged",
+    [
+        True,
+        False,
+    ],
+)
+def test_merge_commit_no_conflict(
+    capsys,
+    tmp_path: Path,
+    with_conflict: bool,
+    secret_location: SecretLocation,
+    merge_skip_unchanged: bool,
+) -> None:
+
+    if (
+        secret_location == SecretLocation.MASTER_BRANCH
+        and with_conflict
+        and not merge_skip_unchanged
+    ):
+        with pytest.raises(CalledProcessError):
+            generate_repo_with_merge_commit(
+                tmp_path,
+                with_conflict=with_conflict,
+                secret_location=secret_location,
+                merge_skip_unchanged=merge_skip_unchanged,
+            )
+
+        # AND the error message contains the Gitlab Token
+        captured = capsys.readouterr()
+        assert "GitLab Token" in captured.err
+    else:
+        generate_repo_with_merge_commit(
+            tmp_path,
+            with_conflict=with_conflict,
+            secret_location=secret_location,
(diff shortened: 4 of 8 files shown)


agateau-gg (commented):
Looks good overall, your approach makes sense.

I am wondering if we always want this behavior or if it should be opt-in 🤔. What do you think?

One remark: calls to `ls-files` and `ls-tree` should use `-z` to handle files with what git calls "unusual names" (*). An easy way to create an "unusual name" is to create one with a double quote in it.

(*): Other parts of ggshield would need to be fixed with regard to that too...

tests/unit/cmd/scan/test_precommit.py
```diff
@@ -0,0 +1,87 @@
...
+
+@pytest.mark.parametrize(
+    "is_merge",
+    [True, False],
+)
+@pytest.mark.parametrize(
+    "with_conflict",
+    [True, False],
+)
+def test_is_merge(tmp_path, is_merge, with_conflict):
```
agateau-gg, line 22: I would split this test in 3:

When is_merge is False, without_conflict does not matter, so the is_merge=False part is executed twice.

The common code between with and without conflict is small enough that it can be duplicated. This common code can even be shortened a bit more: the test does not use "inital.md", so you don't need to create it: `create_commit()` uses `--allow-empty`.

tests/unit/cmd/scan/test_precommit.py
```diff
@@ -0,0 +1,87 @@
...
+        Path(tmp_path / "Other.md").write_text("Other")
+        repo.add(".")
+        repo.create_commit("Commit on master")
+
+        repo.checkout("feature_branch")
+        Path(tmp_path / "Another.md").write_text("Another")
+        repo.add(".")
+        repo.create_commit("Commit on feature_branch")
+
+        os.environ["GIT_REFLOG_ACTION"] = "merge master"  # Simulate merge
```
agateau-gg, line 84: You can use `monkeypatch.setenv()` to avoid permanent changes to the environment.

tests/unit/core/scan/test_commit.py
```diff
@@ -409,3 +409,46 @@ def test_from_staged(tmp_path):
 
     paths_and_modes = [(x.path, x.filemode) for x in files]
     assert paths_and_modes == [(Path("NEW.md"), Filemode.NEW)]
+
+
+def test_from_merge(tmp_path):
+    """
+    GIVEN  Commit instance created from a git merge (after conflict)
```
agateau-gg, line 416: nitpick: typo
```suggestion
    GIVEN a Commit instance created from a git merge (after conflict)
```

tests/functional/secret/test_merge_commit.py
agateau-gg, line 1: I would prefer to have two files here: one for the `generate_repo_with_merge_commit()` function + its wrapper script, and one for the tests using it.

agateau-gg (changes requested):

fnareoh:
@agateau-gg  I started doing something because it looked convenient in my tests and possibly useful but maybe you won't like it: I added an option to `ggshield install` [to install the hook with option](https://github.com/GitGuardian/ggshield/pull/946/files#diff-df89db71222f5e255170e083365b1d9ff6f20b5dd089ff0d579c357da867f871R155). 

For example, `ggshield install -o --json` will install  the following hook: `ggshield secret scan pre-commit --json "$@"`.

But as it seems to fail on windows, I would like to check with you before debugging... :sweat:

agateau-gg:
> @agateau-gg I started doing something because it looked convenient in my tests and possibly useful but maybe you won't like it: I added an option to `ggshield install` [to install the hook with option](https://github.com/GitGuardian/ggshield/pull/946/files#diff-df89db71222f5e255170e083365b1d9ff6f20b5dd089ff0d579c357da867f871R155).
> 
> For example, `ggshield install -o --json` will install the following hook: `ggshield secret scan pre-commit --json "$@"`.
> 
> But as it seems to fail on windows, I would like to check with you before debugging... 😓

The idea sounds good, but it should be in a separate commit or maybe even in a separate PR.

ggshield/core/scan/commit.py
```diff
@@ -73,6 +75,57 @@ def parser(commit: "Commit") -> Iterable[Scannable]:
...
+        exclusion_regexes: Optional[Set[Pattern[str]]] = None,
+        merge_branch: str = "MERGE_HEAD",
+        cwd: Optional[Path] = None,
+    ) -> "Commit":
+
+        def get_file_sha_from_ls_tree(
+            ref: str,
+            list_diff_files: List[str] = [],
+        ) -> Iterable[Tuple[str, str]]:
+            ls_tree = git(["ls-tree", "-z", ref] + list_diff_files, cwd=cwd)
```
agateau-gg, line 89: naming: maybe suffix this one with `_output` to become `ls_tree_output` like you did below for the diff output?

ggshield/core/scan/commit.py
```diff
@@ -73,6 +75,57 @@ def parser(commit: "Commit") -> Iterable[Scannable]:
...
+        list_diff_files = diff_output.split("\x00")[:-1]  # remove the trailing \x00
+
+        dict_path_to_shas = defaultdict(set)
+        for path, sha in chain(
+            get_file_sha_from_ls_tree("HEAD", list_diff_files),
+            get_file_sha_from_ls_tree(merge_branch, list_diff_files),
+        ):
+            dict_path_to_shas[path].add(sha)
+
+        ls_files = git(["ls-files", "--stage", "-z"] + list_diff_files, cwd=cwd)
```
agateau-gg, line 105: Same as `ls_tree` here: `ls_files` → `ls_files_output`.

ggshield/core/scan/commit.py
```diff
@@ -73,6 +75,57 @@ def parser(commit: "Commit") -> Iterable[Scannable]:
...
+            ls_tree = git(["ls-tree", "-z", ref] + list_diff_files, cwd=cwd)
+            for line in ls_tree.split("\x00")[:-1]:
+                mode, type, sha, path = line.split()
+                yield (path, sha)
+
+        # Get files modified and staged
+        diff_output = git(["diff", "--name-only", "--staged", "-z"], cwd=cwd)
+        list_diff_files = diff_output.split("\x00")[:-1]  # remove the trailing \x00
+
+        dict_path_to_shas = defaultdict(set)
```
agateau-gg, line 98: Same as `list_diff_files` here: `path_to_shas` would be nicer, but actually `shas_for_path` would be even nicer, because here `_to_` is used to express a mapping, but further down (`files_to_scan`) `_to_` is used because it's before a verb (I was stuck a bit wondering how a mapping could be a set 😅).

ggshield/core/scan/commit.py
```diff
@@ -73,6 +75,57 @@ def parser(commit: "Commit") -> Iterable[Scannable]:
 
         return Commit(sha, parser, info)
 
+    @staticmethod
+    def from_merge(
+        exclusion_regexes: Optional[Set[Pattern[str]]] = None,
+        merge_branch: str = "MERGE_HEAD",
+        cwd: Optional[Path] = None,
+    ) -> "Commit":
```
agateau-gg, line 83: Overall this function is a bit difficult to read because it mixes low-level aspects (parsing git command output) with high-level aspects (deciding which files to scan).

I made a few naming suggestions but the real readability improvement would probably be to turn `get_file_sha_from_ls_tree()` into a standalone function in this module and to create similar helper functions for the other low-level aspects, so that `from_merge()` only contains the high-level code.

agateau-gg (changes requested):

ggshield/core/scan/commit.py
fnareoh: Sorry and thanks for the pointers on how to improve naming!

ggshield/core/scan/commit.py
fnareoh: Should be better now :)

ggshield/core/scan/commit.py
```diff
@@ -73,6 +76,39 @@ def parser(commit: "Commit") -> Iterable[Scannable]:
...
+
+        files_to_scan = set()
+        for path, sha in get_file_sha_stage(diff_files, cwd=cwd):
+            # The file is either new or different from both HEAD and MERGE_HEAD
+            if sha not in {shas_in_head.get(path), shas_in_merge_branch.get(path)}:
+                files_to_scan.add(path)
+
+        def parser_merge(commit: "Commit") -> Iterable[Scannable]:
+            patch = git(
+                ["diff", "--staged"] + PATCH_COMMON_ARGS + list(files_to_scan), cwd=cwd
```
agateau-gg, line 101: nit, can be simplified to:

```suggestion
                ["diff", "--staged", *PATCH_COMMON_ARGS, *files_to_scan], cwd=cwd
```

This avoids creating an intermediate list for `files_to_scan`.

ggshield/core/scan/commit_utils.py
```diff
@@ -235,3 +235,37 @@ def parse_patch(
...
+    ref: str,
+    files: List[str],
+    cwd: Optional[Path] = None,
+) -> Iterable[Tuple[str, str]]:
+    """
+    Helper function to get the shas of files in the git reference.
+    """
+    ls_tree_output = git(["ls-tree", "-z", ref] + files, cwd=cwd)
+    for line in ls_tree_output.split("\0")[:-1]:
+        mode, type, sha, path = line.split()
```
agateau-gg, line 250: nit: you can omit `mode` and `sha` like you did in `get_file_sha_stage()`.

ggshield/core/scan/commit_utils.py
```diff
@@ -235,3 +235,37 @@ def parse_patch(
...
+
+def get_file_sha_in_ref(
+    ref: str,
+    files: List[str],
+    cwd: Optional[Path] = None,
+) -> Iterable[Tuple[str, str]]:
+    """
+    Helper function to get the shas of files in the git reference.
+    """
+    ls_tree_output = git(["ls-tree", "-z", ref] + files, cwd=cwd)
```
agateau-gg, line 248: super-nit: the nice thing about helper functions is you can use shorter variable names :), here `ls_tree_output` can be `output` without ambiguity.

ggshield/core/scan/commit.py
agateau-gg: Much nicer! ✨

agateau-gg (changes requested):
Some minor remarks, but this is much easier to read!

ggshield/core/scan/commit_utils.py
fnareoh: You are right, so many advantages to helper functions :sparkles:

changelog.d/20240819_162411_fnareoh_scrt_4759_optimize_files_to_scan_on_merge_commits.md
```diff
@@ -0,0 +1,5 @@
+### Added
+
+- The command `ggshield secret scan pre-commit` has a new flag `--merge-skip-unchanged`. It is off by default, if activated,
```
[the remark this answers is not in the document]
fnareoh: I think feared `skip-unchanged-merge` would make it sound like we skip entirely the merge commit when we only skip specific unchanged files ?  
For me as a feature that reduce the amount of what we scan, I would go for safety first and keep it as OFF by default. Plus it would mean keeping the current behavior as default which makes sense to me. 

But I don't have a strong opinion either and it has not been clearly discussed with Mathieu.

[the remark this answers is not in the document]
pierrelalanne: Alright, I understand. This won't be part of the 1.31.0 scheduled for tomorrow anyway, so we can still discuss this with Mathieu before the next release :)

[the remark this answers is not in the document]
agateau-gg: The name is not optimal, but I don't have anything better to propose yet. Need to think this through.

As for it being by default, I think it's better to make it opt-in: this flag is only safe if you are confident that all branches you are about to merge have already been scanned. I don't think this is true in all situations.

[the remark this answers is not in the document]
fnareoh: After discussion with @mathieubellon `--skip-unchanged-merge-files` is good, long but clear and for now we keep it as an option deactivated by default but in the future we may change it to activate it by default (as well as other flags) to improve default performances.

agateau-gg (approved):

PR #947, opened 2024-08-08
agateau/refactor get files from path

agateau-gg:
## Context

`core.scan.file.get_files_from_path()` does too many things, and gets in the way of some cleaning I plan to do on ggshield UI. This PR reduces its scope.

## What has been done

`core.scan.file.get_files_from_path()` scope has been reduced. Most of the work is done in the 1st and 2nd commit. Look at their commit messages for details.

`iac scan` and `sca scan` no longer list skipped binary files. This has been discussed upfront and decided to be OK.

Output of `secret scan path`, `secret scan archive` and `secret scan pypi` is a bit nicer: it now makes use of headings:

image

image

## Validation

Tests still pass. UI looks a bit nicer. Code is nicer.

### DIFF
--- a/ggshield/core/scan/file.py
+++ b/ggshield/core/scan/file.py
@@ -46,14 +46,14 @@ class File(Scannable):
                 )
 
 
-def get_files_from_paths(
+def create_files_from_paths(
     paths: List[Path],
     exclusion_regexes: Set[Pattern[str]],
     list_files_mode: ListFilesMode = ListFilesMode.GIT_COMMITTED_OR_STAGED,
 ) -> Tuple[List[Scannable], List[Path]]:
     """
-    Returns a tuple of a list of scannables for the files, and a list of ignored binary
-    paths
+    Create File instances for `paths` and return them, as well as a list of the ignored
+    paths found in `paths`.
     """
     filepaths = list_files(
         paths,
--- a/ggshield/core/scan/file.py
+++ b/ggshield/core/scan/file.py
@@ -1,12 +1,7 @@
 from pathlib import Path
 from typing import List, Pattern, Set, Tuple, Union
 
-from ggshield.utils.files import (
-    ListFilesMode,
-    get_filepaths,
-    is_path_binary,
-    url_for_path,
-)
+from ggshield.utils.files import ListFilesMode, is_path_binary, list_files, url_for_path
 
 from .scannable import Scannable
 
@@ -60,7 +55,7 @@ def get_files_from_paths(
     Returns a tuple of a list of scannables for the files, and a list of ignored binary
     paths
     """
-    filepaths = get_filepaths(
+    filepaths = list_files(
         paths,
         exclusion_regexes,
         list_files_mode=list_files_mode,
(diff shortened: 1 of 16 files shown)


ggshield/core/scan/file.py
```diff
@@ -57,65 +54,25 @@ def _read_content(self) -> None:
...
-    Create a scan object from files content.
-
-    :param paths: List of file/dir paths from the command
-    :param yes: Skip confirmation option
-    :param display_scanned_files: In some parts of the code (e.g. SCA), we might want
-    to display a processed list instead and set this to False
-    :param display_binary_files: Display all ignored binary files
-    :param ignore_git: Ignore that the folder is a git repository
+    Returns a tuple of a list of scannables for the files, and a list of ignored binary
+    paths
```
gg-mmill, line 61:
```suggestion
    Returns a tuple of a list of scannables for the files, and a list of paths for  binary
    files
```
Nit: at this point, the binary files are not ignored

ggshield/verticals/sca/file_selection.py
```diff
@@ -48,23 +48,14 @@ def get_all_files_from_sca_paths(
...
-            list_files_mode=(
-                ListFilesMode.ALL
-                if ignore_git
-                else ListFilesMode.GIT_COMMITTED_OR_STAGED
-            ),
-        )
-    ]
-
-    return [str(x.relative_to(path)) for x in sorted(paths)]
+    files, _ = get_files_from_paths(
```
gg-mmill, line 51: Ok, so what we're not displaying anymore are the binary files (that are ignored anyway), but returned nonetheless by this function ? (I thought at first that what was not displayed anymore was the *excluded* binary files).

Ok to me !

gg-mmill (approved):
Small suggestion for a docstring, otherwise good to me !

ggshield/core/scan/file.py
agateau-gg: They are ignored in that the function did not create `File` instances for them. So `secret scan` is not going to look at them. This might a bit confusing when looking at `sca|iac scan` because there the code only cares about the paths, it ignores the created File instances.

We should actually be able to make iac|sca code directly use `get_filepaths()` which would be simpler. It only needs to gain support for excluding binaries at that point, but I leave that for future refactorings :).

With this in mind I think the current comment is OK, or maybe it should say "Returns a tuple of a list of **File** instances for the **paths**, and a list of ignored binary paths". What do you think?

gg-mmill: Ok, I think what confuses me is the name of the function, maybe a clearer name would  be `get_scannables_from_paths`  (in this case, the binary files in the output are not scannable, so they can indeed said to be "ignored")

salome-voltz (approved):
I like what you did with `--recursive` option, it's much better this way!

ggshield/core/scan/file.py
agateau-gg: I just went ahead and added a few commits so that iac|sca does not use `get_files_from_paths()` anymore, improved doc and naming. Let me if this looks better. 

(Reviewing commit by commit is going to be simpler)

gg-mmill (approved):
Good to me !

Goop-arsh:
Run

PR #949, opened 2024-08-12
Fix JSON output to remove offset

salome-voltz:
## Context

Bug found in ggshield 1.29.0 in the JSON Output: the position of the secret is incorrect.

### Steps to reproduce

1. Checkout ggshield v1.28.0 / v1.29.0 
2. Run `ggshield secret scan --json path secret.txt` on the following doc:

```yaml
Secret in line 2 :
password = "jdozjizn647483&&dkozep"
```

The outputs are different, and errors are highlighted in red : (only the `occurrences` key is displayed here, as it is the only one that differs.)

**ggshield v1.28.0:**

```json
"occurrences": [
  {
      "match": "jdoz**************ozep",
      "type": "password",
      "line_start": 2,
      "line_end": 2,
      "index_start": 12,
      "index_end": 34,
      "pre_line_start": 2,
      "pre_line_end": 2
  }
],
```

**ggshield v1.29.0:**

```json
"occurrences": [
  {
      "match": "jdoz**************ozep",
      "type": "password",
      "line_start": 1,
      "line_end": 1,
      "index_start": 31,
      "index_end": 52,
      "pre_line_start": 2,
      "pre_line_end": 2
  }
],
```

## What has been done

- Remove line offset previously added when scanning files (not patches). 
- Use line-base indexing instead of file-based indexing.

 ## Validation

-> Pipeline is still passing 

## PR check list


tests/unit/verticals/secret/output/test_json_output.py
```diff
@@ -273,7 +371,72 @@ def test_json_output(client, cache, name, input_patch, expected_exit_code):
...
+        ("single_line_secret", _SINGLE_LINE_SECRET_FILE, 1),
+        ("one_line_and_multiline_secrets", _ONE_LINE_AND_MULTILINE_FILE, 1),
+    ],
+    ids=[
+        "_MULTILINE_SECRET_FILE",
+        "_SINGLE_LINE_SECRET_FILE",
+        "_ONE_LINE_AND_MULTILINE_FILE",
+    ],
+)
+def test_json_output_for_files(
```
agateau-gg, line 399: I think we should be able to avoid the duplication between `test_json_output_for_patches()` and this function by merging `MATCH_INDICES_FOR_FILES` and `MATCH_INDICES_FOR_PATCH` and adding some logic to determine if the content is a file or a patch (either via a boolean argument or by looking at the start of the content)

ggshield/verticals/secret/output/secret_json_output_handler.py
```diff
@@ -109,6 +110,36 @@ def flattened_policy_break(
...
+                "line_end": match.line_end,
+                "index_start": match.span.column_index_start,
+                "index_end": match.span.column_index_end,
+            }
+            if match.pre_line_start is not None and match.pre_line_end is not None:
+                flattened_match_dict["pre_line_start"] = match.pre_line_start
+                flattened_match_dict["pre_line_end"] = match.pre_line_end
+            if match.post_line_start is not None and match.post_line_end is not None:
+                flattened_match_dict["post_line_start"] = match.post_line_start
+                flattened_match_dict["post_line_end"] = match.post_line_end
```
agateau-gg, line 143: I think the CI is unhappy because the `{pre|post}_line_{start|end}` are not always present.

agateau-gg (changes requested):

tests/unit/verticals/secret/output/test_json_output.py
salome-voltz: Added boolean `is_patch` on the test cases :+1:

ggshield/verticals/secret/output/secret_json_output_handler.py
salome-voltz: The CI was unhappy for two reasons : 
1. carriage return on windows
2. MatchSchema removed the key "type" as the expected key was "match_type". 

Regarding `{pre|post}_line_{start|end}`, I thought it was more logical to consistently set them to `None` when scanning a file and not a patch, as the notion of pre/post-commit doesn't exist in this context. This does change the actual behavior of ggshield however. If we don't want to do it, I can simply revert the changes on extended_match.py :smile:

tests/functional/secret/test_scan_path.py
agateau-gg, line 1: These debug comments can be removed

tests/unit/verticals/secret/output/test_json_output.py
```diff
@@ -238,10 +353,22 @@ def check_occurrences_indices(occurrences: List[Dict[str, Any]], patch: str) ->
...
+def test_json_output_for_patch(
+    client, cache, name, input, expected_exit_code, is_patch, tmp_path
+):
+    if is_patch:
+        c = Commit.from_patch(input)
+    else:
+        test_file = tmp_path / "file"
+        with open(test_file, "w", newline="\n") as f:
+            f.write(input)
+        file = File(path=test_file)
```
agateau-gg, line 370: minor: you can avoid the `if is_patch` below by creating a scannable list in each `if` branch here. Something like this:

```python
    if is_patch:
        commit = Commit.from_patch(input)
        scannables = commit.get_files()
    else:
        test_file = tmp_path / "file"
        with open(test_file, "w", newline="\n") as f:
            f.write(input)
        scannables = [File(path=test_file)]
```

Then pass `scannables` to `scanner.scan()`.

changelog.d/20240812_134908_salome.voltz_fix_json_output_handler.md
```diff
@@ -0,0 +1,3 @@
+### Fixed
+
+- JSON Output now displays correct position (line and index) for matches when scanning a file and not a patch.
```
agateau-gg, line 3:
```suggestion
- JSON output: fixed incorrect values for line and index when scanning a file and not a patch.
```

agateau-gg (changes requested):
Minor remarks, looks good overall!

tests/functional/secret/test_scan_path.py
salome-voltz: sorry :sweat:

agateau-gg (approved):
Looks good!

PR #955, opened 2024-09-04
fix(naming): rename option to skip merge files

fnareoh:
## Context

I failed and forgot to push my changes to the naming  :facepalm:

## PR check list

### DIFF
--- a/ggshield/cmd/secret/scan/precommit.py
+++ b/ggshield/cmd/secret/scan/precommit.py
@@ -40,8 +40,7 @@ def check_is_merge_with_conflict(cwd: Path) -> bool:
 
 @click.command()
 @click.option(
-    "--merge-skip-unchanged",
-    "-m",
+    "--skip-unchanged-merge-files",
     is_flag=True,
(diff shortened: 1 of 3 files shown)


ggshield/cmd/secret/scan/precommit.py
```diff
@@ -40,7 +40,7 @@ def check_is_merge_with_conflict(cwd: Path) -> bool:
 
 @click.command()
 @click.option(
-    "--merge-skip-unchanged",
+    "--skip-unchanged-merge-files",
     "-m",
```
agateau-gg, line 44: Not exactly related, but think we should not have a short option for this. Users are unlikely to write this command by hand, so there is no productivity gain there. When using the option in a script file or in a .pre-commit-config.yaml, the long option is more self-documenting.

agateau-gg (approved):
Minor remark, but looks good.

ggshield/cmd/secret/scan/precommit.py
fnareoh: Good point thanks :) I'll remove it.

PR #957, opened 2024-09-10
Fix git parser

agateau-gg:
## Context

This PR fixes two problems:

Problem 1: sometimes the path for a diff is wrong. This is because `commit_utils.parse_patch()` assumes that the diffs are in the same order as the filenames in the commit header, which turns out to not always be true.

Problem 2: turning merge commits into individual commits by calling `git show` with the `-m` option does not work when fetching only some files inside the commit (which we do when scanning huge commits): it always return the changes from the first parent.

## What has been done

For problem 1: get the path name from the `+++ b/path` or `--- a/path` lines of the diff header.

For problem 2: maybe it would be possible to return the changes from the correct parents, but it felt simpler to stop using the `-m` option.

As a bonus, it simplifies `parse_patch()` because it can now assume that there is only one commit per patch.

## Validation

Tests still pass.

## PR check list

### DIFF
--- a/ggshield/core/scan/commit_utils.py
+++ b/ggshield/core/scan/commit_utils.py
@@ -19,7 +19,6 @@ PATCH_PREFIX = "patch"
 HEADER_COMMON_ARGS = [
     "--raw",  # shows a header with the files touched by the commit
     "-z",  # separate file names in the raw header with \0
-    "-m",  # split multi-parent (aka merge) commits into several one-parent commits
 ]
 
 # Command line arguments passed to `git show` and `git diff` to get parsable patches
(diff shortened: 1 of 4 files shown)


ggshield/core/scan/commit_utils.py
```diff
@@ -193,48 +198,107 @@ def parse_patch(
     """
     Parse a patch generated with `git show` or `git diff` using PATCH_COMMON_ARGS.
 
-    If the patch represents a merge commit, then `patch` actually contains multiple
-    commits, one per parent, because we call `git show` with the `-m` option to force it
-    to generate one single-parent commit per parent. This makes later code simpler and
-    ensures we see *all* the changes.
+    patch looks like this:
```
fnareoh, line 201: Nit: "A patch" ? 
Could also be good to add more details about what this function return, maybe something along the line of "this functions extracts the paths of files modified by the commit" ?

tests/unit/core/scan/test_commit.py
```diff
@@ -392,6 +388,30 @@ def test_from_sha(
...
+
+    sha = repo.get_top_sha()
+    commit = Commit.from_sha(sha, cwd=tmp_path)
+    files = list(commit.get_files())
+
+    for file in files:
+        last_line = file.content.splitlines()[-1]
+        assert last_line == f"+{file.path.name}"
+
+
```
fnareoh, line 414: Do you think it would be doable to add a test that in the case of a merge commit we take into account the change from both parents ?

fnareoh (approved):
Thanks for this work !

tests/unit/core/scan/test_commit.py
agateau-gg: I think `scenario_merge_with_changes` does this, but maybe I misunderstand what you are suggesting?

agateau-gg: I just added a test to inspect the content of what we scan in case of a merge commit with conflicts.

PR #960, opened 2024-09-12
Add default behavior to ggshield secret ignore 

salome-voltz:
## Context

Command `ggshield secret ignore` had no default behavior. To use it in the vscode extension, we need to be able to ignore only one secret. 

## What has been done

Default behavior is now : `ggshield secret ignore HASH` to ignore a secret using its hash. There is also an optional option `--name` to name the secret in the `gitguardian.yaml`. 

## Validation

- Get a secret hash
- Run `ggshield secret ignore _hash_ --name _name_`
- The secret and its name are added to the `gitguardian.yaml` 

## PR check list


ggshield/cmd/secret/ignore.py
```diff
@@ -7,41 +7,69 @@
 from ggshield.core.cache import Cache
 from ggshield.core.config import Config
 from ggshield.core.text_utils import pluralize
+from ggshield.core.types import IgnoredMatch
 
 
 @click.command()
+@click.argument(
+    "hash",
```
agateau-gg, line 15: We should call this `match` instead of `hash`, to be consistent with the way it's called in the `.gitguardian.yaml` file.

ggshield/cmd/secret/ignore.py
```diff
@@ -7,41 +7,69 @@
...
+                "Option `--last-found` cannot be used with `HASH` or `--name`."
+            )
         nb = ignore_last_found(config, ctx_obj.cache)
-        path = config.config_path
-        secrets_word = pluralize("secret", nb)
-        click.echo(
-            f"Added {nb} {secrets_word} to the `secret.ignored-matches` section of {path}."
-        )
+    else:
+        match = IgnoredMatch(name=name if name else "", match=hash)
```
agateau-gg, line 64: I think we should make `--name` mandatory to encourage the good practice of naming secrets. Someone who does not want to do so can still use `--name ''`, but at least it's not the path of least resistance.

tests/unit/cmd/test_ignore.py
```diff
@@ -27,6 +28,51 @@ def compare_matches_ignore(match):
...
+    )
+    config = Config()
+
+    with patch("ggshield.cmd.utils.context_obj.ContextObj.get") as mock_get_ctx:
+        mock_get_ctx.return_value.config = config
+
+        cmd = ["secret", "ignore", ignored_match.match]
+        result = cli_fs_runner.invoke(cli, cmd, color=False, catch_exceptions=False)
+
+        assert result.exit_code == 0
```
agateau-gg, line 49: You should use `assert_invoke_ok()` to test the result of invoke(). This ensures that in case of failure the output of the command is printed.

tests/unit/cmd/test_ignore.py
```diff
@@ -27,6 +28,51 @@ def compare_matches_ignore(match):
...
+    """
+    GIVEN any config and cache
+    WHEN I run the command with invalid arguments
+    THEN an error should be raised
+    """
+
+    cmd = ["secret", "ignore", "some_hash", "--last-found"]
+    result = cli_fs_runner.invoke(cli, cmd, color=False, catch_exceptions=False)
+
+    assert result.exit_code == 2
```
agateau-gg, line 70: Here you can use `assert_invoke_exited_with(result, ExitCode.USAGE_ERROR)`.

changelog.d/20240912_120216_salome.voltz_add_ignore_hash_cmd.md
```diff
@@ -0,0 +1,3 @@
+### Added
+
+- Default behavior for `ggshield secret ignore` command to allow ignoring secrets using their hash with optional `--name` argument.
```
agateau-gg, line 3: "Default behavior" is a bit too implementation oriented.

```suggestion
- It is now possible to ignore a secret manually using `ggshield secret ignore MATCH --name NAME`.
```

agateau-gg (changes requested):

ggshield/cmd/secret/ignore.py
salome-voltz: I discussed this with Mathieu, and to summarize, the key point is that the `--name` option is only really helpful when well used. Forcing users to provide a name could lead to frustration, as they may input poorly chosen names just to bypass the requirement. In the end, these poorly named secrets would be as useless as no name.

ggshield/cmd/secret/ignore.py
agateau-gg: Actually I changed my mind: in the text output of scans we use "Secret SHA". Since I expect the use case for this would be users running `ggshield secret scan ...`, getting a list of secrets and wanting to remove them, then it makes sense to use a similar name there.

salome-voltz: Makes sense, do you prefer me to put SHA or SECRET_SHA here?

ggshield/cmd/secret/ignore.py
agateau-gg: @mathieubellon ggshield is, on purpose, a source of frustration 😅. It gets in the way of users to prevent bad practices.

If the option is not mandatory I suspect it will just never be used because users won't even be aware of it.

ggshield/cmd/secret/ignore.py
agateau-gg: SECRET_SHA because it's less ambiguous.

ggshield/cmd/secret/ignore.py
mathieubellon: I would really like to enforce an option that will necessarily provide value (and not maybe provide value) and [our own example](https://github.com/GitGuardian/ggshield/blob/main/.gitguardian.yaml) shows that it is not the case.

If users need it, they will find it (like --ignore-known-secret)

I am also totally for finding the right solution to fix this problem (display secret in clear so it can be found?)

> ggshield is, on purpose, a source of frustration 😅

In its mission maybe, in its usage I would think (and love) not. I think our goal should be to provide the best user experience so it is actually used (and thus enforces good practice). It is the difference between Prettier and Go format for example IMO

mathieubellon: For example that would be better if ggshield actually automatically updates the yaml with useful informations (line number, file/commit, etc ..), maybe from the .cache_ggshield file ?

agateau-gg: > [our own example](https://github.com/GitGuardian/ggshield/blob/main/.gitguardian.yaml) shows that it is not the case

Many `name` fields in our config are empty because the matches were migrated from the v1 config format, which allowed users to use the secret SHA directly, without any `name` field, see this commit which did the migration for more context https://github.com/GitGuardian/ggshield/commit/11e3f0253e2774151feafe7f822d19866fa0d157. It would have been very painful to figure out which secrets the SHA were referring to, so the names were left blank. With hindsights, it would have been better to add a comment in them instead. 

> For example that would be better if ggshield actually automatically updates the yaml with useful informations (line number, file/commit, etc ..), maybe from the .cache_ggshield file ?

That's an interesting idea, the problem is the cache is regularly cleared so it would not be 100% reliable. What could be done is: if `--name` is not passed, prompt the user for a name, with the default value populated from the cache if available. What do you think?

mathieubellon: > if --name is not passed, prompt the user for a name, with the default value populated from the cache if available. What do you think?

That mean also that in case of the cli used in a scripted context the name is mandatory
Ok anyway, let's go with this

ggshield/cmd/secret/ignore.py
```diff
@@ -7,41 +7,74 @@
...
-        secrets_word = pluralize("secret", nb)
-        click.echo(
-            f"Added {nb} {secrets_word} to the `secret.ignored-matches` section of {path}."
-        )
+    else:
+        if not name:
+            raise click.UsageError(
+                "Option `--name` is required when ignoring a secret."
+            )
+        ignored_match = IgnoredMatch(name=name if name else "", match=secret_sha)
```
agateau-gg, line 69:
```suggestion
        ignored_match = IgnoredMatch(name=name, match=secret_sha)
```

ggshield/cmd/secret/ignore.py
```diff
@@ -7,41 +7,74 @@
...
+        config.add_ignored_match(ignored_match)
+        nb = 1
+
+    config.save()
+    secrets_word = pluralize("secret", nb)
+    click.echo(
+        f"Added {nb} {secrets_word} to the secret.ignored-matches section of {path}."
```
agateau-gg, line 76: Sorry, I am going to do a "while you are at it"...

Can you replace `ignored-matches` with `ignored_matches`?

agateau-gg (approved):
2 minor remarks. Feel free to merge once they are fixed.

PR #963, opened 2024-09-20
Port to pdm

agateau-gg:
## Context

This PR makes ggshield use PDM instead of Pipenv.

## What has been done

- Port to PDM
- Update CI
- Update doc

## Validation

- GIVEN a clean environment (no activated virtualenv)
- WHEN following instructions from doc/dev/getting-started.md
- THEN we get a working install
- AND we can run tests
- AND we can run scriv

## PR check list


Mliveon (approved):

.github/workflows/ci.yml
```diff
@@ -34,8 +34,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pipenv==2023.12.1 pre-commit
-          pipenv install --dev --skip-lock
+          python -m pip install --upgrade pdm pre-commit
+          pdm sync --dev
```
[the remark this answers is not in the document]
agateau-gg: Unlike pipenv, pdm lock files should work on all Python versions, so we can make use of them here.

deronnax, line 38: contrary to pipenv, pdm sync `dev` by default, so you don't need the `--dev`.

.github/workflows/build_release_assets.yml
```diff
@@ -26,11 +26,11 @@ jobs:
 
       - name: Install packaging tools
         run: |
-          pip install build
+          pip install --upgrade pdm
```
deronnax, line 29: you could use this [setup PDM ](https://github.com/pdm-project/setup-pdm) GitHub Action that PDM itself uses. And PDM versions should be pinned, I think.

.github/workflows/build_release_assets.yml
```diff
@@ -141,14 +141,8 @@ jobs:
...
           $PYTHON_CMD -m pip install --upgrade pip
-          $PYTHON_CMD -m pip install --upgrade \
-            pipenv==2023.12.1
-
-          pipenv install --dev
-          pipenv run pip install pyinstaller==6.7.0
-        env:
-          # Disable lock otherwise Windows-only dependencies like colorama are not installed
-          PIPENV_SKIP_LOCK: 1
+          $PYTHON_CMD -m pip install --upgrade pdm
```
deronnax, line 144: same, I would say [setup PDM](https://github.com/pdm-project/setup-pdm) + pinned version. In 2 different steps, as PDM is not a dependancy.

Dockerfile
```diff
@@ -19,11 +17,11 @@ RUN \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip3 install pipenv==2023.12.1
+RUN pip3 install --upgrade pdm
```
deronnax, line 20: If you install only prod requirements, you don't need PDM, because runtime dependencies, PEP 621, is a standard that pip undestands

Dockerfile
```diff
@@ -19,11 +17,11 @@ RUN \
...
 
 COPY . .
 
-RUN pipenv install --ignore-pipfile --deploy
+RUN pdm sync --prod --no-editable
```
deronnax, line 24: so you should only need `pip install .` here

Makefile
```diff
@@ -16,32 +16,32 @@ all:
...
 isort:
-	pipenv run isort **/*.py
+	pdm run isort **/*.py
 
 lint: isort black flake8
 
-update-pipfile-lock:
-	scripts/update-pipfile-lock/update-pipfile-lock
+lock:
+	pdm lock --group :all
```
deronnax, line 47: the whole Makefile is basically just providing default arguments to your tools. It could be replaced by PDM user scripts, it would simpler and more pythonic. Or just deleted, it provides no actual value.

doc/dev/getting-started.md
```diff
@@ -11,7 +11,7 @@
 1. Install dev packages and environment
 
    ```sh
-   pipenv install --dev --skip-lock
+   pdm install --dev
```
deronnax, line 14:
```suggestion
   pdm install
```

pyproject.toml
```diff
@@ -1,3 +1,114 @@
...
+    "oauthlib>=3.2.1,<3.3.0",
+    "pygitguardian>=1.17.0,<1.18.0",
+    "pyjwt>=2.6.0,<2.7.0",
+    "python-dotenv>=0.21.0,<0.22.0",
+    "pyyaml>=6.0.1,<6.1",
+    "requests>=2.32.0,<2.33.0",
+    "rich>=12.5.1,<12.6.0",
+    "typing-extensions>=4.12.2,<5.0.0",
+    "urllib3>=2.2.2,<2.3.0",
+]
```
deronnax, line 53: I would take the effort to move these requirements to the `~=`, that's pretty much the norm nowaday. I find the current form hard to read (and I bet a lot of other people do, this form has become pretty uncommon).

pyproject.toml
```diff
@@ -1,3 +1,114 @@
...
+    "flake8-isort",
+    "flake8-quotes",
+    "pre-commit",
+    "pyright==1.1.367",
+    "build>=1.2.2.post1",
+]
+
+tests = [
+    # pinning because of conflicting dependencies with exceptiongroup",
+    "pytest==7.2.1",
```
deronnax, line 80: down to the patch version ? It means even 7.2.2 conflicts?

pyproject.toml
```diff
@@ -1,3 +1,114 @@
...
+    "pytest-mock",
+    "pytest-socket",
+    "pytest-voluptuous",
+    "seed-isort-config",
+    "snapshottest",
+    "import-linter",
+    "jsonschema",
+    "vcrpy>=5.1.0",
+    # Dependency of pytest-voluptuous. Pin it because 0.15.0 does not",
+    # support Python 3.8 anymore.",
```
deronnax, line 90: extra commas, but I would actually reword it to make it one line:
```suggestion
    # pytest-voluptuous dependency . Pinned because 0.15.0 doesn't support Python 3.8 anymore.
```

pyproject.toml
```diff
@@ -1,3 +1,114 @@
...
+    "voluptuous<0.15.0",
+    # Pin because 5.6.0 currently (2024-07-15) does not work on Windows",
+    "pyfakefs>=5.2.0,<5.6.0",
```
deronnax, line 93: this "currently" is 5 month old and many several versions have been released meanwhile, and even so, the [5.2.2](https://pypi.org/project/pyfakefs/5.2.3/) claims to work on Windows, so I don't get it

pyproject.toml
```diff
@@ -1,3 +1,114 @@
...
+    "pyinstaller==6.7.0",
+]
+
+[tool.pdm.resolution.overrides]
+# vcrpy requires urllib3 < 2 if Python is < 3.10, but it works for us with
+# Python 3.9, so force urllib3 version.
+urllib3 = "2.2.3"
+
+# pyinstaller limits Python version to < 3.13, making it incompatible with us
+pyinstaller = "6.7.0"
```
deronnax, line 106: is it needed with the `standalone` group above where pyinstaller is pinned to `pyinstaller==6.7.0`?

deronnax (approved):
few remarks, but you do you

Dockerfile
agateau-gg: Oh, that's a great point! Going to do this.

pyproject.toml
agateau-gg: I don't think that's true anymore, we need to revisit it. But that's out of the scope of this PR.

pyproject.toml
agateau-gg: Oh, my search-and-replace to convert dependencies caused quite a mess with comments 😅.

pyproject.toml
agateau-gg: This comment does not say 5.2.2 does not work on Windows. It says 5.6 does not.

pyproject.toml
agateau-gg: Yes, pyinstaller 6.7.0 wants Python < 3.13. This conflicts with us requiring Python >= 3.8.1

pyproject.toml
agateau-gg: Let's give this a try.

.github/workflows/build_release_assets.yml
agateau-gg: Should be good now, I have changed to the action where it made sense.

PR #965, opened 2024-09-23
Fix output handling of merge commits

agateau-gg:
## Context

GGShield output handlers currently do not support multi-parent commits. When they try to output them, they fail with list index errors:

```
$ ggshield secret scan commit-range HEAD^..
Error: list index out of range
Error: list index out of range
Scanning... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 2 / 2

No secrets have been found
```

This only happens in the `main` branch, not in the current release.

## What has been done

Workaround this by post-processing the diff to turn it into a single-parent commit.

## Validation

Using the same repository as in "Context":

```
$ ggshield secret scan commit-range HEAD^..
Scanning... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 2 / 2

commit ca68e177596982fa38f181aa9944340a359748d2
Author: Aurelien Gateau <aurelien.gateau@gitguardian.com>
Date: Thu Sep 5 18:39:58 2024 +0200

> commit://ca68e177596982fa38f181aa9944340a359748d2/f: 1 incident detected

>> Secret detected: Username Password
   Validity: No Checker
   Occurrences: 1
   Known by GitGuardian dashboard: NO
   Incident URL: N/A
   Secret SHA: b90d21fc2d768dd94ff5fe69afec948ce10dfd53dc56303b5c688db9f40bfcb3

    | @@ -1,1 +1,2 @@
1   | -baz
...
  1 | +username=billybob
                |_username_|
  2 | +password=R34R_cZFFFrzfrzArz
                |____password____|
```

## PR check list

(skip-changelog because this bug is not in the current release)

### DIFF
--- a/ggshield/core/scan/commit_utils.py
+++ b/ggshield/core/scan/commit_utils.py
@@ -39,6 +39,10 @@ _RX_HEADER_FILE_LINE_SEPARATOR = re.compile("[\n\0]:", re.MULTILINE)
 OLD_NAME_RX = re.compile(r"^--- a/(.*?)\t?$", flags=re.MULTILINE)
 NEW_NAME_RX = re.compile(r"^\+\+\+ b/(.*?)\t?$", flags=re.MULTILINE)
 
+MULTI_PARENT_HUNK_HEADER_RX = re.compile(
+    r"^(?P<at>@@+) (?P<from>-\d+(?:,\d+)?) .* (?P<to>\+\d+(?:,\d+)?) @@+$"
+)
+
 
 class PatchParseError(Exception):
     """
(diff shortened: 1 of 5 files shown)


agateau-gg:
> Looks good, just a question about design choices.
> 
> We decided to ignore lines that were removed by another parent. Is there no risk of leak with this decision?

That's a good question. The new merge commit policy assumes parent commits of a merge have already been scanned. This means it would even be safe to ignore all removed lines, since it would be unlikely to find a valid multi-match secret from the combination of lines coming from different commits.

We keep those coming from one of the parents so that the patch we show has consistent line numbers.

fnareoh (approved):
Oh that was bad :cry:, thank you for fixing !

PR #981, opened 2024-10-14
feat: add --no-mirror option

guedou:
## Context

Cloning a repository with `git clone --mirror` may retrieve more git objects, and lead to more secrets being detected.

## What has been done

By default ggshield now used `--mirror`. A new command line option could be used to disable this feature.

## Validation

Clone `https://github.com/nightwatchcybersecurity/gb_testrepo_delete?tab=readme-ov-file`with and without the new option.


agateau-gg (changes requested):
Hi, sorry for the late review. This looks good! It's always nice to be able to spot more secrets!

I think we don't need the `--no-mirror` option. I'd rather not have it for now and add it only if we find a use-case for it: That's one less code path to test.

agateau-gg (approved):
Looks good, thanks!

PR #988, opened 2024-10-22
feat: apply gitignore for file paths when using use-gitignore option

gg-jonathangriffe:
## Context

In the GitGuardian vscode extension, we want to be able to avoid scanning files that are in the .gitignore.

The secret scan path command has an `use-gitignore` option that could do this, however it currently only applies to directories scanned recursively and not single files.

This MR proposes to also apply it to single files, which would allow the use of this argument in the vscode extension to avoid scanning gitignored files. This would also match the behavior of ignored paths in the `.gitguardian.yaml` config, for which an error is raised if an ignored directory is passed to the secret scan path command.

## Validation

- In a repo, put a file in the `.gitignore`
- run `ggshield secret scan path --use-gitignore <your_file>`
No file should be scanned.
-->

## PR check list

### DIFF
--- a/changelog.d/20241022_155052_jonathan.griffe_gitignore_paths.md
+++ b/changelog.d/20241022_155052_jonathan.griffe_gitignore_paths.md
@@ -1,0 +1,3 @@
+### Changed
+
+- The --use-gitignore option now also applies to single files passed as argument.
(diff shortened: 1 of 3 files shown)


salome-voltz (approved):
:heart_eyes:

agateau-gg (approved):
Looks good overall. I am just a bit worried about performances in the case where someone calls `secret scan path` with dozens of file names, because with these changes we run quite a few git commands for each file 😅.

But that's a corner case, just something to keep in mind.

Issue #991, opened 2024-10-25
ggshield crashes when "git merge"ing files with spaces


mherzberg:
## Environment

- ggshield version: 1.32.2
- Operating system (Linux, macOS, Windows): Windows
- Operating system version: 23H2
- Python version: Unknown (self-contained Windows release)

## Describe the bug

ggshield crashes when the user performs a git merge that involves file names with spaces. Some line splitting code will output more results than expected, leading to a crash.

Steps to reproduce:

```
mkdir mergetest; cd mergetest
git init

git checkout -b b1
echo "" > "test file.txt"
git add "test file.txt"; git commit -m "Test"

git checkout -b b2
echo "foo" > "test file.txt"
git add "test file.txt"; git commit -m "Test"

git checkout b1
echo "bar" > "test file.txt"
git add "test file.txt"; git commit -m "Test"

git merge b2
```

Actual result:

Now you can run the following command to see the verbose output of the crash:
```
ggshield secret scan pre-commit --verbose

Error: too many values to unpack (expected 4)

Traceback (most recent call last):
    File "ggshield\cmd\utils\common_decorators.py", line 20, in wrapper
    File "ggshield\cmd\secret\scan\precommit.py", line 85, in precommit_cmd
    File "ggshield\core\scan\commit.py", line 88, in from_merge
    File "ggshield\core\scan\commit_utils.py", line 372, in get_file_sha_in_ref
ValueError: too many values to unpack (expected 4)
```

Users hit the error when they have the pre-commit hook installed and continue their merge:
```
git add "test file.txt"
git merge --continue
```

The last command fails with `ValueError: too many values to unpack (expected 4)`.

Expected result:

The pre-commit hook should scan as normal.

Potential fix:

There are a few lines with the same problem, but to stay with line 372 of this particular error:

This line:
```
_, _, sha, path = line.split()
```
should be replaced with something like:
```
_, _, sha, path = line.split(maxsplit=3)
```
This will stop Python from splitting on spaces etc. in the path.

PR #992, opened 2024-10-28
fix(precommit): handle filenames with spaces in Commit.from_merge

gg-jonathangriffe:
## Context

When using `git merge` with a file which has a space in its name, the command fails with the error : 
- ValueError: too many values to unpack (expected 4)

## What has been done

Since the path is at the end of the parsed line (whose other components do not include spaces), we can simple add `maxsplit=3` as an argument to split.

## Validation

- Make a repo
- Create a file with a space in its name in a first branch
- In another branch, create the same file with a different content
- In a branch run `git merge <other_branch>`
- Resolve the conflict and run `git merge --continue`
If there is no error, it means the bug is fixed.

## PR check list

### DIFF
--- a/changelog.d/20241028_113440_jonathan.griffe_fix_precommit_git_merge_filename_with_spaces.md
+++ b/changelog.d/20241028_113440_jonathan.griffe_fix_precommit_git_merge_filename_with_spaces.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- The ggshield pre-commit hook no longer crashes when merging files with spaces in their names (#991).
--- a/ggshield/core/scan/commit_utils.py
+++ b/ggshield/core/scan/commit_utils.py
@@ -369,7 +369,7 @@ def get_file_sha_in_ref(
     """
     output = git(["ls-tree", "-z", ref] + files, cwd=cwd)
     for line in output.split("\0")[:-1]:
-        _, _, sha, path = line.split()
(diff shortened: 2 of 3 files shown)


changelog.d/20241028_113440_jonathan.griffe_fix_precommit_git_merge_filename_with_spaces.md
```diff
@@ -0,0 +1,3 @@
+### Fixed
+
+- The ggshield pre-commit hook no longer crashes when merging files with spaces in their names.
```
agateau-gg, line 3: This fixes https://github.com/GitGuardian/ggshield/issues/991, right?

If so, it's a good thing to add it here, this way a link to the issue will be created in the release notes:

```suggestion
- The ggshield pre-commit hook no longer crashes when merging files with spaces in their names (#991).
```

agateau-gg (approved):
Super minor remark. I approved nevertheless, feel free to merge when it's fixed.

changelog.d/20241028_113440_jonathan.griffe_fix_precommit_git_merge_filename_with_spaces.md
gg-jonathangriffe: It does ! Thank you for the comment I will change it

salome-voltz (approved):

PR #993, opened 2024-10-28
Fix RPM packages not working on RHEL 8.8

agateau-gg:
## Context

Our RPM packages are built using GitHub Linux runner, which uses libc 2.35. This is too recent for RHEL 8.8, which uses libc 2.28.

## What has been done

This PR reworks the CI to build the RPM and deb packages inside a Rocky Linux 8.8 container.

## Validation

Download the RPM from https://github.com/GitGuardian/ggshield/actions/runs/11552704333 (the RPM is inside the os-packages-ubuntu-22.04 artifact).

Start a rockylinux/rockylinux:8.8 container. The following command assumes $PWD contains code to scan and $GITGUARDIAN_API_KEY contains a valid API key (the `SUDO_UID` is a way to workaround the "dubious ownership" git error):

```
docker run -it --rm \
    -e GITGUARDIAN_API_KEY \
    -e SUDO_UID=$UID \
    -v$PWD:$PWD -w $PWD \
    rockylinux/rockylinux:8.8
```

Inside the container:

```
# Install git
yum install -y git

# Install ggshield
# (assumes the ggshield rpm is in $PWD)
rpm -i ggshield-1.32.2+c48d31d-1.x86_64.rpm

# Run a scan
ggshield secret scan path -ry .
```

## PR check list

### DIFF
--- a/.github/workflows/build_release_assets.yml
+++ b/.github/workflows/build_release_assets.yml
@@ -42,6 +42,7 @@ jobs:
   build_os_packages:
     name: Build packages
     runs-on: ${{ matrix.os }}
+    container: ${{ matrix.os == 'ubuntu-22.04' && 'rockylinux/rockylinux:8.8' || null }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/build_release_assets.yml
```diff
@@ -64,12 +65,44 @@ jobs:
...
+          scripts/download \
+            https://github.com/goreleaser/nfpm/releases/download/v${NFPM_VERSION}/nfpm_${NFPM_VERSION}_Linux_x86_64.tar.gz \
+            nfpm.tar.gz \
+            $NFPM_CHECKSUM
+
+          tar xf nfpm.tar.gz nfpm
+          cp nfpm /usr/local/bin
+
+          # Avoid "dubious permission" git error
+          git config --global --add safe.directory '*'
```
salome-voltz, line 104: Why do we need this? We don't use git to build the binary, or am I missing something?

salome-voltz (approved):

.github/workflows/build_release_assets.yml
agateau-gg: Good question, we probably don't need it anymore. It was required before I replaced the `--git-version` arg with `--suffix`. Going to check if we can do without it.

agateau-gg: You were right: not necessary anymore :+1:.

PR #994, opened 2024-10-28
fix(config_list): Add JSON output

salome-voltz:
## What has been done

This MR contains two commits: 
-  [feat(config_list): Add JSON output] : Adds JSON output to command `ggshield config list`
- [fix(api-status): Update JSON output] : Updates JSON output for command `ggshield api-status` to include the instance which is checked. (This info is present in the text output). 

Also, [some stages name for pre-commit have been changed](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages).
 
## PR check list

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-default_stages: [commit]
+default_stages: [pre-commit]
 repos:
   - repo: https://github.com/ambv/black
     rev: 24.3.0
@@ -68,7 +68,7 @@ repos:
         name: GitGuardian Shield
         entry: pipenv run ggshield secret scan pre-commit
         language: system
-        stages: [commit]
+        stages: [pre-commit]
 
   - repo: local
     hooks:
@@ -77,14 +77,14 @@ repos:
         entry: pipenv run ggshield secret scan pre-push
         language: system
         pass_filenames: false
-        stages: [push]
+        stages: [pre-push]
 
   - repo: https://github.com/gitguardian/ggshield
     rev: v1.32.0
     hooks:
       - id: ggshield
         language_version: python3
-        stages: [commit]
+        stages: [pre-commit]
 
   - repo: local
     hooks:
@@ -94,7 +94,7 @@ repos:
         language: system
         pass_filenames: false
         types: [python]
-        stages: [commit]
+        stages: [pre-commit]
 
       - id: import-linter
         name: Import Linter
@@ -102,4 +102,4 @@ repos:
         language: system
         pass_filenames: false
         types: [python]
-        stages: [commit]
+        stages: [pre-commit]
--- a/ggshield/cmd/config/config_list.py
+++ b/ggshield/cmd/config/config_list.py
@@ -1,8 +1,9 @@
+import json
 from typing import Any, Tuple
 
 import click
 
-from ggshield.cmd.utils.common_options import add_common_options
+from ggshield.cmd.utils.common_options import add_common_options, json_option
 from ggshield.cmd.utils.context_obj import ContextObj
 
 from .constants import DATETIME_FORMAT, FIELDS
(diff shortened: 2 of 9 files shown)


.pre-commit-config.yaml
agateau-gg, line 1: The changes in this file have nothing to do with adding JSON output 😅. They should at least be in their own commit.

ggshield/cmd/status.py
```diff
@@ -29,8 +30,12 @@ def status_cmd(ctx: click.Context, **kwargs: Any) -> int:
     if not isinstance(response, HealthCheckResponse):
         raise UnexpectedError("Unexpected health check response")
 
+    # Add the API URL to the JSON output
+    json_output = response.to_dict()
+    json_output["instance"] = client.base_uri
+
     click.echo(
-        response.to_json()
+        json.dumps(json_output, indent=2)
```
gg-jonathangriffe, line 38: It would maybe be better to change `HealthCheckResponse` and `health_check` to keep `response.to_json()` here.

tests/unit/cmd/test_status.py
```diff
@@ -36,6 +36,7 @@ def test_api_status(cli_fs_runner, api_status_json_schema):
             All(
                 {
                     "detail": "Valid API key.",
+                    "instance": Match(r"https://[^\s]+\.com"),
```
gg-jonathangriffe, line 39: I'm not sure we want to match the `.com` here as it will fail when running the tests with a non default instance.
Maybe it's fine since other tests would also fail ? WDYT ?

ggshield/cmd/config/config_list.py
```diff
@@ -1,35 +1,37 @@
...
     Print the list of configuration keys and values.
     """
-    config = ContextObj.get(ctx).config
+    ctx_obj = ContextObj.get(ctx)
+    config = ctx_obj.config
     default_token_lifetime = config.auth_config.default_token_lifetime
 
-    message_lines = []
+    # Initialize the structure for JSON output
+    config_data = {"instances": [], "global_values": {}}
```
gg-jonathangriffe, line 25: It would maybe be clearer if we use a schema here, wdyt ?

ggshield/cmd/status.py
salome-voltz: `HealthCheckResponse` is defined in pygitguardian. We can change it but it requires more work. 
I agree that it would have been more understandable, but I'm not entirely sure of where `HealthCheckResponse` is used. We might need to check the imports before modifying it :)

.pre-commit-config.yaml
salome-voltz: :innocent: fixed!

ggshield/cmd/config/config_list.py
```diff
@@ -1,35 +1,37 @@
...
 
-    def add_entries(*entries: Tuple[str, Any]):
+    def add_global_entries(*entries: Tuple[str, Any]):
         for key, value in entries:
-            message_lines.append(f"{key}: {value}")
+            config_data["global_values"][key] = value
```
agateau-gg, line 29: This function is only used once: it can be inlined

tests/unit/cmd/test_config.py
```diff
@@ -45,7 +50,7 @@
 
 
 class TestConfigList:
-    def test_valid_list(self, cli_fs_runner):
+    def test_valid_list(self, cli_fs_runner, config_list_json_schema):
```
agateau-gg, line 53: I would prefer if there was two separate tests: one for text and one for the JSON output. This way we have better diagnostics if one of them fails (and each test is shorter).

You can put the setup in a class-level fixture to avoid code duplication.

ggshield/cmd/status.py
```diff
@@ -29,8 +30,12 @@ def status_cmd(ctx: click.Context, **kwargs: Any) -> int:
     if not isinstance(response, HealthCheckResponse):
         raise UnexpectedError("Unexpected health check response")
 
+    # Add the API URL to the JSON output
+    json_output = response.to_dict()
+    json_output["instance"] = client.base_uri
```
agateau-gg, line 35: This code is only for the JSON output, I think it's better to split the `click.echo()` call instead. Something like this:

```python
if ctx_obj.use_json:
    json_output = response.to_dict()
    json_output["instance"] = client.base_uri
    click.echo(json.dumps(json_output))
else:
    # text output
    ...
```

ggshield/cmd/config/config_list.py
```diff
@@ -1,35 +1,37 @@
...
-from ggshield.cmd.utils.common_options import add_common_options
+from ggshield.cmd.utils.common_options import add_common_options, json_option
 from ggshield.cmd.utils.context_obj import ContextObj
 
 from .constants import DATETIME_FORMAT, FIELDS
 
 
 @click.command()
 @click.pass_context
+@json_option
```
agateau-gg, line 14: Can you also add `@text_json_format_option`? It adds the `--format` option which is now provided by all commands with a `--json` option.

ggshield/cmd/config/config_list.py
```diff
@@ -54,16 +56,34 @@ def add_entries(*entries: Tuple[str, Any]):
...
+            "token": token,
+            "token_name": token_name,
+            "expiry": expiry,
+        }
+        config_data["instances"].append(
+            {"instance_name": instance_name, **instance_info}
         )
+
+    if ctx_obj.use_json:
+        click.echo(json.dumps(config_data, indent=2))
```
agateau-gg, line 72: No need for `indent=2`: this is for machines, and non machines can pipe it to `jq` :)

agateau-gg (changes requested):

agateau-gg (approved):
Looks good!

PR #996, opened 2024-10-29
feat(secret_scan): add all-secrets option

gg-mmill:
Allows to display all secrets, even those excluded. Also adds a message when excluded secrets are found, to advertise the feature

## Context

Allows to display all secrets, even those excluded.

Also adds a message when excluded secrets are found, to advertise the feature

## What has been done

- add `--all-secret` option
- update output handlers
    - json output is always extended
    - text output
         -   when option is not passed, add line below file info showing number of hidden secrets
         - when option is passed, show all secrets, and add `Exclude reason` when appropriate
    - sarif output: ???

## Validation

Run ggshield scans on a file containing secrets with ignored / resolved issues, with and without `--all-secrets` option

## PR check list


ggshield/cmd/secret/scan/secret_scan_common_options.py
```diff
@@ -139,6 +139,17 @@ def _banlist_detectors_callback(
 )
 
 
+_all_secrets = click.option(
+    "--all-secrets",
+    is_flag=True,
```
agateau-gg, line 144: You need to add a `default=None` otherwise the option won't work when it appears early on the command line (`secret scan path --all-secrets .` is OK, but `secret --all-secrets scan path .` is not).

ggshield/cmd/secret/scan/secret_scan_common_options.py
```diff
@@ -139,6 +139,17 @@ def _banlist_detectors_callback(
...
+    help=(
+        "Display all secrets, including those that have been ignored for example. "
+        "Note that all returned secrets are considered as alerts."
```
agateau-gg, line 147: What is an "alert" in this context?

ggshield/core/text_utils.py
```diff
@@ -136,15 +136,28 @@ def file_info(
     filename: str,
     incident_count: int,
     incident_status: IncidentStatus = IncidentStatus.DETECTED,
+    number_of_hidden_secrets: int = 0,
```
agateau-gg, line 139: naming: I would name this `hidden_secret_count`, it's shorter and consistent with `incident_count`.

ggshield/verticals/secret/output/schemas.py
```diff
@@ -14,6 +14,8 @@ class FlattenedPolicyBreak(BaseSchema):
     incident_url = fields.String(required=True, dump_default="")
     incident_details = fields.Nested(SecretIncidentSchema)
     known_secret = fields.Bool(required=True, dump_default=False)
+    is_excluded = fields.Bool(required=False)
+    exclude_reason = fields.String(required=False)
```
agateau-gg, line 18: It's probably a bit late, but I just realized `exclude_reason` does not sound like correct English. I think it should be `exclusion_reason`.

agateau-gg (changes requested):

PR #999, opened 2024-10-30
 feat: scan .git/config @gg-gvaladon

guedou:
## Context

Currently `ggshield` excludes the `.git/config` file that may include secrets.

## What has been done

The files exclusion regexp was changed.

## Validation

The new unit test passes.


agateau-gg (approved):
Thanks for this new feature!

PR #1003, opened 2024-11-07
Fix with-incident-details option

pierrelalanne:
## The Fix
If no default value is defined for the option `with-incident-details`, the behavior of ggshield can be misleading and actually bugged.
Indeed, each command or subcommand uses the decorator "add_secret_scan_common_options" to
collect values of several command options. This lets the user place the option --with-incident-details
at several level in the command line.

If they do so, the command or subcommand that does not have the flag will set the value to False
which can prevent the option from working correctly. For instance:
`ggshield secret scan --with-incident-details path dummy.py` => The option is set to False in the end because the path subcommand does not have the option defined.
`ggshield secret scan path --with-incident-details dummy.py` => The optiton is set to True, because defined in the path subcommand.

We make the option default to None. This does not override the value defined earlier in the command line.

## Remarks
1. Let me know if the comment or the commit is too verbose.

### DIFF
--- a/changelog.d/20241107_194448_pierrelalanne_fix_with_incident_details_option.md
+++ b/changelog.d/20241107_194448_pierrelalanne_fix_with_incident_details_option.md
@@ -1,0 +1,41 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+<!--
+### Changed
+
+- A bullet item for the Changed category.
+
+-->
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -5,7 +5,9 @@ verify_ssl = true
 
 [packages]
 # Do not add dependencies here, but list them in the setup.py instead
-ggshield = { editable = true, path = "." }
+ggshield = {file = "."}
+#TODO Replace with the newest version of py-gitguardian when it's released.
+pygitguardian = {editable = true, ref = "b567ad4faa707d7b9b99de713d2d615a2baf59e5", git = "git+https://github.com/GitGuardian/py-gitguardian.git"}
```
agateau-gg, line 10: These changes should not be necessary. What happens for you without them?

agateau-gg (changes requested):

Pipfile
agateau-gg: The simplest solution for this is to do `pip install -e ../path/to/your/py-gitguardian`. Then you don't even need to commit changes.

agateau-gg (approved):

PR #1004, opened 2024-11-08
Policy breaks deprecation

elacaille18:
## Context

This PR handles the deprecation of policy breaks

## What has been done

- correct the Readme

## PR check list


actions/secret/action.yml
```diff
@@ -1,5 +1,5 @@
 name: 'ggshield scan'
-description: 'Scan commits for hardcoded secrets and security policy breaks.'
+description: 'Scan commits for hardcoded secrets.'
```
elacaille18, line 2: Shouldn't we remove completely everything about the old and deprecated `ggshield scan` commands?

agateau-gg: The "ggshield scan" name here refers to the name of the GitHub action, not the executed ggshield command (which is `ggshield secret scan`). But it should be safe to rename the action to "ggshield secret scan".

agateau-gg (approved):
Looks good, thanks!

PR #1005, opened 2024-11-12
fix(scan): instantiate client after computing options

gg-jonathangriffe:
## Context

The `--allow-self-signed` option can theoretically be placed at any level in all `ggshield secret scan` commands, but in practice does not work when passed at the last level (ex: `ggshield secret scan path --allow-self-signed`). This is due to the GGClient being instantiated before taking into account higher level options.

## What has been done

This MR makes the GGClient be instantiated in the command function to fix this issue.

## Validation

Set up a proxy, for example with mitmproxy. To verify that it works, run `ggshield secret scan path <file>`, this should return a certificate error.
Run `ggshield secret scan path <file> --allow-self-signed`. It should succeed.

## PR check list

### DIFF
--- a/changelog.d/20241112_170752_jonathan.griffe_fix_secret_scan_option.md
+++ b/changelog.d/20241112_170752_jonathan.griffe_fix_secret_scan_option.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Take into account the `--allow-self-signed` option at all levels in `ggshield secret scan` commands
--- a/tests/unit/cmd/scan/test_path.py
+++ b/tests/unit/cmd/scan/test_path.py
@@ -233,6 +233,18 @@ secret:
             _, kwargs = client_mock.call_args
             assert kwargs["base_uri"] == "https://dashboard.my-instance.com/exposed"
 
+    @pytest.mark.parametrize("position", [0, 1, 2, 3, 4])
+    def test_ssl_verify(self, cli_fs_runner, position):
+        self.create_files()
+
+        cmd = ["secret", "scan", "path", "file1"]
+        cmd.insert(position, "--allow-self-signed")
+
+        with patch("ggshield.core.client.GGClient") as client_mock:
+            cli_fs_runner.invoke(cli, cmd)
+            _, kwargs = client_mock.call_args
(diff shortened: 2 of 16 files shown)


changelog.d/20241112_170752_jonathan.griffe_fix_secret_scan_option.md
```diff
@@ -0,0 +1,3 @@
+### Fixed
+
+- Take into account --allow-self-signed options at all levels in `ggshield secret scan` commands
```
agateau-gg, line 3:
```suggestion
- Take into account the `--allow-self-signed` option at all levels in `ggshield secret scan` commands
```

tests/unit/cmd/scan/test_path.py
```diff
@@ -216,6 +216,19 @@ def test_scan_ignored_file(self, scan_mock, cli_fs_runner):
         assert "An ignored file or directory cannot be scanned." in result.stdout
         scan_mock.assert_not_called()
 
+    @pytest.mark.parametrize("verify", [True, False])
```
agateau-gg, line 219: Do we care about the case where verify is False? I think it could be more useful to test all possible positions of `--allow-self-signed`.

agateau-gg (changes requested):
Must have been a bit tedious to do 😅.

Looks good overall, just some minor remarks.

tests/unit/cmd/scan/test_path.py
```diff
@@ -216,6 +216,19 @@ def test_scan_ignored_file(self, scan_mock, cli_fs_runner):
         assert "An ignored file or directory cannot be scanned." in result.stdout
         scan_mock.assert_not_called()
 
+    @pytest.mark.parametrize("position", [0, 1, 2, 3, 4])
+    def test_ssl_verify(self, cli_fs_runner, position):
+        self.create_files()
+
+        cmd = ["secret", "scan", "path", "file1"]
+        cmd.insert(position, "--allow-self-signed")
+        print(cmd)
```
salome-voltz, line 225: :bomb:

salome-voltz (approved):

agateau-gg (approved):

PR #1006, opened 2024-11-12
feat(cmd): add the instance option to more commands

gg-jonathangriffe:
## Context

Currently, we can login to a specific instance by using the --instance option.
We want to also be able to use a specific instance for other commands

## What has been done

The `--instance` option has been added to all `secret scan` command, as well as `api-status` and `quota` commands

## Validation

Use commands specified above using the `--instance` with a non default instance and an API KEY for the instance in the `.env`. Verify that the command succeeds.

## PR check list

### DIFF
--- a/changelog.d/20241112_172503_jonathan.griffe_expand_instance_option.md
+++ b/changelog.d/20241112_172503_jonathan.griffe_expand_instance_option.md
@@ -1,0 +1,3 @@
+### Added
+
+- Added the `--instance` option to allow using a different instance to all `secret scan` commands as well as the `api-status` and `quota` commands.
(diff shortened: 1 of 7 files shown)


pierrelalanne:
Hey @gg-jonathangriffe,
I saw you opened this PR and it was somehow related to options and where they are propagated in the command. Just wanted to share with you the [bug I encountered and fixed last week](https://github.com/GitGuardian/ggshield/pull/1003). It may also bite you in this case (I did not test though). Cheers

gg-jonathangriffe:
Thank you @pierrelalanne  for your comment, I didn't know that !
It seems not to have an impact here, I believe that is because my option is not a flag, therefore it defaults to None and not False.

agateau-gg (approved):
Looks good :+1:

PR #1007, opened 2024-11-13
feat(api-status): add api key and instance sources to the api-status command output

gg-jonathangriffe:
## Context

There are a lot of ways for ggshield users to set authentication credentials. When these credentials are invalid, it can be difficult to search where they were set. When multiple sources are set, it can also be complicated to find the source that was used.

## What has been done

This MR makes the api-status command return the source of the instance and API key used.

## Validation

Run `ggshield api-status` and verify that the sources are the correct ones.
Repeat for all possible api key and instance locations.

## PR check list

### DIFF
--- a/doc/schemas/api-status.json
+++ b/doc/schemas/api-status.json
@@ -13,6 +13,14 @@
       "format": "uri",
       "description": "URL of the GitGuardian instance"
     },
+    "instance_source": {
+      "enum": ["CMD_OPTION", "DOTENV", "ENV_VAR", "USER_CONFIG", "DEFAULT"],
+      "description": "Source the instance was read from"
+    },
+    "api_key_source": {
+      "enum": ["DOTENV", "ENV_VAR", "USER_CONFIG"],
+      "description": "Source the API key was read from"
+    },
     "detail": {
       "type": "string",
       "description": "Human-readable version of the status"
--- a/ggshield/core/config/config.py
+++ b/ggshield/core/config/config.py
@@ -1,7 +1,8 @@
 import logging
 import os
+from enum import Enum
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Optional, Set, Tuple
 
 import click
 
@@ -17,6 +18,19 @@ from ggshield.core.url_utils import (
 )
 
 
+class ConfigSource(Enum):
+    """
+    Enum of the different sources of configuration
+    where an API key or instance URL can come from
+    """
+
+    CMD_OPTION = "command line option"
+    DOTENV = ".env file"
+    ENV_VAR = "environment variable"
+    USER_CONFIG = "user config"
+    DEFAULT = "default"
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -26,7 +40,13 @@ class Config:
     AuthConfig.
     """
 
-    __slots__ = ["user_config", "auth_config", "_cmdline_instance_name", "_config_path"]
+    __slots__ = [
+        "user_config",
+        "auth_config",
+        "_cmdline_instance_name",
+        "_config_path",
+        "_dotenv_vars",
+    ]
 
     user_config: UserConfig
     auth_config: AuthConfig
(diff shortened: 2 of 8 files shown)


ggshield/core/config/config.py
```diff
@@ -40,6 +55,7 @@ def __init__(self, config_path: Optional[Path] = None):
         self.user_config, self._config_path = UserConfig.load(config_path=config_path)
         self.auth_config = AuthConfig.load()
         self._cmdline_instance_name = None
+        self.dotenv_vars: Set[str] = set()
```
agateau-gg, line 58: Can you rename `dotenv_vars` to `_dotenv_vars` and declare its type with its `_cmdline_instance_name` and `_config_path` friends at the top of the class?

Also adding a comment explaining what this variable is for would be useful.

ggshield/core/config/config.py
```diff
@@ -51,17 +67,27 @@ def config_path(self) -> Path:
 
     @property
     def instance_name(self) -> str:
+        return self.get_instance()[0]
+
+    @property
+    def instance_source(self) -> str:
+        return self.get_instance()[1].value
+
+    def get_instance(self) -> Tuple[str, ConfigSource]:
```
agateau-gg, line 76: I think we don't need the `instance_source` property and I would rename `get_instance()` to `get_instance_name_and_source()`.

doc/schemas/api-status.json
```diff
@@ -13,6 +13,14 @@
...
       "description": "URL of the GitGuardian instance"
     },
+    "instance_source": {
+      "type": "string",
+      "description": "Source the instance was read from"
+    },
+    "api_key_source": {
+      "type": "string",
+      "description": "Source the API key was read from"
+    },
```
agateau-gg, line 23: The `type` for these fields should be `enum` and the values should be the internal enum values, not their user-friendly texts (so `CMD_OPTION`, `ENV_VAR`...). This way we can reword the user-friendly texts without breaking anything.

ggshield/core/config/config.py
```diff
@@ -17,6 +18,14 @@
 )
 
 
+class ConfigSource(Enum):
```
agateau-gg, line 21: Can you add a short docstring?

agateau-gg (changes requested):

ggshield/core/config/config.py
gg-jonathangriffe: I think it is clearer to have the `instance_source` property, so that the syntax is similar to that of the api key, wdyt ?

doc/schemas/api-status.json
agateau-gg: Yes, the test would need to check that we get human-friendly names in text mode, but machine-friendly enum values in JSON mode.

ggshield/core/config/config.py
agateau-gg: I am not fond of the existing `Config.instance` property: I think it's wrong to have properties which do complicated work and can raise exceptions. It's already there so I haven't changed it, but I'd rather avoid spreading the problem. Especially for a property which is only going to be called once.

agateau-gg: (should have explained my reasons for not liking properties)

A property is meant to represent a computed value that can be used the same way one uses a class member variable. The assumption is that it's cheap and reliable to use it, just like it's cheap and reliable to read a member variable.

Having properties do complex work breaks that assumption.

gg-jonathangriffe: Thanks for the clarification :)

salome-voltz (approved):

ggshield/core/config/config.py
```diff
@@ -51,17 +77,23 @@ def config_path(self) -> Path:
...
+    def get_instance_name_and_source(self) -> Tuple[str, ConfigSource]:
         """
+        Return the instance name and source of the selected instance.
+
         The instance name (defaulting to URL) of the selected instance
         priority order is:
         - set from the command line (by setting cmdline_instance_name)
-        - env var (in auth_config.current_instance)
+        - GITGUARDIAN_INSTANCE env var
+        - GITGUARDIAN_API_URL env var
```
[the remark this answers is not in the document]
gg-jonathangriffe: We do not differentiate here between system env var or `.env` vars, since at this point, `load_dot_env` has been called so `.env` vars have overriden env vars.

ggshield/core/config/config.py
```diff
@@ -123,7 +165,16 @@ def dashboard_url(self) -> str:
 
     @property
     def api_key(self) -> str:
+        return self.get_api_key()[0]
+
+    @property
+    def api_key_source(self) -> ConfigSource:
+        return self.get_api_key()[1]
+
+    def get_api_key(self) -> Tuple[str, ConfigSource]:
```
agateau-gg, line 174: I think you should do the same here as for the instance name: remove the property and rename `get_api_key()` to `get_api_key_and_source()`.

agateau-gg (changes requested):
Minor remark. Nice work on restricting the valid values for `api_key_source` in the JSON schema :+1:.

agateau-gg (approved):
Nice!

PR #1010, opened 2024-11-14
feat(output): Only fail secret scans when the secret is introduced

### LINKED ISSUE #1001: [FEATURE] Fail only upon adding a secret (not when removing or when a secret is on the line above or below).
## Is your feature request related to a problem? Please describe.

I am using ggshield in a few projects pipeline, and first of all *thanks for this great tool* !

I am annoyed with a behaviour of ggshield that AFAIK, cannot be customized. Since I've started a work of taking secrets out of my project repo, I am removing secrets one by one.

Each time I am **removing a secret**, ggshield job is failing and need to use `SKIP=ggshield` as a workaround.

Example 1 (original names and value were changed) :
```
 10 | +define('ENV', $_ENV['ENV']);
...
25    | -$_ENV['USER'] = 'token';
26    | -$_ENV['PASSWORD'] = 'g42wyu2zA*********************************xyzxyzk7g';
27    | -$_ENV['X-OVH-TOKEN'] = '8au611**-****-****-****-******21xyq6';
                                 |______________apikey______________|
28 11 |  

>> Secret detected: Generic Password
```

Similar issue occurs when **modifying a line below or above a secret**. GGshield job fails even though the line you modified has nothing to do with a secret.

Example 2 (original names and value were changed) : 
```
27    | -$_ENV['ELASTIC_ALIAS'] = 'logs-xt-04221-a-dev-logs';
...
   27 | +$_ENV['ELASTIC_ALIAS'] = 'logs-xt-04221-a-tests-logs';
28 28 |  $_ENV['AN_API_TOKEN'] = '6yz828**-****-****-****-******2ufg2';
                                 |______________apikey_______________|
29 29 |  
30 30 |  /**

>> Secret detected: Generic Password
```
Here I modified **line 27** and got the job failed for an existing secret on **line 28**.

## Describe the solution you'd like

To ensure **tests hermeticity** ggshield should be able to fail **only when a secret has been added** on a specific commit. 
* Not when removed.
* Not when modified.
* Not when the secret is not part of the commited changes.

It is even more relevant when several developers are working on the same repo, a developer should not be blocked for a vulnerability he did not introduced.

## Describe alternatives you've considered

In case the behaviour of ggshield could not be changed (for backward compatibility), implement an option to behave the way described above.

Walz:
## Context

Related to SPI-526 and will close  #1001

We want to only fail secret scans when the secret is introduced.  
New fields have been added to the API to automatically detect if the content is a diff and if the secret has been added, deleted or in the context. (https://github.com/GitGuardian/py-gitguardian/pull/122)

- Add `not_introduced` as an ignore reason
- Make use of `diff_kind` to ignore policy breaks

`diff_kind` can be `null` (outside of commit) or `ADDITION`/`DELETION`/`CONTEXT`. We want to keep only the policy breaks with `diff_kind`:
- `null` => outside of commits, when scanning files or docker layers
- `ADDITION` => introduced by a commit

In #1024, a `--all-secrets` option will allow user to display ignored secrets.

## Validation

- Create a repository
- Add a secret in commit A
- Edit the line below the secret in commit B
- Remove the secret in commit C

When scanning the repository with `main`: 3 policy breaks are found in commit A, B and C
When scanning with this PR: 1 policy break is found in commit A

## PR check list

### DIFF
--- a/ggshield/verticals/secret/secret_scan_collection.py
+++ b/ggshield/verticals/secret/secret_scan_collection.py
@@ -28,6 +28,7 @@ class IgnoreReason(Enum):
     IGNORED_MATCH = "ignored_match"
     IGNORED_DETECTOR = "ignored_detector"
     KNOWN_SECRET = "known_secret"
+    NOT_INTRODUCED = "not_introduced"
 
 
 class Result:
--- a/changelog.d/20241129_155452_samuel.guillaume_spi_526_implement_deletion_commit_identification_in_ggshield.md
+++ b/changelog.d/20241129_155452_samuel.guillaume_spi_526_implement_deletion_commit_identification_in_ggshield.md
@@ -1,0 +1,41 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+-->
+
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+
+### Changed
+
+- When scanning commits, ggshield will ignore by default secrets which are removed or contextual to the patch.
+
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category.
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category.
+
+-->
--- a/ggshield/verticals/secret/secret_scanner.py
+++ b/ggshield/verticals/secret/secret_scanner.py
@@ -7,7 +7,13 @@ from concurrent.futures import Future
 from typing import Dict, Iterable, List, Optional, Union
 
 from pygitguardian import GGClient
-from pygitguardian.models import APITokensResponse, Detail, MultiScanResult, TokenScope
+from pygitguardian.models import (
+    APITokensResponse,
+    Detail,
+    DiffKind,
+    MultiScanResult,
+    TokenScope,
+)
 
 from ggshield.core import ui
 from ggshield.core.cache import Cache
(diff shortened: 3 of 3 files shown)


ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -70,6 +77,18 @@ def censor(self) -> None:
...
+        self, *, diff_kinds: Optional[List[DiffKind]] = None
+    ) -> List[PolicyBreak]:
+        policy_breaks = self.scan.policy_breaks
+        if diff_kinds is not None and self.scan.is_diff:
+            policy_breaks = [
+                policy_break
+                for policy_break in self.scan.policy_breaks
+                if policy_break.diff_kind in diff_kinds
+            ]
+        return policy_breaks
```
agateau-gg, line 90: nit: can be slightly simplified to:

```suggestion
        if diff_kinds is not None and self.scan.is_diff:
            return [
                policy_break
                for policy_break in self.scan.policy_breaks
                if policy_break.diff_kind in diff_kinds
            ]
        return self.scan.policy_breaks
```

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -70,6 +77,18 @@ def censor(self) -> None:
     def has_policy_breaks(self) -> bool:
         return self.scan.has_policy_breaks
 
+    def policy_breaks(
+        self, *, diff_kinds: Optional[List[DiffKind]] = None
```
agateau-gg, line 81: Would be nicer if diff_kinds was a set instead of a list, since we don't want to have duplicate kinds.

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -70,6 +77,18 @@ def censor(self) -> None:
     def has_policy_breaks(self) -> bool:
         return self.scan.has_policy_breaks
 
+    def policy_breaks(
```
agateau-gg, line 80: naming: this should be `get_policy_breaks()`.

agateau-gg (changes requested):

Walz:
Thanks to https://github.com/GitGuardian/ggshield/issues/1001, this PR is a **lot** simpler now.

agateau-gg (approved):
Nice! And thanks for the great PR description ✨.

PR #1013, opened 2024-11-19
fix(error): Catch error when merging two configs

salome-voltz:
## Context

GGshield was crashing when merging two badly formed user config. This was returning exit code 1 instead of 128. 

## PR check list

### DIFF
--- a/tests/unit/core/config/test_user_config.py
+++ b/tests/unit/core/config/test_user_config.py
@@ -613,3 +613,16 @@ class TestUserConfig:
         )
         config, _ = UserConfig.load(local_config_path)
         assert config.secret.ignore_known_secrets
+
+    def test_bad_local_config(self, local_config_path, global_config_path):
+        """
+        GIVEN a malformed .gitguardian.yaml, with a list of instance
+        WHEN loading the user condiguration
+        THEN an error is returned
+        """
+        write_yaml(global_config_path, {"instance": "https://test.gitguardian.com/"})
+        write_yaml(
+            local_config_path, {"instance": ["https://dashboard.gitguardian.com/"]}
+        )
+        with pytest.raises(UnexpectedError):
+            UserConfig.load()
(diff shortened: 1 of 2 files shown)


ggshield/core/config/user_config.py
```diff
@@ -320,30 +320,33 @@ def load(cls, config_path: Optional[Path] = None) -> Tuple["UserConfig", Path]:
...
+            local_config_path = find_local_config_path()
+            if local_config_path:
+                dct = _load_config_dict(local_config_path, deprecation_messages)
+                update_dict_from_other(user_config_dict, dct)
+                config_path = local_config_path
+                logger.debug("Loaded local config from %s", local_config_path)
+            else:
+                logger.debug("No local config")
+        except AttributeError:
+            raise UnexpectedError("Failed to load user configuration.")
```
agateau-gg, line 349: If it's possible, I would prefer if this was handled in `update_dict_from_other()`:
- we can include in the error message the name of the failing key
- it avoids having a long block of code inside the try...except block

tests/unit/core/config/test_user_config.py
```diff
@@ -613,3 +613,16 @@ def test_can_load_ignored_known_secrets_from_secret(self, local_config_path):
...
+        GIVEN a malformed .gitguardian.yaml, with a list of instance
+        WHEN loading the user condiguration
+        THEN an error is returned
+        """
+        write_yaml(global_config_path, {"instance": "https://test.gitguardian.com/"})
+        write_yaml(
+            local_config_path, {"instance": ["https://dashboard.gitguardian.com/"]}
+        )
+        with pytest.raises(UnexpectedError):
+            _, _ = UserConfig.load()
```
agateau-gg, line 628: I think you can just do `UserConfig.load()`, no need for the `_, _ = `.

agateau-gg (changes requested):

ggshield/core/config/user_config.py
salome-voltz: I understand what you mean, let me know if this new version works for you. 

I initally made this choice because `update_dict_from_other()` is a utils function, so I wanted to avoid raising specific errors for a function that could be reused.

agateau-gg: That was good reasoning: I was worried about `update_dict_from_other()` being an utils function as well, but then I saw it was only used for config, and that it was in core.config.utils, so we can accept this specialization.

PR #1015, opened 2024-11-20
fix(with-incident-details): Ensure the token has the correct scope

salome-voltz:
## Context

ggshield 1.32.0 can now return the details of an incident when a scan is run with the --with-incident-details option. For this feature to work, the token used to authenticate with the API must have the 'incidents:read' scope. This is currently not documented and if the token does not have the required scope, the flag will be silently ignored.

## What has been done

Ensure the token has the required scope before running a command with `--with-incident-details` flag. This MR goes with https://github.com/GitGuardian/py-gitguardian/pull/126

## PR check list

- [X] As much as possible, the changes include tests (unit and/or functional)
- [X] If the changes affect the end user (new feature, behavior change, bug fix) then the PR has a changelog entry (see doc/dev/getting-started.md). If the changes do not affect the end user, then the `skip-changelog` label has been added to the PR.

### DIFF
--- a/ggshield/cmd/secret/scan/secret_scan_common_options.py
+++ b/ggshield/cmd/secret/scan/secret_scan_common_options.py
@@ -126,6 +126,7 @@ _banlist_detectors_option = click.option(
     metavar="DETECTOR",
 )
 
+
 _with_incident_details_option = click.option(
     "--with-incident-details",
     is_flag=True,
@@ -134,7 +135,7 @@ _with_incident_details_option = click.option(
     # If the option is placed early in the command line, the value may be overridden
     # later on with False if no default is defined.
     default=None,
-    help="Display full details about the dashboard incident if one is found (JSON and SARIF formats only).",
+    help="""Display full details about the dashboard incident if one is found (JSON and SARIF formats only). Requires the 'incidents:read' scope.""",  # noqa
     callback=create_config_callback("secret", "with_incident_details"),
 )
 
--- a/ggshield/core/errors.py
+++ b/ggshield/core/errors.py
@@ -7,12 +7,13 @@ import logging
 import platform
 import traceback
 from enum import IntEnum
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 import click
 from marshmallow import ValidationError
-from pygitguardian.models import Detail
+from pygitguardian.models import Detail, TokenScope
 
+from ggshield.core.text_utils import pluralize
 from ggshield.utils.git_shell import GitError, InvalidGitRefError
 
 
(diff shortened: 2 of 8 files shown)


ggshield/cmd/secret/scan/secret_scan_common_options.py
```diff
@@ -126,6 +129,31 @@ def _banlist_detectors_callback(
...
+        elif isinstance(response, Detail):
+            raise UnexpectedError(response.detail)
+
+        if TokenScopes.INCIDENTS_READ not in response.scopes:
+            raise MissingScopeError(
+                "Token is missing the scope: 'incidents:read' to retrieve incident details."
+            )
+
+        ctx_obj.config.user_config.secret.with_incident_details = value
+    return value
```
agateau-gg, line 154: It's not a good idea to create a client and send HTTP requests while parsing the command-line because some not-yet-parsed CLI flags might affect it, for example  `--allow-self-signed`.

In general we should try to run as little code as possible while parsing the command-line.

You should be able to move the logic of your check to `SecretScanner.__init__()`.

ggshield/core/errors.py
```diff
@@ -66,6 +66,15 @@ def __init__(self, message: str):
         super().__init__(ExitCode.UNEXPECTED_ERROR, message)
 
 
+class MissingScopeError(_ExitError):
+    """
+    Token does not have the required scope
+    """
+
+    def __init__(self, message: str):
+        super().__init__(ExitCode.UNEXPECTED_ERROR, message)
```
agateau-gg, line 75: Maybe you can make this exception take the scope as argument and do the formatting itself? This would make it more reusable in other verticals like honeytoken (currently honeytoken commands do not check if the token has the right scope, they just fail and say "check your personal access token has the `honeytokens:write` scope.")

The wording would have to be a bit more generic though, something like "Your token needs the `{scope}` scope to perform this operation.".

agateau-gg (changes requested):

ggshield/core/errors.py
```diff
@@ -66,6 +67,19 @@ def __init__(self, message: str):
         super().__init__(ExitCode.UNEXPECTED_ERROR, message)
 
 
+class MissingScopeError(_ExitError):
```
agateau-gg, line 70: naming: should be `MissingscopesError` since it supports multiple scopes.

agateau-gg (changes requested):
One minor remark, but reviewing your work prompted me to create https://github.com/GitGuardian/py-gitguardian/pull/128, which is going to create more work for you :/.

agateau-gg (approved):

PR #1016, opened 2024-11-20
chore: move secret ignoring logic inside the scanner

gg-mmill:
## Context

Currently, there are logic related to ignoring secrets in two places:
- in the secret scanner
- in the output handlers

## What has been done

Move all the logic to the scanner:
- "ignore known issues" now works in the same way than other ignore featues
- Result objects now contain the list of PolicyBreaks (instead of ScanResult), as well as a dictionary containing the counts of ignored PolicyBreak by reason
- As a result, there is no more result filtering in output handlers - they can still access the options and count of ignored issues if needed

## Validation

~~Need to add tests for the scanner~~ Done
Will also perform some manual validation

## PR check list


ggshield/verticals/secret/secret_scanner.py
```diff
@@ -40,6 +40,16 @@
     ScanFuture = Future
 
 
+def remove_known_secrets_from_result(scan_result: ScanResult):
+    scan_result.policy_breaks = [
+        policy_break
+        for policy_break in scan_result.policy_breaks
+        if policy_break.known_secret
```
agateau-gg, line 47: Should be `if not`, no?

agateau-gg (changes requested):
Minor remarks, but this looks much cleaner! ✨

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -129,18 +160,7 @@ def __init__(
...
-        ) = self._get_known_new_secrets_count()
-
-    @property
-    def has_new_secrets(self) -> bool:
-        return self.new_secrets_count > 0
-
-    @property
-    def has_secrets(self) -> bool:
-        return (self.new_secrets_count + self.known_secrets_count) > 0
+        self.all_results = list(self.get_all_results())
```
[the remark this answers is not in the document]
gg-mmill: We can do without - I feel like the interface of SecretScanCollection is kinda flawed no matter what we do though, because of how it's mixing up results, nested scan results, etc.

Changed to `total_policy_breaks_count` -> more correct, as Results can now be empty of policy breaks

ggshield/verticals/secret/output/secret_json_output_handler.py
```diff
@@ -27,7 +27,7 @@ def create_scan_dict(
         if scan.extra_info:
             scan_dict["extra_info"] = scan.extra_info
 
-        if top and scan.has_results:
+        if top:
```
agateau-gg, line 30: I am not sure why the original code behaved that way, but is there any reason for this change?

agateau-gg (changes requested):
Minor remarks.

ggshield/verticals/secret/output/secret_json_output_handler.py
gg-mmill: Yes, I removed the `has_result` attribute of `ScanCollection`, which wasn't used anywhere else (and was a bit bugged, since it did not consider sub-scans)

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -62,13 +81,28 @@ def enrich_matches(self, lines: List[Line]) -> None:
...
     @property
     def has_policy_breaks(self) -> bool:
-        return self.scan.has_policy_breaks
+        return len(self.policy_breaks) > 0
+
+    def apply_ignore_function(
+        self, reason: IgnoreReason, ignore_function: Callable[[PolicyBreak], bool]
+    ):
+        if reason in self.ignored_policy_breaks_count_by_reason:
+            raise Exception(f"Ignore was already computed for {IgnoreReason}")
```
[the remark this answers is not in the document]
gg-mmill: I had to change this part again when implementing the `--all-secrets`, so this will probably be removed anyway

gg-mmill (commented):
Thanks for the review !
About the ignoring, I had to change what I did here when implementing the `all-secrets` option (future MR, not ready yet): separate the computation of the ignore reason, and the actual ignoring, see https://github.com/GitGuardian/ggshield/pull/1024/files#diff-37638c3845d9f616b7031dec851a57324bec777285097c841b887d2a1f9dd34eR236

agateau-gg (approved):
Looks better now, thanks for this work!

PR #1020, opened 2024-11-25
Update py-gitguardian

agateau-gg:
Update to the latest release of py-gitguardian.


salome-voltz (approved):

sevbch (approved):
Why wasn't it targeting 1.17.0?

agateau-gg:
> Why wasn't it targeting 1.17.0?

Because we were making use of features we added in 1.18.0. When a ggshield change requires a  change in py-gitguardian, we make the change in py-gitguardian and make ggshield's pyproject.toml point to the py-gitguardian commit containing the change.

We never release a version of ggshield that depend on a py-gitguardian commit: when it's time to make a ggshield release, if it uses a py-gitguardian commit, we make a py-gitguardian release and update ggshield's pyproject.toml. This last step is what this PR does.

PR #1022, opened 2024-11-26
Fix patch parser crash

agateau-gg:
## Context

Sometimes a patch contains content on the line of the hunk header, like this:

```
@@ -3,4 +3,4 @@ I'm on the hunk header
 a line
 some other line

-old line
+new line
```

This caused our patch parser to fail because the line on the hunk header was not attributed pre or post line numbers.

## What has been done

Take this case into account, add a test for it.

## Validation

Create a commit with this kind of patch, try to scan it.

## PR check list


sevbch (approved):

ggshield/core/lines.py
```diff
@@ -7,8 +7,8 @@
 from ggshield.utils.git_shell import Filemode
 
 
-REGEX_PATCH_HEADER = re.compile(
-    r"^(?P<line_content>@@ -(?P<pre_index>\d+),?\d* \+(?P<post_index>\d+),?\d* @@(?: .+)?)"  # noqa
+REGEX_HUNK_HEADER = re.compile(
+    r"^@@ -(?P<pre_index>\d+),?\d* \+(?P<post_index>\d+),?\d* @@(?P<trailing_content> .+)?"  # noqa
```
salome-voltz, line 11:
```suggestion
    r"^@@ -(?P<pre_index>\d+),?\d* \+(?P<post_index>\d+),?\d* @@(?P<trailing_content>.+)?"  # noqa
```
Is there constantly a space here? Otherwise, this could also create an error I guess

salome-voltz (approved):

ggshield/core/lines.py
agateau-gg: Yes, the space is always there.

PR #1026, opened 2024-11-28
Fix functional tests

agateau-gg:
## Context

I merged a PR which came from a fork. In this conditions functional tests are skipped (they can't be run because they don't have access to secrets). This broke `main` because there was a small bug in a test added by the PR.

## What has been done

Fix the test, and actually merge it with an existing one.

## Validation

Tests pass!

## PR check list


gg-mmill (approved):
Looks good !

sevbch (approved):

PR #1031, opened 2024-12-10
chore: make ignore reason human readable in text output

gg-mmill:
## Context

Following https://github.com/GitGuardian/ggshield/pull/1024 ,

-  ~~small~~ fix to make ignore reasons human readable.
- add ignore reason to json output schema

## Validation

Perform scans with ignored secrets (either from config or  backend), with the `--all-secrets` options.

Ensure the output contains ignore info (both for text and json output)

## PR check list

Note: skipping changelog as it was included in previous MR.

### DIFF
--- a/ggshield/verticals/secret/output/schemas.py
+++ b/ggshield/verticals/secret/output/schemas.py
@@ -12,7 +12,7 @@ class IgnoreReasonSchema(BaseSchema):
 class FlattenedPolicyBreak(BaseSchema):
     policy = fields.String(required=True)
     occurrences = fields.List(fields.Nested(ExtendedMatchSchema), required=True)
-    break_type = fields.String(data_key="type", required=True)
+    detector = fields.String(data_key="type", required=True)
     validity = fields.String(required=False, allow_none=True)
     ignore_sha = fields.String(required=True)
     total_occurrences = fields.Integer(required=True)
--- a/ggshield/verticals/secret/output/schemas.py
+++ b/ggshield/verticals/secret/output/schemas.py
@@ -4,6 +4,11 @@ from pygitguardian.models import BaseSchema, SecretIncidentSchema
 from ggshield.verticals.secret.extended_match import ExtendedMatchSchema
 
 
+class IgnoreReasonSchema(BaseSchema):
+    kind = fields.String(required=True)
+    detail = fields.String()
+
+
 class FlattenedPolicyBreak(BaseSchema):
     policy = fields.String(required=True)
     occurrences = fields.List(fields.Nested(ExtendedMatchSchema), required=True)
@@ -14,6 +19,7 @@ class FlattenedPolicyBreak(BaseSchema):
     incident_url = fields.String(required=True, dump_default="")
     incident_details = fields.Nested(SecretIncidentSchema)
     known_secret = fields.Bool(required=True, dump_default=False)
+    ignore_reason = fields.Nested(IgnoreReasonSchema, dump_default=None)
 
 
 class JSONResultSchema(BaseSchema):
--- a/ggshield/verticals/secret/secret_scan_collection.py
+++ b/ggshield/verticals/secret/secret_scan_collection.py
@@ -1,3 +1,5 @@
+import hashlib
+import operator
 from dataclasses import dataclass, field
 from enum import Enum
 from pathlib import Path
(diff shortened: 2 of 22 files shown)


agateau-gg (changes requested):
This change makes it hard to get a machine-friendly constant, which would be handy for machine-friendly outputs like JSON.

At the minimum I would remove the `__str__()` override: code that wants the human-readable version can call `.value` to get it.

But even this way using human-friendly values makes recreating the enum from a string more difficult:

```python
class MyEnum(str, Enum):
    FOO = "This is foo"
    BAR = "This is bar"

# Fails
MyEnum("FOO")

# Fails
MyEnum("MyEnum.FOO")

# Works
MyEnum("This is foo")
```

That's why I would advocate for something like this instead:

```python
class MyEnum(str, Enum):
    FOO = "FOO"
    BAR = "BAR"

    @property
    def display_value(self) -> str:
        # Or use a dict
        if self == self.FOO:
            return "This is foo"
        elif self == self.BAR:
            return "This is bar"
        raise ValueError(self)
```

With this we can do:

```python
>>> MyEnum("FOO")
<MyEnum.FOO: 'FOO'>
>>> MyEnum.FOO.value
'FOO'
>>> MyEnum.FOO.display_value
'This is foo'
```

tests/unit/cmd/test_ignore.py
```diff
@@ -212,26 +213,14 @@ def test_do_not_duplicate_last_found_secrets(client, isolated_fs):
     )
     cache = Cache()
 
-    cache.add_found_policy_break(policy_break, "a")
-    cache.add_found_policy_break(policy_break, "b")
-
-    assert len(cache.last_found_secrets) == 1
-
-
-def test_do_not_add_policy_breaks_to_last_found(client, isolated_fs):
```
gg-mmill, line 221: Not clear in the diff: I removed `test_do_not_add_policy_breaks_to_last_found`, which was not useful anymore.

What was tested was that non-secret policy breaks were not added to the cache (GIVEN/WHEN/THEN was not accurate).

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -17,48 +19,111 @@
...
     return ignore_reason
 
 
+@dataclass
+class Secret:
+    """GGShield specific model to handle policy-breaks.
+    Named Secret since we are dropping other kind of policy breaks.
+    """
+
+    break_type: str
```
agateau-gg, line 89: At some point we should rename this "detector" (maybe not today, though 😉).

ggshield/verticals/secret/secret_scan_collection.py
```diff
@@ -70,25 +135,13 @@ class Result:
     filemode: Filemode
     path: Path
     url: str
-    policy_breaks: List[PolicyBreak]
-    ignored_policy_breaks_count_by_reason: Counter[str]
+    policy_breaks: List[Secret]
+    ignored_policy_breaks_count_by_kind: Counter[IgnoreKind]
```
agateau-gg, line 145: Since we are already renaming this field, I think it would make sense to name it `ignored_secrets_count_by_kind`, but only if we rename `policy_breaks` to `secrets`. That would actually not be that invasive: it's only used in 11 places in the code (and 31 in tests).

ggshield/verticals/secret/output/schemas.py
```diff
@@ -14,6 +14,8 @@ class FlattenedPolicyBreak(BaseSchema):
     incident_url = fields.String(required=True, dump_default="")
     incident_details = fields.Nested(SecretIncidentSchema)
     known_secret = fields.Bool(required=True, dump_default=False)
+    is_ignored = fields.Bool(required=False)
+    ignore_reason = fields.String(required=False)
```
agateau-gg, line 18: I would prefer if we used a separate schema for IgnoreReason. If we do so we don't need the `is_ignored` field: either the `ignore_reason` field is there and the secret has been ignored, or it's not and the secret has not been ignored.

agateau-gg (changes requested):

ggshield/verticals/secret/secret_scan_collection.py
gg-mmill: Yes I've also thought of it, even started doing so, but it was reverted at some point (probably after I got tangled up in cyclic import issues)

ggshield/verticals/secret/output/schemas.py
gg-mmill: Made it a separate schema

ggshield/verticals/secret/secret_scan_collection.py
gg-mmill: Added rename commit.

ggshield/verticals/secret/secret_scan_collection.py
gg-mmill: Done as well :)

agateau-gg (approved):
Looks good! Love the stroke "small" word in the PR description :sweat_smile:.

Thanks for this work. I think the code-base is less confusing thanks to these renamings.

PR #1034, opened 2024-12-13
add IPython in dev dependencies

sevbch:
IPython is needed when running the debugger

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,18 +63,6 @@ requires = ["pdm-backend"]
 build-backend = "pdm.backend"
 
 [tool.pdm.dev-dependencies]
-dev = [
-    "scriv[toml]",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -63,18 +63,6 @@ requires = ["pdm-backend"]
 build-backend = "pdm.backend"
 
 [tool.pdm.dev-dependencies]
-dev = [
```
sevbch, line 66: `dev` group was present 2 times. It was removed automatically when running `pdm add --dev ipython`

agateau-gg: Must have been messed up when pdm moved it to use the new "dependency-groups" section. Looks like the `standalone` group has been duplicated too.

agateau-gg (approved):

PR #1041, opened 2025-01-06
chore(cmd): improve the exit-zero option help message

gg-jonathangriffe:
## Context

The help message for the `exit-zero` option can be slightly misleading, as one could understand that it will make the command exit with the 0 exit code, regardless of the error raised.

## What has been done

The `exit-zero` option help message was modified to make it clear it only applies when a secret is detected.


sevbch (approved):

mathieubellon (commented):
LGTM @sevbch

PR #1045, opened 2025-01-08
fix: make API urls from GitGuardian work with GGShield

sevbch:
## Context

See https://github.com/GitGuardian/gitguardian-vscode/pull/67

## What has been done

When running `ggshield cmd --instance ${INSTANCE}`, if `INSTANCE` matches any of the URLs mentioned https://api.eu1.gitguardian.com/docs, it is replaced by the relevant URL.

### Validate with GGShield
- Checkout this branch
- Run any `ggshield` command having the `--instance` option available with `--instance https://api.gitguardian.com/v1` or `--instance https://api.eu1.gitguardian.com/v1`
- Command should work the same as if you put `--instance https://dashboard.gitguardian.com/` or `--instance https://dashboard.eu1.gitguardian.com/`

### Validate with VSCode Extension
- Checkout this branch
- Checkout `main` at https://github.com/GitGuardian/gitguardian-vscode
- Replace [this line](https://github.com/GitGuardian/gitguardian-vscode/blob/816b609fba6dc09617216b2e3c9f22d14a9a3c83/src/lib/ggshield-configuration-utils.ts#L21) with path to ggshield in your local repo
- Launch the extension in debug mode
- Set extension API url param to https://api.gitguardian.com/v1 or https://api.eu1.gitguardian.com/v1
- Reload window
- You should have logged in perfectly
- You can try adding a secret in a file: when saving it the extension will warn you

## PR check list

### DIFF
--- a/ggshield/core/config/config.py
+++ b/ggshield/core/config/config.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import re
 from enum import Enum
 from pathlib import Path
 from typing import Any, Optional, Set, Tuple
@@ -94,10 +95,10 @@ class Config:
         - the default instance
         """
         if self._cmdline_instance_name:
-            if (
-                "api" in self._cmdline_instance_name
-                and "gitguardian" in self._cmdline_instance_name
-            ):
+            if re.match(
+                r"^https:\/\/api(\.[a-z0-9]+)?\.gitguardian\.com",
+                self._cmdline_instance_name,
+            ) or re.match(r"/exposed/?$", self._cmdline_instance_name):
                 return (
                     api_to_dashboard_url(self._cmdline_instance_name),
                     ConfigSource.CMD_OPTION,
--- a/ggshield/core/config/config.py
+++ b/ggshield/core/config/config.py
@@ -86,6 +86,7 @@ class Config:
         The instance name (defaulting to URL) of the selected instance
         priority order is:
         - set from the command line (by setting cmdline_instance_name)
+          - in case the user set the api url instead of dashboard url, we replace it
         - GITGUARDIAN_INSTANCE env var
         - GITGUARDIAN_API_URL env var
         - in local user config (in user_config.dashboard_url)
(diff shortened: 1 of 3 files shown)


ggshield/core/config/config.py
```diff
@@ -86,13 +86,22 @@ def get_instance_name_and_source(self) -> Tuple[str, ConfigSource]:
...
         - GITGUARDIAN_API_URL env var
         - in local user config (in user_config.dashboard_url)
         - in global user config (in user_config.dashboard_url)
         - the default instance
         """
         if self._cmdline_instance_name:
+            if (
+                "api" in self._cmdline_instance_name
+                and "gitguardian" in self._cmdline_instance_name
+            ):
```
gg-jonathangriffe, line 100: Can the same issue arise in an on-prem instance for which there might not be "gitguardian" in the url ?

gg-jonathangriffe (commented):
Looks good to me ! Thanks for the improvement, I just have a small comment.

ggshield/core/config/config.py
sevbch: It's a good question, I need to investigate

agateau-gg, line 100: This looks a bit too generic to me. We could run into a corner case where an on-prem URL contains both `api` and `gitguardian`. I'd rather use a regex matching on `^https://api(\.[a-z0-9]+)?\.gitguardian\.com`.

agateau-gg (changes requested):

sevbch:
to answer you both @gg-jonathangriffe and @agateau-gg:

I think on-prem urls should be handled as well, as the issue encountered by one of GitGuardian customer with the VSCode Extension could happen too.

The API url looks like [this](https://docs.gitguardian.com/self-hosting/management/application-management/api-configuration) for an on-prem customer.

I'm going to adapt this MR and add a test for this use case.

sevbch:
@gg-jonathangriffe & @agateau-gg : I finished manual checks, this PR is ready for re-review

agateau-gg (approved):

Issue #1032, opened 2024-12-13
`secret scan pre-commit` crashing on Windows when doing big merges


mherzberg:
## Environment

- ggshield version: 1.34.0
- Operating system (Linux, macOS, Windows): Windows
- Operating system version: Windows 11 Pro 23H2
- Python version: 3.12

## Describe the bug

When the user is about to commit a merge with many files and long file names, `ggshield secret scan pre-commit` crashes with `[WinError 206] The filename or extension is too long`. This is because `ggshield` is building and executing a git command that can be longer then the [maximum path length on Windows](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation), which is approximately 32,767 with "long paths" enabled.

Steps to reproduce:

You can use the following bash script in a Git shell on Windows to reproduce the error:

```bash
#!/usr/bin/env bash

cd $(mktemp -d)
git init

function commit_files() {
    for i in $(seq -f "%0200g" 200)
    do
        echo $1 > $i.txt
    done
    git add .
    git commit -m "msg" --no-verify
}

git checkout -b a
commit_files x

git checkout -b b
commit_files foo

git checkout a
commit_files bar

git merge b
git add .

ggshield secret scan pre-commit --verbose --debug
```

This script creates 200 files with file names that are a bit longer than 200 characters each, creates some changes for each of them across two branches, and finally merges the 200 changes from both branches. The kind of merge within this script is a bit artifical but demonstrates the issue that we observed during real merges.

Actual result:

```
Traceback (most recent call last):
  File "C:\Users\User\pipx\venvs\ggshield\Lib\site-packages\ggshield\cmd\utils\common_decorators.py", line 18, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\User\pipx\venvs\ggshield\Lib\site-packages\ggshield\cmd\secret\scan\precommit.py", line 85, in precommit_cmd
    commit = Commit.from_merge(ctx_obj.exclusion_regexes)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\User\pipx\venvs\ggshield\Lib\site-packages\ggshield\core\scan\commit.py", line 87, in from_merge
    shas_in_merge_branch = dict(
                           ^^^^^
[...]

PR #1048, opened 2025-01-10
fix: secret scan pre-commit crashing on big merges

salome-voltz:
## Context

`ggshield secret scan pre-commit` crashes on Windows when committing a merge with many long file names with `[WinError 206] The filename or extension is too long`. 
This happens because ggshield is building a git command that is longer than the maximum path length on Windows. (see #1032 ).

## What has been done

Functions that execute git commands on all files in a merge were updated to process files in batches, respecting the limit defined by `MAX_FILES_PER_GIT_COMMAND`.

## Validation

On Windows:
- Runs the script provided in the issue description. ggshield should crash
- Test with the updated ggshield version. ggshield should proceed as expected.

## PR check list

### DIFF
--- a/changelog.d/20250110_162407_salome.voltz_scrt_5247_ggshield_crashed_on_windows_because_of_git_command_too_long.md
+++ b/changelog.d/20250110_162407_salome.voltz_scrt_5247_ggshield_crashed_on_windows_because_of_git_command_too_long.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Fix `secret scan pre-commit` crashing on big merges (#1032).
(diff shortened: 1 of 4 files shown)


sevbch (commented):
It's looking good to me!

As our CI is running tests in Windows environment, is it possible to design a test which would crash before and now pass? Or are there blockers for such a test?

salome-voltz:
I added a test that crashed on Windows before the change and passes now. The test is a bit slow, however, as it generates a lot of files, and I can't use cassettes to speed it up so I don't know if we want to keep it. WDYT @sevbch ?

sevbch:
> I added a test that crashed on Windows before the change and passes now. The test is a bit slow, however, as it generates a lot of files, and I can't use cassettes to speed it up so I don't know if we want to keep it. WDYT @sevbch ?

I tried locally launching your test and it took 0.2s. It's indeed a bit slow for a unit test but it seems acceptable to me. I'd rather keep the test to prevent us from future regressions.

sevbch (approved):

PR #1053, opened 2025-02-04
Move ScannerUI outside of core.ui

agateau-gg:
While investigating a potential issue in the way we parse commits, I realized I could not import Commit because of an import loop error.

This PR fixes that by removing the dependency from `ui` to `ScannerUI` (which itself rightfully depends on `Commit`). Now it's the other way around: `ScannerUI` depends on `ui`.


salome-voltz (approved):
LGTM

PR #1054, opened 2025-02-05
chore: restrict auth_config file permissions

salome-voltz:
## Context
File `auth_config.yaml` contains sensitive information, but had group and world read permission. 

This MR restricts the file's permission to user-ony read/write permissions.

## PR check list


agateau-gg (changes requested):
Minor remarks.

agateau-gg (approved):

PR #1057, opened 2025-02-25
fix: enable core.longpaths on git commands

sevbch:
## Context

We have been informed of an issue with ggshield and longpaths on Windows in https://github.com/GitGuardian/ggshield/issues/1032.

## What has been done

As the issuer reported, setting `core.longpaths` to true fixes the issue.

ℹ️  Note that this option is only available on the [git-for-windows fork](https://github.com/search?q=repo%3Agit-for-windows%2Fgit%20longpaths&type=code). With the standard git cli, this option has no effect.

## Validation

It was not possible to replicate the encountered issue, but the issuer confirmed [it worked](https://github.com/GitGuardian/ggshield/issues/1032#issuecomment-2694550173) with their original case.

## PR check list

### DIFF
--- a/tests/unit/utils/test_git_shell.py
+++ b/tests/unit/utils/test_git_shell.py
@@ -1,10 +1,10 @@
 import os
 import platform
-import subprocess
 import tarfile
 from io import BytesIO
 from pathlib import Path
 from typing import Optional
+from unittest.mock import Mock, patch
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


tests/unit/utils/test_git_shell.py
```diff
@@ -442,3 +444,37 @@ def test_git_ls_unstaged(tmp_path):
     # as relative to repo.path
     expected_paths = {x.relative_to(repo.path) for x in (repo_file, submodule_file)}
     assert {Path(x) for x in unstaged_files} == expected_paths
+
+
+def test_git_command_includes_longpaths_on_windows(monkeypatch):
+    # GIVEN a mock for subprocess.run
+    commands_run = []
+
+    def mock_subprocess_run(args, **kwargs):
```
salome-voltz, line 453: I think you could do this with a simple mock? 
```
    mock_run = Mock(return_value=Mock(stdout=b""))
    monkeypatch.setattr(subprocess, "run", mock_run)
    git(["status"])

    mock_run.assert_called_once()
    command = mock_run.call_args[0][0]
    ```

salome-voltz (commented):
nit: I am not used to this kind of mock, but I trust you if you think it's better this way

tests/unit/utils/test_git_shell.py
sevbch: much simpler indeed 👍

salome-voltz (approved):

PR #1060, opened 2025-02-27
Prepare release

agateau-gg:
This PR fixes a few issues that prevent running tests without cassettes:

- Docker tests: some tests were reusing a cassette from another test. That fails now that we do different tests depending on whether we ignore files.
- HMSL tests: these tests were getting rate-limited because they were running without an API key.

It also makes `scripts/release` more robust: it always restore cassettes and stops the test Vault server, even when tests fail.


sevbch (approved):
Sorry about the docker tests cassettes 🤦‍♀️ , thanks for fixing it!

Just wondering: which api key are the HSML tests using now? The one from the repo?

PR #1062, opened 2025-03-03
feat: do not block pre-receive when server isn't responding (503)

sevbch:
## Context

Today, when running `ggshield secret scan pre-receive`, if GitGuardian server is temporarily unavailable (returns a response with 503 error code), an error is raised and the command returns `ExitCode.UNEXPECTED_ERROR`, which prevents all commit pushes to get through.

## What has been done

In this particular case, we now display an error in the UI but return 0, allowing for the pushes to proceed.

## Validation

Tested with the new test added to the suite.

## PR check list

### DIFF
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -8,7 +8,12 @@ from requests import Session
 
 from .config import Config
 from .constants import DEFAULT_INSTANCE_URL
-from .errors import APIKeyCheckError, UnexpectedError, UnknownInstanceError
+from .errors import (
+    APIKeyCheckError,
+    ServiceUnavailableError,
+    UnexpectedError,
+    UnknownInstanceError,
+)
 from .ui.client_callbacks import ClientCallbacks
 
 
@@ -103,7 +108,11 @@ def check_client_api_key(client: GGClient) -> None:
         raise APIKeyCheckError(client.base_uri, "Invalid API key.")
     elif response.status_code == 404:
         raise UnexpectedError(
-            "The server returned a 404 error. Check your instance URL" " settings."
+            "The server returned a 404 error. Check your instance URL settings.",
+        )
+    elif response.status_code == 503:
+        raise ServiceUnavailableError(
+            message=f"Server is not responding.\nDetails: {response.detail}",
         )
     else:
         raise UnexpectedError(
--- a/tests/unit/cmd/scan/test_prereceive.py
+++ b/tests/unit/cmd/scan/test_prereceive.py
@@ -2,6 +2,7 @@ from unittest.mock import ANY, Mock, patch
 
 import pytest
 from click.testing import CliRunner
+from requests import Response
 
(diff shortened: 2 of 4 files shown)


ggshield/core/client.py
```diff
@@ -103,7 +108,11 @@ def check_client_api_key(client: GGClient) -> None:
         raise APIKeyCheckError(client.base_uri, "Invalid API key.")
     elif response.status_code == 404:
         raise UnexpectedError(
-            "The server returned a 404 error. Check your instance URL" " settings."
+            "The server returned a 404 error. Check your instance URL settings.",
+        )
+    elif response.status_code == 503:
```
agateau-gg, line 113: Maybe we should expand this to all 5xx errors?

tests/unit/cmd/scan/test_prereceive.py
```diff
@@ -363,3 +364,35 @@ def test_stdin_input_deletion(
...
+        """
+        GIVEN a repo on which the command is ran
+        WHEN the server is not responding (503)
+        THEN it should return 0
+        AND display an error message
+        """
+        # Set up the mock to return a 503 response
+        response = Response()
+        response.status_code = 503
+        response.detail = "Service Temporarily Unavailable"
```
agateau-gg, line 384: It does no matter much here, but `Response` does not have a `detail` attribute. You can use `create_html_response()` to create this response if you want (it's in `tests.unit.request_mock`).

tests/unit/cmd/scan/test_prereceive.py
```diff
@@ -363,3 +364,35 @@ def test_stdin_input_deletion(
...
+        old_sha = repo.get_top_sha()
+        shas = [repo.create_commit() for _ in range(3)]
+        with cd(repo.path):
+            result = cli_fs_runner.invoke(
+                cli,
+                ["-v", "secret", "scan", "pre-receive"],
+                input=f"{old_sha} {shas[-1]} origin/main\n",
+            )
+        assert "Server is not responding" in result.output
+        assert_invoke_exited_with(result, 0)
```
agateau-gg, line 398: We usually call `assert_invoke*()` before other asserts. This makes debugging easier because if the command did not exit with the expected exit code, `assert_invoke*()` prints the command stdout and stderr.

Also there is `assert_invoke_ok()` for the case where the exit code is 0.

agateau-gg (changes requested):

tests/unit/cmd/scan/test_prereceive.py
sevbch: Actually, you comment made me realize I wasn't using the relevant return type (should be `Detail` instead of `Response`)

agateau-gg (approved):
Minor remark on the error message, but this looks good to me.

I didn't know one can set the return value in the `@patch()` decorator, that's nice!

PR #1063, opened 2025-03-06
Fix ggshield error with emojis in file name

sevbch:
## Context

Today, when running `ggshield secret scan pre-commit` on a patch modifying a file which has an emoji in its name, the command fails and raises an error.

## What has been done

Surprisingly, the issue is coming from the git command itself. The header and content are not encoded the same in the output of below command (cf screenshot).

image

We can see that
- in the header the emoji is there
- in the diff part, the emoji was replaced by its utf-8 octal representation (`\360\237\230\212`).

In ggshield internals, we encode both header and content in utf-8. But the emoji in the header is getting encoding by an utf-8 byte (hexadecimal) representation: `\xf0\x9f\x98\x8a`. At some point we check filenames against each other and this is were the issue arises.

From what I understood, octal-representation occurs in some Unix/Linux environments or older programming languages, this is probably why git uses it.

Anyway, proposed fix rely on [core.quotePath](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath) param. See related info on the doc link.

Using this param fixes the issue:
image

## Validation
- commit a file with an emoji in its name
- edit some content in the file
- stage it (git add)
- run `ggshield secret scan pre-commit`

Before: fails, after: works fine.

## PR check list

### DIFF
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -204,7 +204,7 @@ def git(
     try:
         logger.debug("command=%s timeout=%d", command, timeout)
         result = subprocess.run(
-            [_get_git_path()] + ["-c", "core.quotePath=false"] + command,
+            [_get_git_path(), "-c", "core.quotePath=false"] + command,
             check=check,
             capture_output=True,
             timeout=timeout,
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -204,7 +204,7 @@ def git(
     try:
         logger.debug("command=%s timeout=%d", command, timeout)
         result = subprocess.run(
-            [_get_git_path()] + command,
+            [_get_git_path()] + ["-c", "core.quotePath=false"] + command,
             check=check,
             capture_output=True,
             timeout=timeout,
--- a/tests/unit/cmd/scan/test_precommit.py
+++ b/tests/unit/cmd/scan/test_precommit.py
@@ -3,12 +3,15 @@ from pathlib import Path
 
 import pytest
 
+from ggshield.__main__ import cli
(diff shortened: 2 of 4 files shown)


ggshield/utils/git_shell.py
```diff
@@ -204,7 +204,7 @@ def git(
     try:
         logger.debug("command=%s timeout=%d", command, timeout)
         result = subprocess.run(
-            [_get_git_path()] + command,
+            [_get_git_path()] + ["-c", "core.quotePath=false"] + command,
```
agateau-gg, line 207: Can you add a comment explaining why we set `core.quotePath` to false?

nit, the two lists can be merged:
```suggestion
            [_get_git_path(), "-c", "core.quotePath=false"] + command,
```

tests/unit/cmd/scan/test_precommit.py
```diff
@@ -98,3 +101,30 @@ def test_get_merge_branch_from_reflog(monkeypatch):
...
+
+@my_vcr.use_cassette("test_emoji_filename")
+def test_precommit_with_emoji_filename(tmp_path, cli_fs_runner):
+    """
+    GIVEN a repository with a staged diff on a file with an emoji in its name
+    WHEN the precommit command is run
+    THEN it executes successfully
+    """
+    # Set up repository
+    repo = Repository.create(tmp_path, initial_branch="main")
```
agateau-gg, line 114: nit: I am not sure why the other tests use `initial_branch="master"`, but `"main"` is the default value for `initial_branch`, so it's not necessary to set it.

agateau-gg (approved):
Nice!
Just some minor remarks.

PR #1068, opened 2025-03-19
chore: remove support for python 3.8

gg-jonathangriffe:
## Context

We have previously added deprecation warning when the python version was 3.8, to prepare removing its support in GGShield.

## What has been done
This MR removes support for GGShield 3.8

## PR check list

### DIFF
--- a/tests/unit/utils/test_files.py
+++ b/tests/unit/utils/test_files.py
@@ -1,5 +1,4 @@
 import re
-import sys
 import tarfile
 from io import BytesIO
 from pathlib import Path, PurePath, PurePosixPath, PureWindowsPath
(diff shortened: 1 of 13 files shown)


tests/unit/utils/test_files.py
```diff
@@ -25,8 +24,7 @@ def test_get_empty_tar():
 
     # THEN the file is considered as a .tar
     # `tarfile.is_tarfile` won't work until Python 3.9
-    if sys.version_info >= (3, 9):
-        assert tarfile.is_tarfile(tar_stream)
+    assert tarfile.is_tarfile(tar_stream)
```
agateau-gg, line 27: nit: can you remove the comment about 3.9?

agateau-gg (approved):
One tiny remark, but looks good! Approving in advance.

PR #1072, opened 2025-03-21
Use system certificates

### LINKED ISSUE #1067: RFE: Support OS Trust Store
## Is your feature request related to a problem? Please describe.

GGShield uses the Python `certifi` package as the source of the SSL trust store.  `certifi` bundles its own trust store (does not use the OS trust store) and provides no means to alter this behavior.

Corporate environment regularly deploy SSL decryption / MITM security solutions for HTTPS traffic inspection which use SSL certificates signed by some authority that is not included in `certifi`'s list (such as a private corporate CA).

## Describe the solution you'd like

Make use of the Python package `truststore` to create an SSL context that trusts the OS trust store, hence putting control back into the customer's hands.

https://pypi.org/project/truststore/

## Describe alternatives you've considered

Manual update of the embedded `cacert.pem` file that ships with `certifi` (boo).

agateau-gg:
## Context

This PR makes ggshield use the system certificates instead of the bundled ones.

Fixes #1067.

## What has been done

Integrate [truststore](https://pypi.org/project/truststore/) and use it. Note that this only works with Python >= 3.10.

The 2nd commit changes the CI so that the Linux standalone installer uses Python 3.10 instead of 3.9.

## Validation

I tested this using `mitmproxy`:
- Install it (`pipx install mitmproxy` for example)
- Follow [the documentation](https://docs.mitmproxy.org/stable/overview-getting-started/) to generate and install its certificate
- Tell ggshield to use the proxy (`export https_proxy=http://localhost:8080`)
- Run `ggshield api-status` using the main branch → it crashes with SSL errors
- Run `ggshield api-status` using this branch → it does not crash and you can see the exchange in mitmproxy UI

## PR check list

### DIFF
--- a/.github/workflows/build_release_assets.yml
+++ b/.github/workflows/build_release_assets.yml
@@ -54,6 +54,8 @@ jobs:
           - os: ubuntu-22.04
             # When building on Linux we use a container to build using an old enough version
             container: rockylinux/rockylinux:8
+            sha256sum:
+              python: e8d7ed8c6f8c6f85cd083d5051cafd8c6c01d09eca340d1da74d0c00ff1cb897
           - os: windows-2022
           - os: macos-13
             arch: x86_64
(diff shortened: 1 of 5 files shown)


sevbch (approved):
LGTM! Nice workaround for the python installation in CI, it must have been a bit of a pain to settle

PR #1075, opened 2025-03-28
Jgriffe/fix update vscode ggshield

gg-jonathangriffe:
## Context

Last release the "Update VSCode Extension" job failed. This MR fixes it


sevbch (approved):
Nice catch indeed

PR #1078, opened 2025-04-02
chore(ci): add job to push new version to chocolatey

gg-jonathangriffe:
## What has been done

A job was added to automatically publish new ggshield versions on chocolatey.
This MR also introduces the ggshield.nuspec used for the ggshield chocolatey package.

## Validation

- Test the push-to-chocolatey script
- Download the nupkg and install it on a windows VM with "choco install ggshield -source ."

### DIFF
--- a/scripts/push-to-chocolatey/push-to-chocolatey
+++ b/scripts/push-to-chocolatey/push-to-chocolatey
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+die() {
+    echo $* >&2
+    exit 1
+}
+
+if [ -z "${CHOCOLATEY_API_KEY:-}" ] ; then
+    die '$CHOCOLATEY_API_KEY is not set'
+fi
+
+if ! command -v unzip > /dev/null ; then
+    apt update
+    apt install unzip
+fi
+
+zip_name=$(ls packages)
+unzip packages/$zip_name -d packages
+
+version=$(echo "$zip_name" | grep -oP '(?<=ggshield-)[0-9.]+')
+
+mkdir ggshield-package
+mkdir ggshield-package/tools
+
+mv packages/*/_internal ggshield-package/tools
+mv packages/*/ggshield.exe ggshield-package/tools
+cp scripts/push-to-chocolatey/ggshield.nuspec ggshield-package
+sed -i "s/__VERSION__/$version/" ggshield-package/ggshield.nuspec
+
+cd ggshield-package
+choco pack
+
+choco push ggshield.$version.nupkg --source https://push.chocolatey.org/ --api-key $CHOCOLATEY_API_KEY
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -175,3 +175,23 @@ jobs:
           scripts/push-to-cloudsmith
         env:
           CLOUDSMITH_API_KEY: ${{ secrets.CLOUDSMITH_API_KEY }}
+
+  push_to_chocolatey:
+    needs: build_release_assets
+    name: Push to Chocolatey
+    runs-on: chocolatey/choco
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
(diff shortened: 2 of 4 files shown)


scripts/push-to-chocolatey/push-to-chocolatey
```diff
@@ -0,0 +1,32 @@
...
+zip_name=$(ls packages)
+unzip packages/$zip_name -d packages
+
+version=$(echo "$filename" | grep -oP '(?<=ggshield-)[0-9.]+')
+
+mkdir ggshield-package
+mkdir ggshield-package/tools
+
+mv  packages/*/_internal ggshield-package/tools
+mv  packages/*/ggshield.exe ggshield-package/tools
```
agateau-gg, line 25: nit:

```suggestion
mv packages/*/_internal ggshield-package/tools
mv packages/*/ggshield.exe ggshield-package/tools
```

scripts/push-to-chocolatey/push-to-chocolatey
```diff
@@ -0,0 +1,32 @@
...
+    echo $* >&2
+    exit 1
+}
+
+if [ -z "${CHOCOLATEY_API_KEY:-}" ] ; then
+    die '$CHOCOLATEY_API_KEY is not set'
+fi
+
+apt update
+apt install unzip
```
agateau-gg, line 14: Maybe wrap this in an `if` so that it's skipped if unzip is there (useful when debugging: it's faster and removes the need to run the script as root):

```suggestion
if ! command -v unzip > /dev/null ; then
    apt update
    apt install unzip
fi
```

scripts/push-to-chocolatey/push-to-chocolatey
```diff
@@ -0,0 +1,32 @@
...
+
+mv  packages/*/_internal ggshield-package/tools
+mv  packages/*/ggshield.exe ggshield-package/tools
+cp scripts/push-to-chocolatey/ggshield.nuspec ggshield-package
+sed -i "s/__VERSION__/$version/" ggshield-package/ggshield.nuspec
+
+cd ggshield-package
+choco pack
+
+choco push ggshield.$version.nupkg --source https://push.chocolatey.org/ --api-key $CHOCOLATEY_API_KEY
```
agateau-gg, line 32: It's not a good practice to pass secrets via the CLI (arguments are visible to however can run `ps` on the machine). Is it possible to pass it via a config file or an environment variable instead?

agateau-gg (changes requested):

scripts/push-to-chocolatey/push-to-chocolatey
gg-jonathangriffe: I haven't found a way to do so.
To the best of my knowledge there are two ways of passing the API key :
- With --api-key
- By running beforehand choco apikey add -s source -k key

agateau-gg: OK, let's keep it this way then.

scripts/push-to-chocolatey/push-to-chocolatey
```diff
@@ -0,0 +1,34 @@
...
+
+if ! command -v unzip > /dev/null ; then
+    apt update
+    apt install unzip
+fi
+
+zip_name=$(ls packages)
+unzip packages/$zip_name -d packages
+
+version=$(echo "$filename" | grep -oP '(?<=ggshield-)[0-9.]+')
```
salome-voltz, line 21: What does filename refers to here ?

sevbch:
I didn't actually review, but I was thinking this could be definitely worth a changelog entry, to say ggshield will now be available on chocolatey! 🍫 

WDTY?

scripts/push-to-chocolatey/push-to-chocolatey
gg-jonathangriffe: good catch this should be zip_name

salome-voltz (approved):

PR #1080, opened 2025-04-15
chore: fix chocolatey package to follow guidelines

gg-jonathangriffe:
## Context

The chocolatey package failed verification in the package repository

## What has been done

A VERIFICATION.txt has been added with instructions on how to verify the binaries's authenticity.
a packageSourceUrl field has been added.

### DIFF
--- a/scripts/push-to-chocolatey/VERIFICATION.txt
+++ b/scripts/push-to-chocolatey/VERIFICATION.txt
@@ -1,0 +1,31 @@
+VERIFICATION
+Verification is intended to assist the Chocolatey moderators and community
+in verifying that this package's contents are trustworthy.
+
+1. Go to the releases page and download the windows package for the relevant version:
+https://github.com/GitGuardian/ggshield/releases
+
+2. Extract the archive
+
+3. Get the hashes from both the downloaded release and the choco package. The hashes should match.
+
+To verify all files, you can use the following powershell script:
+
+$folder1 = "C:\Absolute\Path\To\Folder1"
+$folder2 = "C:\Absolute\Path\To\Folder2"
+
+$hashes1 = Get-ChildItem -Path $folder1 -Recurse -File | ForEach-Object {
+    [PSCustomObject]@{
+        RelativePath = $_.FullName.Substring($folder1.Length).TrimStart('\')
+        Hash = (Get-FileHash $_.FullName).Hash
+    }
+}
+
+$hashes2 = Get-ChildItem -Path $folder2 -Recurse -File | ForEach-Object {
+    [PSCustomObject]@{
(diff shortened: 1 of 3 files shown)


scripts/push-to-chocolatey/VERIFICATION.txt
```diff
@@ -0,0 +1,13 @@
...
+Verification is intended to assist the Chocolatey moderators and community
+in verifying that this package's contents are trustworthy.
+
+1. Go to the releases page and download the windows package for the relevant version:
+https://github.com/GitGuardian/ggshield/releases
+
+2. Extract the archive
+
+3. Get the hashes from both the downloaded release and the choco package: 
+powershell -command Get-FileHash *
```
sevbch, line 11: Did you test it on a local windows machine?

scripts/push-to-chocolatey/ggshield.nuspec
```diff
@@ -11,6 +11,7 @@
     <licenseUrl>https://raw.githubusercontent.com/GitGuardian/ggshield/refs/heads/main/LICENSE</licenseUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <projectSourceUrl>https://github.com/GitGuardian/ggshield</projectSourceUrl>
```
sevbch, line 13: I think you may need to remove `ProjectSourceUrl` balise because it's the same as `ProjectUrl` (but it's unclear as the doc isn't filled out).
> ProjectUrl and ProjectSourceUrl are typically different, but not always. Please ensure that projectSourceUrl is pointing to software source code or remove the field from the nuspec. [More...](https://docs.chocolatey.org/en-us/community-repository/moderation/package-validator/rules/cpmr0041)

Also, it looks like it's not blocking but could you also address this icon topic (about `iconUrl` balise)? I'm afraid it could result in a bad rendering once published. We probably have some good quality images in marketing materials.
> As per the [packaging guidelines](https://docs.chocolatey.org/en-us/create/create-packages#package-icon-guidelines) icons should be either a png or svg file. [More...](https://docs.chocolatey.org/en-us/community-repository/moderation/package-validator/rules/cpmr0058)

sevbch: Alright then; thank you for looking into it!

scripts/push-to-chocolatey/VERIFICATION.txt
gg-jonathangriffe: I did, it works but does not allow verifying all files recursively.
I now added a new script to allow automatically verifying all files.

sevbch: nice, thanks!

sevbch (approved):

PR #1083, opened 2025-04-22
feat: Add extra scan results information

salome-voltz:
## PR check list


agateau-gg (changes requested):

I am wondering whether it would make sense to add this to our JSON and SARIF outputs too, though. WDYT?

salome-voltz:
> I am wondering whether it would make sense to add this to our JSON and SARIF outputs too, though. WDYT?

I asked Mathieu about it, and he favored keeping this text output only. Since JSON is primarily made for machines, it's probably not very useful to add a URL there. I'm not sure about SARIF, though, since it could end up in UI.

ggshield/verticals/secret/output/secret_text_output_handler.py
```diff
@@ -308,6 +308,8 @@ def secret_header(
 {indent}Incident URL: {secrets[0].incident_url if known_secret and secret.incident_url else "N/A"}
 {indent}Secret SHA: {ignore_sha}
 """
+    if secret.documentation_url is not None:
+        message += f"{indent}Detector documentation: {secret.documentation_url}#revoke-the-secret\n"
```
sevbch, line 312: The url being already quite long, and also to be consistent with the detector wording, it would be better to remove the trailing `#revoke-the-secret` IMO

agateau-gg:
> > I am wondering whether it would make sense to add this to our JSON and SARIF outputs too, though. WDYT?
> 
> I asked Mathieu about it, and he favored keeping this text output only. Since JSON is primarily made for machines, it's probably not very useful to add a URL there.

But it could then be used in the VS Code extension, no? I think this would be nice.

agateau-gg (approved):

PR #1084, opened 2025-04-22
fix: let tarfile auto-detect archive type

sevbch:
## Context

When scanning a docker image, one could encounter this error when scanning layers files:

`tarfile.ReadError: invalid header`

## What has been done
This happened because when dealing with layers files, code was implicitly assuming they were `.tar` files by using `tarfile.TarFile`.

Instead, we use `tarfile.open` which automatically detects the compression format (based on file extension, or in our case headers) and returns the relevant `TarFile` instance (matching the appropriate compression/decompression algorithm).

## Validation
Added a non-regression test. It contains a layer with a `.tar.gz` file. Parsing failed before, doesn't fail anymore.

### Note
In docker image `manifest.json` file, sometimes there is a `mediaType` field indicating the content type of each component. However this field isn't mandatory (it's part of the Docker Registry API V2 specs) so we can't rely on it to choose the relevant decompression method.

Also it's simpler to let `tarfile` module handle it for us.

## PR check list


changelog.d/20250422_185842_severine.bonnechere_handle_tar_gz_in_docker_images.md
```diff
@@ -0,0 +1,3 @@
+### Fixed
+
+- Fixed error when scanning `.tar.gz` files inside docker images.
```
agateau-gg, line 3: nit: it's not really .tar.gz files inside docker images (as in a Dockerfile with a `COPY archive.tar.gz` command), it's more about tar.gz-compressed layers.

agateau-gg (approved):
nitpicking on the changelog, but looks good, approved.

PR #1085, opened 2025-04-24
Jgriffe/add choco package icon

gg-jonathangriffe:
## Context

In order for the ggshield chocolatey package to follow the community repository guidelines, we need change the iconUrl so that the icon is served by a static CDN. We also need to add a LICENSE file. 

## What has been done

The push-to-chocolatey script now copies the ggshield LICENSE file to the tools folder of the package so that it is included.

The icon is included in ggshield (but not in the chocolatey package), and the iconUrl in the nuspec points to an url of statically.io that will cache the icon from the ggshield github repo and and serve it.

As per the chocolatey community repository guidelines, the ref used is a commit and not a branch, so that the icon can be changed by changing the iconUrl, without having to wait for the statically.io cache to be invalidated (if it ever is).


scripts/push-to-chocolatey/README.md
```diff
@@ -0,0 +1,7 @@
+# Push to Chocolatey
+
+This folder contains everything necessary to build and publish the chocolatey package.
+
+The icon is not included in the package, but instead it is served through a static CDN and its url is in the iconUrl tag in the .nuspec.
```
agateau-gg, line 5: Is it possible to add the icon to the package itself and refer to it? It would avoid relying on a third-party service.

Not a blocker if it's not possible, just be careful not to rebase this PR so that the URL remains valid.

gg-jonathangriffe: Unfortunately this is not in accordance with the chocolatey community repo guidelines, here is the feedback we received :
> Can you use a static CDN for your <iconUrl> - githack.com and jsdelivr.com are popular. See https://docs.chocolatey.org/en-us/create/create-packages#package-icon-guidelines

We do have a static CDN, but using it for this would be a diverted usage and it might break unexpectedly.
This solution at least keeps both the icon and the url used in the repo.

PR #1095, opened 2025-05-07
feat(nhi): add secret is_vaulted info in ggshield output

amascia-gg:
Use the newly introduced `is_vaulted` field to modify text and json outputs of ggshield.
Works in staging:

Screenshot 2025-05-07 at 14 31 01

With this invalid, dummy secret:

Screenshot 2025-05-07 at 14 31 57


ggshield/verticals/secret/output/secret_json_output_handler.py
```diff
@@ -141,6 +141,8 @@ def serialized_secret(
                 secrets[0].ignore_reason
             )
 
+        if secrets[0].is_vaulted:
+            flattened_dict["secret_vaulted"] = secrets[0].is_vaulted
```
xblanchot-gg, line 145: Not sure I understand why we only check the first element of the list for all the fields here, but that seem consistent with the rest of the implem

xblanchot-gg (approved):
That looks good to me, thanks !

ggshield/verticals/secret/output/secret_json_output_handler.py
amascia-gg: The list contain occurrences of the same secret if my understanding is correct. That's why we can take the metadata from only the first occurrence

tests/unit/verticals/secret/output/test_text_output.py
```diff
@@ -264,3 +264,37 @@ def test_ignore_reason(ignore_reason):
...
+            id="scan",
+            type="scan",
+            results=Results(results=[result], errors=[]),
+        )
+    )
+
+    if is_vaulted:
+        assert "Secret in Secrets Manager: True" in output
+    else:
+        assert "Secret in Secrets Manager: False" in output
```
sevbch, line 300: what about  
```suggestion
        assert f"Secret in Secrets Manager: {is_vaulted}" in output
```

sevbch (commented):
LGTM!

Did you look into adding this to the sarif output? For reference we had this MR https://github.com/GitGuardian/ggshield/pull/1083/commits/8ac31b6b1dfa80d32116c4de3ceecd570196b684 recently which interacted with it.

salome-voltz:
You can update py-gitguardian version in the pyproject.toml, we'll update the version once py-gitguardian is released :) 
see [this MR for reference](https://github.com/GitGuardian/ggshield/pull/1083/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711)

agateau-gg (changes requested):
Looks good! My only request would be to add the field you added in the JSON output to https://github.com/GitGuardian/ggshield/blob/main/doc/schemas/secret.json.

PR #1096, opened 2025-05-14
Make `scan docker` scan files in /usr/src/app

agateau-gg:
## Context

`/usr/src/app` is often used to store the containerized app, so we should scan it.

## What has been done

Tune the file ban list to ensure we scan the wanted directory.

## Validation

- Put a secret in a file called config.py
- Create a Dockerfile with this content:
```
FROM scratch

COPY config.py /usr/src/app
```
- Build an image for it: `docker build -t ggshield-src-app .`
- Scan it with ggshield: `ggshield secret scan docker ggshield-src-app` → it finds the secret

## PR check list

### DIFF
--- a/ggshield/verticals/secret/docker.py
+++ b/ggshield/verticals/secret/docker.py
@@ -33,7 +33,7 @@ if TYPE_CHECKING:
     from ggshield.core.scan import ScanContext
 
 FILEPATH_BANLIST = [
-    r"^/?usr/(?!share/nginx)",
+    r"^/?usr/(?!share/nginx|src/app)",
     r"^/?lib/",
     r"^/?share/",
     r"^/?bin/",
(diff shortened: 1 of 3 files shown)


sevbch (approved):

ggshield/verticals/secret/docker.py
```diff
@@ -33,7 +33,7 @@
     from ggshield.core.scan import ScanContext
 
 FILEPATH_BANLIST = [
-    r"^/?usr/(?!share/nginx)",
+    r"^/?usr/(?!share/nginx|src/app)",
```
sevbch, line 36: Actually, just wondering, why not just add the `src` part? Is `src/app` typical of docker containers?

agateau-gg: According to Guillaume, it is indeed typical. Furthermore, some packages install files in /usr/src (random example: [linux-headers--6.14.0-1002-realtime](https://packages.ubuntu.com/plucky/amd64/linux-headers-6.14.0-1002-realtime/filelist)), and it's better not to waste time scanning these files.

PR #1098, opened 2025-05-15
Do not require git

agateau-gg:
## Context

When running on a machine where git is not in $PATH, ggshield would crash because there are a few places that tries to use git at startup. This is annoying because there are valid situations (such being installed inside a Docker image) where git is not available but we would like to use non-git-related scan commands such as `secret scan path`.

## What has been done

- Add a test that tries to run `secret scan path` when git is not available
- Fix all crashes so that the test pass

## Validation

- Install ggshield in a Docker image that does not have git installed.
- Try to scan a path, it should not crash.

## PR check list


sevbch (approved):
LGTM! Definitely an improvement

PR #1105, opened 2025-06-20
chore: fix Secret typing following py-gitguardian changes

gg-jonathangriffe:
## Context

Changes have been made in py-gitguardian to fix the compatibility with older GitGuardian API versions. 

The typing in ggshield must now be updated


agateau-gg (changes requested):
Can you update the pyproject.toml to use the git version of py-gitguardian until it's released?

PR #1106, opened 2025-06-23
fix: install libstdcpp for clearlinux

sevbch:
## Context

[CI is failing](https://github.com/GitGuardian/ggshield/actions/runs/15779418933/job/44482026043) on `clearlinux` tests because libstdcpp is not installed in the docker image for `clearlinux`. This is most likely new and coming from the latest version of the [docker image](https://hub.docker.com/_/clearlinux)  but we can't know for sure as the image isn't versioned. 

## What has been done
- install libstdcpp lib for clearlinux
- cosmetic changes: move all git installs in the new requirements step

## Validation

CI failed before, succeeds after.

## PR check list

### DIFF
--- a/.github/workflows/build_release_assets.yml
+++ b/.github/workflows/build_release_assets.yml
@@ -292,6 +292,14 @@ jobs:
           - clearlinux:latest
 
     steps:
+      - name: Install requirements
+        run: |
+          case "${{ matrix.image }}" in
+            clearlinux*)
+              swupd bundle-add libstdcpp
+              ;;
+          esac
+
       - name: Download OS packages
         uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/build_release_assets.yml
```diff
@@ -292,6 +292,25 @@ jobs:
...
+              apt install --no-install-recommends -y git
+              ;;
+            rockylinux*)
+              yum install -y git-core
+              ;;
+            opensuse*)
+              zypper install -y git-core
+              ;;
+            clearlinux*)
+              swupd bundle-add libstdcpp
```
agateau-gg, line 311: Maybe add a comment explaining why we need to install this package, otherwise future-us are going to scratch their heads :).

agateau-gg (approved):
Approved with a request for comment :)

.github/workflows/build_release_assets.yml
sevbch: done in b2c3343f

PR #1109, opened 2025-06-25
feat: add new parameter source_uuid to secret scan

alina-tuholukova-gg:
Add new parameter source_uuid, when provided, the scan will also create incidents on GIM dashboard

## Context

When calling `ggshield secret scan` with --source-uuid, the scan will also create incidents on GIM dashboard.

## Validation

To validate change:
- Create a custom source on GIM
- Copy its uuid
- Call `ggshield secret scan` with --source-uuid on some documents that have secrets
- Check that the corresponding issues are created on GIM

## PR check list


pdm.lock
alina-tuholukova-gg, line 1: this is just to check the pipeline with the new py-gitguardian

ggshield/verticals/secret/secret_scanner.py
```diff
@@ -70,6 +70,16 @@ def __init__(
...
 
+        if secret_config.source_uuid:
+            response = self.client.api_tokens()
+
+            if not isinstance(response, (Detail, APITokensResponse)):
+                raise UnexpectedError("Unexpected api_tokens response")
+            elif isinstance(response, Detail):
+                raise UnexpectedError(response.detail)
+            if TokenScope.SCAN_CREATE_INCIDENTS not in response.scopes:
+                raise MissingScopesError([TokenScope.SCAN_CREATE_INCIDENTS])
```
agateau-gg: In certain situations, for example when scanning commit ranges, ggshield creates multiple SecretScanner instances. This is why we have the `if check_api_key:` at the start of the method: we only check the API key validity if it has not already be done by the caller.

I think these new checks should be done in `check_client_api_key()`. It does not have access to the config though, so you need to extend the function to accept a `source_uuid` optional argument.

agateau-gg (changes requested):

agateau-gg (changes requested):
Sorry, looking at the code, I realize my suggestion was not very good. Plus there is already some code that calls `client.api_tokens()` in `SecretScanner.__init__()`, which is not ideal.

Here is a better (I think, but do tell me if you disagree!) suggestion:

- Change `check_client_api_key()` to take a mandatory second argument that would be a set of the required scopes.

- Change `check_client_api_key()` callers to pass this second argument:

    - oauth.py can pass an empty set (it only cares if the key is valid)
    - secret_scanner.py and repo.py can generate the required set doing something like this:

```python
scopes = {TokenScope.SCAN}
if secret_config.with_incident_details:
    scopes.add(TokenScope.INCIDENTS_READ)
if secret_config.source_uuid:
    scopes.add(TokenScope.SCAN_CREATE_INCIDENTS)
check_client_api_key(client, scopes)
```

What do you think?

changelog.d/20250625_161705_alina.tuholukova.md
```diff
@@ -0,0 +1,4 @@
+### Changed
+
+- `ggshield secret scan` now provides an `--source-uuid` option. When this option is set, it will create the incidents on the GIM
+  dashboard on the correspoding source. Note, that the token should have the scope `scan:create-incidents`.
```
agateau-gg: nitpick: typos
```suggestion
  dashboard on the corresponding source. Note that the token should have the scope `scan:create-incidents`.
```

corenting: do we want to officially document the option for now, or do we prefer to mark it as beta or hidden?

ggshield/core/client.py
```diff
@@ -118,3 +122,38 @@ def check_client_api_key(client: GGClient) -> None:
...
+            raise UnexpectedError(response.detail)
+
+        missing_scopes = required_scopes - set(
+            TokenScope(scope) for scope in response.scopes
+        )
+        if missing_scopes:
+            raise MissingScopesError(list(missing_scopes))
+
+
+def get_required_token_scopes_from_config(
```
agateau-gg: It's a good idea to factorize this code, but I would prefer if it were kept in the secret vertical. Maybe you can move this function in verticals/secret_scanner.py?

agateau-gg (changes requested):
Looks nicer now. Just one last change and this can go in.

changelog.d/20250625_161705_alina.tuholukova.md
mathieubellon: I prefer we officially document and mark it as beta

agateau-gg (approved):
Looks good, thanks!

Issue #952, opened 2024-08-22
Please provide an ARM based Docker image


testworksau:
## Is your feature request related to a problem? Please describe.

GitHub now supports ARM based Actions runners, and we'd like to make use of them as they are 37% cheaper than x64 runners.

https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/

## Describe the solution you'd like

Provide a Docker image for the ARM platform.

iress-ac:
@sevbch Thanks for your effort - it seems `:unstable` works but `:latest` hasn't been updated yet

agateau-gg:
> [@sevbch](https://github.com/sevbch) Thanks for your effort so far - it seems `:unstable` works but `:latest` hasn't been updated yet

Indeed, `:latest` will be updated on the next ggshield release.

PR #1112, opened 2025-07-09
feat: push linux/arm64 docker images

sevbch:
## Context

Solves #952 

- update docker build and push flow in CI: we were using action docker/build-push-action v1 while latest version is v6. Multi-platform isn't supported in v1.
- Provide multi-platform images for linux/amd64 and linux/arm64.

## What has been done
Multi-plaform support is ensured by [this line](https://github.com/GitGuardian/ggshield/pull/1112/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR198)

Most of the work was about updating the jobs to latest version.

Main points of attention / notes:
- ghcr.io is the same as docker.pkg.github.com which is deprecated, see [here](https://docs.github.com/en/packages/working-with-a-github-packages-registry/migrating-to-the-container-registry-from-the-docker-registry)
- There is 1 unique job pushing to both docker and github registries (instead of 2 previously). This is the standard way of doing as per [docker docs](https://docs.docker.com/build/ci/github-actions/push-multi-registries/)
- Main difference between ci job and tag job is the use of docker/metadata-action [here](https://github.com/GitGuardian/ggshield/pull/1112/commits/62169571accd2daed033bf66cc28e6a8654f656e#diff-84dff8d1094ca39c02ac0e48d951ca22f4da29c76b50ae517f5bd2d50f94c2f6R100). This is required to replace previous option `tag_with_ref: true` which doesn't exist anymore. The writing is largely inspired by [this example](https://docs.docker.com/build/ci/github-actions/manage-tags-labels/) and the [action doc](https://github.com/docker/metadata-action?tab=readme-ov-file#typeref).

## Validation
Commit 2 of this PR shows that the job is working for the `ci` workflow (commit will be revert before merge)
- see job [here](https://github.com/GitGuardian/ggshield/actions/runs/16195470450/job/45723620174)
- see `unstable` version on [docker hub](https://hub.docker.com/r/gitguardian/ggshield) updated today
- see `unstable` version on github hub updated today:
```bash
$ docker pull ghcr.io/gitguardian/ggshield/ggshield:unstable
$ docker inspect ghcr.io/gitguardian/ggshield/ggshield:unstable | grep Created 
"Created": "2025-07-10T09:34:20.054697506Z",
```

⚠️ Unfortunately I have no mean of checking that the `tag` job will work as expected (regarding the version tagging part, using docker/metadata-action).
I think it should work as I followed the doc and example cited above and also saw [other examples](https://github.com/mlflow/mlflow/blob/master/.github/workflows/push-images.yml#L33) on github which seem to work.
If you have an idea how to test it I'm interested.

Regarding the [`latest` tag](https://github.com/GitGuardian/ggshield/pull/1112/commits/62169571accd2daed033bf66cc28e6a8654f656e#diff-84dff8d1094ca39c02ac0e48d951ca22f4da29c76b50ae517f5bd2d50f94c2f6R136), I'm confident it will work as expected because it's exactly the same logic as in the `ci` flow tested here.

## PR check list

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,8 +164,8 @@ jobs:
           GITGUARDIAN_API_KEY: ${{ secrets.GITGUARDIAN_API_KEY }}
           GITGUARDIAN_API_URL: ${{ secrets.GITGUARDIAN_API_URL }}
 
-  dockerhub-unstable:
-    name: Push Docker image to Docker Hub
+  push_docker_images-unstable:
+    name: Push Docker image to Docker Hub and GitHub Packages
     runs-on: ubuntu-22.04
     if: github.ref == 'refs/heads/main' && github.event_name == 'push'
     needs:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -164,41 +164,39 @@ jobs:
           GITGUARDIAN_API_KEY: ${{ secrets.GITGUARDIAN_API_KEY }}
           GITGUARDIAN_API_URL: ${{ secrets.GITGUARDIAN_API_URL }}
 
-  dockerhub-unstable:
-    name: Push Docker image to Docker Hub
+  dockerhub-github_packages-unstable:
+    name: Push Docker image to Docker Hub and GitHub Packages
```
agateau-gg: nit: maybe shorten the job name to something like `publish_docker_images`? Long names are cropped in the sidebar:

image

agateau-gg (approved):
Approved, but maybe it would make sense to create a separate workflow file like the build_release_assets.yml and update_vscode_extension.yml workflows?

agateau-gg (approved):

PR #1113, opened 2025-07-17
feat: remove filepaths from docker banlist

guedou:
## Context

This PR adds filepath that likely contain application code and should be scanned.


gg-gvaladon:
@agateau-gg done

PR #1115, opened 2025-07-25
chore: update number of detectors

sevbch:
Note: I already updated the "About" section of the repo


salome-voltz (approved):
:partying_face:

Issue #832, opened 2024-01-10
GGShield test suite fails with Python 3.12


agateau-gg:
GGShield test suite currently does not pass with Python 3.12. It fails with that error message:

```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/pytest/__main__.py", line 5, in <module>
    raise SystemExit(pytest.console_main())
                     ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/_pytest/config/__init__.py", line 190, in console_main
    code = main()
           ^^^^^^
[...]
  File "/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/snapshottest/__init__.py", line 3, in <module>
    from .module import assert_match_snapshot
  File "/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/_pytest/assertion/rewrite.py", line 168, in exec_module
    exec(co, module.__dict__)
  File "/opt/hostedtoolcache/Python/3.12.1/x64/lib/python3.12/site-packages/snapshottest/module.py", line 4, in <module>
    import imp
ModuleNotFoundError: No module named 'imp'
[...]
Error: Process completed with exit code 1.
```

This is because the `snapshottest` package uses the deprecated-and-now-gone `imp` module. There is a PR to fix it [here](https://github.com/syrusakbary/snapshottest/pull/168) but it is not moving. It looks like snapshottest is unmaintained. We need to either use the code from the PR (ideally from our own fork) or replace snapshottest with a maintained Python package (https://github.com/joseph-roitman/pytest-snapshot could be a good candidate).

GG-HH:
Hello @agateau-gg, I tried to give a shot by replacing it with [syrupy](https://github.com/syrupy-project/syrupy) as it sound like an almost drop-in replacement. Can I get your opinion on this ?
I gave it a go through this [PR](https://github.com/GitGuardian/ggshield/pull/1118)

agateau-gg:
> Hello [@agateau-gg](https://github.com/agateau-gg), I tried to give a shot by replacing it with [syrupy](https://github.com/syrupy-project/syrupy) as it sound like an almost drop-in replacement. Can I get your opinion on this ? I gave it a go through this [PR](https://github.com/GitGuardian/ggshield/pull/1118)

[...]

PR #1118, opened 2025-08-07
chore: replace snapshottest with syrupy and support py3.12

GG-HH:
## Context

An attempt to fix https://github.com/GitGuardian/ggshield/issues/832 and bring python3.12 support.

## What has been done

1. Replace snapshottest with [syrupy](https://github.com/syrupy-project/syrupy)
2. Add py3.12 in the test matrix

## Validation

Run tests with a python3.12 environment
Ensure there is nowhere where I should update to python3.12 that was forgotten

## PR check list


tests/unit/core/test_filter.py
```diff
@@ -69,7 +68,7 @@ def test_get_ignore_sha(
     policy_breaks: List[PolicyBreak],
     duplicates: bool,
     expected_shas: Set[str],
-    snapshot: Snapshot,
+    snapshot,
```
agateau-gg: Looks like this test function does not actually uses snapshots, so we can remove this argument, unless I am missing something.

GG-HH: Indeed! I did not read thoroughly enough this change.
I removed it and indeed, it seemed unused :tada:

agateau-gg (approved):
Thanks for taking care of this!

PR #1120, opened 2025-08-13
fix: update error messages to specify "Invalid GitGuardian API key"

mathieubellon:
## Context

Our error message ("Error: Invalid API key.") is too generic. In some cases, it confuses developers, who may suspect that their Git credentials or token are the source of the problem, leading to a costly waste of time.

## What has been done

We changed the error strings to be more specific.

## PR check list


sevbch (approved):

sevbch:
To fix the CI, run this:
- `pre-commit install`
- `pre-commit run --all-files`

PR #1123, opened 2025-08-20
chore(git_shell): bypass git dubious ownership errors

salome-voltz:
## Context

Git has a security feature that prevents executing commands in repositories owned by different users. This causes "dubious ownership" errors. This MR adds `safe.directory=*` configuration to all git commands to bypass ownership checks.

## Validation

Create a repository as user A and run ggshield scan as user B. No error should be displayed.


ggshield/utils/git_shell.py
```diff
@@ -214,7 +214,13 @@ def git(
         logger.debug("command=%s timeout=%d", command, timeout)
         result = subprocess.run(
             (
-                [_get_git_path(), "-c", "core.quotePath=false"]
+                [
+                    _get_git_path(),
+                    "-c",
+                    "core.quotePath=false",
+                    "-c",
+                    "safe.directory=*",
```
agateau-gg, line 222: Would it work to use `.` instead of `*` or does `safe.directory` have to be the root of the working tree? It would feel safer to use `.`.

salome-voltz: The `.` is not working, neither is `cwd`, so I think it has to be the root of the working tree

agateau-gg: OK, let's keep that way then.

I just realized however that this changes makes it possible to simplify our docker config quite a bit: we don't need [docker/entrypoint.sh](https://github.com/GitGuardian/ggshield/blob/main/docker/entrypoint.sh) to create a custom config file with `safe.directory` in it anymore, which means I think we  don't need entrypoint.sh anymore at all!

Can you try removing it? Removing it as the additional advantage that the "dubious permission" error is checked for Docker scans (by tests/functional/test_docker_image.py), so this would ensure your changes are tested.

docker/actions-secret-entrypoint.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -euo pipefail
 args=("$@")
-exec /app/docker/entrypoint.sh ggshield secret scan -v ${args[@]} ci
+ggshield secret scan -v ${args[@]} ci
```
agateau-gg: You can keep the `exec` here, it's a small optimization that makes `ggshield` replace the bash interpreter.

changelog.d/20250820_163548_salome.voltz_scrt_5812_investigate_ggshield_running_into_an_ownership_issue_in_git.md
```diff
@@ -0,0 +1,42 @@
...
+<!--
+### Added
+
+- A bullet item for the Added category.
+
+-->
+
+### Changed
+
+- Bypass git dubious ownership errors by default.
```
agateau-gg: I think this can be considered as a bug fix.

agateau-gg (changes requested):

docker/actions-secret-entrypoint.sh
salome-voltz: Oh, learned something today

PR #1125, opened 2025-08-26
Add curl and jq to Docker image

phililippe:
## Context

With the new Bring Your Own Source feature in GIM, we want to propose a simple way to scan CI logs.
But in order to allow this kind of jobs, we must interact with the CI platform to fetch the logs and analyse them.

This PR adds minimal tooling to the ggshield image (curl, jq) to allow  listing, fetching CI logs.
 

## What has been done

Added curl and jq packages to the ggshield image.

## Validation

See the associated GIM MR.

## PR check list

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ WORKDIR /app
 RUN \
     apt-get update \
     && apt-get dist-upgrade -y --no-install-recommends \
-    && apt-get install -y --no-install-recommends git openssh-client \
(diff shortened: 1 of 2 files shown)


mathieubellon:
As seen with @phililippe the image shoudl stay as lean as possible- Closing this PR

mathieubellon:
As seen with @phililippe we prefer the image to stay as lean as possible

PR #1132, opened 2025-10-16
feat: add --insecure option and deprecate --allow-self-signed

guischguardian:
## Context

We received a report about ggshield's `--allow-self-signed` option. The option name is misleading: it suggests that self-signed certificates are validated, when in reality it completely disables SSL verification, making connections vulnerable to man-in-the-middle attacks.

This PR addresses the issue by:
1. Adding a clearer `--insecure` option (matching curl's convention)
2. Deprecating `--allow-self-signed` with prominent warnings
3. Adding security warnings explaining the risks and secure alternatives

See SCRT-5952.

### New Features
- Added `--insecure` CLI option and `insecure` configuration setting as clearer alternatives to `--allow-self-signed`
- Added prominent warning messages when SSL verification is disabled (via either option), explaining:
  - The security risks (MITM attacks, traffic interception)
  - The recommended secure alternative (system certificate trust store with Python >= 3.10)
  - Link to documentation for more information
- Help text for `--insecure` matches curl's standard wording for consistency

### Deprecation
- The `--allow-self-signed` CLI option is now deprecated with a warning displayed on every use
- The `allow_self_signed` config setting is now deprecated with a warning displayed when loading config files
- Both options remain fully functional for backward compatibility
- Deprecation warnings guide users to use `--insecure` instead

### Implementation Details
- Both `--insecure` and `--allow-self-signed` set both internal flags for consistency
- Warnings use `ui.display_warning()` for proper user-facing output (not `logger.warning()`)
- Comprehensive test coverage for both old and new options

## Validation

To validate the new `--insecure` option:
```bash
# Test the new --insecure flag works
ggshield --insecure secret scan path .

# Verify deprecation warning is shown for --allow-self-signed
ggshield --allow-self-signed secret scan path .
```

Expected behavior:
- Both commands disable SSL verification
- `--allow-self-signed` displays a deprecation warning guiding to `--insecure`
- Both display security warnings about MITM risks and secure alternatives

To validate config file deprecation:
```bash
# Create a config with allow_self_signed
echo "version: 2" > .gitguardian.yaml
echo "allow_self_signed: true" >> .gitguardian.yaml

# Run any ggshield command
ggshield secret scan path .
```

Expected: Deprecation warning about `allow_self_signed` config key

## PR check list

### DIFF
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -93,7 +93,7 @@ def create_session(allow_self_signed: bool = False) -> Session:
         ui.display_warning(
             "To securely use self-signed certificates with Python >= 3.10, disable this option and "
             "install your certificate in your system's trust store. "
-            "See: https://docs.gitguardian.com/ggshield-docs/configuration#allow_self_signed"
+            "See: https://docs.gitguardian.com/ggshield-docs/configuration#support-for-self-signed-certificates"
         )
         urllib3.disable_warnings()
         session.verify = False


ggshield/cmd/utils/common_options.py
```diff
@@ -136,12 +136,32 @@ def log_file_callback(
...
 
 
+def allow_self_signed_callback(
+    ctx: click.Context, param: click.Parameter, value: Optional[bool]
+) -> Optional[bool]:
+    if value:
+        ui.display_warning(
+            "The --allow-self-signed option is deprecated and will be removed in a future version. "
+            "Use --insecure instead, which more accurately describes what this option does "
+            "(it disables all certificate validation, not just allows self-signed certificates)."
```
agateau-gg: This is quite verbose. It produces a wall of text when used:

```
$ ggshield --allow-self-signed api-status
Warning: The --allow-self-signed option is deprecated and will be removed in a future version. Use --insecure instead, which more accurately describes what this option does (it disables all certificate validation, not just allows 
self-signed certificates).
Warning: SSL verification is disabled. Your connection to the GitGuardian API is NOT encrypted and is vulnerable to man-in-the-middle attacks. Traffic, including API keys and scan results, can be intercepted and modified.
Warning: To securely use self-signed certificates with Python >= 3.10, disable this option and install your certificate in your system's trust store. See: https://docs.gitguardian.com/ggshield-docs/configuration#allow_self_signed
API URL: https://dashboard.gitguardian.com
Status: healthy
App version: v2.315.0
Secrets engine version: 2.149.0

Instance source: default
API key source: user config
```

I'd rather have a shorter version, especially since this is followed by the --insecure warning.

```suggestion
            "Use --insecure instead."
```

ggshield/cmd/utils/common_options.py
```diff
@@ -136,12 +136,32 @@ def log_file_callback(
...
+    return create_config_callback("allow_self_signed")(ctx, param, value)
+
+
 _allow_self_signed_option = click.option(
     "--allow-self-signed",
     is_flag=True,
     default=None,
-    help="Ignore ssl verification.",
-    callback=create_config_callback("allow_self_signed"),
+    help="(Deprecated: use --insecure) Ignore ssl verification.",
```
agateau-gg:
```suggestion
    help="Deprecated: use --insecure.",
```

ggshield/cmd/utils/common_options.py
```diff
@@ -136,12 +136,32 @@ def log_file_callback(
...
+    callback=allow_self_signed_callback,
+)
+
+_insecure_option = click.option(
+    "--insecure",
+    is_flag=True,
+    default=None,
+    help="WARNING: using this option makes the transfer insecure, by skipping all certificate's validation checks.",
```
agateau-gg: We should state what the option does *before* warning the user about its danger:

```suggestion
    help="Skip all certificate verification checks. WARNING: this option makes the transfer insecure.",
```

ggshield/core/config/user_config.py
```diff
@@ -88,6 +88,7 @@ class UserConfig(FilteredConfig):
     exit_zero: bool = False
     verbose: bool = False
     allow_self_signed: bool = False
+    insecure: bool = False
```
agateau-gg, line 90: I'd rather have `insecure` replace `allow_self_signed` here to keep the code simple and make inconsistent state impossible. See my other comment below for details.

ggshield/core/config/user_config.py
```diff
@@ -226,6 +228,18 @@ def _fix_ignore_known_secrets(data: Dict[str, Any]) -> None:
...
+def _check_allow_self_signed_deprecation(
+    data: Dict[str, Any], config_path: Path, deprecation_messages: List[str]
+) -> None:
+    """Check if allow_self_signed is used and add a deprecation warning."""
+    if data.get("allow_self_signed"):
+        deprecation_messages.append(
+            f"{config_path}: The 'allow_self_signed' option is deprecated and will be removed in a future version. "
+            "Use 'insecure: true' instead, which more accurately describes what this option does "
+            "(it disables all certificate validation, not just allows self-signed certificates)."
+        )
```
agateau-gg: As stated in my other comment, I'd rather get rid of `allow_self_signed` in UserConfig. This function can do the conversion:

```suggestion
    if insecure := data.pop("allow_self_signed", None):
        ui.display_warning(
            f"{config_path}: The 'allow_self_signed' option is deprecated and will be removed in a future version. "
            "Use 'insecure: true' instead, which more accurately describes what this option does "
            "(it disables all certificate validation, not just allows self-signed certificates)."
        )
        data["insecure"] = insecure
```

You should also rename the function `_fix_allow_self_signed()` for consistency.

Note: I made the function use `ui.display_warning()` instead of `deprecation_messages` because `deprecation_messages` was only useful to avoid showing them when running config migrations.

ggshield/cmd/utils/common_options.py
guischguardian: Solved in last commit

ggshield/cmd/utils/common_options.py
guischguardian: Solved in last commit

ggshield/core/config/user_config.py
guischguardian: Solved in last commit

ggshield/core/config/user_config.py
guischguardian: Solved in last commit

ggshield/__main__.py
```diff
@@ -97,11 +97,11 @@ def cli(
...
 
-    # Update allow_self_signed in the config
+    # Update SSL verification settings in the config
     # TODO: this should be reworked: if a command which writes the config is called with
-    # --allow-self-signed, the config will contain `allow_self_signed: true`.
-    if allow_self_signed:
-        user_config.allow_self_signed = allow_self_signed
+    # --insecure, the config will contain `insecure: true`.
+    if insecure:
+        user_config.insecure = insecure
```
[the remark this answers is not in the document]
agateau-gg: The bot is right: this breaks when `--allow-self-signed` is first:

- `ggshield api-status --allow-self-signed` disables SSL verifications
- `ggshield --allow-self-signed api-status` does not disable SSL verifications

This is because in the 2nd case the callback in `common_options.py` can't update the config. The following patch fixes this:

```
diff --git a/ggshield/__main__.py b/ggshield/__main__.py
index 7dbd983c..32a2bdcf 100644
--- a/ggshield/__main__.py
+++ b/ggshield/__main__.py
@@ -81,6 +81,7 @@ def exit_code(ctx: click.Context, exit_code: int, **kwargs: Any) -> int:
 def cli(
     ctx: click.Context,
     *,
+    allow_self_signed: Optional[bool],
     insecure: Optional[bool],
     config_path: Optional[Path],
     **kwargs: Any,
@@ -100,8 +101,8 @@ def cli(
     # Update SSL verification settings in the config
     # TODO: this should be reworked: if a command which writes the config is called with
     # --insecure, the config will contain `insecure: true`.
-    if insecure:
-        user_config.insecure = insecure
+    if insecure or allow_self_signed:
+        user_config.insecure = True
 
     ctx_obj.config._dotenv_vars = load_dot_env()
```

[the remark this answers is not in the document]
guischguardian: Fixed in last Commit

agateau-gg (approved):

PR #1133, opened 2025-10-20
fix(secret): skip non-seekable files during scanning

salome-voltz:
## Context

Ggshield crashes on scanning non-seekable files, to reproduce (only on linux): 
```
> cd test_folder/
> ln -s /proc/self/mounts ./mtab
> ls -l mtab #verify if the symbolic link is correctly set up
> ggshield secret scan path ./mtab
```
Should return: 
```
Scanning... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% 0 / 1

Error: [Errno 22] Invalid argument
```

Explicitly skip them with a message: 
image

### DIFF
--- a/ggshield/core/scan/scannable.py
+++ b/ggshield/core/scan/scannable.py
@@ -31,6 +31,12 @@ class DecodeError(Exception):
     pass
 
 
+class NonSeekableFileError(Exception):
+    """Raised when a file cannot be seeked"""
+
+    pass
+
+
 class Scannable(ABC):
     """Base class for content that can be scanned by GGShield"""
 
@@ -143,8 +149,12 @@ class Scannable(ABC):
         Raises DecodeError if the file cannot be decoded.
         """
         # Get the byte size
-        assert fp.seekable()
-        byte_size = fp.seek(0, SEEK_END)
+        # Note: IOBase.seekable() returns True on some non-seekable files like /proc/self/mounts
+        try:
+            byte_size = fp.seek(0, SEEK_END)
+            fp.seek(0, SEEK_SET)
+        except OSError as exc:
+            raise NonSeekableFileError() from exc
 
         if byte_size > max_utf8_encoded_size * UTF8_TO_WORSE_OTHER_ENCODING_RATIO:
             # Even if the file used the worst encoding (UTF-32), encoding the content of
@@ -153,7 +163,6 @@ class Scannable(ABC):
             return True, None, None
 
         # Determine the encoding
-        fp.seek(0, SEEK_SET)
         charset_matches = charset_normalizer.from_fp(fp)
         charset_match = charset_matches.best()
         if charset_match is None:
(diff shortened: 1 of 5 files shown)


ggshield/core/scan/scannable.py
```diff
@@ -153,7 +162,10 @@ def _is_file_longer_than(
             return True, None, None
 
         # Determine the encoding
-        fp.seek(0, SEEK_SET)
+        try:
+            fp.seek(0, SEEK_SET)
+        except OSError:
+            raise NonSeekableFileError("File does not support seeking operations")
```
agateau-gg: The goal of this call to `seek()` is to rewind after the first `seek()`, which moved to the end of the file.

Instead of wrapping this one separately, we can move it right after the first seek, that would make the code simpler. 

Something like this:

```python
        try:
            byte_size = fp.seek(0, SEEK_END)
            fp.seek(0, SEEK_CUR)
        except OSError:
            raise NonSeekableFileError(...)
```

ggshield/core/scan/scannable.py
```diff
@@ -144,7 +150,10 @@ def _is_file_longer_than(
         """
         # Get the byte size
         assert fp.seekable()
```
agateau-gg: That assert should have failed 😞. Looks like we can remove it and add a comment saying `IOBase.seekable()` returns True on some non-seekable files like `/proc/self/mounts`.

ggshield/core/scan/scannable.py
```diff
@@ -144,7 +150,10 @@ def _is_file_longer_than(
...
-        byte_size = fp.seek(0, SEEK_END)
+        try:
+            byte_size = fp.seek(0, SEEK_END)
+        except OSError:
+            raise NonSeekableFileError("File does not support seeking operations")
```
agateau-gg: I think you can remove the error message: it's never shown.

ggshield/core/scan/file.py
```diff
@@ -30,13 +30,16 @@ def is_longer_than(self, max_utf8_encoded_size: int) -> bool:
...
+        try:
+            with self.path.open("rb") as fp:
+                (
+                    result,
+                    self._content,
+                    self._utf8_encoded_size,
+                ) = Scannable._is_file_longer_than(fp, max_utf8_encoded_size)
+            return result
+        except NonSeekableFileError:
+            raise
```
agateau-gg: I think we don't need this try-except block.

agateau-gg (changes requested):

ggshield/core/scan/scannable.py
salome-voltz: I'm not sure I understand your suggestion. Using `fp.seek(0, SEEK_CUR)`, while the current position is the end of the file, wouldn't rewind here. 
I have pushed a solution with `fp.seek(0, SEEK_SET)`, let me know I missed something.

agateau-gg: Oups, you are correct, should have written `SEEK_SET` indeed.

PR #1137, opened 2025-10-28
chore: merge 1.44.x hotfix changes

6d7a:
Hotfix changes to fix CI incompatibility between pdm and Python 3.14.


.github/workflows/build_release_assets.yml
```diff
@@ -18,7 +18,8 @@ on:
         default: false
 
 env:
-  PDM_VERSION: 2.22.4
+  PDM_VERSION: 2.26.0
+  DEFAULT_PYTHON_VERSION: '3.12'
```
agateau-gg: This is not the default Python version we are going to use to build our assets, it's the version used by pdm. As such I think it should be called `PDM_PYTHON_VERSION` to avoid confusion.

6d7a: I decided to use `DEFAULT_PYTHON_VERSION` mainly for consistency, because we use the variable in our other workflows (see [here](https://github.com/GitGuardian/ggshield/blob/3616c444dbc3903f5523b87ae82666c3e2ee8a21/.github/workflows/ci.yml#L22) for example). If that should be changed, I think we should open a separate PR off main instead of changing it here.

agateau-gg: Oh right, I just had a look at doc for setup-pdm and it is indeed the version we use for jobs unless specified, so the name makes sense. But I think we should try to keep those values in sync between the different workflows.

ggshield/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "1.44.0"
```
sevbch, line 1: I'm wondering about this and the changes in `actions/secret/action.yml`. 

I think those changes are supposed to happen automatically through the CI. See [here](https://github.com/GitGuardian/ggshield/blob/b2160673e209827a3597f9a2d31dd460f1d051db/scripts/release#L36)

This, in addition to the branch protection rules I saw make me think that you should open a MR toward `1.44.x` instead of merging `1.44.x` into `main`.

Tagging @agateau-gg here as I don't see where `script/release` is actually used in the CI. This is very strange.

sevbch (changes requested):
I'm getting confused about the hotfix release process. I don't see how the tag and release assets will be built 🤔 

Did you find out how it's supposed to be triggered?

sevbch (approved):

PR #1141, opened 2025-10-29
chore: add HTTPAdapter to request session

sevbch:
## Context

A customer reported this issue when scanning with ggshield, on a self-hosted instance (`XXX`= host url).

```bash
59089:605129 [E] ggshield.core.errors:209 status_code=None detail=HTTPSConnectionPool(host='XXX', port=443): Read timed out. (read
timeout=60)
2025-10-28 11:13:04 59089:605240 [W] urllib3.connectionpool:329 Connection pool is full, discarding connection: XXX. Connection pool size: 10
2025-10-28 11:13:23 59089:605238 [W] urllib3.connectionpool:329 Connection pool is full, discarding connection: XXX. Connection pool size: 10
Scanning... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╺━━ 93% 529 / 569
Error: Scanning failed: HTTPSConnectionPool(host='XXX', port=443): Read timed out. (read timeout=60)
```

## What has been done
Adding a `HTTPAdapter` to the session configuration to better handle many parallel requests. The risk is that it could put additional pressure on the server. After reading [this article](https://stackoverflow.com/questions/34837026/whats-the-meaning-of-pool-connections-in-requests-adapters-httpadapter), I left `pool_connections` to default value, as it shouldn't have much of an impact for our case, but increased `pool_maxsize` as we can see in the logs that this is the param that was blocking for the customer.

Another option was to increase the 60s timeout, but I'm afraid it would have too much impact. For instance, when a ggshield command fail, it could result in a longer waiting time for the caller to fetch the output.

I'm open to suggestions if you see another way of handling this.

## Validation
Validated with the unit test.

## PR check list

### DIFF
--- a/tests/unit/core/test_client.py
+++ b/tests/unit/core/test_client.py
@@ -9,7 +9,11 @@ from pyfakefs.fake_filesystem import FakeFilesystem
 from pygitguardian import GGClient
 from pygitguardian.models import APITokensResponse, Detail, TokenScope
 
-from ggshield.core.client import check_client_api_key, create_client_from_config
+from ggshield.core.client import (
+    check_client_api_key,
+    create_client_from_config,
+    create_session,
+)
 from ggshield.core.config import Config
 from ggshield.core.errors import (
     APIKeyCheckError,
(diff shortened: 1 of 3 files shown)


tests/unit/core/test_client.py
```diff
@@ -239,3 +243,64 @@ def test_retrieve_client_unknown_custom_dashboard_url(isolated_fs: FakeFilesyste
...
+
+    # Verify adapters are mounted
+    assert "http://" in session.adapters
+    assert "https://" in session.adapters
+
+    # Verify it's an HTTPAdapter (not the default)
+    from requests.adapters import HTTPAdapter
+
+    assert isinstance(session.get_adapter("http://example.com"), HTTPAdapter)
+    assert isinstance(session.get_adapter("https://example.com"), HTTPAdapter)
```
agateau-gg: This test does not test much: Session.adapters contains HTTPAdapter instances by default, for both http and https:

```
>>> from requests import Session
>>> s = Session()
>>> s.adapters
OrderedDict({'https://': <requests.adapters.HTTPAdapter object at 0x78f4aca515b0>, 'http://': <requests.adapters.HTTPAdapter object at 0x78f4ab918e30>})
```

I think it can be dropped.

sevbch: I only saw your review now, I removed the test.

PR #1166, opened 2026-01-08
feat(cache): automatically add .cache_ggshield to .gitignore

6d7a:
## Context

`.cache_ggshield` generated files are finding their ways to remote repositories after developers accidentally commit them.

## What has been done

This PR ~~adds a warning that will be displayed~~ has `ggshield` add `.cache_ggshield` automatically to `.gitignore` if `.cache_ggshield` is not ignored in a git repository (and `git` is available). If `.gitignore` does not exist, `ggshield` will try to create it.

## PR check list

### DIFF
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -489,16 +489,39 @@ def get_default_branch(wd: Optional[Union[str, Path]] = None) -> str:
     return default_branch
 
 
-def is_gitignored(path: Path, wd: Optional[Union[str, Path]] = None) -> bool | None:
+def is_gitignored(path: Path) -> Optional[bool]:
     """
     Returns True if the path is ignored by git, False if it is not,
         or None if git is not available or the directory is not a git repository.
     :param path: path to the file or directory to check
     """
-    if wd is None:
-        wd = Path.cwd()
-    if not is_git_available() or not is_git_dir(wd):
+    working_dir = Path.cwd()
+    if not is_git_available() or not is_git_dir(working_dir):
         return None
     else:
-        res = git(["check-ignore", "--", path.as_posix()], cwd=wd, check=False)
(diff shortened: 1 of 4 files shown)


ggshield/utils/git_shell.py
```diff
@@ -487,3 +487,18 @@ def get_default_branch(wd: Optional[Union[str, Path]] = None) -> str:
...
     return default_branch
+
+
+def is_gitignored(path: Path, wd: Optional[Union[str, Path]] = None) -> bool | None:
+    """
+    Returns True if the path is ignored by git, False if it is not,
+        or None if git is not available or the directory is not a git repository.
+    :param path: path to the file or directory to check
+    """
+    if wd is None:
```
sevbch: Nit: not a fan of abbreviations, I prefer word variable names
```suggestion
    if working_directory is None:
```

sevbch: Is `wd` really relevant? It's not used currently, so it's dead code so far

6d7a: Updated it 👍

6d7a: It is needed for testing.

sevbch: can't you find a solution mocking the working directory in the test? we shouldn't have to modify a function for tests

sevbch: I don't think this comment has been addressed. You should be able to mock the working directory instead of adding a function parameter just for testing purpose

6d7a: Sorry, I overlooked your comment. I updated the code accordingly.

agateau-gg: I think the previous version was better because using `cd()` in tests is a bad idea: it changes the working directory of the whole process and that goes in the way of running tests in parallel (a work-in-progress project of mine that has been stalled for too long 😞).

Adding a way to override the default path to use does not sound that bad to me: it's a form of dependency injection.

PR #1167, opened 2026-01-16
feat: add enterprise plugin system for ggshield

clement-tourriere:
## Context

This PR introduces an enterprise plugin system for ggshield, allowing organizations to extend ggshield with additional capabilities through plugins distributed by GitGuardian.

## What has been done

New Plugin Commands

  Added a new ggshield plugin command group with the following subcommands:

  - ggshield plugin status - Show available plugins for your GitGuardian account based on your plan and entitlements
  - ggshield plugin install <name> - Download and install a plugin from GitGuardian
  - ggshield plugin list - List installed plugins with their version and enabled/disabled status
  - ggshield plugin update [<name>|--all|--check] - Update installed plugins to the latest version
  - ggshield plugin enable <name> - Enable an installed plugin
  - ggshield plugin disable <name> - Disable a plugin without uninstalling
  - ggshield plugin uninstall <name> - Remove an installed plugin

  Core Plugin Infrastructure

  - Entitlements client: Fetches available plugins from GitGuardian API based on the user's account/plan
  - Plugin downloader: Securely downloads plugin wheels with SHA256 verification
  - Plugin loader: Discovers and loads plugins from local wheels or pip-installed entry points
  - Enterprise config: Persists plugin settings (enabled/disabled state, versions)
  - Plugin registry: Central registry for loaded plugins to register their capabilities

## PR check list

### DIFF
--- a/ggshield/cmd/plugin/install.py
+++ b/ggshield/cmd/plugin/install.py
@@ -29,10 +29,6 @@ from ggshield.core.plugin.downloader import (
 )
 
 
-# Backward-compatible alias for existing tests/imports.
-SourceType = PluginSourceType
-
-
 def detect_source_type(plugin_source: str) -> PluginSourceType:
     """
     Detect the type of plugin source from the argument.
--- a/tests/unit/cmd/plugin/test_install.py
+++ b/tests/unit/cmd/plugin/test_install.py
@@ -6,9 +6,14 @@ from pathlib import Path
 from unittest import mock
 
 from ggshield.__main__ import cli
-from ggshield.cmd.plugin.install import SourceType, detect_source_type
+from ggshield.cmd.plugin.install import detect_source_type
 from ggshield.core.errors import ExitCode
-from ggshield.core.plugin.client import PluginCatalog, PluginDownloadInfo, PluginInfo
+from ggshield.core.plugin.client import (
+    PluginCatalog,
+    PluginDownloadInfo,
+    PluginInfo,
+    PluginSourceType,
+)
 
 
 class TestPluginInstall:
@@ -541,38 +546,43 @@ class TestDetectSourceType:
     def test_detect_github_artifact(self) -> None:
         """Test detection of GitHub artifact URLs."""
         url = "https://github.com/owner/repo/actions/runs/123456/artifacts/789"
-        assert detect_source_type(url) == SourceType.GITHUB_ARTIFACT
+        assert detect_source_type(url) == PluginSourceType.GITHUB_ARTIFACT
 
     def test_detect_github_release(self) -> None:
         """Test detection of GitHub release URLs."""
         url = "https://github.com/owner/repo/releases/download/v1.0.0/plugin.whl"
-        assert detect_source_type(url) == SourceType.GITHUB_RELEASE
+        assert detect_source_type(url) == PluginSourceType.GITHUB_RELEASE
 
     def test_detect_generic_url(self) -> None:
         """Test detection of generic HTTPS URLs."""
-        assert detect_source_type("https://example.com/plugin.whl") == SourceType.URL
         assert (
-            detect_source_type("https://pypi.org/packages/plugin.whl") == SourceType.URL
+            detect_source_type("https://example.com/plugin.whl") == PluginSourceType.URL
+        )
+        assert (
+            detect_source_type("https://pypi.org/packages/plugin.whl")
+            == PluginSourceType.URL
         )
 
     def test_detect_http_url(self) -> None:
         """Test detection of HTTP URLs (will be rejected later)."""
-        assert detect_source_type("http://example.com/plugin.whl") == SourceType.URL
+        assert (
+            detect_source_type("http://example.com/plugin.whl") == PluginSourceType.URL
+        )
 
     def test_detect_local_file(self, tmp_path: Path) -> None:
         """Test detection of local wheel files."""
         wheel_path = tmp_path / "plugin.whl"
         wheel_path.touch()
-        assert detect_source_type(str(wheel_path)) == SourceType.LOCAL_FILE
+        assert detect_source_type(str(wheel_path)) == PluginSourceType.LOCAL_FILE
 
     def test_detect_local_file_nonexistent(self) -> None:
         """Test non-existent local wheel paths are treated as local sources."""
-        assert detect_source_type("./nonexistent.whl") == SourceType.LOCAL_FILE
+        assert detect_source_type("./nonexistent.whl") == PluginSourceType.LOCAL_FILE
 
     def test_detect_plugin_name(self) -> None:
         """Test plugin names default to GitGuardian API."""
-        assert detect_source_type("tokenscanner") == SourceType.GITGUARDIAN_API
-        assert detect_source_type("my-plugin") == SourceType.GITGUARDIAN_API
+        assert detect_source_type("tokenscanner") == PluginSourceType.GITGUARDIAN_API
+        assert detect_source_type("my-plugin") == PluginSourceType.GITGUARDIAN_API
 
 
 class TestInstallFromLocalWheel:
@@ -596,7 +606,7 @@ class TestInstallFromLocalWheel:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.LOCAL_FILE,
+                return_value=PluginSourceType.LOCAL_FILE,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -639,7 +649,7 @@ class TestInstallFromLocalWheel:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.LOCAL_FILE,
+                return_value=PluginSourceType.LOCAL_FILE,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -681,7 +691,7 @@ class TestInstallFromUrl:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.URL,
+                return_value=PluginSourceType.URL,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -729,7 +739,7 @@ class TestInstallFromUrl:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.URL,
+                return_value=PluginSourceType.URL,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -769,7 +779,7 @@ class TestInstallFromUrl:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.URL,
+                return_value=PluginSourceType.URL,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -808,7 +818,7 @@ class TestInstallFromGitHubRelease:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.GITHUB_RELEASE,
+                return_value=PluginSourceType.GITHUB_RELEASE,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -855,7 +865,7 @@ class TestInstallFromGitHubArtifact:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.GITHUB_ARTIFACT,
+                return_value=PluginSourceType.GITHUB_ARTIFACT,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -898,7 +908,7 @@ class TestInstallFromGitHubArtifact:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.GITHUB_ARTIFACT,
+                return_value=PluginSourceType.GITHUB_ARTIFACT,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -942,7 +952,7 @@ class TestInstallFromGitHubArtifact:
             ) as mock_config_class,
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.GITHUB_ARTIFACT,
+                return_value=PluginSourceType.GITHUB_ARTIFACT,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -981,7 +991,7 @@ class TestInstallErrorHandling:
 
         with mock.patch(
             "ggshield.cmd.plugin.install.detect_source_type",
-            return_value=SourceType.LOCAL_FILE,
+            return_value=PluginSourceType.LOCAL_FILE,
         ):
             result = cli_fs_runner.invoke(
                 cli,
@@ -1011,7 +1021,7 @@ class TestInstallErrorHandling:
             mock.patch("ggshield.cmd.plugin.install.EnterpriseConfig"),
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.LOCAL_FILE,
+                return_value=PluginSourceType.LOCAL_FILE,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -1046,7 +1056,7 @@ class TestInstallErrorHandling:
             mock.patch("ggshield.cmd.plugin.install.EnterpriseConfig"),
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.LOCAL_FILE,
+                return_value=PluginSourceType.LOCAL_FILE,
             ),
         ):
             mock_downloader = mock.MagicMock()
@@ -1076,7 +1086,7 @@ class TestInstallErrorHandling:
             mock.patch("ggshield.cmd.plugin.install.EnterpriseConfig"),
             mock.patch(
                 "ggshield.cmd.plugin.install.detect_source_type",
-                return_value=SourceType.URL,
+                return_value=PluginSourceType.URL,
             ),
         ):
             mock_downloader = mock.MagicMock()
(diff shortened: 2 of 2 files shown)


ggshield/cmd/plugin/install.py
```diff
@@ -0,0 +1,160 @@
...
+    "version",
+    default=None,
+    help="Specific version to install (defaults to latest)",
+)
+@add_common_options()
+@click.pass_context
+def install_cmd(
+    ctx: click.Context,
+    plugin_name: Optional[str],
+    install_all: bool,
```
agateau-gg: Are you sure we want an option to install *all* plugins? That is surprising.

ggshield/cmd/plugin/install.py
```diff
@@ -0,0 +1,160 @@
...
+    enterprise_config.save()
+
+    # Summary
+    if success_count > 0:
+        ui.display_info("")
+        if success_count == 1:
+            ui.display_info("Plugin installed successfully.")
+        else:
+            ui.display_info(f"{success_count} plugins installed successfully.")
+        ui.display_info("Restart ggshield to use the new plugins.")
```
agateau-gg: ???

ggshield/cmd/plugin/install.py
```diff
@@ -0,0 +1,160 @@
...
+    # Save config
+    enterprise_config.save()
+
+    # Summary
+    if success_count > 0:
+        ui.display_info("")
+        if success_count == 1:
+            ui.display_info("Plugin installed successfully.")
+        else:
+            ui.display_info(f"{success_count} plugins installed successfully.")
```
agateau-gg: You can use `pluralize()` here.

ggshield/core/plugin/loader.py
```diff
@@ -0,0 +1,313 @@
...
+                wheel_path.stat().st_mtime > extract_dir.stat().st_mtime
+            ):
+                import shutil
+
+                if extract_dir.exists():
+                    shutil.rmtree(extract_dir)
+                extract_dir.mkdir(parents=True)
+
+                with ZipFile(wheel_path, "r") as zf:
+                    zf.extractall(extract_dir)
```
agateau-gg: ZipFile.extractall() does not sanitize paths. A malicious wheel or artifact ZIP could contain entries like `../../etc/cron.d/evil` that write outside the intended directory.

Python only added built-in protection against this in 3.12. Since ggshield must support older versions of Python, use `safe_unpack()` from `utils.archive` instead.

ggshield/cmd/plugin/update.py
```diff
@@ -0,0 +1,359 @@
...
+    Args:
+        release: GitHub release data.
+
+    Returns:
+        Download URL for the wheel or None.
+    """
+    assets = release.get("assets", [])
+    for asset in assets:
+        name = asset.get("name", "")
+        if name.endswith(".whl"):
```
agateau-gg: This picks the first .whl file from a GitHub release's assets. If a release publishes multiple wheels for different platforms (e.g., manylinux_x86_64, macosx_arm64), it may pick an incompatible one.

ggshield/core/config/enterprise_config.py
```diff
@@ -0,0 +1,114 @@
...
+        if plugin_name not in self.plugins:
+            self.plugins[plugin_name] = PluginConfig(enabled=False)
+        else:
+            self.plugins[plugin_name].enabled = False
+
+    def is_plugin_enabled(self, plugin_name: str) -> bool:
+        """Check if a plugin is enabled."""
+        plugin_config = self.plugins.get(plugin_name)
+        # Default: enabled if not explicitly configured
+        return plugin_config.enabled if plugin_config else True
```
agateau-gg: This returns True for plugins not in the config. This means any pip package that registers a ggshield.plugins entry point is automatically loaded and executed at ggshield startup without user consent. This looks dangerous.

ggshield/__main__.py
```diff
@@ -106,9 +143,31 @@ def cli(
...
+    """Register plugin commands with the CLI."""
+    try:
+        registry = _load_plugins()
+        for cmd in registry.get_commands():
+            cli.add_command(cmd)
+    except Exception as e:
+        logger.debug("Failed to register plugin commands: %s", e)
+
+
+_register_plugin_commands()
```
agateau-gg: Can we do this in `main()` instead? That would help keeping import fast (useful for unit-tests).

ggshield/__main__.py
```diff
@@ -106,9 +143,31 @@ def cli(
...
+# Register plugin commands at module load time
+# This must happen after the cli group is defined
+def _register_plugin_commands() -> None:
+    """Register plugin commands with the CLI."""
+    try:
+        registry = _load_plugins()
+        for cmd in registry.get_commands():
+            cli.add_command(cmd)
+    except Exception as e:
+        logger.debug("Failed to register plugin commands: %s", e)
```
agateau-gg: I think this should be a `logger.warning()`, but ideally this should be stored and sent to the logger later because the logger is not configured at this point (even with the change I suggest below).

ggshield/cmd/plugin/update.py
```diff
@@ -0,0 +1,359 @@
...
+                if not installed or not available:
+                    continue
+
+                installed_version = installed.version
+                latest_version = available.latest_version
+
+                if (
+                    installed_version
+                    and latest_version
+                    and installed_version != latest_version
```
agateau-gg: Using `!=` here means the existing version could be replaced by a lower version. You could extract some of the code to split versions from `core/check_updates.py` and use it here.

ggshield/core/plugin/downloader.py
```diff
@@ -0,0 +1,650 @@
...
+                source=source,
+            )
+
+            logger.info("Installed %s v%s", plugin_name, download_info.version)
+
+            return wheel_path
+
+        except requests.RequestException as e:
+            if temp_path.exists():
+                temp_path.unlink()
```
agateau-gg, line 118: This should be in a `finally` clause.

doc/dev/enterprise-plugin-system.md
```diff
@@ -0,0 +1,155 @@
...
+├── cmd/
+│   └── plugin/
+│       ├── __init__.py      # plugin_group
+│       ├── install.py       # install command
+│       ├── plugin_list.py   # list command
+│       ├── status.py        # status command
+│       ├── manage.py        # enable/disable/uninstall
+│       └── update.py        # update command
+└── __main__.py              # Plugin loading at startup
+```
```
agateau-gg: Experience tells me the probability of this becoming outdated is close to 100%. Better remove it.

doc/dev/enterprise-plugin-system.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Architecture
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│                        ggshield CLI                          │
+├─────────────────────────────────────────────────────────────┤
+│  ┌──────────────┐    ┌──────────────┐    ┌──────────────┐   │
+│  │ Command Layer│◄───│ Plugin System│◄───│ Plugin Cmds  │   │
+│  │ (cmd/)       │    │ (core/plugin)│    │ (cmd/plugin) │   │
```
agateau-gg: What does the arrow mean? and why is there a lower-level layer part (anything in core) sandwiched between two higher-level layer parts (cmd)?

doc/dev/enterprise-plugin-system.md
```diff
@@ -0,0 +1,155 @@
...
+class MyPlugin(GGShieldPlugin):
+    @property
+    def metadata(self) -> PluginMetadata:
+        return PluginMetadata(
+            name="myplugin",
+            version="1.0.0",
+            display_name="My Plugin",
+            description="Does something useful",
+            min_ggshield_version="1.45.0",
+            capabilities=[PluginCapability.COMMAND],
```
agateau-gg: This document should explain what capabilities are and which are available.

Update: after checking it seems there is only one capability, so maybe just drop it for now?

doc/dev/enterprise-plugin-system.md
```diff
@@ -0,0 +1,155 @@
...
+Downloaded plugins are stored in:
+
+- Linux: `~/.local/share/ggshield/plugins/<plugin_name>/`
+- macOS: `~/Library/Application Support/ggshield/plugins/<plugin_name>/`
+- Windows: `%LOCALAPPDATA%/ggshield/plugins/<plugin_name>/`
+
+Each plugin directory contains:
+
+- The wheel file (`.whl`)
+- A `manifest.json` with metadata
```
agateau-gg: Should we document what's in this file?

doc/dev/enterprise-plugin-system.md
```diff
@@ -0,0 +1,155 @@
...
+ggshield plugin disable <plugin_name>
+
+# Update plugins
+ggshield plugin update --check
+ggshield plugin update <plugin_name>
+ggshield plugin update --all
+
+# Uninstall
+ggshield plugin uninstall <plugin_name>
+```
```
agateau-gg: The target audience of documents in doc/dev are developers. This is good documentation, but it's more for the user level. It should go in the user doc (that lives in our private doc repo).

ggshield/cmd/plugin/update.py
```diff
@@ -0,0 +1,382 @@
...
+
+            elif source_type == PluginSourceType.GITHUB_RELEASE:
+                # Download from GitHub release URL
+                download_url = update.get("download_url")
+                github_repo = update.get("github_repo")
+                if download_url and github_repo:
+                    downloader.download_from_github_release(download_url)
+
+            # Update config
+            enterprise_config.enable_plugin(name, version=latest_version)
```
agateau-gg, line 309: We should stop before enabling the plugin if source_type is something we did not handle, or if we skipped the download (did not enter the `if` on line 277)

ggshield/core/plugin/loader.py
```diff
@@ -0,0 +1,313 @@
...
+                    shutil.rmtree(extract_dir)
+
+                from ggshield.utils.archive import safe_unpack
+
+                safe_unpack(wheel_path, extract_dir)
+
+            # Add extracted directory to sys.path
+            extract_str = str(extract_dir)
+            if extract_str not in sys.path:
+                sys.path.insert(0, extract_str)
```
agateau-gg: Claude tells me: "Extracted directories are inserted at the front of sys.path with sys.path.insert(0, ...) and never removed. A malicious wheel could include modules shadowing stdlib names (json.py, logging.py, os.py), causing arbitrary code execution elsewhere in ggshield."

ggshield/__main__.py
```diff
@@ -106,9 +145,33 @@ def cli(
...
 
 
+# Register plugin commands with the CLI group.
+# Called from main() to avoid import-time side effects.
+def _register_plugin_commands() -> None:
+    """Register plugin commands with the CLI."""
+    try:
+        registry = _load_plugins()
+        for cmd in registry.get_commands():
+            cli.add_command(cmd)
```
agateau-gg, line 183: Claude is worried this can shadow built-in commands. I don't think we want to allow that?

ggshield/core/plugin/downloader.py
```diff
@@ -0,0 +1,650 @@
...
+        self.plugins_dir = get_plugins_dir()
+
+    def download_and_install(
+        self,
+        download_info: PluginDownloadInfo,
+        plugin_name: str,
+        source: Optional[PluginSource] = None,
+    ) -> Path:
+        """Download a plugin wheel and install it locally."""
+        plugin_dir = self.plugins_dir / plugin_name
```
agateau-gg, line 74: Claude is worried this could overwrite files outside the plugin dir if `plugin_name` is something like `../../foo`. Sounds unlikely, but maybe add a check that `plugin_name` does not contain `os.sep`?

ggshield/cmd/plugin/install.py
```diff
@@ -0,0 +1,377 @@
...
+            r"https://github\.com/[^/]+/[^/]+/releases/download/",
+            plugin_source,
+        ):
+            return SourceType.GITHUB_RELEASE
+        # Any other URL
+        return SourceType.URL
+
+    # Check if it's a local wheel file
+    path = Path(plugin_source)
+    if path.exists() and plugin_source.endswith(".whl"):
```
agateau-gg: Claude thinks if the plugin_source ends with ".whl" but the file does not exist we should report an error and not mark it as a GITGUARDIAN_API plugin.

ggshield/core/plugin/downloader.py
```diff
@@ -0,0 +1,650 @@
...
+            with open(temp_path, "wb") as f:
+                for chunk in response.iter_content(chunk_size=8192):
+                    f.write(chunk)
+                    sha256_hash.update(chunk)
+
+            computed_hash = sha256_hash.hexdigest()
+            if computed_hash.lower() != download_info.sha256.lower():
+                raise ChecksumMismatchError(download_info.sha256, computed_hash)
+
+            temp_path.rename(wheel_path)
```
agateau-gg: You should move this below the _write_manifest() call so that if writing the manifest fails the temp file is deleted.

ggshield/core/plugin/loader.py
```diff
@@ -0,0 +1,313 @@
...
+            dist = importlib.metadata.distribution(package_name)
+            return dist.version
+        except (importlib.metadata.PackageNotFoundError, IndexError):
+            return None
+
+    def _check_version_compatibility(self, metadata: PluginMetadata) -> bool:
+        """Check if plugin is compatible with current ggshield version."""
+        from packaging import version
+
+        from ggshield import __version__ as ggshield_version
```
agateau-gg: Can these be top-level imports? At least I think the packaging one can be moved (but maybe do it as `import packaging.version` to avoid adding a `version` symbol to the module scope).

Also if you are using this to parse versions, maybe use it in the other places where you need to parse versions?

ggshield/core/plugin/loader.py
```diff
@@ -0,0 +1,313 @@
...
+
+        try:
+            current = version.parse(ggshield_version)
+            required = version.parse(metadata.min_ggshield_version)
+            return current >= required
+        except Exception:
+            return True
```
agateau-gg: I think it would be safer to return False here and log an error.

ggshield/core/config/enterprise_config.py
```diff
@@ -0,0 +1,114 @@
...
+        if plugin_name not in self.plugins:
+            self.plugins[plugin_name] = PluginConfig()
+
+        self.plugins[plugin_name].enabled = True
+        if version:
+            self.plugins[plugin_name].version = version
+
+    def disable_plugin(self, plugin_name: str) -> None:
+        """Disable a plugin."""
+        if plugin_name not in self.plugins:
```
agateau-gg, line 93: Should it be an error to disable a plugin that is not present?

I can imagine someone wanting to disable the plugin, making a typo in the plugin name, not getting any error message and thinking it worked.

ggshield/core/plugin/wheel_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    return WheelMetadata(
+        name=name,
+        version=version,
+        summary=msg.get("Summary"),
+        author=msg.get("Author"),
+        license=msg.get("License"),
+    )
+
+
+def normalize_plugin_name(name: str) -> str:
```
agateau-gg: This function is never called.

ggshield/core/plugin/downloader.py
```diff
@@ -0,0 +1,650 @@
...
+    """Error downloading GitHub artifact."""
+
+    pass
+
+
+def get_plugins_dir() -> Path:
+    """Get the directory where plugin wheels are stored."""
+    plugins_dir = get_data_dir() / "plugins"
+    plugins_dir.mkdir(parents=True, exist_ok=True)
+    return plugins_dir
```
agateau-gg: This function is duplicated in loader.py (minus the mkdir() call but it can be removed because download_and_install() takes care of creating the dir). Can you move it to `core.dirs`?

ggshield/cmd/plugin/install.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+class SourceType:
+    """Plugin source types for detection."""
+
+    GITGUARDIAN_API = "gitguardian_api"
+    LOCAL_FILE = "local_file"
+    URL = "url"
+    GITHUB_RELEASE = "github_release"
+    GITHUB_ARTIFACT = "github_artifact"
```
agateau-gg: This is a duplication of PluginSourceType in plugin/client.py.

.importlinter
```diff
@@ -47,6 +48,8 @@ ignore_imports =
 	ggshield.cmd.honeytoken.** -> ggshield.verticals.honeytoken.**
 	ggshield.cmd.install.** -> ggshield.verticals.install
 	ggshield.cmd.install.** -> ggshield.verticals.install.**
+	ggshield.cmd.plugin.** -> ggshield.verticals.plugin
+	ggshield.cmd.plugin.** -> ggshield.verticals.plugin.**
```
agateau-gg: There is no such thing as `ggshield.verticals.plugin`. This should be `ggshield.core.plugin`.

agateau-gg (changes requested):

ggshield/core/plugin/downloader.py
agateau-gg: Any opinion on this one?

ggshield/core/plugin/loader.py
```diff
@@ -0,0 +1,323 @@
...
+from ggshield.core.config.enterprise_config import EnterpriseConfig
+from ggshield.core.dirs import get_plugins_dir
+from ggshield.core.plugin.base import GGShieldPlugin, PluginMetadata
+from ggshield.core.plugin.registry import PluginRegistry
+
+
+try:
+    from packaging import version as packaging_version
+except ImportError:  # pragma: no cover
+    packaging_version = None
```
agateau-gg: Is there a reason for doing that instead of adding `packaging` to the list of dependencies?

ggshield/core/plugin/downloader.py
```diff
@@ -0,0 +1,680 @@
...
+                        in_section = False
+                        for line in content.splitlines():
+                            line = line.strip()
+                            if line == f"[{PLUGIN_ENTRY_POINT_GROUP}]":
+                                in_section = True
+                            elif line.startswith("["):
+                                in_section = False
+                            elif in_section and "=" in line:
+                                ep_name, _ = line.split("=", 1)
+                                return ep_name.strip()
```
agateau-gg: This code is a duplication of `loader._parse_entry_point_name()`. Consider factorizing it, and possibly using ConfigParser instead of parsing by hand.

ggshield/cmd/plugin/install.py
```diff
@@ -0,0 +1,371 @@
...
+    ChecksumMismatchError,
+    DownloadError,
+    GitHubArtifactError,
+    InsecureSourceError,
+    PluginDownloader,
+)
+
+
+# Backward-compatible alias for existing tests/imports.
+SourceType = PluginSourceType
```
agateau-gg: There is no need to be backward-compatible for brand new code. Just use PluginSourceType everywhere.

PR #1170, opened 2026-02-03
HMSL - Add a length filter client side

pierrelalanne:
## Summary
 
Add a minimum length filter (secrets must be at least 6 characters) to exclude obvious non-secrets before sending them to the HMSL API. This improves UX by reducing false positive matches from the HMSL database.   

**Note 1:** This [commit](https://github.com/GitGuardian/ggshield/pull/1167/changes/ffeb2476e5f3a8b44ad92c1b4f5baa80e869fa85) from another PR will fix the failing test in the CI.
**Note 2:** in a perfect world, these secrets should never end up in HMSL DB in the first place. Unfortunately this happens. This PR is a client side fix to address this.   

### How to test

```
echo -e "abc\n123\nshort\sup3rs3cr3t" > /tmp/secrets.txt
# Both should only prepare or check only one secret
pdm run ggshield hmsl fingerprint /tmp/secrets.txt
pdm run ggshield hmsl check /tmp/secrets.txt
```

## Architecture Overview

After exploring the codebase, here's how secrets flow through HMSL:

```
Input Sources:
├── check.py          → collect() → prepare() → check_secrets()
├── fingerprint.py    → collect() → prepare() → write_outputs()
└── hashicorp_vault.py → collect_list() → prepare() → check_secrets()
```

The key file is `ggshield/verticals/hmsl/collection.py` which contains:
- `collect()` - Collects secrets from file/env inputs
- `collect_list()` - Collects secrets from list of tuples (used by Vault)
- `prepare()` - Hashes secrets and builds the payload

### Two Collection Functions

**`collect()`** handles two input types:
- **FILE** (e.g., `secrets.txt`): Plain text file with one secret per line. No keys, just values.
  ```
  my-api-key-abc123
  database-password-456
  ```
  → yields `SecretWithKey(value="my-api-key-abc123", key=None)`

- **ENV** (e.g., `.env` file): Key-value pairs parsed with `dotenv`.
  ```
  DB_PASSWORD=super-secret-123
  API_KEY=sk-live-abc123
  ```
  → yields `SecretWithKey(value="super-secret-123", key="DB_PASSWORD")`

**`collect_list()`** handles HashiCorp Vault input:
- Vault stores secrets as key-value pairs in its KV store
- The Vault API client returns `List[Tuple[str, str]]` like `[("DB_PASSWORD", "secret123"), ...]`
- Both key and value are used for filtering (e.g., skip keys like `HOST`, `PORT`)

For our length filter, we only care about the **value** in all cases.

### Existing Filtering: EXCLUDED_KEYS and EXCLUDED_VALUES

The codebase already has filtering mechanisms in `ggshield/verticals/hmsl/utils.py`. These are **separate filters** applied to different fields of `SecretWithKey`:

| Filter | Applies to | Purpose |
|--------|-----------|---------|
| `EXCLUDED_KEYS` | `key` (variable name) | Skip non-secret env vars like `HOST`, `PORT` |
| `EXCLUDED_VALUES` | `value` (the secret) | Skip placeholder/boolean values like `"true"`, `"null"` |

The filtering happens during `collect()` and `collect_list()`:

```python
# Both conditions checked (key OR value triggers exclusion)
if key.upper() in EXCLUDED_KEYS or value.lower() in EXCLUDED_VALUES:
    continue
```

Our new **length filter** applies to the **value** only (like `EXCLUDED_VALUES`).

---

### Refactoring: Extract a Unified Validation Utility

Currently, the filtering logic is **inline and duplicated** in both `collect()` and `collect_list()`:

**Proposal**: Extract this into a single validation function in `utils.py` that handles all exclusion logic, including the new length check.

Usage in `collect()` and `collect_list()`:
```python
if not should_process_secret(value, key):
    continue
```

### Q1: Should the minimum length be configurable?

**Decision**: No. Hardcoded as a constant (`MIN_SECRET_LENGTH = 6`) that maintainers can modify in code, but not exposed as a CLI option.

### Q2: What should the CLI message say?

**Decision**: No message unless verbose mode is enabled.

### Q3: Should this apply to all HMSL commands?

**Decision**: Yes, apply to all HMSL commands (`check`, `fingerprint`, `hashicorp_vault`).

### Q4: Should filtered secrets appear in JSON output?

**Decision**: No change to JSON output. Filtered secrets are simply not processed, so they won't appear. This is consistent with how `EXCLUDED_VALUES` works - those values are silently skipped.

### Q5: What about the edge case where ALL secrets are filtered?

**Decision**: Short-circuit. Don't call the API if there are no secrets to check. 

### Q6: Should we add the length to `EXCLUDED_VALUES` logic or keep it separate?

**Decision**: Consolidate into a single `should_process_secret()` utility function. While `EXCLUDED_VALUES` (exact match) and length filtering are different criteria, they serve the same purpose: filtering out non-secrets. A unified function:

### Q7: Should `should_process_secret()` return detailed exclusion reasons?

**Decision**: No, keep it simple with a boolean return.

### DIFF
--- a/tests/functional/hmsl/test_hmsl_check.py
+++ b/tests/functional/hmsl/test_hmsl_check.py
@@ -1,0 +1,52 @@
+"""
+Functional tests for the HMSL check command, focusing on secret filtering.
+"""
+
+from pathlib import Path
+
+from tests.functional.utils import run_ggshield
+
+
+def test_check_short_circuits_when_all_filtered(tmp_path: Path) -> None:
+    """
+    GIVEN a file where all secrets are too short or excluded
+    WHEN running the check command
+    THEN the command short-circuits without calling the API
+    """
+    secrets_file = tmp_path / "secrets.txt"
+    secrets_file.write_text(
+        "abc\n"  # 3 chars - filtered
+        "12345\n"  # 5 chars - filtered
+        "changeme\n"  # excluded value
+    )
+
+    result = run_ggshield(
+        "hmsl", "check", str(secrets_file), cwd=tmp_path, expected_code=0
+    )
+
+    # Should indicate no secrets found (short-circuit, no API call)
+    assert "Collected 0 secrets" in result.stderr
+    assert "No leaked secret has been found" in result.stderr
+
+
+def test_check_env_filters_excluded_keys_and_values(tmp_path: Path) -> None:
+    """
+    GIVEN an env file with only excluded keys and values
+    WHEN running the check command with --type env
+    THEN the command short-circuits without calling the API
+    """
+    env_file = tmp_path / ".env"
+    env_file.write_text(
+        "HOST=192.168.1.1\n"  # excluded key
+        "PORT=8080\n"  # excluded key
+        "DEBUG=placeholder\n"  # excluded value
+        "MODE=changeme\n"  # excluded value
+    )
+
+    result = run_ggshield(
+        "hmsl", "check", "--type", "env", str(env_file), cwd=tmp_path, expected_code=0
+    )
+
+    # Should indicate no secrets found (short-circuit, no API call)
+    assert "Collected 0 secrets" in result.stderr
+    assert "No leaked secret has been found" in result.stderr
--- a/tests/functional/hmsl/test_hmsl_fingerprint.py
+++ b/tests/functional/hmsl/test_hmsl_fingerprint.py
@@ -1,0 +1,103 @@
+"""
+Functional tests for the HMSL fingerprint command, focusing on secret filtering.
+"""
+
+from pathlib import Path
+
+from tests.functional.utils import run_ggshield
+
+
+def test_fingerprint_filters_short_secrets(tmp_path: Path) -> None:
+    """
+    GIVEN a file with secrets of various lengths
+    WHEN running the fingerprint command
+    THEN only secrets >= 6 characters are processed
+    """
+    secrets_file = tmp_path / "secrets.txt"
+    secrets_file.write_text(
+        "abc\n"  # 3 chars - filtered
+        "12345\n"  # 5 chars - filtered
+        "123456\n"  # 6 chars - kept
+        "validSecretHere\n"  # 15 chars - kept
+    )
+
+    result = run_ggshield(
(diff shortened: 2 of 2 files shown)


xblanchot-gg:
Looks good to me thanks ! I never heard about this hashicorp vault thing, I don't really understand what it does, but that's unrelated

pierrelalanne:
> Looks good to me thanks ! I never heard about this hashicorp vault thing, I don't really understand what it does, but that's unrelated

At some point we built a mini integration with Vault to pull your secrets hash and check them against HMSL IIRC. This could totally be removed as I think it is obsolete and not used.

tests/unit/verticals/hmsl/test_hmsl_utils.py
```diff
@@ -169,3 +173,90 @@ def test_good_token_validation(monkeypatch):
     )
     monkeypatch.setattr(time, "time", lambda: 1680000001)
     assert is_token_valid(token, "https://hasmysecretleaked.gitguardian.com")
+
+
+class TestShouldProcessSecret:
```
sevbch: You can refactor this class into 1 test using parametrized

pierrelalanne: Thx, I proposed a parametrized test instead.

agateau-gg:
> With this, you can almost not read the diff.

I certainly want to read the diff. The PR description should help me do this, not replace the diff. It should add what I can't get from reading the diff:
- The motivation behind the change
- The main architectural changes
- Are the separate commits additive changes that make sense reviewing one by one or do I risk commenting on issues that are fixed in later commits?
- Where to start: give me a guided tour of the changes. Maybe many changes are a reaction to an API change (the signature of a function changed for example). It would make the review easier to point out where the API change was made so that I can start reviewing this, and quickly go over where the changes are "boilerplaty".

I certainly don't want to read a wall of text. Before AI we would get one-line descriptions. Now with AI we get generated walls of text that we are not sure the author actually wrote or even read, but definitely didn't take the time to trim. The only conclusion is that people don't like the act of writing prose 😢.

sevbch (approved):

PR #1172, opened 2026-02-09
fix: keep docker stdout off JSON output

clement-tourriere:
## Summary
- Prevents docker scan stdout from leaking into JSON output, ensuring clean parseable JSON when using `--json` flag.

## Test plan
- [x] Unit tests updated in `tests/unit/verticals/secret/test_scan_docker.py`

### DIFF
--- a/tests/unit/verticals/secret/test_scan_docker.py
+++ b/tests/unit/verticals/secret/test_scan_docker.py
@@ -1,5 +1,6 @@
 import re
 import subprocess
+import sys
 from pathlib import Path
 from typing import Dict, List
 from unittest.mock import patch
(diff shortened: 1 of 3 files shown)


tests/unit/verticals/secret/test_scan_docker.py
```diff
@@ -121,11 +122,13 @@ class TestDockerPull:
     def test_docker_pull_image_success(self):
         with patch("subprocess.run") as call:
             docker_pull_image("ggshield-non-existant", DOCKER_TIMEOUT)
-            call.assert_called_once_with(
-                ["docker", "pull", "ggshield-non-existant"],
-                check=True,
-                timeout=DOCKER_TIMEOUT,
-            )
+            call.assert_called_once()
```
agateau-gg: Do we need to replace `assert_called_once_with()` with `assert_called_once()`? It seems to me that something like this would work fine:

```python
             call.assert_called_once_with(
                 ["docker", "pull", "ggshield-non-existant"],
                 check=True,
                 stdout=sys.stderr,
                 stderr=sys.stderr,
                 timeout=DOCKER_TIMEOUT,
             )
```

agateau-gg (approved):
Approved, but some changes seem unnecessary.

PR #1177, opened 2026-02-17
fix(ci): fetch target branch before computing MR commit range

alexpasmantier:
## Context

  `ggshield secret scan ci` computes the commit range using `origin/<target_branch>..<CI_COMMIT_SHA>` without fetching the target branch first. In CI environments with cached repos (`GIT_STRATEGY=fetch`) or shallow clones, `origin/<target_branch>` can be stale, causing ggshield to scan unrelated target branch commits. This led to 700+ API calls per MR and secrets reported in files not modified by the MR.

  The parallel code path in `previous_commit.py` already fetches the target branch, but `commit_range.py` did not.

  ## What has been done

  - Added `_fetch_target_branch()` in `commit_range.py` to fetch the target branch before computing the commit range, with error handling for network failures and timeouts.
  - Updated tests to mock and assert on the new fetch call.

  ## Validation

  - `pdm run pytest tests/unit/cmd/scan/test_ci.py -v`
  - In a real CI: run `ggshield secret scan ci --verbose` on an MR pipeline with `GIT_STRATEGY=fetch`. Output should show `Fetched <branch> from <remote>` and the commit count should  match the MR only.

  ## PR check list

### DIFF
--- a/ggshield/core/git_hooks/ci/commit_range.py
+++ b/ggshield/core/git_hooks/ci/commit_range.py
@@ -51,8 +51,8 @@ def _fetch_target_branch(remote: str, branch: str) -> None:
     computation to include unrelated commits from the target branch.
     """
     try:
+        ui.display_verbose(f"\tFetching {branch} from {remote}")
         git(["fetch", remote, branch])
-        ui.display_verbose(f"\tFetched {branch} from {remote}")
     except (subprocess.CalledProcessError, GitCommandTimeoutExpired):
         logger.warning("Failed to fetch %s from %s", branch, remote)
 


ggshield/core/git_hooks/ci/commit_range.py
```diff
@@ -32,6 +42,21 @@
...
+    """
+    Fetch the target branch from the remote to ensure the local ref is up to date.
+
+    In CI environments with cached repos (GIT_STRATEGY=fetch) or shallow clones,
+    the local origin/<target_branch> ref can be stale, causing the commit range
+    computation to include unrelated commits from the target branch.
+    """
+    try:
+        git(["fetch", remote, branch])
+        ui.display_verbose(f"\tFetched {branch} from {remote}")
```
agateau-gg: I think it would be better to show the message before fetching the changes, so that it's there even if the fetching fails.

alexpasmantier: That makes sense indeed 👍🏻 
Pushed the updated version.

PR #1180, opened 2026-02-19
Plugin signing

clement-tourriere:
## Context

  The plugin system shipped in 1.48.0 used a --force flag to skip security warnings when installing from non-GitGuardian sources. This was a placeholder — it displayed a text warning but performed no actual
  cryptographic verification.

  This PR replaces that mechanism with real sigstore-based signature verification for plugin wheels, using OIDC identity-based trust (keyless). It also fixes a bug where the verification mode was not forwarded
  through the GitHub release and artifact download paths.

  What has been done

  - Add ggshield.core.plugin.signature module: keyless signature verification using sigstore bundles. Wheels are verified against a configurable set of trusted GitHub Actions workflow identities (OIDC). Three modes:
   STRICT (block unsigned/invalid), WARN (log and continue), DISABLED (skip).
  - Replace --force with --allow-unsigned: the new flag downgrades verification from STRICT to WARN instead of silently skipping a text warning. Verification mode is also configurable via plugin_signature_mode in
  enterprise config.
  - Forward signature_mode through all download paths: download_from_github_release and download_from_github_artifact now accept and forward signature_mode. The artifact path also calls verify_wheel_signature before
   writing the manifest (was missing entirely).
  - Show signature status in plugin status: installed plugins now display their signature status (e.g. valid (GitGuardian/satori), missing).
  - Suppress signature/loader log noise at startup: logger levels are temporarily raised during plugin loading in __main__.py to avoid noisy output before logging is configured.
  - Collapse repetitive tests in test_install.py: 12 near-duplicate error-handling tests reduced to 4 parameterized tests + 1 standalone. Merged the two test_install_unavailable_plugin variants into a single
  parameterized test.
  - Add sigstore dependency to pyproject.toml.

  Validation

  pdm run pytest tests/unit/core/plugin/ tests/unit/cmd/plugin/ -x -q

  All 246 tests pass. To manually test signature verification:

  - ggshield plugin install tokenscanner — should verify signature in STRICT mode (default)
  - ggshield plugin install tokenscanner --allow-unsigned — should warn but proceed
  - ggshield plugin install ./unsigned.whl — should fail in STRICT mode with a clear error pointing to --allow-unsigned
  - ggshield plugin status — should show signature status for installed plugins

  PR check list

  - As much as possible, the changes include tests (unit and/or functional)
  - If the changes affect the end user (new feature, behavior change, bug fix) then the PR has a changelog entry (see doc/dev/getting-started.md). If the changes do not affect the end user, then the skip-changelog
  label has been added to the PR.

### DIFF
--- a/changelog.d/20260219_155531_clement.tourriere_plugin_signing.md
+++ b/changelog.d/20260219_155531_clement.tourriere_plugin_signing.md
@@ -1,0 +1,7 @@
+### Added
+
+- Add sigstore signature verification for plugin wheels, enforcing identity-based trust via OIDC. Plugins can be verified in STRICT, WARN, or DISABLED mode, configurable through enterprise settings or the `--allow-unsigned` flag.
+
+### Fixed
+
+- Forward `signature_mode` through GitHub release and GitHub artifact download paths, ensuring signature verification is applied consistently across all install sources.
--- a/changelog.d/20260219_155531_clement.tourriere_plugin_signing.md
+++ b/changelog.d/20260219_155531_clement.tourriere_plugin_signing.md
@@ -1,6 +1,6 @@
 ### Added
 
-- Add sigstore signature verification for plugin wheels, enforcing identity-based trust via OIDC. Plugins can be verified in STRICT, WARN, or DISABLED mode, configurable through enterprise settings or the `--allow-unsigned` flag.
+- Add sigstore signature verification for plugin wheels, enforcing identity-based trust via OIDC. Install and update operations are strict by default, while `--allow-unsigned` persists an explicit trust exception for the exact wheel hash so explicitly accepted unsigned plugins can still load at runtime.
 
 ### Fixed
 
--- a/ggshield/__main__.py
+++ b/ggshield/__main__.py
@@ -70,6 +70,15 @@ def _load_plugins() -> PluginRegistry:
     """Load plugins at module level so commands are available."""
     global _plugin_registry
     if _plugin_registry is None:
+        # Suppress signature/loader loggers during startup to avoid noisy output
+        # before logging is configured
+        sig_logger = logging.getLogger("ggshield.core.plugin.signature")
+        loader_logger = logging.getLogger("ggshield.core.plugin.loader")
+        orig_sig_level = sig_logger.level
+        orig_loader_level = loader_logger.level
+        sig_logger.setLevel(logging.CRITICAL)
+        loader_logger.setLevel(logging.CRITICAL)
+
         try:
             enterprise_config = EnterpriseConfig.load()
             plugin_loader = PluginLoader(enterprise_config)
@@ -77,6 +86,9 @@ def _load_plugins() -> PluginRegistry:
         except Exception as e:
             _deferred_warnings.append(f"Failed to load plugins: {e}")
             _plugin_registry = PluginRegistry()
+        finally:
+            sig_logger.setLevel(orig_sig_level)
+            loader_logger.setLevel(orig_loader_level)
 
         # Make registry available to hooks module
         from ggshield.core.plugin.hooks import set_plugin_registry
--- a/ggshield/__main__.py
+++ b/ggshield/__main__.py
@@ -70,15 +70,6 @@ def _load_plugins() -> PluginRegistry:
     """Load plugins at module level so commands are available."""
     global _plugin_registry
     if _plugin_registry is None:
-        # Suppress signature/loader loggers during startup to avoid noisy output
-        # before logging is configured
-        sig_logger = logging.getLogger("ggshield.core.plugin.signature")
-        loader_logger = logging.getLogger("ggshield.core.plugin.loader")
-        orig_sig_level = sig_logger.level
-        orig_loader_level = loader_logger.level
-        sig_logger.setLevel(logging.CRITICAL)
-        loader_logger.setLevel(logging.CRITICAL)
-
         try:
             enterprise_config = EnterpriseConfig.load()
             plugin_loader = PluginLoader(enterprise_config)
@@ -86,9 +77,6 @@ def _load_plugins() -> PluginRegistry:
         except Exception as e:
             _deferred_warnings.append(f"Failed to load plugins: {e}")
             _plugin_registry = PluginRegistry()
-        finally:
-            sig_logger.setLevel(orig_sig_level)
-            loader_logger.setLevel(orig_loader_level)
 
         # Make registry available to hooks module
         from ggshield.core.plugin.hooks import set_plugin_registry
@@ -286,13 +274,13 @@ def main(args: Optional[List[str]] = None) -> Any:
 
     `args` is only used by unit-tests.
     """
+    log_utils.disable_logs()
+
     _register_plugin_commands()
 
     # Required by pyinstaller when forking.
     # See https://pyinstaller.org/en/latest/common-issues-and-pitfalls.html#multi-processing
     multiprocessing.freeze_support()
-
-    log_utils.disable_logs()
 
     if not os.getenv("GG_PLAINTEXT_OUTPUT", False) and sys.stderr.isatty():
         ui.set_ui(RichGGShieldUI())
--- a/ggshield/cmd/plugin/install.py
+++ b/ggshield/cmd/plugin/install.py
@@ -27,6 +27,10 @@ from ggshield.core.plugin.downloader import (
     InsecureSourceError,
     PluginDownloader,
 )
+from ggshield.core.plugin.signature import (
+    SignatureVerificationError,
+    SignatureVerificationMode,
+)
 
 
 def detect_source_type(plugin_source: str) -> PluginSourceType:
@@ -64,12 +68,6 @@ def detect_source_type(plugin_source: str) -> PluginSourceType:
     return PluginSourceType.GITGUARDIAN_API
 
 
-def _display_unsigned_warning() -> None:
-    """Display warning about installing unsigned plugins."""
-    ui.display_warning("This plugin is not from GitGuardian and has not been verified.")
-    ui.display_warning("Only install plugins from sources you trust.")
-
-
 @click.command()
 @click.argument("plugin_source")
 @click.option(
@@ -85,10 +83,10 @@ def _display_unsigned_warning() -> None:
     help="Expected SHA256 checksum for URL verification",
 )
 @click.option(
-    "--force",
-    "force",
+    "--allow-unsigned",
+    "allow_unsigned",
     is_flag=True,
-    help="Skip security warnings for non-GitGuardian sources",
+    help="Allow installing plugins without valid signatures (overrides strict mode)",
 )
 @add_common_options()
 @click.pass_context
@@ -97,7 +95,7 @@ def install_cmd(
     plugin_source: str,
     version: Optional[str],
     sha256: Optional[str],
-    force: bool,
+    allow_unsigned: bool,
     **kwargs: Any,
 ) -> None:
     """
@@ -127,24 +125,31 @@ def install_cmd(
 
         ggshield plugin install https://github.com/owner/repo/actions/runs/123/artifacts/456
     """
+    # Determine signature verification mode
+    enterprise_config = EnterpriseConfig.load()
+    signature_mode = enterprise_config.get_signature_mode()
+    if allow_unsigned:
+        signature_mode = SignatureVerificationMode.WARN
+
     source_type = detect_source_type(plugin_source)
 
     if source_type == PluginSourceType.GITHUB_ARTIFACT:
-        _install_from_github_artifact(ctx, plugin_source, force)
+        _install_from_github_artifact(ctx, plugin_source, signature_mode)
     elif source_type == PluginSourceType.GITHUB_RELEASE:
-        _install_from_github_release(ctx, plugin_source, sha256, force)
+        _install_from_github_release(ctx, plugin_source, sha256, signature_mode)
     elif source_type == PluginSourceType.URL:
-        _install_from_url(ctx, plugin_source, sha256, force)
+        _install_from_url(ctx, plugin_source, sha256, signature_mode)
     elif source_type == PluginSourceType.LOCAL_FILE:
-        _install_from_local_wheel(ctx, plugin_source, force)
+        _install_from_local_wheel(ctx, plugin_source, signature_mode)
     else:
-        _install_from_gitguardian(ctx, plugin_source, version)
+        _install_from_gitguardian(ctx, plugin_source, version, signature_mode)
 
 
 def _install_from_gitguardian(
     ctx: click.Context,
     plugin_name: str,
     version: Optional[str],
+    signature_mode: SignatureVerificationMode = SignatureVerificationMode.STRICT,
 ) -> None:
     """Install a plugin from GitGuardian API."""
     ctx_obj = ContextObj.get(ctx)
@@ -192,7 +197,9 @@ def _install_from_gitguardian(
         )
 
         # Download and install
-        downloader.download_and_install(download_info, plugin_name)
+        downloader.download_and_install(
+            download_info, plugin_name, signature_mode=signature_mode
+        )
 
         # Enable in config
         enterprise_config.enable_plugin(plugin_name, version=download_info.version)
@@ -202,6 +209,12 @@ def _install_from_gitguardian(
 
         ui.display_info(f"Installed {plugin_name} v{download_info.version}")
 
+    except SignatureVerificationError as e:
+        ui.display_error(f"Signature verification failed for {plugin_name}: {e}")
+        ui.display_info(
+            "Use --allow-unsigned to install without signature verification"
+        )
+        ctx.exit(ExitCode.UNEXPECTED_ERROR)
     except PluginNotAvailableError as e:
         ui.display_error(f"Failed to install {plugin_name}: {e}")
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
@@ -216,7 +229,7 @@ def _install_from_gitguardian(
 def _install_from_local_wheel(
     ctx: click.Context,
     wheel_path_str: str,
-    force: bool,
+    signature_mode: SignatureVerificationMode = SignatureVerificationMode.STRICT,
 ) -> None:
     """Install a plugin from a local wheel file."""
     wheel_path = Path(wheel_path_str)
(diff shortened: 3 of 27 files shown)


ggshield/core/config/enterprise_config.py
```diff
@@ -4,7 +4,11 @@
 
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, Optional
```
sevbch: You can put this under the `if TYPE_CHECKING:` below and add `from __future__ import annotations` at the top of the file

ggshield/core/config/enterprise_config.py
```diff
@@ -54,7 +59,9 @@ def load(cls) -> "EnterpriseConfig":
             else:
                 plugins[name] = PluginConfig(enabled=True)
 
-        return cls(plugins=plugins)
+        plugin_signature_mode = data.get("plugin_signature_mode", "strict")
```
sevbch: "plugin_signature_mode" should probably be a const and "strict" part of an enum

ggshield/core/config/enterprise_config.py
```diff
@@ -72,13 +79,24 @@ def save(self) -> None:
...
         # Remove None values for cleaner YAML
         for plugin_data in data["plugins"].values():
             if plugin_data["version"] is None:
                 del plugin_data["version"]
 
         save_yaml_dict(data, config_path)
 
+    def get_signature_mode(self) -> "SignatureVerificationMode":
+        """Get the signature verification mode."""
+        from ggshield.core.plugin.signature import SignatureVerificationMode
```
sevbch: to move at the top

changelog.d/20260219_155531_clement.tourriere_plugin_signing.md
```diff
@@ -0,0 +1,7 @@
+### Added
+
+- Add sigstore signature verification for plugin wheels, enforcing identity-based trust via OIDC. Plugins can be verified in STRICT, WARN, or DISABLED mode, configurable through enterprise settings or the `--allow-unsigned` flag.
```
sevbch: `--allow-unsigned` is yet another option: did you look into merging it with the already existing `--insecure` option?

ggshield/__main__.py
```diff
@@ -70,13 +70,25 @@ def _load_plugins() -> PluginRegistry:
     """Load plugins at module level so commands are available."""
     global _plugin_registry
     if _plugin_registry is None:
+        # Suppress signature/loader loggers during startup to avoid noisy output
+        # before logging is configured
```
sevbch: This looks like a workaround for sth that should be addressed at the source: configure properly the concerned loggers from the start

sevbch (dismissed):
I only reviewed files outside of `**/plugin/*` paths, since those will be reviewed by @GitGuardian/nhi-governance

ggshield/core/config/enterprise_config.py
clement-tourriere: Done — `from __future__ import annotations` is at the top of the file and `SignatureVerificationMode` is now under the `if TYPE_CHECKING:` block.

ggshield/core/config/enterprise_config.py
clement-tourriere: Good call. I introduced `_PLUGIN_SIGNATURE_MODE_KEY` and `_DEFAULT_SIGNATURE_MODE` as module-level constants, and `SignatureVerificationMode` is already an `enum.Enum` in `plugin/signature.py`.

  I kept the dataclass field itself as `str` because typing it as `SignatureVerificationMode` would pull `signature.py` into the module's import graph and trigger the circular import `enterprise_config →
  plugin/__init__ → loader → enterprise_config`. `get_signature_mode()` converts to the enum at the usage boundary. Happy to invert that dependency in a follow-up if you'd prefer the dataclass to hold the enum
  directly.

ggshield/core/config/enterprise_config.py
clement-tourriere: This one is a deferred import on purpose — moving it to the top triggers `enterprise_config → plugin/__init__ → loader → enterprise_config` at import time. I've left a comment right above the import explaining
  why. The type annotation on `get_signature_mode()` still resolves correctly thanks to `from __future__ import annotations` + the `TYPE_CHECKING` import, so the public signature isn't affected.

changelog.d/20260219_155531_clement.tourriere_plugin_signing.md
clement-tourriere: I considered it but ruled it out — the two flags target different trust layers:

  - `--insecure` disables **TLS certificate verification** (transport-layer trust). It's a client-wide flag that affects every HTTPS call: GitGuardian API, OAuth, etc.
  - `--allow-unsigned` downgrades **sigstore wheel signature verification** (artifact/identity trust) from STRICT to WARN, scoped to plugin installation only.

  Conflating them would mean a user on an internal CA (a legitimate reason for `--insecure`) is silently opted into accepting unsigned plugins — a strictly weaker security posture for no reason. It would also widen
  `--insecure`'s blast radius well beyond its current documented meaning ("Skip all certificate verification checks"). This matches how e.g. `pip`'s `--trusted-host` (TLS) and artifact-integrity flags stay distinct.
   I'd rather keep the semantics orthogonal.

ggshield/__main__.py
clement-tourriere: Agreed — suppressing loggers was the wrong fix. I reworked it: `_load_plugins()` no longer touches logger levels and no longer emits log records during plugin import. Instead, any warning is collected into a
  `_deferred_warnings: List[str]`, and `cli()` flushes them via `logger.warning(...)` once the UI / log level has been configured from user config. No logger mutation anywhere. See `ggshield/__main__.py` around L66,
   L69–85, and L155–158.

ggshield/core/plugin/signature.py
```diff
@@ -0,0 +1,174 @@
...
+
+import enum
+import logging
+from dataclasses import dataclass
+from pathlib import Path
+from typing import List, Optional
+
+from sigstore.models import Bundle
+from sigstore.verify import Verifier
+from sigstore.verify.policy import AllOf, GitHubWorkflowRepository, OIDCIssuer
```
HelifeWasTaken, line 23:
```suggestion
from sigstore.errors import VerificationError
from sigstore.models import Bundle
from sigstore.verify import Verifier
from sigstore.verify.policy import AllOf, GitHubWorkflowRepository, OIDCIssuer
```
Import `VerificationError` from `sigstore.errors` so the `except` below can be narrowed
to that specific type instead of bare `Exception`.

ggshield/core/plugin/signature.py
```diff
@@ -0,0 +1,174 @@
...
+            logger.info(
+                "Signature valid for %s (repository: %s)",
+                wheel_path.name,
+                trusted.repository,
+            )
+            return SignatureInfo(
+                status=SignatureStatus.VALID,
+                identity=trusted.repository,
+            )
+        except Exception as e:
```
HelifeWasTaken:
```suggestion
except VerificationError as e:
```
`except Exception` swallows any unexpected error — an `IOError` reading the wheel,
a `MemoryError`, a sigstore API change — and silently treats it as "this identity
didn't match", continuing to the next identity. The debug log message is then
misleading. Narrow to `VerificationError` so only expected sigstore verification
failures are caught here; everything else propagates normally.

ggshield/core/plugin/downloader.py
```diff
@@ -425,6 +479,9 @@ def download_from_github_artifact(
             # Compute SHA256
             sha256 = self._compute_sha256(dest_wheel_path)
 
+        # Verify signature
+        sig_info = verify_wheel_signature(dest_wheel_path, signature_mode)
```
HelifeWasTaken, line 496: Bundle files (`.sigstore` / `.sigstore.json`) inside the artifact zip are never
extracted, so `verify_wheel_signature` will always return `MISSING` for artifact
installs — causing every artifact install to fail in STRICT mode (the default).

The fix belongs a few lines earlier, still inside the `with tempfile.TemporaryDirectory()` block
where `temp_wheel_path` is in scope (around the `shutil.copy2` call for the wheel):

```python
shutil.copy2(temp_wheel_path, dest_wheel_path)

# Copy bundle alongside the wheel if present in the artifact
for ext in (".sigstore", ".sigstore.json"):
    bundle_src = temp_wheel_path.parent / (temp_wheel_path.name + ext)
    if bundle_src.exists():
        shutil.copy2(bundle_src, plugin_dir / bundle_src.name)
        break
```

By the time we reach `verify_wheel_signature` here the temp dir is already cleaned up,
so there is no bundle to find.

ggshield/core/config/enterprise_config.py
```diff
@@ -91,15 +119,16 @@ def enable_plugin(self, plugin_name: str, version: Optional[str] = None) -> None
...
 
         self.plugins[plugin_name].enabled = False
 
     def is_plugin_enabled(self, plugin_name: str) -> bool:
         """Check if a plugin is enabled."""
         plugin_config = self.plugins.get(plugin_name)
-        # Default: disabled if not explicitly configured
-        return plugin_config.enabled if plugin_config else False
+        # Default: enabled if not explicitly configured
+        return plugin_config.enabled if plugin_config else True
```
HelifeWasTaken: This changed the default from `False` to `True`: plugins are now **enabled by
default** for any user who has never touched `enterprise_config.yaml`. Previously
they were disabled unless explicitly enabled. That is a silent breaking change —
a user who installed a plugin but deliberately didn't enable it will find it
loading automatically after this update. Was this intentional? If so, it should
be called out in the changelog.

ggshield/core/plugin/downloader.py
```diff
@@ -103,38 +134,42 @@ def download_and_install(
...
-            temp_path.rename(wheel_path)
-
             logger.info("Installed %s v%s", plugin_name, download_info.version)
 
             return wheel_path
 
         except requests.RequestException as e:
+            self._cleanup_failed_install(wheel_path)
             raise DownloadError(f"Failed to download plugin: {e}") from e
+        except SignatureVerificationError:
```
6d7a, line 147: Shouldn't the cleanup be triggered on arbitrary exceptions?

6d7a (commented):
I guess Mattis was faster than me 😁
Just one additional suggestion.

ggshield/cmd/plugin/install.py
```diff
@@ -192,7 +197,9 @@ def _install_from_gitguardian(
         )
 
         # Download and install
-        downloader.download_and_install(download_info, plugin_name)
+        downloader.download_and_install(
```
6d7a, line 200: Sorry for missing this yesterday: I noticed that this PR is not updating the plugin update command. This means that when updating, we always download with strict signature verification without taking the enterprise config into account.

6d7a (approved):
LGTM. Thank you for the changes!

tests/unit/core/plugin/test_signature.py
```diff
@@ -0,0 +1,290 @@
...
+        wheel = self._setup_wheel_and_bundle(tmp_path)
+
+        mock_verifier_cls = MagicMock()
+        mock_bundle_cls = MagicMock()
+        mock_all_of_cls = MagicMock()
+        mock_oidc_issuer_cls = MagicMock()
+        mock_gh_repo_cls = MagicMock()
+
+        mock_verifier = MagicMock()
+        mock_verifier_cls.production.return_value = mock_verifier
```
HelifeWasTaken, line 180:
```suggestion
mock_verifier.verify_artifact.side_effect = VerificationError("Verification failed")
```
This test (and the two below with the same pattern) passes today because `verify_wheel_signature`
catches bare `Exception`. But the sibling review comment suggests narrowing that to
`VerificationError` from `sigstore.errors`. After that fix is applied, `Exception("Verification failed")`
would **not** be caught and would propagate instead of being treated as a failed identity match —
breaking these tests.

Update the side effect to `VerificationError` (and add the import at the top of the file):

```python
from sigstore.errors import VerificationError
```

pyproject.toml
```diff
@@ -44,8 +44,9 @@ dependencies = [
     "python-dotenv~=0.21.0",
     "pyyaml~=6.0.1",
     "requests~=2.32.0",
-    "rich~=12.5.1",
-    "typing-extensions~=4.12.2",
+    "rich~=13.0",
+    "sigstore>=3.0.0",
```
HelifeWasTaken:
```suggestion
"sigstore>=3.0.0,<4",
```
`sigstore>=3.0.0` has no upper bound. The code calls sigstore internals directly
(`Bundle.from_json`, `Verifier.production()`, `verify_artifact`, policy classes in
`sigstore.verify.policy`). A future sigstore 4.x could rename or remove any of these
and silently break signature verification at runtime, with no pip constraint to warn about it.
Pin to `<4` to get a clear pip error on the next major instead.

tests/unit/core/plugin/test_downloader.py
```diff
@@ -278,17 +287,19 @@ def test_download_and_install_manifest_failure_cleans_temp_file(
...
+                    return_value=MOCK_SIG_INFO,
                 ):
-                    with pytest.raises(OSError):
-                        downloader.download_and_install(download_info, "testplugin")
+                    downloader = PluginDownloader()
+
+                    with patch.object(
+                        downloader, "_write_manifest", side_effect=OSError("disk full")
+                    ):
+                        with pytest.raises(OSError):
```
HelifeWasTaken, line 402: This test now only checks that `temp_path` is gone — but it used to also assert
`assert not wheel_path.exists()`. The assertion was removed because `temp_path.rename(wheel_path)`
now happens **before** `_write_manifest`. If `_write_manifest` raises `OSError` (or any other
exception that isn't `requests.RequestException` or `SignatureVerificationError`), the wheel is
left at `wheel_path` without cleanup.

Consider adding `OSError` to the cleanup handlers in `download_and_install`, or wrapping the
post-rename block in a `try/finally` that calls `_cleanup_failed_install`. Restoring the original
assertion here would catch the regression.

tests/unit/cmd/plugin/test_install.py
```diff
@@ -1025,330 +901,410 @@ def test_install_local_wheel_download_error(
...
-    def test_install_github_artifact_generic_error(self, cli_fs_runner) -> None:
+        assert call_kwargs["signature_mode"] == SignatureVerificationMode.STRICT
+
+    def test_allow_unsigned_flag(self, cli_fs_runner, tmp_path: Path) -> None:
         """
-        GIVEN a GitHub artifact URL with unexpected error
-        WHEN running 'ggshield plugin install <url>'
-        THEN it shows an error
+        GIVEN --allow-unsigned flag
+        WHEN installing a plugin
```
HelifeWasTaken, line 1275: `SignatureVerificationMode` is imported mid-function. It is already imported (indirectly) by
the production code under test, and importing inside a function body makes the dependency
less visible and slows the test down marginally on repeated calls.
Move it to the module-level imports alongside the other `ggshield` imports.

ggshield/core/plugin/loader.py
```diff
@@ -221,6 +260,32 @@ def _load_from_wheel(self, wheel_path: Path) -> Optional[GGShieldPlugin]:
             logger.warning("Failed to load wheel %s: %s", wheel_path, e)
             return None
 
+    def _load_mode_for(self, wheel_path: Path) -> SignatureVerificationMode:
```
6d7a: Doesn't this behavior defeat the purpose of signature verification at load-time? If the verification override lives alongside the wheel, an attacker can simply drop a manifest along with the wheel to trigger skipping the signature verification.

clement-tourriere: Good catch ! I changed that: the manifest no longer affects runtime verification. Load-time verification is STRICT again by default, and --allow-unsigned now stores an explicit trust record outside the plugin directory, tied to the exact wheel SHA256. So an unsigned plugin only loads if the user explicitly trusted that exact wheel at install time; if the wheel changes later, it’s rejected again.

6d7a (approved):

mattisdalleau-gg (approved):
LGTM !

ggshield/cmd/plugin/install.py
```diff
@@ -202,6 +211,12 @@ def _install_from_gitguardian(
 
         ui.display_info(f"Installed {plugin_name} v{download_info.version}")
 
+    except SignatureVerificationError as e:
+        ui.display_error(f"Signature verification failed for {plugin_name}: {e}")
+        ui.display_info(
+            "Use --allow-unsigned to install without requiring a valid signature"
```
benjaminrigaud-gg: Should we really recommend bypassing? We may want to be more cautious there.

ggshield/cmd/plugin/install.py
```diff
@@ -318,6 +336,12 @@ def _install_from_github_release(
 
         ui.display_info(f"Installed {plugin_name} v{version}")
 
+    except SignatureVerificationError as e:
+        ui.display_error(f"Signature verification failed: {e}")
+        ui.display_info(
+            "Use --allow-unsigned to install without requiring a valid signature"
```
benjaminrigaud-gg: same.

ggshield/cmd/plugin/install.py
```diff
@@ -335,27 +359,33 @@ def _install_from_github_release(
...
         # Enable in config
         enterprise_config.enable_plugin(plugin_name, version=version)
         enterprise_config.save()
 
         ui.display_info(f"Installed {plugin_name} v{version}")
 
+    except SignatureVerificationError as e:
+        ui.display_error(f"Signature verification failed: {e}")
+        ui.display_info(
+            "Use --allow-unsigned to install without requiring a valid signature"
```
benjaminrigaud-gg: again.

ggshield/core/plugin/downloader.py
```diff
@@ -600,19 +709,111 @@ def _write_manifest(
...
 
+    def _sync_trust_record(
+        self,
+        plugin_name: str,
+        sha256: str,
+        signature_info: SignatureInfo,
+    ) -> None:
+        """Persist or revoke trust for a plugin based on install-time verification."""
+        if signature_info.status == SignatureStatus.VALID:
+            self.trust_store.revoke_plugin(plugin_name)
```
benjaminrigaud-gg, line 757: Do we need to revoke before trusting it again?

benjaminrigaud-gg (approved):

PR #1181, opened 2026-02-20
fix: assorted bug and security fixes

sevbch:
## Context

This PR addresses a collection of bugs and security issues found during a code review pass. Each fix is in its own commit.

A changelog entry was added only when the change isn't too much of an edge case.

There are many commits, but it should actually be a pretty quick review.

## What has been done

See commit messages.

## Validation

All unit tests pass, and all added non-reg tests pass.

## PR check list

### DIFF
--- a/ggshield/core/config/user_config.py
+++ b/ggshield/core/config/user_config.py
@@ -72,7 +72,9 @@ class SecretConfig(FilteredConfig):
                     self.prereceive_remediation_message
                 ),
                 "all_secrets": self.all_secrets,
-                "source_uuid": self.source_uuid,
+                "source_uuid": (
+                    str(self.source_uuid) if self.source_uuid is not None else None
+                ),
             }
         )
 
--- a/tests/unit/core/config/test_user_config.py
+++ b/tests/unit/core/config/test_user_config.py
@@ -1,10 +1,16 @@
+import json
 from datetime import datetime, timezone
 from typing import Optional
+from uuid import UUID
 
 import pytest
 
 from ggshield.core.config import Config
-from ggshield.core.config.user_config import CURRENT_CONFIG_VERSION, UserConfig
+from ggshield.core.config.user_config import (
+    CURRENT_CONFIG_VERSION,
+    SecretConfig,
+    UserConfig,
+)
 from ggshield.core.errors import ParseError, UnexpectedError
 from ggshield.core.types import IgnoredMatch
 from tests.unit.conftest import write_text, write_yaml
@@ -313,6 +319,19 @@ class TestUserConfig:
         assert "allow_self_signed" in captured.err
         assert "deprecated" in captured.err
         assert "insecure" in captured.err
+
+    def test_dump_for_monitoring_serializes_uuid_as_string(self):
+        """
+        GIVEN a SecretConfig with a UUID source_uuid
+        WHEN dump_for_monitoring() is called
+        THEN source_uuid is serialized as a plain string, not a UUID object
+        """
+        uid = UUID("12345678-1234-5678-1234-567812345678")
+        config = SecretConfig()
+        config.source_uuid = uid
+        result = json.loads(config.dump_for_monitoring())
+        assert isinstance(result["source_uuid"], str)
+        assert result["source_uuid"] == str(uid)
 
     def test_bad_local_config(self, local_config_path, global_config_path):
         """


ggshield/core/config/utils.py
```diff
@@ -64,17 +65,18 @@ def save_yaml_dict(
...
+        if restricted:
+            # Create file with owner-only permissions (0o600) atomically
+            fd = os.open(str(p), os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+            with os.fdopen(fd, "w") as f:
+                f.write(stream)
+        else:
+            with p.open("w") as f:
+                f.write(stream)
+    except Exception as e:
+        raise UnexpectedError(f"Failed to save config to {path}:\n{str(e)}") from e
```
alexpasmantier: It seems like this won't enforce the 600 permissions anymore when the file already exists which looks like a little regression.

ggshield/verticals/hmsl/client.py
```diff
@@ -176,9 +180,15 @@ def _query(self, endpoint: str, body: Dict[str, List[str]]) -> Any:
...
         response.raise_for_status()
         now = datetime.now().replace(microsecond=0)
-        self._quota = Quota(
-            remaining=int(response.headers["RateLimit-Remaining"]),
-            limit=int(response.headers["RateLimit-Limit"]),
-            reset=now + timedelta(seconds=1 + int(response.headers["RateLimit-Reset"])),
-        )
+        try:
+            self._quota = Quota(
+                remaining=int(response.headers.get("RateLimit-Remaining", 0)),
```
6d7a: I wonder if we should silently fall back to 0 in these cases. Since HSML is maintained by us, I feel we would want to know about any unexpected header state.

ggshield/core/errors.py
```diff
@@ -210,7 +210,8 @@ def handle_exception(exc: Exception) -> int:
 def handle_api_error(detail: Detail) -> None:
     # Use %s for status_code because it can be None. Logger is OK with an int being
     # passed for a %s placeholder.
-    logger.error("status_code=%s detail=%s", detail.status_code, detail.detail)
+    logger.error("status_code=%s", detail.status_code)
+    logger.debug("detail=%s", detail.detail)
     if detail.status_code == 401:
         raise click.UsageError(detail.detail)
```
6d7a, line 216: Do we need to obfuscate `detail.detail` here, too? Or can we assume that `401` responses do not contain sensitive data?

ggshield/verticals/hmsl/secret_manager/hashicorp_vault/api_client.py
```diff
@@ -35,15 +36,22 @@ def __init__(self, vault_url: str, api_token: str) -> None:
...
-        )
+        try:
+            api_res = self.session.request(
+                method,
+                f"{self.vault_url}/{api_version}/{endpoint}",
+                timeout=self._DEFAULT_TIMEOUT,
+            )
+        except requests.exceptions.RequestException as e:
+            raise VaultAPIError(f"Network error contacting Vault: {e}") from e
         api_res.raise_for_status()
```
6d7a, line 55: Shouldn't this be part of the `try` block?

ggshield/verticals/hmsl/client.py
```diff
@@ -176,9 +180,15 @@ def _query(self, endpoint: str, body: Dict[str, List[str]]) -> Any:
...
+                limit=int(response.headers.get("RateLimit-Limit", 0)),
+                reset=now
+                + timedelta(
+                    seconds=1 + int(response.headers.get("RateLimit-Reset", 0))
+                ),
+            )
+        except (ValueError, TypeError):
+            logger.warning("Could not parse RateLimit headers from HMSL response")
```
xblanchot-gg: That surprizes me, HMSL should have reliable output, did we already encountered this case ?

tests/unit/verticals/hmsl/secret_manager/hashicorp_vault/test_api_client.py
```diff
@@ -506,3 +506,18 @@ def test_make_request_network_error(vault_api_client):
...
+    WHEN repr() is called
+    THEN the secrets field is not included in the output
+    """
+    vs = VaultSecrets(
+        secrets=[("MY_KEY", "super_secret_value")],
+        not_fetched_paths=[],
+    )
+    representation = repr(vs)
+    assert "super_secret_value" not in representation
+    assert "MY_KEY" not in representation
```
xblanchot-gg: Doesn't this breaks the purpose of the integration ? We want to show which secrets have been publicly leaked, if i remember well, so we need to be able to identity at least the path and the key of a result

ggshield/core/tar_utils.py
```diff
@@ -44,7 +44,8 @@ def tar_from_ref_and_filepaths(
             raw_file_content = read_git_file(ref, path, wd)
             data = BytesIO(raw_file_content.encode())
 
-            tarinfo = tarfile.TarInfo(str(path))
+            safe_name = Path(path).as_posix().lstrip("/")
```
6d7a: This doesn't prevent the `..` path traversal case though. Could we use the `_is_bad_path()` util here?

ggshield/verticals/hmsl/client.py
sevbch: As discussed since we own the API we don't need to add those extra safety nets. I removed this commit

ggshield/verticals/hmsl/client.py
sevbch: See https://github.com/GitGuardian/ggshield/pull/1181/changes/a0fde5c85626bea1fc70133addd8acfce738dbd0#r2878864943
We discussed this with Xavier and decided to remove the commit

ggshield/core/errors.py
sevbch: I think it's fine because 401 is an authentication error, so at this stage, nothing has been scanned and there shouldn't be sensitive content displayed.

Also, displaying the details at this point can be helpful for the user to address their authentication issue.

ggshield/core/config/utils.py
sevbch: good catch! I amended the commit, see 17de70edac9290d0b35b3753a02133d528621098

ggshield/core/tar_utils.py
sevbch: good catch, I updated this in e17008f0d416fb2a47757ca80b932a40701dbf69 .

tests/unit/verticals/hmsl/secret_manager/hashicorp_vault/test_api_client.py
sevbch: good point, I removed the commit

ggshield/verticals/hmsl/secret_manager/hashicorp_vault/api_client.py
sevbch: `raise_for_status` raises `HTTPError` which are dealt with separately in the codebase (line 69, 92, and 122 in this file). So I think it makes sense to keep it separate and not encapsulate it within the VaultAPIError

ggshield/core/config/utils.py
sevbch: I actually ended up actually dropping this as the fix didn't work across all platforms and really address a concern of milli-seconds timeframe

PR #1183, opened 2026-03-03
feat(nhi,ai-agents): Add support for Cursor and Claude secrets hooks

xblanchot-gg:
This PR adds support for Cursor and Claude secret detection hooks.
It's from Paul Petit, waiting for his account to be re-activated, the PR description will be detailed today.

### DIFF
--- a/ggshield/cmd/secret/scan/ai_hook.py
+++ b/ggshield/cmd/secret/scan/ai_hook.py
@@ -1,0 +1,51 @@
+import sys
+from typing import Any
+
+import click
+
+from ggshield.cmd.secret.scan.secret_scan_common_options import (
+    add_secret_scan_common_options,
+)
+from ggshield.cmd.utils.context_obj import ContextObj
+from ggshield.core import ui
+from ggshield.core.client import create_client_from_config
+from ggshield.core.scan import ScanContext, ScanMode
+from ggshield.verticals.secret import SecretScanner
+from ggshield.verticals.secret.ai_hook import AIHookScanner
+from ggshield.verticals.secret.ai_hook.models import MAX_READ_SIZE
+
+
+@click.command()
+@add_secret_scan_common_options()
+@click.pass_context
+def ai_hook_cmd(
+    ctx: click.Context,
+    **kwargs: Any,
+) -> int:
+    """
+    Scan AI tool interactions for secrets.
+
+    Reads a hook event from stdin as JSON, processes it based on the
+    event type and mode, and outputs the response to stdout as JSON.
+    """
+    ctx_obj = ContextObj.get(ctx)
+    config = ctx_obj.config
+    ctx_obj.client = create_client_from_config(config)
+    scanner = SecretScanner(
+        client=ctx_obj.client,
+        cache=ctx_obj.cache,
+        scan_context=ScanContext(
+            scan_mode=ScanMode.AI_HOOK,
+            command_path=ctx.command_path,
+        ),
+        secret_config=config.user_config.secret,
+    )
+
+    # Read input from stdin
+    stdin_content = sys.stdin.read(MAX_READ_SIZE).strip()
+
+    try:
+        return AIHookScanner(scanner).scan(stdin_content)
+    except ValueError as e:
+        ui.display_error(str(e.args[0]))
+        return 1
--- a/ggshield/verticals/secret/ai_hook/copilot.py
+++ b/ggshield/verticals/secret/ai_hook/copilot.py
@@ -1,0 +1,32 @@
+import json
+from pathlib import Path
+from typing import Any
+
+import click
+
+from .claude_code import Claude
+from .models import Result
+
+
+class Copilot(Claude):
+    """Behavior specific to Copilot Chat.
+
+    Inherits most of its behavior from Claude Code.
+    """
+
+    name = "Copilot"
+
+    def output_result(self, result: Result) -> int:
+        response: dict[str, Any] = {
+            "continue": not result.block,
+        }
+        if result.block:
+            # Having stopReason blocks Copilot, whether continue is true or false.
+            response["stopReason"] = result.message
+
+        click.echo(json.dumps(response))
+        return 0
+
+    @property
+    def settings_path(self) -> Path:
+        return Path(".github") / "hooks" / "hooks.json"
--- a/ggshield/verticals/secret/ai_hook/installation.py
+++ b/ggshield/verticals/secret/ai_hook/installation.py
@@ -1,0 +1,160 @@
+import json
+from copy import deepcopy
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Callable, Dict, List, Literal, Optional
+
+import click
+
+from ggshield.core.dirs import get_user_home_dir
+from ggshield.core.errors import UnexpectedError
+
+from .claude_code import Claude
+from .copilot import Copilot
+from .cursor import Cursor
+
+
+AI_FLAVORS = {
+    "cursor": Cursor,
+    "claude-code": Claude,
+    "copilot": Copilot,
+}
+
+
+@dataclass
+class InstallationStats:
+    added: int
+    already_present: int
+
+
+def install_hooks(
+    name: str, mode: Literal["local", "global"], force: bool = False
+) -> int:
+    """Install the hooks for the AI hook.
+
+    Args:
+        name: Name of the AI coding tool
+        mode: Mode of the hook installation
+        force: Whether to force the installation
+
+    Returns an error code (0 on success, 1 on failure)
+    """
+
+    try:
+        flavor = AI_FLAVORS[name]()
+    except KeyError:
+        raise ValueError(f"Unsupported tool name: {name}")
+
+    base_dir = get_user_home_dir() if mode == "global" else Path(".")
+    settings_path = base_dir / flavor.settings_path
+
+    command = "ggshield secret scan ai-hook"
+
+    # Load existing config or create new one
+    existing_config: dict = {}
+
+    if settings_path.exists():
+        try:
+            with settings_path.open("r", encoding="utf-8") as f:
+                existing_config = json.load(f)
+        except json.JSONDecodeError as e:
+            raise UnexpectedError(
+                f"Failed to parse {settings_path}: {e}. "
+                "Please fix or remove the file before installing hooks."
+            )
+
+    # Track what we did for reporting
+    stats = InstallationStats(
+        added=0,
+        already_present=0,
+    )
+
+    stats = _fill_dict(
+        config=existing_config,
+        template=flavor.settings_template,
+        command=command,
+        overwrite=force,
+        stats=stats,
+        locator=flavor.settings_locate,
+    )
+
+    # Ensure parent directory exists
+    settings_path.parent.mkdir(parents=True, exist_ok=True)
+
+    # Write the updated config
+    with settings_path.open("w", encoding="utf-8") as f:
+        json.dump(existing_config, f, indent=2)
+        f.write("\n")
+
+    # Report what happened
+    styled_path = click.style(settings_path, fg="yellow", bold=True)
+    if stats.added == 0 and stats.already_present > 0:
+        click.echo(f"{flavor.name} hooks already installed in {styled_path}")
+    elif stats.added > 0 and stats.already_present > 0:
+        click.echo(f"{flavor.name} hooks updated in {styled_path}")
+    else:
+        click.echo(f"{flavor.name} hooks successfully added in {styled_path}")
+
+    return 0
+
+
+def _fill_dict(
+    config: Dict[str, Any],
+    template: Dict[str, Any],
+    command: str,
+    overwrite: bool,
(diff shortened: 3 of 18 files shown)


ggshield/verticals/secret/ai_hook/constants.py
```diff
@@ -0,0 +1,106 @@
+from dataclasses import dataclass
+from enum import Enum, auto
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+import click
+
+
+MAX_FILE_READ_SIZE = 1024 * 1024 * 10  # 10MB
```
xblanchot-gg: What is the rationale behind that constant ?

ggshield/verticals/secret/ai_hook/models.py
```diff
@@ -0,0 +1,106 @@
...
+class Result:
+    """Result of a scan: allow or not."""
+
+    event_type: EventType
+    block: bool
+    message: str
+    nbr_secrets: int
+
+
+class Flavor:
```
xblanchot-gg, line 42: I would not put that in constants as it's a base class for handling data

ggshield/verticals/secret/ai_hook/copilot.py
```diff
@@ -0,0 +1,16 @@
+from pathlib import Path
+
+from .claude_code import Claude
+
+
+class Copilot(Claude):
+    """Behavior specific to Copilot Chat.
+
+    Inherits most of its behavior from Claude Code.
```
xblanchot-gg, line 14: Then should this rather be defined as the base behavior in Flavor, and just overriden in Cursor ?

ggshield/verticals/secret/ai_hook/installation.py
```diff
@@ -0,0 +1,152 @@
...
+
+    Returns an error code (0 on success, 1 on failure)
+    """
+
+    try:
+        flavor = AI_FLAVORS[name]()
+    except KeyError:
+        raise ValueError(f"Unsupported tool name: {name}")
+
+    base_dir = get_user_home_dir() if mode == "global" else Path(".")
```
xblanchot-gg, line 48: Does that works on Windows ?

ggshield/verticals/secret/ai_hook/installation.py
```diff
@@ -0,0 +1,152 @@
...
+        click.echo(f"{flavor.name} hooks already installed in {styled_path}")
+    elif stats["added"] > 0 and stats["already_present"] > 0:
+        click.echo(f"{flavor.name} hooks updated in {styled_path}")
+    else:
+        click.echo(f"{flavor.name} hooks successfully added in {styled_path}")
+
+    return 0
+
+
+def _fill_dict(
```
xblanchot-gg, line 101: That's quite hard to read, and feels even too generic in the principle. I think I would be more confident with one ad-hoc install process for each kind of configurations.

ggshield/verticals/secret/ai_hook/scanner.py
```diff
@@ -0,0 +1,276 @@
...
+                file = Path(identifier)
+                if file.is_file() and file.stat().st_size <= MAX_FILE_READ_SIZE:
+                    try:
+                        content = file.read_text()
+                    except (UnicodeDecodeError, OSError):
+                        content = ""
+                else:
+                    content = ""
+            else:
+                content = ""
```
xblanchot-gg: We would like to not return a payload in those case, to avoid hitting the API on an empty file to scan no ?

ggshield/verticals/secret/ai_hook/constants.py
paulpetit-gg-ext: This was suggested in Achille's automated review: put a limit on the size of read files. This was one of the two "open questions" at the end of my last comment: do we want such a limit? If yes, which one?

ggshield/verticals/secret/ai_hook/models.py
paulpetit-gg-ext: Makes sense. I will move it elsewhere.

xblanchot-gg: Yes indeed better name !

ggshield/verticals/secret/ai_hook/installation.py
paulpetit-gg-ext: I empathize with the feeling that it looks over-engineered, but to this implementation's defense:
- ad-hoc install processes would still have to deal with non-trivial logic: how to add hooks without disturbing existing ones, dealing with overriding or not an existing ggshield hook, etc... So we would have two 80% as complex implementations instead of this one.
- once we trust this function does its job, the variants become self-documenting: you basically have to just write a template as a Python dict. So adding support for future code assistants will be easy (granted, there is a good chance they will be exactly like one already implemented so it would be easy anyway).

ggshield/verticals/secret/ai_hook/scanner.py
paulpetit-gg-ext: I don't know about returning a `Payload` or not, but you are right we could avoid requesting the API with empty content.

ggshield/verticals/secret/ai_hook/models.py
paulpetit-gg-ext: After consideration, would it make more sense to let it where it is but rename the file `constants.py` in `model.py`?

ggshield/cmd/secret/scan/ai_hook.py
```diff
@@ -0,0 +1,49 @@
...
+        secret_config=config.user_config.secret,
+    )
+
+    # Read input from stdin
+    stdin_content = sys.stdin.read()
+
+    try:
+        return AIHookScanner(scanner).scan(stdin_content)
+    except ValueError as e:
+        click.echo(str(e.args[0]), err=True)
```
agateau-gg: You can use `ui.display_error()` here.

ggshield/verticals/secret/ai_hook/models.py
agateau-gg: I think `models.py` (with an `s`) would be nice. We already have  `verticals/hmsl/secret_manager/hashicorp_vault/models.py` so it's consistent.

ggshield/verticals/secret/ai_hook/models.py
```diff
@@ -0,0 +1,106 @@
...
+            result: the result of the scan.
+
+        Returns: the exit code.
+        """
+        if result.block:
+            click.echo(result.message, err=True)
+            return 2
+        else:
+            click.echo("No secrets found. Good to go.")
+            return 0
```
agateau-gg, line 67: Since every class inheriting from Flavor override this method, I think it should be abstract.

ggshield/verticals/secret/ai_hook/installation.py
```diff
@@ -0,0 +1,152 @@
...
+            # but first, make sure we only have one object in the template
+            if len(value) != 1:
+                raise ValueError(f"Expected only one object in template for {key}")
+
+            config_list = config.setdefault(key, [])
+            existing_value = locator(config_list, value[0])
+            if existing_value is not None:
+                # Found it. Continue with this object.
+                _fill_dict(existing_value, value[0], command, overwrite, stats, locator)
+                break
```
amascia-gg: why do we need the break here?

ggshield/verticals/secret/ai_hook/scanner.py
```diff
@@ -0,0 +1,276 @@
...
+    def scan(self, content: str) -> int:
+        """Scan the content, print the result and return the exit code."""
+
+        payload = self._parse_input(content)
+        result = self._scan_content(payload)
+
+        # Special case: in post-tool use, the action is already done, we can only notify the user
+        if result.block and payload.event_type == EventType.POST_TOOL_USE:
+            # Too late, but at least notify the user
+            self._send_secret_notification(
```
amascia-gg, line 47: can't this crash on headless/CI environments? should we use try/except and switch to best effort notification?

ggshield/verticals/secret/ai_hook/installation.py
```diff
@@ -0,0 +1,152 @@
...
+
+    return 0
+
+
+def _fill_dict(
+    config: Dict[str, Any],
+    template: Dict[str, Any],
+    command: str,
+    overwrite: bool,
+    stats: Dict[str, int],
```
agateau-gg: This could be a dataclass, and it could be returned by the function instead of being an in-out parameter.

ggshield/cmd/secret/scan/ai_hook.py
```diff
@@ -0,0 +1,49 @@
...
+        cache=ctx_obj.cache,
+        scan_context=ScanContext(
+            scan_mode=ScanMode.AI_HOOK,
+            command_path=ctx.command_path,
+        ),
+        secret_config=config.user_config.secret,
+    )
+
+    # Read input from stdin
+    stdin_content = sys.stdin.read()
```
amascia-gg: this is unbounded. Should we consider sys.stdin.read(MAX_SIZE) with a reasonable limit (e.g., 10-50MB) to avoid OOMs? from adversarial attacks for example

ggshield/verticals/secret/ai_hook/scanner.py
```diff
@@ -0,0 +1,276 @@
...
+
+from .claude_code import Claude
+from .constants import EventType, Flavor, Payload, Result, Tool
+from .cursor import Cursor
+
+
+class AIHookScanner:
+    """AI hook scanner.
+
+    It is called with the paylod of a hook event.
```
agateau-gg: nit: typo
```suggestion
    It is called with the payload of a hook event.
```

ggshield/verticals/secret/ai_hook/scanner.py
```diff
@@ -0,0 +1,276 @@
...
+            raise ValueError(f"Error: Failed to parse JSON from stdin: {e}") from e
+
+        # Infer the event type
+        event_name = lookup(data, ["hook_event_name", "hookEventName"], None)
+        if event_name is None:
+            raise ValueError("Error: couldn't find event type")
+        event_type = HOOK_NAME_TO_EVENT_TYPE.get(event_name.lower(), EventType.OTHER)
+
+        # Extract the identifier and content based on the event type
+        identifier = ""
```
agateau-gg, line 77: If you add `content = ""` and `tool = None` here then you can significantly simplify the `if`s below by remove a bunch of `content = ""` and `tool = None` lines (and many `else` blocks).

ggshield/verticals/secret/ai_hook/copilot.py
paulpetit-gg-ext: Yes, from a POO perspective that would make sense... But the behavior is not exactly a "sensible default", it's just that they chose to do exactly like Claude.

ggshield/verticals/secret/ai_hook/installation.py
paulpetit-gg-ext: Ok I think I finally understood comment n°1 of the previous review :sweat: 
Yes, this is an error, fixed and a test is added.

ggshield/cmd/secret/scan/ai_hook.py
paulpetit-gg-ext: In that case I suggest to use the same limit than for reading files. Let's say 50M to avoid choking on big (but legitimate) files/data, wdyt?

paulpetit-gg-ext: (Also, note that we don't have such limits for pre-push/pre-receive hooks)

amascia-gg: Looks good to me!

ggshield/verticals/secret/ai_hook/scanner.py
paulpetit-gg-ext: I'm not sure to see a scenario where a user would trigger a post-tool use hook with its AI agent while in a headless environment... But I guess it doesn't hurt to add a try/except anyway :)

ggshield/verticals/secret/ai_hook/installation.py
paulpetit-gg-ext: I confirm it works for Cursor/Claude. Not sure for VSCode which doesn't seem to look for "global" hooks by default, but if it did this looks like the proper folder to put hooks into. Hooks are still a "Preview" feature in VSCode/Copilot.

ggshield/verticals/secret/ai_hook/models.py
paulpetit-gg-ext: Currently if we are not able to identify which code assistant triggered the hook, we fallback to a base instance with this default implementation (returning 2 is a "universal" convention among coding tools, but prevents from having a deeper integration, like surfacing the error to the user, etc.)

So we can't make the class abstract with the current logic.

ggshield/verticals/secret/ai_hook/copilot.py
xblanchot-gg: Ok for me then

agateau-gg (approved):

PR #1185, opened 2026-03-05
fix(syntax): fix pyright errors raised for Python3.9

paulpetit-gg-ext:
## Context

Some previous commits introduced Pyright errors in `main` that somehow weren't caught by our pre-commit hooks or the CI.

This PR fixes them.

## Step to reproduce

```bash
pdm use 3.9
pdm sync
pdm run pre-commit run -a pyright
```

## What has been done

1. `manage.py`: Python>=3.10 annotation style replaced with Python3.9 compatible annotation
2. `loader.py`: change the the type error catch by a version check.
3. `secret_scanner.py` use the `__future__` package to fix the type error and get rid of the version check.

## :warning:  EDIT: CI/pyproject.toml mismatch
(see comment below, detailed here for clarity)

The problem is more general: most of our CI is configured to use Python3.10, but GGShield is still supposed to support Python 3.9 (see `pyproject.toml`). There are indeed jobs that run tests for each supported Python version, but the lint job uses Python 3.10.

I think we should downgrade at least this job to 3.9, otherwise we may again have invalid Python 3.9 code that are merged in main. I will try this in this PR to see how things go.

The only issue I see is that devs with local environments using more recent Python version will start having linting issues until they recreate their environment with Python 3.9.
Conversely, devs that use Python 3.9 in their dev environment shouldn't be blocked by pre-commit hooks that disagree with the CI (which is what started me on this issue).

This may be an opportunity to introduce some tooling (pyenv/mise/...) to force a more uniform DX.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ on:
 
 env:
   PDM_VERSION: 2.26.1
-  DEFAULT_PYTHON_VERSION: '3.10'
+  DEFAULT_PYTHON_VERSION: '3.9'
 
 jobs:
   lint:


alexpasmantier (approved):
lgtm

PR #1190, opened 2026-03-17
fix: prevent duplicate .cache_ggshield entries in .gitignore

### LINKED ISSUE #1189: ggshield appends duplicate .cache_ggshield entries to .gitignore on every run
## Environment

  - ggshield version: 1.48.0
  - Installed via: Homebrew
  - OS: macOS (Darwin 25.3.0, Apple Silicon)
  - Hook setup: global core.hooksPath with pre-commit hook calling ggshield secret scan pre-commit

## Describe the bug

Each time ggshield secret scan pre-commit runs, it appends the following to .gitignore without checking if the entry already exists:

  ```
# Added by ggshield
  .cache_ggshield

  After several commits, .gitignore accumulates duplicate entries:

  # Added by ggshield
  .cache_ggshield

  # Added by ggshield
  .cache_ggshield

  # Added by ggshield
  .cache_ggshield
```

  This causes a real workflow problem: since .gitignore is a tracked file, each ggshield run creates unstaged changes, which blocks
  operations like git checkout or git switch without stashing or discarding.

  

Steps to reproduce:

Install ggshield in this environment.  Issue a git commands that trigger the pre-commit hook

Actual result:

After several commits, .gitignore accumulates duplicate entries:

  ```
# Added by ggshield
  .cache_ggshield

  # Added by ggshield
  .cache_ggshield

  # Added by ggshield
  .cache_ggshield
```

Expected result:

ggshield should check if .cache_ggshield is already present in .gitignore before appending.

If applicable, add logs or screenshots to help explain your problem.

sunilhonest:
## Summary

Fixes #1189

The `gitignore()` function in `ggshield/utils/git_shell.py` blindly appends `.cache_ggshield` to `.gitignore` on every invocation without checking if the entry already exists. This causes duplicate entries to accumulate, which in turn causes pre-commit to fail with "files were modified by this hook".

### Root cause

The `is_gitignored()` check in `cache.py:load_cache()` uses `git check-ignore`, which reports tracked files as **not ignored** even when a matching `.gitignore` pattern exists. So if `.cache_ggshield` was ever accidentally tracked (added to the index), `is_gitignored()` returns `False`, and `gitignore()` appends a duplicate entry every time.

### Fix

Before appending to `.gitignore`, read the file and check if the entry already exists as a line. If it does, skip writing.

### Tests

- `test_no_duplicate_gitignore_entry`: verifies that multiple `Cache()` instantiations don't duplicate the entry
- `test_no_duplicate_gitignore_entry_tracked_file`: reproduces the exact bug — a tracked `.cache_ggshield` file with an existing `.gitignore` entry should not cause duplication

### DIFF
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -516,16 +516,17 @@ def gitignore(path: Path):
     repo_root = _git_rev_parse(option="--show-toplevel", wd=working_dir)
     if repo_root is not None:
         gitignore_path = Path(repo_root) / ".gitignore"
+        posix_path = path.as_posix()
         try:
             # Check if the entry already exists in .gitignore
             if gitignore_path.is_file():
                 existing_content = gitignore_path.read_text(encoding="utf-8")
                 for line in existing_content.splitlines():
-                    if line.strip() == path.as_posix():
+                    if line.strip() == posix_path:
                         return
             with open(gitignore_path, "a") as f:
                 f.write("\n# Added by ggshield\n")
-                f.write(path.as_posix() + "\n")
+                f.write(posix_path + "\n")
         except OSError:
             logger.debug(
                 "Failed to add %s to .gitignore in %s",


ggshield/utils/git_shell.py
```diff
@@ -515,8 +515,15 @@ def gitignore(path: Path):
...
     repo_root = _git_rev_parse(option="--show-toplevel", wd=working_dir)
     if repo_root is not None:
+        gitignore_path = Path(repo_root) / ".gitignore"
         try:
-            with open(Path(repo_root) / ".gitignore", "a") as f:
+            # Check if the entry already exists in .gitignore
+            if gitignore_path.is_file():
+                existing_content = gitignore_path.read_text(encoding="utf-8")
+                for line in existing_content.splitlines():
+                    if line.strip() == path.as_posix():
```
6d7a: nit: could you move the `as_posix` call out of the for loop to avoid calling it for every line?

PR #1192, opened 2026-03-19
[AI Hook] Read files that may be omitted by Claude

paulpetit-gg-ext:
### Claude "misbehavior"
While dogfooding the new AI hooks, we noticed that Claude Code doesn't always trigger the PreToolUse hook to read files if they were included in the prompt with the "@filepath" convention.

The only way we found to still be able to block the model from reading the file content (and possibly leaking a secret), is to parse these filepaths and read it beforehand.

The only "architectural" change is that as a consequence, parsing a hook input can yield multiple payloads to scan. I cascaded a few changes to keep the code clean. The rest is mostly test cases for the utility function that parses user prompts.

The AI hook command has not been released yet, so I skip the changelog for this PR.

### Better file management
A second commit also improves consistency in file scanning by reusing utils and classes already made for scanning a given file.

## PR check list

### DIFF
--- a/ggshield/verticals/secret/ai_hook/scanner.py
+++ b/ggshield/verticals/secret/ai_hook/scanner.py
@@ -1,13 +1,11 @@
 import hashlib
 import json
 import re
-from pathlib import Path
 from typing import Any, Dict, List, Sequence, Set
 
 from notifypy import Notify
 
 from ggshield.core.filter import censor_match
-from ggshield.core.scan import StringScannable
 from ggshield.core.scanner_ui import create_message_only_scanner_ui
 from ggshield.core.text_utils import pluralize, translate_validity
 from ggshield.verticals.secret import SecretScanner
@@ -16,7 +14,7 @@ from ggshield.verticals.secret.secret_scan_collection import Secret
 
 from .claude_code import Claude
 from .cursor import Cursor
-from .models import MAX_READ_SIZE, EventType, Flavor, Payload, Result, Tool
+from .models import EventType, Flavor, Payload, Result, Tool
 
 
 # Regex (and method) to look for any @file_path in the prompt.
@@ -24,7 +22,7 @@ from .models import MAX_READ_SIZE, EventType, Flavor, Payload, Result, Tool
 _FILE_PATH_REGEX = re.compile(
     r'@"((?:[^"\\]|\\.)*)"'  # quoted: @"..."
     r"|"
-    r"(?:\W|^)@([\w/\\\-\.]+)",  # unquoted: @path
+    r"(?:\W|^)@([\w/\\.-]+)",  # unquoted: @path
     re.MULTILINE,
 )
 
@@ -77,7 +75,7 @@ class AIHookScanner:
 
         return payload.flavor.output_result(result)
 
-    def _parse_input(self, content: str) -> list[Payload]:
+    def _parse_input(self, raw_content: str) -> list[Payload]:
         """Parse the input content. Raises a ValueError if the input is not valid.
 
         Returns:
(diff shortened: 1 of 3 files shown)


ggshield/verticals/secret/ai_hook/scanner.py
```diff
@@ -18,6 +19,30 @@
 from .models import MAX_READ_SIZE, EventType, Flavor, Payload, Result, Tool
 
 
+# Regex (and method) to look for any @file_path in the prompt.
+# A list of test cases can be found in test_hooks.py.
+_FILE_PATH_REGEX = re.compile(
+    r'@"((?:[^"\\]|\\.)*)"'  # quoted: @"..."
+    r"|"
+    r"(?:\W|^)@([\w/\\\-\.]+)",  # unquoted: @path
```
agateau-gg: nit: can be simplified to `r"(?:\W|^)@([\w/\\.-]+)"` (no need to escape `.` inside `[]`, move `-` at the end to remove the need to escape it)

ggshield/verticals/secret/ai_hook/scanner.py
```diff
@@ -110,71 +136,103 @@ def _parse_input(self, content: str) -> Payload:
...
+        # unnecessarily scan files multiple times.
+        if isinstance(flavor, Claude):
+            # match multiple @file_path in the prompt
+            matches = find_filepaths(content)
+            for match in matches:
+                payloads.append(
+                    Payload(
+                        event_type=EventType.PRE_TOOL_USE,
+                        tool=Tool.READ,
+                        content="",
```
agateau-gg, line 229: Why is content empty here?

ggshield/verticals/secret/ai_hook/scanner.py
```diff
@@ -91,14 +123,8 @@ def _parse_input(self, content: str) -> Payload:
                 content = tool_input.get("command", "")
                 identifier = content
             elif tool == Tool.READ:
+                # We only need to deal with the identifier, the content will be read by the Scannable
                 identifier = lookup(tool_input, ["file_path", "filePath"], "")
-                # Read the file before the AI tool
-                file = Path(identifier)
-                if file.is_file() and file.stat().st_size <= MAX_READ_SIZE:
```
agateau-gg, line 97: MAX_READ_SIZE is no longer checked. If it's not relevant anymore, you can remove the constant.

paulpetit-gg-ext: It still limits the size of stdin (last use of this constant).

ggshield/verticals/secret/ai_hook/scanner.py
paulpetit-gg-ext: we no longer need to read the content ourselves, as the `.scannable` property will return a `File`.

`Payload.content` could have been changed into an `Optional[str]` but I figured it would have lead to even more changes in this PR.

ggshield/verticals/secret/ai_hook/scanner.py
```diff
@@ -15,7 +14,31 @@
...
 from .claude_code import Claude
 from .cursor import Cursor
-from .models import MAX_READ_SIZE, EventType, Flavor, Payload, Result, Tool
+from .models import EventType, Flavor, Payload, Result, Tool
+
+
+# Regex (and method) to look for any @file_path in the prompt.
+# A list of test cases can be found in test_hooks.py.
+_FILE_PATH_REGEX = re.compile(
+    r'@"((?:[^"\\]|\\.)*)"'  # quoted: @"..."
```
xblanchot-gg, line 23: We should check that there is the beginning or a space before the @ no ? Otherwise this will match emails for example

paulpetit-gg-ext: Here I thing it is superfluous, because it matches a string with a `"` after the `@`, which can't be an email. But if you think it would be better to have the same rule before the `@` I can add it.

xblanchot-gg: Indeed

agateau-gg (approved):

PR #1193, opened 2026-03-24
Improve ai hooks

paulpetit-gg-ext:
## Context

This MR brings two improvements  on the `ggshield secret scan ai-hook` command:
- Hooks response are more customized depending on the detected code assistant and hook: this brings slightly better integrated messages in some cases, and more future-proof response in others.
- I ended up getting the same "bug" fixed by #1192 with Cursor. So a second commit also reads "@files" regardless of the code assistant. This can only help reduce false negatives, at the expense of a bit more traffic with GIM (some files will be read twice).
- Command help message is more helpful.

## PR check list

### DIFF
--- a/ggshield/cmd/secret/scan/ai_hook.py
+++ b/ggshield/cmd/secret/scan/ai_hook.py
@@ -25,8 +25,13 @@ def ai_hook_cmd(
     """
     Scan AI tool interactions for secrets.
 
-    Reads a hook event from stdin as JSON, processes it based on the
-    event type and mode, and outputs the response to stdout as JSON.
+    Reads a hook event from stdin as JSON, processes it,
+    and outputs a response to stdout as JSON that blocks the action
+    if secrets are detected.
+
+    To install the hook for your AI coding tool, run:
+
+    `ggshield install -t <your-code-assistant> -m [local|global]`
     """
     ctx_obj = ContextObj.get(ctx)
     config = ctx_obj.config
--- a/ggshield/verticals/secret/ai_hook/scanner.py
+++ b/ggshield/verticals/secret/ai_hook/scanner.py
@@ -257,9 +257,7 @@ class AIHookScanner:
                 f"  - {secret.detector_display_name} ({validity}): {match_str}"
             )
 
-        if payload.event_type == EventType.USER_PROMPT:
-            message = "Please remove the secrets from your prompt before submitting."
-        elif payload.tool == Tool.BASH:
+        if payload.tool == Tool.BASH:
             if payload.event_type == EventType.POST_TOOL_USE:
                 message = "Secrets detected in the command output."
             else:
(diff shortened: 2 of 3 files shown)


ggshield/cmd/secret/scan/ai_hook.py
```diff
@@ -25,8 +25,13 @@ def ai_hook_cmd(
...
 
-    Reads a hook event from stdin as JSON, processes it based on the
-    event type and mode, and outputs the response to stdout as JSON.
+    Reads a hook event from stdin as JSON, processes it,
+    and outputs a response to stdout as JSON that blocks the action
+    if secrets are detected.
+
+    To install the hook for your AI coding tool, run:
+
+    `ggshield install -t [cursor|claude-code|copilot] -m [local|global]`
```
agateau-gg: I would not list the possible tools here, because it's easy to forget to update this list when a new tool is added.

ggshield/verticals/secret/ai_hook/claude_code.py
```diff
@@ -13,10 +13,29 @@ class Claude(Flavor):
...
+            elif result.payload.event_type == EventType.PRE_TOOL_USE:
+                response["hookSpecificOutput"] = {
+                    "hookEventName": "PreToolUse",
+                    "permissionDecision": "deny",
+                    "permissionDecisionReason": result.message,
+                }
+            else:
+                # Should not happen, but just in case use Claude's "universal" fields.
+                response = {
+                    "continue": not result.block,
```
agateau-gg: `result.block` is always True, so this can be `"continue": False`.

ggshield/verticals/secret/ai_hook/claude_code.py
```diff
@@ -13,10 +13,29 @@ class Claude(Flavor):
...
-        }
+        response = {}
+        if result.block:
+            if result.payload.event_type in [
+                EventType.USER_PROMPT,
+                EventType.POST_TOOL_USE,
+            ]:
+                response["decision"] = "block"
+                response["reason"] = result.message
+                response["additionalContext"] = result.message
```
agateau-gg, line 24: `reason` and `additionalContext` are the same. Is it on purpose?

ggshield/verticals/secret/ai_hook/copilot.py
```diff
@@ -17,12 +16,22 @@ class Copilot(Claude):
...
+                response["hookSpecificOutput"] = {
+                    "hookEventName": "PreToolUse",
+                    "permissionDecision": "deny",
+                    "permissionDecisionReason": result.message,
+                }
+            elif result.payload.event_type == EventType.POST_TOOL_USE:
+                response["decision"] = "block"
+                response["reason"] = result.message
+            else:
+                response["continue"] = not result.block
```
agateau-gg: Same thing here: `response["continue"] = False`.

raw_hooks.jsonl
agateau-gg, line 1: Was this file supposed to be committed? I don't understand what it is used for.

ggshield/verticals/secret/ai_hook/claude_code.py
paulpetit-gg-ext: yes, `reason` is shown to the user, `additionalContext` is for the agent (if the doc is reliable).

PR #1195, opened 2026-03-26
[Refactoring] Create a new "AI" vertical

paulpetit-gg-ext:
## Context

Because we will add new AI governance capabilities to GIM, this PR is a refactoring that extracts what has been done for AI hooks into a new `ai` vertical that will be expanded upon later (new commands will arrive in the `ai` vertical, and hooks won't only serve to scan secrets).

It is quite a lot of change, but this is 99% moving code from a place to another and fixing imports.
I think it is worth having it in a distinct PR so that later work is more easily reviewed.

There is only one notable change that is not purely code shuffling: `Flavor` has been renamed to the less cryptic `Agent` and has become an abstract class. The rationale is to keep everything specific to an AI agent (Cursor, Claude Code, etc.) into a single class/file even when the number of AI governance features will grow.

Note that we still need at some point to be able to reuse both logic from the scan vertical (to scan secrets) and from the new ai vertical (for hooks, payload model, etc.).
Instead of adding coupling between verticals (ai -> scan or scan -> ai), this PR chose to have commands import what they need from each vertical in order to work. This is what I think maintains the clearer boundary between modules.

## PR check list


ggshield/verticals/ai/installation.py
```diff
@@ -89,11 +80,11 @@ def install_hooks(
     # Report what happened
     styled_path = click.style(settings_path, fg="yellow", bold=True)
     if stats.added == 0 and stats.already_present > 0:
-        click.echo(f"{flavor.name} hooks already installed in {styled_path}")
+        click.echo(f"{agent.name} hooks already installed in {styled_path}")
```
6d7a: We should use `agent.display_name` here.

ggshield/verticals/secret/ai_hook.py
6d7a, line 1: The coupling between the `ai` and `secret` vertical doesn't feel right to me. I wonder if we rather should move `ggshield/verticals/secret/secret_scan_collection.py` and `ggshield/verticals/secret/secret_scanner.py` into `ggshield/core/scan` and this file into the `ai` vertical. That way we could ensure a clean separation between verticals.

6d7a:
```suggestion
    we instead have a single class which detects which protocol to use (called "agent").
```

6d7a: I was talking to Aurélien about this and he had some interesting insights: The verticals represent GitGuardian business verticals (although `auth` is a pretty wide stretch for a vertical 😅). Therefore, the AI hooks' logic should reside in the `secret` vertical.

paulpetit-gg-ext: Interesting. I suppose leaving the hook part in `secret/` makes sense, but there are still a few questions:
- Adding an `ggshield ai` subcommand is still relevant for us (for instance we are developing a `ggshield ai discover` command that lists agents and MCP servers configured on a developer's machine and which wouldn't make much sense under `ggshield secret`). Does it mean we can create the vertical or should we have a new command subgroup but not a new `vertical/` subfolder?
- In case we have part of the new AI governance code under a specific folder, it means two things for the hook code:
  - either it imports model from it (coupling)
  - either everything useful for hooks stays under `secret/` but we may have code duplication, and the logic specific to a given agent like Cursor is split between multiple folders.

I think in this case some form of coupling may be warranted. What do you think?

paulpetit-gg-ext: After discussion with @agateau-gg , I pushed a new solution: both verticals are now independent, and the commands are able (with documented exceptions in the import linter) to pull the logic they need from both verticals (`scan` and `ai`).

6d7a: Sounds good to me 👍

agateau-gg (approved):

PR #1203, opened 2026-04-01
feat(auth): use keyring to store gitguardian api token

6d7a:
## Context

`ggshield` used to store the auth token in cleartext in the global YAML config.

## What has been done

This commit adds support for using the system keyring. If not available, the previous cleartext storage will be used as a fallback.

## Validation
- check out `main`, sync and run `uv run ggshield auth login` if you are not logged in with `ggshield`, yet
- inspect `ggshield`'s global `auth_config.yaml` - the token is stored as a clear-text value
- run `uv run ggshield api-status` to make sure your token is valid and used
- check out this branch, sync and run `uv run ggshield auth login`
- inspect `auth_config.yaml`, the clear-text value should have been replaced with a `__KEYRING__` sentinel
- run `uv run ggshield api-status` to make sure your token is accessible and valid

## PR check list

### DIFF
--- a/ggshield/core/config/auth_config.py
+++ b/ggshield/core/config/auth_config.py
@@ -1,3 +1,4 @@
+import logging
 from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
@@ -6,6 +7,11 @@ from typing import Any, Dict, List, Optional, cast
 import marshmallow_dataclass
 from pygitguardian.models_utils import FromDictMixin, ToDictMixin
 
+from ggshield.core.config.token_store import (
+    KEYRING_SENTINEL,
+    TokenStore,
+    get_token_store,
+)
 from ggshield.core.config.utils import (
     get_auth_config_filepath,
     load_yaml_dict,
@@ -18,6 +24,9 @@ from ggshield.core.errors import (
     UnknownInstanceError,
 )
 from ggshield.utils.datetime import datetime_from_isoformat
+
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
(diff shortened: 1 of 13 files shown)


ggshield/core/config/auth_config.py
```diff
@@ -168,9 +217,73 @@ def get_instance_token(self, instance_name: str) -> str:
...
+            return
+        try:
+            store.store_token(inst.url, inst.account.token)
+            logger.info(
+                "Token for %s migrated to OS credential store",
+                inst.url,
+            )
+        except Exception:
+            logger.warning(
+                "Failed to store token in keyring for %s, " "storing in config file",
```
agateau-gg: nit: odd formatting
```suggestion
                "Failed to store token in keyring for %s, storing in config file",
```

ggshield/core/config/token_store.py
```diff
@@ -0,0 +1,127 @@
...
+_token_store: Optional[TokenStore] = None
+
+
+def get_token_store() -> TokenStore:
+    """Return the active token store, selecting keyring when available."""
+    global _token_store
+    if _token_store is not None:
+        return _token_store
+
+    if os.environ.get("GGSHIELD_NO_KEYRING", "").lower() in ("1", "true", "yes"):
```
agateau-gg: You should use `ggshield.os.getenv_bool()` instead of implementing it by hand.

agateau-gg (changes requested):
Minor remarks, but I have a few deeper comments:

- ~I am not sure I like the command transparently moving the secrets without me asking for it to be migrated. I can see some customers being surprised / annoyed by it.~ Nevermind, I realized I had to logout and log back in for it to work. That's OK (the only surprising part is that it converted all my config entries, not just the one I logged out from, but it sounds OK)

- `ggshield api-status` prints the API key source:
```
$ uv run ggshield api-status
API URL: https://dashboard.gitguardian.com
Status: healthy
App version: v2.429.0
Secrets engine version: 2.160.1

Instance source: default
API key source: user config
Token scopes: scan
```

It would be nice to use a different value for `API key source` (but note that could require an update in the VSCode extension)

6d7a:
@agateau-gg I created a new key source for keyring-based configs. [This](https://github.com/GitGuardian/gitguardian-vscode/pull/126) is the companion PR for the extension.

agateau-gg (approved):

PR #1205, opened 2026-04-02
MCP 3/3 - add MCP usage monitoring

paulpetit-gg-ext:
## Context

See Linear issue https://linear.app/gitguardian/issue/NHI-1513/mcp-discovery-and-usage-in-ggshield

## What has been done

- The `ggshield secret scan ai-hook` now additionally send the activity of MCP PreToolUse hooks to a specific GIM endpoint. In addition to logging the activity, we receive whether we should block the call or not.
- The non-trivial part is building a unified payload, more precisely getting the server and tool used. As usual, most of the heavy-lifting is found in the specific `Agent`s implementations. 

## Validation

- Install AI hooks (`ggshield install -t {cursor|claude-code|copilot}`)
- Have your agent use an MCP tool
- See the activity in GIM.

## PR check list

### DIFF
--- a/changelog.d/20260421_191828_paul.petit.ext_HEAD.md
+++ b/changelog.d/20260421_191828_paul.petit.ext_HEAD.md
@@ -1,0 +1,42 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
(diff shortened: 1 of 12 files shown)


xblanchot-gg (approved):
Peer reviewed and validated ! ✔️

PR #1208, opened 2026-04-10
6d7a/skip platform specific api tests

6d7a:
## Context

Currently, functional tests that make requests to gitguardian's API are run for every platform. This leads to rate-limiting issues.

## What has been done

This PR introduces a differentiation between functests that use the API and those that do not. The tests that use the API are run only once in CI (on a linux runner). The platform-specific runs only execute the functests that do not use the API.

## Validation

- Check that CI only runs the functional tests that do not use the API in platform-specific runs
- Run `make functest` locally to ensure the command behaves as expected/before

## PR check list


.github/workflows/ci.yml
```diff
@@ -127,13 +127,39 @@ jobs:
...
+
+  functest_api:
+    name: Functional tests (API)
+    # See note about steps requiring the GITGUARDIAN_API at the top of this file
+    if: ${{ !github.event.pull_request.head.repo.fork }}
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 10
```
alexpasmantier, line 146: any particular reason for this?

alexpasmantier (approved):
Very nice, this should greatly reduce flakyness 🥳 
A minor question, pre-approving 👍🏻

.github/workflows/ci.yml
6d7a: I should add the comment from above here, too:
`# Get enough commits to run `ggshield secret scan commit-range` on ourselves`

.github/workflows/ci.yml
clement-tourriere, line 201: Do we need to add in the `needs` list the `functest_api` ? Without that we could publish docker image with failing tests

clement-tourriere (commented):
Also in `tests/functional/secret/test_scan_invalid_api_key.py` , we are also calling the GitGuardian API via the SecretScanner => `secret scan path -> SecretScanner(...) -> check_client_api_key() -> client.read_metadata()`.

Maybe it's ok, but with the matrix of all the tests, we could want to set the new flag !

PR #1209, opened 2026-04-15
ci: mock get API calls

6d7a:
## Context

The test mock of the GitGuardian API forwards GET requests implicitly to the public API. This can cause rate-limiting issues on many concurrent CI runs.

## What has been done

This commit introduces mocks for the GET endpoints used in tests. As far as I can see from git logs, we used to forward get requests to the public API in order to prevent having to update the mock frequently. Since the GET requests `ggshield` barely changed, we may assume that breaking the mock is less of a concern than rate-limiting. This PR also requires functional tests to succeed in order to push the unstable docker image.

## PR check list

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -1,16 +1,14 @@
 import abc
 import http.server
+import json
 import shutil
 import socketserver
 import time
 from multiprocessing import Event, Process, Value
 from pathlib import Path
 from typing import Generator
-from urllib.parse import urlparse
 
 import pytest
-import requests
-from pygitguardian.config import DEFAULT_BASE_URI
 
 from tests.repository import Repository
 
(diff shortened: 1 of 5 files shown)


tests/functional/conftest.py
```diff
@@ -36,32 +34,57 @@
...
 
-        self.wfile.write(response.content)
+        self.send_response(200)
+        self.send_header("content-type", "application/json")
+        self.send_header("Content-Length", str(len(content)))
+        self.end_headers()
+        self.wfile.write(content)
 
     @abc.abstractmethod
     def do_POST(self):
```
alexpasmantier, line 90: the `responses` package might come in handy for these kinds of tests

https://github.com/getsentry/responses

alexpasmantier (approved):
A suggestion to reduce boilerplate, otherwise LGTM 👍🏻

tests/functional/conftest.py
6d7a: Cool! I'll keep it in mind.

PR #1210, opened 2026-04-16
feat: do not block on connection or server error

6d7a:
## Context

Currently, `ggshield` blocks in pre-push, pre-commit, and pre-receive hooks as well as in CI if the gitguardian server does not respond. In pre-push and pre-commit hooks and CI `ggshield` also blocks if the server reponds with an internal server error. 
In order to reduce dev friction, ~we should not block in any of the cases above neither on server error responses or connection errors.~ we should introduce a parameter to configure the failure mode. 

## What has been done

This PR adds a new `secret.fail_on_server_error` configuration option (default `True`), available as the `--fail-on-server-error/--no-fail-on-server-error` flag or `GITGUARDIAN_FAIL_ON_SERVER_ERROR` environment variable. When set to `False`, `secret scan pre-commit`, `secret scan pre-push`, `secret scan pre-receive`, and `secret scan ci` exit with code `0` and display a warning instead of blocking the git operation when the GitGuardian server is unreachable or returns a 5xx response. The default preserves the previous blocking behavior.

## Validation

- make sure your command-line environment variables aren't overridden by another source
- `GITGUARDIAN_FAIL_ON_SERVER_ERROR=False GITGUARDIAN_API_URL=https://localhost:1 GITGUARDIAN_API_KEY=fake uv run ggshield -v secret scan pre-commit` should return exit code 0
- `GITGUARDIAN_API_URL=https://localhost:1 GITGUARDIAN_API_KEY=fake uv run ggshield -v secret scan pre-commit` should return exit code 127

## PR check list

### DIFF
--- a/ggshield/cmd/secret/scan/prereceive.py
+++ b/ggshield/cmd/secret/scan/prereceive.py
@@ -12,6 +12,7 @@ from ggshield.cmd.secret.scan.secret_scan_common_options import (
     add_secret_scan_common_options,
     create_output_handler,
 )
+from ggshield.cmd.utils.common_decorators import fail_on_server_error_option
 from ggshield.cmd.utils.context_obj import ContextObj
 from ggshield.core import ui
 from ggshield.core.cache import ReadOnlyCache
@@ -79,6 +80,7 @@ def _execute_prereceive(
     hidden=True,
 )
 @add_secret_scan_common_options()
+@fail_on_server_error_option
 @click.pass_context
 def prereceive_cmd(
     ctx: click.Context, web: bool, prereceive_args: List[str], **kwargs: Any
@@ -142,6 +144,9 @@ def prereceive_cmd(
         process.kill()
         return 0
     if process.exitcode == ExitCode.GITGUARDIAN_SERVER_UNAVAILABLE:
+        if config.user_config.secret.fail_on_server_error:
+            ui.display_error("\nGitGuardian server is not responding.")
+            return ExitCode.GITGUARDIAN_SERVER_UNAVAILABLE
         ui.display_error("\nGitGuardian server is not responding. Skipping checks.")
         return 0
 
--- a/changelog.d/20260416_100952_6d7a_do_not_block_on_connection_or_server_error.md
+++ b/changelog.d/20260416_100952_6d7a_do_not_block_on_connection_or_server_error.md
@@ -1,0 +1,3 @@
+### Changed
+
+- `secret scan pre-commit`, `secret scan pre-push`, `secret scan ci`, and `secret scan pre-receive` no longer block git operations when the GitGuardian server is unreachable or returns a server error. A warning is displayed and the command exits with code 0.
--- a/changelog.d/20260416_100952_6d7a_do_not_block_on_connection_or_server_error.md
+++ b/changelog.d/20260416_100952_6d7a_do_not_block_on_connection_or_server_error.md
@@ -1,3 +1,7 @@
+### Added
+
+- Added a new `secret.fail_on_server_error` configuration option (default `True`), available as the `--fail-on-server-error/--no-fail-on-server-error` flag or `GITGUARDIAN_FAIL_ON_SERVER_ERROR` environment variable. When set to `False`, `secret scan pre-commit`, `secret scan pre-push`, `secret scan pre-receive`, and `secret scan ci` exit with code `0` and display a warning instead of blocking the git operation when the GitGuardian server is unreachable or returns a 5xx response. The default preserves the previous blocking behavior.
+
 ### Changed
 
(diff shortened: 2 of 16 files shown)


clement-tourriere:
What bothers me is that we're changing failure semantics on a security tool for everyone, without a way to opt out. On upgrade, teams that rely on secret scan ci to actually block a leaked secret will silently
  start fail-opening every time the server blips. That feels less like a UX fix and more like a policy decision we're making on their behalf.

  The three hooks also aren't really the same shape:
  - pre-commit / pre-push — fail-open is fine here, devs can --no-verify anyway and there's a server gate behind them
  - secret scan ci — this is the gate in most pipelines. A transient 5xx or a flaky proxy and the secret ships through with just a warning in the logs
  - pre-receive — you already did the right thing (connection errors only, not 5xx). Same reasoning kind of applies to CI

  Could we put this behind a config flag instead? exit_zero is a clean precedent — field on UserConfig, CLI in common_options.py:174-186, GITGUARDIAN_EXIT_ZERO env var, YAML entry, all wired through
  create_config_callback. Same shape would work here as secret.fail_on_server_error (default True). Decorator reads it before swallowing; pre-receive can read the env var directly from its subprocess like it already
   does with GITGUARDIAN_TIMEOUT. That way today's behavior holds on upgrade and fail-open becomes a deliberate choice — and we can always flip the default later if support data shows nobody wants strict.

  A few smaller things while I'm here:

  - We lose the exit code signal. ExitCode.GITGUARDIAN_SERVER_UNAVAILABLE = 4 already exists (errors.py:23-43) and pre-receive uses it internally. Collapsing to 0 means CI monitoring can't tell "clean scan" from
  "skipped because server was down" — could be weeks before anyone notices scans haven't actually run. Either keep a distinct code or at least emit something structured that scrapers can catch.
  - No negative test. All three new tests mock ServiceUnavailableError and assert 0. None check that a 4xx (401, quota reached) still fails. If someone later widens the caught type, CI won't catch it.
  - Adjacent case falls through. handle_api_error with status_code=None (errors.py:217-218) raises UnexpectedError, not ServiceUnavailableError — so a proxy that returns an HTML error page still crashes the hook.
  Worth picking a side: either any infra weirdness is non-blocking, or only real 5xx is.
  - display_warning vs display_error. Pre-receive uses display_error for the same condition (prereceive.py:144-146); the new decorator uses display_warning. Minor, but worth aligning so all four hooks log
  consistently.

6d7a:
@clement-tourriere I agree. I made the failure mode in hooks configurable, with failure on 5xx and connection errors as the default. This way, customers that want to reduce dev friction as much as possible have a way to do so, while avoiding any unexpected changes in the hook behavior for all other users that stick with the default.

ggshield/cmd/secret/scan/prereceive.py
```diff
@@ -142,6 +142,9 @@ def prereceive_cmd(
         process.kill()
         return 0
     if process.exitcode == ExitCode.GITGUARDIAN_SERVER_UNAVAILABLE:
+        if config.user_config.secret.fail_on_server_error:
```
clement-tourriere, line 147: Pre-receive default flips from fail-open to fail-closed. On main, prereceive.py:144-146 unconditionally returns 0 on server unavailable. This PR makes it return ExitCode.GITGUARDIAN_SERVER_UNAVAILABLE by
  default. The changelog says "the default preserves the previous blocking behavior" — true for ci/pre-commit/pre-push, but the opposite for pre-receive. Either default pre-receive to fail_on_server_error=False to
  preserve existing behavior, or add a prominent "Breaking" note to the changelog.

ggshield/cmd/secret/scan/secret_scan_common_options.py
```diff
@@ -185,6 +202,7 @@ def decorator(cmd: AnyFunction) -> AnyFunction:
         instance_option(cmd)
         _all_secrets(cmd)
         _source_uuid_option(cmd)
+        _fail_on_server_error_option(cmd)
```
clement-tourriere: --fail-on-server-error leaks to 11 commands where it's a no-op. _fail_on_server_error_option is wired in add_secret_scan_common_options() (secret_scan_common_options.py:202), so it's exposed on path, repo,
  docker, dockerarchive, archive, pypi, docset, changes, range, ai_hook, and the scan group — none of which apply @non_blocking_server_error. A user running ggshield secret scan path --no-fail-on-server-error will
  think they're fail-opening and still get blocked. Suggest a dedicated decorator that bundles the option + the handler and apply it only on the 4 hook/CI commands.

clement-tourriere (commented):
handle_api_error with status_code is None still raises UnexpectedError (errors.py:217-218), which the decorator doesn't catch. A proxy returning malformed HTML still blocks even with --no-fail-on-server-error.
  One-line fix: raise ServiceUnavailableError there too. Naturally pairs with `Suggest a dedicated decorator that bundles the option + the handler and apply it only on the 4 hook/CI commands.` if you consolidate into a single decorator.

clement-tourriere (approved):

PR #1213, opened 2026-04-16
MCP 2/3 - new `ai discover` command

paulpetit-gg-ext:
## Context

This PR is part of the new "Agentic AI" stack of PRs and introduces the `ggshield ai discover` command, which is the biggest part of the stack.

## What has been done

The PR is split in 7 commits:

+ The first adds a few tests to the existing model to increase coverage.
+ The second is the main one and adds the new `ai` group and the `ai discovery` command.
+ The "capability discovery" (detecting MCP tools/resources/prompts) is added (Cursor only).
+ The user info is filled using the same logic than our machine scan plugin (should be merged in the future).
+ The result of the discovery is cached in a local file to avoid unnecessary roundtrips with GIM.
+ The raw output of the command is replaced with a proper summary.
+ Changelog.

## Validation

- Clone repository
- cd into `repo`.
- run `ggshield ai discover`, it should show you all the MCP servers and configurations detected.

## PR check list

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -9,7 +9,7 @@ name = ggshield-layers
 type = layers
 layers = 
 	ggshield.__main__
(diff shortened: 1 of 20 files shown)


xblanchot-gg (approved):
Peer validated, all good for me !

PR #1216, opened 2026-04-23
feat: cache auth check api calls for `secret scan`s

6d7a:
## Context

`check_client_api_key` hits `/v1/metadata` and `/v1/api_tokens/self` on every invocation. When a caller runs bursty scans — notably the GitGuardian VSCode extension, which rescans on every save — this is two API round-trips per scan even though the key state rarely changes between them.

## What has been done

This PR caches a successful auth check on disk for a short TTL (300s) keyed on `(instance_url, api_key_hash)`. While the entry is fresh, `check_client_api_key` returns without contacting the API, provided the cached scopes cover the required ones. A cache entry seeded without scopes (e.g. from `auth login`) still skips `/v1/metadata` but falls through to `/v1/api_tokens/self` when the caller needs scopes.

Trade-off: a server-side key revocation may not be observed for up to the TTL. This is self-healing — the next API call returns 401, which runs through `handle_api_error` and invalidates the cache.

## Validation

- create a test file to scan `echo "my secret" > /tmp/scan-me.txt`
- run `uv run ggshield --debug secret scan path /tmp/scan-me.txt` twice in succession and confirm the second invocation does not log the metadata/api_tokens round-trips

## PR check list

### DIFF
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -10,7 +10,7 @@ from pygitguardian.models import APITokensResponse, Detail, TokenScope
 from requests import Session
 from requests.adapters import HTTPAdapter
 
-from . import ui
+from . import auth_check_cache, ui
 from .config import Config
 from .constants import DEFAULT_INSTANCE_URL
 from .errors import (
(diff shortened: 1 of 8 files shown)


ggshield/core/client.py
```diff
@@ -117,32 +117,52 @@ def check_client_api_key(client: GGClient, required_scopes: set[TokenScope]) ->
...
+        client.secrets_engine_version = cached.secrets_engine_version
+    if cached is not None and (
+        not required_scopes
+        or (cached.scopes is not None and required_scopes <= cached.scopes)
+    ):
+        return
+
+    # The cache either missed or does not prove the scopes we need. We still get to
+    # skip /v1/metadata when the cache told us the key was recently verified.
+    if cached is None or not cached.metadata_verified:
```
benjaminrigaud-gg: The branch `not cached.metadata_verified` is unreachable

benjaminrigaud-gg (dismissed):

clement-tourriere (approved):

benjaminrigaud-gg (approved):
QA looking good.

PR #1219, opened 2026-04-28
feat(auth): add --no-browser flag for browser-less OAuth login

timothedelion:
## Summary

Adds `ggshield auth login --no-browser` for environments without a browser (SSH sessions, headless servers).

When the flag is set, ggshield:
1. Builds the authorize URL with `redirect_uri=urn:ietf:wg:oauth:2.0:oob` (the standard OAuth out-of-band sentinel).
2. Prints the URL to stderr — the user opens it on another device.
3. Reads the displayed code from stdin via `click.prompt`.
4. Exchanges it at `/v1/oauth/token` with the same OOB redirect_uri and the existing `code_verifier`.

PKCE still binds the exchange to the local CLI. The localhost listener is not started in OOB mode. State validation is intentionally absent (no redirect to inspect) — documented inline.

Server side: MR in GIM : https://gitlab.gitguardian.ovh/gg-code/prm/ward-runs-app/-/merge_requests/24425

image
image
image

Loom video : https://www.loom.com/share/4b358456740b4905a16ee9d2af2a8c6f

## Test plan

- [x] New unit tests cover: URL contains the sentinel + no browser opened + no listener started, POST body has the sentinel, code-stripping, empty-code error, 4xx surfacing, and `--no-browser` rejection with `--method=token`.
- [x] All pre-existing auth tests still pass (358/358).
- [x] `black --check` / `isort --check` clean on changed files.
- [ ] Manual smoke against a backend with OOB support.

Refs: SI-2994

### DIFF
--- a/ggshield/verticals/auth/oauth.py
+++ b/ggshield/verticals/auth/oauth.py
@@ -28,6 +28,13 @@ from ggshield.utils.datetime import get_pretty_date
 CLIENT_ID = "ggshield_oauth"
 SCAN_SCOPE = "scan"
 
+# Sentinel `redirect_uri` value used by the out-of-band (browser-less) OAuth
+# flow. Mirrors gcloud's `--no-launch-browser` and the historical Google
+# `urn:ietf:wg:oauth:2.0:oob` pattern. When the CLI sends this value, the
+# backend renders a page displaying the authorization code as text instead of
+# redirecting to localhost; the user pastes the code back into the terminal.
+OOB_REDIRECT_URI = "urn:ietf:wg:oauth:2.0:oob"
+
 # potential port range to be used to run local server
 # to handle authorization code callback
 # this is the largest band of not commonly occupied ports
@@ -74,6 +81,12 @@ class OAuthClient:
         self._port = USABLE_PORT_RANGE[0]
         self.server: Optional[HTTPServer] = None
 
+        # When True, run the browser-less (out-of-band) flow: the CLI prints
+        # the authorize URL and prompts the user to paste back the code shown
+        # by the dashboard, instead of opening a browser and listening on
+        # localhost.
+        self._no_browser: bool = False
+
         self._generate_pkce_pair()
 
     def oauth_process(
@@ -82,11 +95,19 @@ class OAuthClient:
         lifetime: Optional[int] = None,
         login_path: Optional[str] = None,
         extra_scopes: Optional[List[str]] = None,
+        no_browser: bool = False,
     ) -> None:
         """
-        Handle the whole oauth process which includes
-        - opening the user's webbrowser to GitGuardian login page
-        - open a server and wait for the callback processing
+        Handle the whole oauth process.
+
+        Two modes are supported:
+        - localhost flow (default): open the user's webbrowser to
+          GitGuardian login page and wait for the OAuth callback on a
+          temporary local HTTP server.
+        - out-of-band flow (``no_browser=True``): print the authorize URL,
+          ask the user to open it and paste back the code shown by the
+          dashboard. No browser is opened and no localhost listener is
+          started.
         """
         # enable redirection to http://localhost
         os.environ["OAUTHLIB_INSECURE_TRANSPORT"] = str(True)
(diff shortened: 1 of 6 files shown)


ggshield/verticals/auth/oauth.py
```diff
@@ -104,11 +139,65 @@ def oauth_process(
...
+            raise
+        if not authorization_code:
+            raise UnexpectedError("No authorization code was provided.")
+        click.echo(f"Authorization code: {_mask_code(authorization_code)}", err=True)
+
+        try:
+            self._claim_token(authorization_code)
+        except OAuthError as error:
+            raise UnexpectedError(error.message)
+        token_data = self._validate_access_token()
```
6d7a: `self._validate_access_token()` may raise an `OAuthError` as well, so we should move it into the `try` block.

6d7a (changes requested):
~Thank you for the PR and sorry for the long wait. The changes look solid, I only have one  change request.
I have one concern though about the general scope of this: We already have `ggshield auth login --method token` which provides a completely headless auth flow. AFAIK OAuth 2.0 OOB is a legacy out-of-band flow, so if we want to adopt a standard for replacing our custom headless auth flow, shouldn't we go with a future-proof protocol?~
Saw the discussion on the sibling MR 😅

6d7a (changes requested):
Thank you for the PR and sorry for the long wait. The changes look solid, I only have one general suggestion: We already have the `--method` CLI parameter, could you wire up the OOB flow to it so that we can use it like `ggshield auth login --method oob`. That would be more consistent with the existing interface.

tests/unit/cmd/auth/test_login.py
```diff
@@ -1010,3 +1010,201 @@ def test_api_instance_url(self, instance_url, cli_fs_runner, monkeypatch):
         exit_code, output = self.run_cmd(cli_fs_runner)
         assert exit_code == ExitCode.SUCCESS, output
         self._webbrowser_open_mock.assert_called()
+
+
+OOB_REDIRECT_URI = "urn:ietf:wg:oauth:2.0:oob"
```
HelifeWasTaken: `OOB_REDIRECT_URI` is already exported from `ggshield.verticals.auth.oauth`.
Redefining it here creates a silent drift risk — if the value ever changes in
the source, this copy stays wrong while tests keep passing.
Add `from ggshield.verticals.auth.oauth import OOB_REDIRECT_URI` to the
import block at the top of the file and remove this local definition.

ggshield/verticals/auth/oauth.py
```diff
@@ -36,6 +43,20 @@
...
 logger = logging.getLogger(__name__)
 
+# Number of leading characters of an authorization code we keep visible when
+# echoing it back; the rest is replaced by `*` so the full code never lands
+# in the terminal scrollback in plain text.
+_OOB_CODE_VISIBLE_PREFIX = 4
+
+
+def _mask_code(code: str) -> str:
+    """Return ``code`` with all but the first 4 characters replaced by ``*``."""
```
HelifeWasTaken:
```suggestion
"""Return ``code`` with all but the first 4 characters replaced by ``*`` (fully masked if ≤ 4 chars)."""
```
The docstring says "all but the first 4 characters replaced by `*`", but the
guard `<= _OOB_CODE_VISIBLE_PREFIX` means codes whose length is exactly 4
(or shorter) have ALL characters masked — not just the suffix. The docstring
should reflect that edge case.

tests/unit/cmd/auth/test_login.py
```diff
@@ -1010,3 +1010,201 @@ def test_api_instance_url(self, instance_url, cli_fs_runner, monkeypatch):
...
+        assert "Success! You are now authenticated" in result.output
+        # Stderr should mention pasting the code.
+        assert (
+            "paste" in result.output.lower() or "paste" in (result.stderr or "").lower()
+        )
+
+        # Pasted code must not appear in plain text in the terminal output.
+        # Only the first 4 chars are echoed back, followed by `*` placeholders.
+        assert "some_authorization_code" not in result.output
+        assert "some" + "*" * (len("some_authorization_code") - 4) in result.output
```
HelifeWasTaken:
```suggestion
assert (
    "some" + "*" * (len("some_authorization_code") - 4) in result.output
    or "some" + "*" * (len("some_authorization_code") - 4)
    in (result.stderr or "")
)
```
This assertion only checks `result.output`, relying on `CliRunner`'s default
`mix_stderr=True` to include the stderr echo in output. The "paste" check
just above already defensively handles both `mix_stderr` settings with an `or`
fallback. Apply the same pattern here for consistency.

HelifeWasTaken:
Just nitpicks I could find because I was doing a review not really security issues if @6d7a concerns are addressed

6d7a (approved):
Thank you for the changes!

PR #1224, opened 2026-04-28
[MCP] Parse claude plugins and claude.ai servers

paulpetit-gg-ext:
## Context

Linear issue: NHI-1619

We didn't detect MCP servers installed as Claude plugins (or through Claude.ai)

## What has been done

Add their detection

## Validation

- Install a plugin with claude that has a MCP server (e.g. Context7)
- `ggshield ai discover`

## PR check list

### DIFF
--- a/ggshield/verticals/ai/agents/copilot.py
+++ b/ggshield/verticals/ai/agents/copilot.py
@@ -7,7 +7,7 @@ from pygitguardian.models import AIDiscovery, MCPActivityRequest
 
 from ggshield.core.dirs import get_user_home_dir
 
-from ..models import EventType, HookPayload, HookResult
+from ..models import Agent, EventType, HookPayload, HookResult, MCPConfiguration
 from .claude_code import Claude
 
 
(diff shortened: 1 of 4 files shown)


ggshield/verticals/ai/agents/copilot.py
```diff
@@ -66,6 +66,9 @@ def discover_project_directories(self) -> Iterator[Path]:
                 if path.is_dir():
                     yield path.resolve()
 
+    def _get_user_mcp_configurations(self) -> Iterator[MCPConfiguration]:
+        yield from []
```
clement-tourriere: `yield from []` here drops the base `Agent._get_user_mcp_configurations` behaviour, which would read `self.config_folder / "mcp.json"` — i.e. `~/.config/Code/User/mcp.json`, the standard VS Code user-level MCP
  file. As a result, user-scoped MCP servers configured globally in VS Code are no longer discovered for Copilot.

  I think the intent here is to *stop* inheriting Claude's `~/.claude.json` reader (which was always wrong for Copilot pre-PR), but the override should fall through to the grandparent rather than to nothing:

  ```python
  from ..models import Agent  # add to imports

  def _get_user_mcp_configurations(self) -> Iterator[MCPConfiguration]:
      yield from Agent._get_user_mcp_configurations(self)
  ```

  Test: drop a fake `<config_folder>/mcp.json` under a tmp config folder and assert the entry is discovered with `Scope.USER` and `agent="copilot"`.

clement-tourriere (approved):
Just a small comments, but otherwise seems good!

PR #1229, opened 2026-04-30
feat(mcp): detect cursor plugins and extensions

paulpetit-gg-ext:
## Context

Following the detection of MCP servers installed via plugins in Claude, this PR adds the same improvement for Cursor.

## What has been done

- detect servers declared in both `plugins/` and `extensions/`.
- the capability discovery logic has been changed to be able to detect tools for these servers.

## Validation

- install a plugin in Cursor (and activate the installed MCP in a project)
- it should appear in `ggshield ai discover`

## PR check list

### DIFF
--- a/changelog.d/20260430_074704_paul.petit.ext_HEAD.md
+++ b/changelog.d/20260430_074704_paul.petit.ext_HEAD.md
@@ -1,0 +1,43 @@
+<!--
+A new scriv changelog fragment.
+
+Uncomment the section that is right (remove the HTML comment wrapper).
+For top level release notes, leave all the headers commented out.
+-->
+
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+
+### Added
+
+- detection of MCP servers installed with Cursor plugins or Cursor extensions
+<!--
+
+### Changed
+
+- A bullet item for the Changed category.
+
+-->
+
+<!--
+### Deprecated
+
+- A bullet item for the Deprecated category.
+
+-->
+<!--
+### Fixed
+
+- A bullet item for the Fixed category.
+
+-->
+<!--
+### Security
+
+- A bullet item for the Security category.
+
+-->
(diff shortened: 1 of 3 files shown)


ggshield/verticals/ai/agents/cursor.py
```diff
@@ -79,6 +88,92 @@ def settings_locate(
...
+                    mcp_data = {"mcpServers": mcp_data}
+                yield from self._parse_servers_block(mcp_data, Scope.USER, None)
+                return
+
+        # Fallback: inline mcpServers in the plugin manifest.
+        manifest = self._load_json_file(install_dir / ".cursor-plugin" / "plugin.json")
+        if not manifest:
+            return
+        inline = manifest.get("mcpServers")
+        if isinstance(inline, dict):
```
6d7a, line 135: I noticed in the cursor docs that this field apparently can also be [an array of inline servers](https://cursor.com/docs/reference/plugins#optional-fields), although I couldn't find an example in the docs of that structure.

changelog.d/20260430_074704_paul.petit.ext_HEAD.md
```diff
@@ -0,0 +1,43 @@
...
+<!--
+### Removed
+
+- A bullet item for the Removed category.
+
+-->
+
+### Added
+
+- detection of MCP servers installed with Cursor plugins or Cursor extensions
```
6d7a: nit:
```suggestion
- Detect MCP servers installed with Cursor plugins or Cursor extensions.
```

6d7a (approved):
Looks good, just non-blocking nits.

ggshield/verticals/ai/agents/cursor.py
paulpetit-gg-ext: Interesting! I never encountered this either. It could also be a path to another file according to the doc. I will open another issue to investigate, as this could be a non-trivial change to this MR, and is more of a corner case anyway.

PR #1230, opened 2026-04-30
chore: pass ty checks

paulpetit-gg-ext:
## Context

[Ty](https://docs.astral.sh/ty/) is a type checker gaining popularity that is extremely fast.

GGshield's codebase (tests excluded) currently almost passes all ty's checks, and most of the errors are already silenced for pyright.

This PR adds the necessary fixes (that also work with Pyright) to have ty's checks pass. This could allow developers to work with type checking enabled on the whole codebase at all times, detecting errors even in files they don't change almost instantly.

Notes:
- this PR *does not* remove Pyright from the pre-commit checks.
- this PR *does not* try to fix type checking in tests (could be a later goal)

## What has been done

- added ty in the pre-commit hooks
- fixed check errors

## Validation

`pre-commit run -a ty`

## PR check list

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -96,7 +96,7 @@ repos:
     hooks:
       - id: import-linter-config
         name: Import Linter Config Check
-        entry: ./scripts/check-import-linter-config.sh
+        entry: uv run ./scripts/check-import-linter-config.sh
         language: system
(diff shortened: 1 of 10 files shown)


benjaminrigaud-gg (approved):
Let's TrY!

PR #1233, opened 2026-05-01
Add Codex AI hook support

trickyfalcon:
## Summary

Adds native Codex support to `ggshield secret scan ai-hook`.

## Changes

- Adds a Codex agent implementation with native Codex hook JSON output.
- Detects Codex hook payloads via `turn_id` or `.codex` transcript paths.
- Scans Codex `Bash` and `apply_patch` PreToolUse payloads.
- Adds `ggshield install -t codex` support and enables `codex_hooks` in Codex config.
- Updates README supported AI assistants.
- Adds focused tests for Codex parsing, output behavior, and installation.

## Validation

- `uv run --frozen --group tests pytest tests/unit/verticals/ai/test_hooks.py tests/unit/verticals/ai/test_installation.py`
- Locally verified Codex accepts `permissionDecision: "deny"` from a PreToolUse hook and blocks the Bash command before execution.

### DIFF
--- a/ggshield/verticals/ai/agents/codex.py
+++ b/ggshield/verticals/ai/agents/codex.py
@@ -1,0 +1,141 @@
+import json
+from pathlib import Path
+from typing import Any, Dict, Iterator, List, Literal, Optional
+
+import click
+from pygitguardian.models import AIDiscovery, MCPActivityRequest
+
+from ggshield.core.dirs import get_user_home_dir
+
+from ..models import Agent, EventType, HookPayload, HookResult
+from .claude_code import _mangle_server_name
+
+
+class Codex(Agent):
+    """Behavior specific to OpenAI Codex."""
+
+    @property
+    def name(self) -> str:
+        return "codex"
+
+    @property
+    def display_name(self) -> str:
+        return "Codex"
+
+    @property
+    def config_folder(self) -> Path:
+        return get_user_home_dir() / ".codex"
+
+    def output_result(self, result: HookResult) -> int:
+        response: Dict[str, Any] = {}
+        if result.block:
+            if result.payload.event_type == EventType.PRE_TOOL_USE:
+                response["hookSpecificOutput"] = {
+                    "hookEventName": "PreToolUse",
+                    "permissionDecision": "deny",
+                    "permissionDecisionReason": result.message,
+                }
+            elif result.payload.event_type in [
+                EventType.USER_PROMPT,
+                EventType.POST_TOOL_USE,
+            ]:
+                response["decision"] = "block"
+                response["reason"] = result.message
+            else:
+                click.echo(result.message, err=True)
+                return 2
+
+        click.echo(json.dumps(response))
+        return 0
+
+    def settings_path(self, mode: Literal["local", "global"]) -> Path:
+        return Path(".codex") / "hooks.json"
+
+    @property
+    def settings_template(self) -> Dict[str, Any]:
+        return {
+            "hooks": {
+                "PreToolUse": [
+                    {
+                        "matcher": ".*",
+                        "hooks": [
+                            {
+                                "type": "command",
+                                "command": "<COMMAND>",
+                            }
+                        ],
+                    }
+                ],
+                "PostToolUse": [
+                    {
+                        "matcher": ".*",
+                        "hooks": [
+                            {
+                                "type": "command",
+                                "command": "<COMMAND>",
+                            }
+                        ],
+                    }
+                ],
+                "UserPromptSubmit": [
+                    {
+                        "hooks": [
+                            {
+                                "type": "command",
+                                "command": "<COMMAND>",
+                            }
+                        ],
+                    }
+                ],
+            }
+        }
+
+    def settings_locate(
+        self, candidates: List[Dict[str, Any]], template: Dict[str, Any]
+    ) -> Optional[Dict[str, Any]]:
+        if "matcher" in template:
+            for obj in candidates:
+                if obj.get("matcher") == template["matcher"]:
+                    return obj
+            return None
+        for obj in candidates:
+            command = obj.get("command", "")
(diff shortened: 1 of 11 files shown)


ggshield/verticals/ai/installation.py
```diff
@@ -77,6 +78,9 @@ def install_hooks(
         json.dump(existing_config, f, indent=2)
         f.write("\n")
 
+    if agent.name == "codex":
+        _enable_codex_hooks_feature(base_dir)
+
```
paulpetit-gg-ext: It seems that since the most recent versions, it is no longer necessary to explicitly enable the feature. We should remove this specific logic.

ggshield/verticals/ai/installation.py
```diff
@@ -89,6 +93,38 @@ def install_hooks(
...
+            config_path.write_text("\n".join(lines) + "\n", encoding="utf-8")
+            return
+
+    separator = "\n" if content.endswith("\n") else "\n\n"
+    config_path.write_text(
+        f"{content}{separator}[features]\ncodex_hooks = true\n",
+        encoding="utf-8",
+    )
+
+
```
paulpetit-gg-ext: (same here)

ggshield/verticals/ai/models.py
```diff
@@ -38,6 +38,7 @@ class Tool(Enum):
 
     BASH = auto()
     READ = auto()
+    EDIT = auto()
```
paulpetit-gg-ext: For now the goal of the feature was to prevent secrets from leaking to the LLM. We don't monitor edit tools for any current Agent as they can't introduce secrets the Agent didn't already know about. We should keep it outside of this PR.

ggshield/verticals/ai/hooks.py
```diff
@@ -109,6 +110,9 @@ def parse_hook_input(raw_content: str) -> list[HookPayload]:
         if tool == Tool.BASH:
             content = tool_input.get("command", "")
             identifier = content
+        elif tool == Tool.EDIT:
+            content = tool_input.get("command", "")
+            identifier = content
```
paulpetit-gg-ext: See comment below: we don't need to track editions so this should be kept out of this PR.

ggshield/verticals/ai/hooks.py
```diff
@@ -27,6 +27,7 @@
     "shell": Tool.BASH,  # Cursor
     "bash": Tool.BASH,  # Claude Code
     "run_in_terminal": Tool.BASH,  # Copilot
+    "apply_patch": Tool.EDIT,  # Codex
```
paulpetit-gg-ext: See comment below: we don't need to track editions so this should be kept out of this PR.

ggshield/verticals/ai/hooks.py
```diff
@@ -317,6 +323,11 @@ def _message_from_secrets(
                     "Please remove the secrets from the command before executing it. "
                     "Consider using environment variables or a secrets manager instead."
                 )
+        elif payload.tool == Tool.EDIT:
+            message = (
+                "Please remove the secrets from the file edit before applying it. "
+                "Consider using environment variables or a secrets manager instead."
+            )
```
paulpetit-gg-ext: See comment below: we don't need to track editions so this should be kept out of this PR.

ggshield/verticals/ai/hooks.py
```diff
@@ -347,6 +358,8 @@ def _send_secret_notification(
             source = "reading a file"
         elif tool == Tool.BASH:
             source = "running a command"
+        elif tool == Tool.EDIT:
+            source = "editing a file"
```
paulpetit-gg-ext: See comment below: we don't need to track editions so this should be kept out of this PR.

ggshield/verticals/ai/agents/codex.py
```diff
@@ -0,0 +1,141 @@
...
+                        "hooks": [
+                            {
+                                "type": "command",
+                                "command": "<COMMAND>",
+                            }
+                        ],
+                    }
+                ],
+            }
+        }
```
paulpetit-gg-ext, line 92: nitpick: this is the third Agent (all but Cursor, in fact) that has the same template, maybe we should make this the default implementation instead.
This is not blocking and can be done in a later commit.

ggshield/verticals/ai/agents/codex.py
```diff
@@ -0,0 +1,141 @@
...
+        for obj in candidates:
+            command = obj.get("command", "")
+            if "ggshield" in command or "<COMMAND>" in command:
+                return obj
+        for obj in candidates:
+            for hook in obj.get("hooks", []):
+                command = hook.get("command", "")
+                if "ggshield" in command or "<COMMAND>" in command:
+                    return obj
+        return None
```
paulpetit-gg-ext, line 111: same remark, maybe this should be the default.

ggshield/verticals/ai/agents/codex.py
```diff
@@ -0,0 +1,141 @@
...
+    def display_name(self) -> str:
+        return "Codex"
+
+    @property
+    def config_folder(self) -> Path:
+        return get_user_home_dir() / ".codex"
+
+    def output_result(self, result: HookResult) -> int:
+        response: Dict[str, Any] = {}
+        if result.block:
```
paulpetit-gg-ext, line 31: We should probably fill the generic "systemMessage" field every time to surface the message to the UI.

paulpetit-gg-ext (changes requested):
Thank you for your contribution! There are a few things that should be removed.
We will also need to do more testing with Codex but this is a good foundation.

paulpetit-gg-ext (approved):
Thanks for the fixes!

Blaxav (approved):
Looks good to me, thanks a lot for the contribution, that's really appreciated !

xblanchot-gg (approved):

PR #1235, opened 2026-05-05
feat(plugin): install plugins from the GitGuardian platform

benjaminrigaud-gg:
## Context

This PR is the ggshield side of the GitGuardian-platform plugin install flow. With the matching backend deployed, `ggshield plugin install/update/status` discover and pull plugins from the user's GitGuardian instance instead of a hard-coded GitHub release URL.

## What has been done

**Core install flow**
- `PluginAPIClient` rewritten to call `/v1/endpoints/plugins` (list / detail / streaming download / signature) with API-key auth. The client no longer sees the upstream wheel URL: the platform proxies bytes from a GitGuardian-hosted repository and exposes `X-Plugin-SHA256`, `X-Plugin-Version`, `Content-Length` (and `X-Plugin-Signature-URL`) on the download response.
- `download_and_install` accepts a byte stream rather than fetching a URL — keeps the verify-sha256 / extract pipeline pure.
- `install` / `update` / `status` commands wired to the new client surface. `update` calls the new `installed` endpoint best-effort. `status` surfaces a dedicated `PluginsNotEnabledError` when the workspace doesn't have the feature.

**Install-directory naming**
- `download_and_install` (catalog flow) now names the on-disk plugin directory after the wheel's PEP 503-normalised distribution name, the same way `install_from_wheel` (local-file flow) already does. The catalog reference is still passed in as `plugin_name` and stored in the manifest, but the on-disk dir comes from the wheel filename.
- Convergent naming means a user who installed the local wheel and then runs `plugin install <ref>` ends up with **one** plugin directory flipped from `source.type=local_file` to `platform` in place — not two side-by-side directories backed by the same wheel bytes.
- `update --check` pairing keys on entry-point name = catalog reference (independent of the on-disk dir name), so the rename above doesn't change the pairing logic. `_resolve_plugin_dir`'s wheel-based entry-point fallback already handles the dir-name / reference divergence for `plugin uninstall`.

**Manifest**
- New `PluginSourceType.PLATFORM` (value `"platform"`) replaces `GITGUARDIAN_API`. A `_missing_` classmethod accepts the legacy `"gitguardian_api"` value so manifests written by pre-1.50 ggshield keep parsing.
- `ggshield plugin list` now reads the install source from the manifest and renders it verbatim (`platform`, `local file`, `url`, `github release`, `github artifact`); falls back to `pip` for entry-point-only plugins and `on-disk` for the wheel-without-manifest case.

**Smaller fixes**
- variable shadowing in `download_plugin` (`version` → `resolved_version`)
- missing `return` after `ctx.exit()` in `update.py` error handlers (was hitting `UnboundLocalError`)
- removed redundant try/except around `report_installation` (method already swallows + logs)
- `.worktrees/` added to `.gitignore`

## Validation

Unit and functional tests in `tests/unit/cmd/plugin/` and `tests/unit/core/plugin/` cover the new client, the rewritten command flows, and the source-type compat layer (parametrized matrix across all PluginSourceType values plus the two fallback paths).

End-to-end against a live backend: a separate harness drives ggshield as a real user would against a GitGuardian instance with the plugin feature enabled, then asserts on real CLI output and on-disk manifests. 14 scenarios passing as of the latest push:

\`\`\`
status / install pinned / install latest / list /
status post-install (update available) / update --check / update --all /
re-install / uninstall / unknown-plugin error / unknown-version error /
strict-sig install (sigstore bundle proxied + verified) /
local→local wheel upgrade / local-wheel-then-platform install (in-place)
\`\`\`

Manual smoke test against a GitGuardian instance with the plugin feature enabled:

1. \`ggshield auth login --instance <instance-url>\`.
2. \`ggshield plugin status\` → catalog plugins listed as available.
3. \`ggshield plugin install <ref>\` → wheel + sigstore bundle streamed via the platform; manifest records \`source.type == platform\`.
4. \`ggshield plugin list\` → \`<entry-point-name>: v…, enabled, platform, signature: signed (…)\`.
5. \`ggshield plugin update --check\` then \`--all\` → upgrades to the catalog's latest.

## PR check list

- [x] If the changes affect the end user (new feature, behavior change, bug fix) then the PR has a changelog entry (see doc/dev/getting-started.md). If the changes do not affect the end user, then the \`skip-changelog\` label has been added to the PR.

### DIFF
--- a/ggshield/cmd/plugin/install.py
+++ b/ggshield/cmd/plugin/install.py
@@ -17,16 +17,17 @@ from ggshield.core.errors import ExitCode
 from ggshield.core.plugin.client import (
     PluginAPIClient,
     PluginAPIError,
-    PluginNotAvailableError,
+    PluginsNotEnabledError,
     PluginSourceType,
 )
 from ggshield.core.plugin.downloader import (
     ChecksumMismatchError,
-    DownloadError,
     GitHubArtifactError,
     InsecureSourceError,
     PluginDownloader,
 )
+from ggshield.core.plugin.loader import resolve_config_key
+from ggshield.core.plugin.platform import get_platform_info
 from ggshield.core.plugin.signature import (
     SignatureVerificationError,
     SignatureVerificationMode,
@@ -65,7 +66,7 @@ def detect_source_type(plugin_source: str) -> PluginSourceType:
         return PluginSourceType.LOCAL_FILE
 
     # Default: assume it's a plugin name for GitGuardian API
-    return PluginSourceType.GITGUARDIAN_API
+    return PluginSourceType.PLATFORM
 
 
 @click.command()
@@ -153,15 +154,20 @@ def _install_from_gitguardian(
     version: Optional[str],
     signature_mode: SignatureVerificationMode = SignatureVerificationMode.STRICT,
 ) -> None:
-    """Install a plugin from GitGuardian API."""
+    """Install a plugin from the GitGuardian platform."""
     ctx_obj = ContextObj.get(ctx)
     config = ctx_obj.config
 
-    # Fetch available plugins
     try:
         client = create_client_from_config(config)
         plugin_api_client = PluginAPIClient(client)
         catalog = plugin_api_client.get_available_plugins()
+    except PluginsNotEnabledError:
+        ui.display_error(
+            "Plugin system is not available on this workspace. "
+            "Contact your administrator."
+        )
+        ctx.exit(ExitCode.UNEXPECTED_ERROR)
     except PluginAPIError as e:
         ui.display_error(str(e))
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
@@ -169,11 +175,9 @@ def _install_from_gitguardian(
         ui.display_error(f"Failed to connect to GitGuardian: {e}")
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
 
-    # Check if plugin is available
     available_plugins = {p.name: p for p in catalog.plugins if p.available}
 
     if plugin_name not in available_plugins:
-        # Check if plugin exists but is not available
         unavailable = next((p for p in catalog.plugins if p.name == plugin_name), None)
         if unavailable:
             ui.display_error(
@@ -186,30 +190,49 @@ def _install_from_gitguardian(
             ui.display_info("Use 'ggshield plugin status' to see available plugins")
         ctx.exit(ExitCode.USAGE_ERROR)
 
-    # Install the plugin
     downloader = PluginDownloader()
     enterprise_config = EnterpriseConfig.load()
+    platform_info = get_platform_info()
 
     ui.display_info(f"Installing {plugin_name}...")
 
     try:
-        # Get download info
-        download_info = plugin_api_client.get_download_info(
-            plugin_name, version=version
-        )
+        with plugin_api_client.download_plugin(
+            plugin_name, platform_info=platform_info, version=version
+        ) as (info, chunks):
+            bundle_bytes: Optional[bytes] = None
+            if info.signature_url:
+                try:
+                    bundle_bytes = plugin_api_client.download_signature_bundle(
+                        info.signature_url
+                    )
+                except PluginAPIError as bundle_err:
+                    # In STRICT mode, an unreachable bundle is a hard
+                    # failure — the user expects signature verification.
+                    # In WARN mode they have opted into accepting
+                    # unsigned plugins, so a bundle the proxy couldn't
+                    # serve is equivalent to "no signature available"
+                    # rather than a reason to abort an otherwise valid
+                    # wheel download.
+                    if signature_mode == SignatureVerificationMode.STRICT:
+                        raise
+                    ui.display_warning(
+                        f"Could not fetch signature bundle for {plugin_name}: "
+                        f"{bundle_err}. Continuing without verification "
+                        f"(--allow-unsigned)."
+                    )
+            wheel_path = downloader.download_and_install(
+                info,
+                chunks,
+                plugin_name,
+                signature_mode=signature_mode,
+                bundle_bytes=bundle_bytes,
+            )
 
-        # Download and install
-        downloader.download_and_install(
-            download_info, plugin_name, signature_mode=signature_mode
-        )
-
-        # Enable in config
-        enterprise_config.enable_plugin(plugin_name, version=download_info.version)
-
-        # Save config
+        config_key = resolve_config_key(wheel_path, fallback=plugin_name)
+        enterprise_config.enable_plugin(config_key, version=info.version)
         enterprise_config.save()
-
-        ui.display_info(f"Installed {plugin_name} v{download_info.version}")
+        ui.display_info(f"Installed {plugin_name} v{info.version}")
 
     except SignatureVerificationError as e:
         ui.display_error(f"Signature verification failed for {plugin_name}: {e}")
@@ -219,15 +242,13 @@ def _install_from_gitguardian(
             "pass --allow-unsigned."
         )
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
-    except PluginNotAvailableError as e:
-        ui.display_error(f"Failed to install {plugin_name}: {e}")
-        ctx.exit(ExitCode.UNEXPECTED_ERROR)
-    except DownloadError as e:
-        ui.display_error(f"Failed to install {plugin_name}: {e}")
-        ctx.exit(ExitCode.UNEXPECTED_ERROR)
     except Exception as e:
         ui.display_error(f"Failed to install {plugin_name}: {e}")
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
+
+    plugin_api_client.report_installation(
+        plugin_name, info.version, platform_info.os, platform_info.arch
+    )
 
 
 def _install_from_local_wheel(
@@ -248,12 +269,12 @@ def _install_from_local_wheel(
     ui.display_info(f"Installing from {wheel_path.name}...")
 
     try:
-        plugin_name, version, _ = downloader.install_from_wheel(
+        plugin_name, version, installed_wheel = downloader.install_from_wheel(
             wheel_path, signature_mode=signature_mode
         )
 
-        # Enable in config
-        enterprise_config.enable_plugin(plugin_name, version=version)
+        config_key = resolve_config_key(installed_wheel, fallback=plugin_name)
+        enterprise_config.enable_plugin(config_key, version=version)
         enterprise_config.save()
 
         ui.display_info(f"Installed {plugin_name} v{version}")
@@ -265,9 +286,6 @@ def _install_from_local_wheel(
             "If you trust its origin and still want to install it, "
             "pass --allow-unsigned."
         )
-        ctx.exit(ExitCode.UNEXPECTED_ERROR)
-    except DownloadError as e:
-        ui.display_error(f"Failed to install from wheel: {e}")
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
     except Exception as e:
         ui.display_error(f"Failed to install from wheel: {e}")
@@ -287,12 +305,12 @@ def _install_from_url(
     ui.display_info("Installing from URL...")
 
     try:
-        plugin_name, version, _ = downloader.download_from_url(
+        plugin_name, version, installed_wheel = downloader.download_from_url(
             url, sha256, signature_mode=signature_mode
         )
 
-        # Enable in config
-        enterprise_config.enable_plugin(plugin_name, version=version)
+        config_key = resolve_config_key(installed_wheel, fallback=plugin_name)
+        enterprise_config.enable_plugin(config_key, version=version)
         enterprise_config.save()
 
         ui.display_info(f"Installed {plugin_name} v{version}")
@@ -310,9 +328,6 @@ def _install_from_url(
         ctx.exit(ExitCode.USAGE_ERROR)
     except ChecksumMismatchError as e:
         ui.display_error(f"Checksum verification failed: {e}")
-        ctx.exit(ExitCode.UNEXPECTED_ERROR)
-    except DownloadError as e:
-        ui.display_error(f"Failed to install from URL: {e}")
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
     except Exception as e:
         ui.display_error(f"Failed to install from URL: {e}")
@@ -332,12 +347,12 @@ def _install_from_github_release(
     ui.display_info("Installing from GitHub release...")
 
     try:
-        plugin_name, version, _ = downloader.download_from_github_release(
+        plugin_name, version, installed_wheel = downloader.download_from_github_release(
             url, sha256, signature_mode=signature_mode
         )
 
-        # Enable in config
-        enterprise_config.enable_plugin(plugin_name, version=version)
+        config_key = resolve_config_key(installed_wheel, fallback=plugin_name)
+        enterprise_config.enable_plugin(config_key, version=version)
         enterprise_config.save()
 
         ui.display_info(f"Installed {plugin_name} v{version}")
@@ -355,9 +370,6 @@ def _install_from_github_release(
         ctx.exit(ExitCode.USAGE_ERROR)
     except ChecksumMismatchError as e:
         ui.display_error(f"Checksum verification failed: {e}")
-        ctx.exit(ExitCode.UNEXPECTED_ERROR)
-    except DownloadError as e:
-        ui.display_error(f"Failed to install from GitHub release: {e}")
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
     except Exception as e:
         ui.display_error(f"Failed to install from GitHub release: {e}")
@@ -378,12 +390,12 @@ def _install_from_github_artifact(
     ui.display_info("Installing from GitHub artifact...")
 
     try:
-        plugin_name, version, _ = downloader.download_from_github_artifact(
-            url, signature_mode=signature_mode
+        plugin_name, version, installed_wheel = (
+            downloader.download_from_github_artifact(url, signature_mode=signature_mode)
         )
 
-        # Enable in config
-        enterprise_config.enable_plugin(plugin_name, version=version)
+        config_key = resolve_config_key(installed_wheel, fallback=plugin_name)
+        enterprise_config.enable_plugin(config_key, version=version)
         enterprise_config.save()
 
         ui.display_info(f"Installed {plugin_name} v{version}")
@@ -398,9 +410,6 @@ def _install_from_github_artifact(
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
     except GitHubArtifactError as e:
         ui.display_error(str(e))
-        ctx.exit(ExitCode.UNEXPECTED_ERROR)
-    except DownloadError as e:
-        ui.display_error(f"Failed to install from GitHub artifact: {e}")
         ctx.exit(ExitCode.UNEXPECTED_ERROR)
     except Exception as e:
         ui.display_error(f"Failed to install from GitHub artifact: {e}")
--- a/ggshield/cmd/plugin/update.py
+++ b/ggshield/cmd/plugin/update.py
@@ -4,6 +4,7 @@ Plugin update command - updates installed plugins.
 
 import logging
 import os
+from dataclasses import dataclass
 from typing import Any, Dict, List, Optional
 
 import click
@@ -19,16 +20,32 @@ from ggshield.core.errors import ExitCode
 from ggshield.core.plugin.client import (
     PluginAPIClient,
     PluginAPIError,
-    PluginNotAvailableError,
+    PluginsNotEnabledError,
     PluginSourceType,
 )
 from ggshield.core.plugin.downloader import DownloadError, PluginDownloader
-from ggshield.core.plugin.loader import PluginLoader
+from ggshield.core.plugin.loader import PluginLoader, resolve_config_key
+from ggshield.core.plugin.platform import get_platform_info
 from ggshield.core.plugin.signature import SignatureVerificationMode
 from ggshield.core.text_utils import pluralize
 
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class _InstallationReport:
+    """Args for one deferred ``report_installation`` call.
+
+    Buffered during the update loop and fired after ``enterprise_config.save()``
+    so a stalled ``/installed`` endpoint can't desync wheel-on-disk from
+    the on-disk version record.
+    """
+
+    name: str
+    version: str
+    platform: str
+    arch: str
 
 
 def _is_newer_version(installed_version: str, candidate_version: str) -> bool:
@@ -141,9 +158,9 @@ def update_cmd(
     for name in plugins_to_process:
         source = downloader.get_plugin_source(name)
         if source is None:
-            # Legacy manifest or unknown source - try GitGuardian API
+            # Legacy manifest or unknown source - try platform
             gitguardian_plugins.append(name)
-        elif source.type == PluginSourceType.GITGUARDIAN_API:
+        elif source.type == PluginSourceType.PLATFORM:
             gitguardian_plugins.append(name)
         elif source.type == PluginSourceType.GITHUB_RELEASE:
             github_release_plugins.append(name)
@@ -152,20 +169,29 @@ def update_cmd(
             non_updatable_plugins.append(
                 {
                     "name": name,
-                    "source_type": source.type.value,
+                    "source_type": source.type,
                 }
             )
 
     # Check GitGuardian API for updates
     updates_available: List[Dict[str, Any]] = []
     plugin_api_client: Optional[PluginAPIClient] = None
+    # Tracks whether we asked the platform for catalog updates and got a
+    # usable answer. Stays False when there were no platform plugins in
+    # the batch OR when the catalog fetch failed and we degraded to
+    # github-only checks. Drives the empty-result message at the bottom
+    # so we don't claim "All updatable plugins are up to date" when half
+    # the picture was actually missing.
+    platform_checked = False
 
     if gitguardian_plugins:
+        catalog_error: Optional[str] = None
         try:
             client = create_client_from_config(config)
             plugin_api_client = PluginAPIClient(client)
             catalog = plugin_api_client.get_available_plugins()
             available_plugins = {p.name: p for p in catalog.plugins if p.available}
+            platform_checked = True
 
             for name in gitguardian_plugins:
                 installed = installed_plugins.get(name)
@@ -187,16 +213,36 @@ def update_cmd(
                             "name": name,
                             "installed_version": installed_version,
                             "latest_version": latest_version,
-                            "source_type": PluginSourceType.GITGUARDIAN_API,
+                            "source_type": PluginSourceType.PLATFORM,
                         }
                     )
 
+        except PluginsNotEnabledError:
+            catalog_error = (
+                "Plugin system is not available on this workspace. "
+                "Contact your administrator."
+            )
         except PluginAPIError as e:
-            ui.display_error(str(e))
-            ctx.exit(ExitCode.UNEXPECTED_ERROR)
+            catalog_error = str(e)
         except Exception as e:
-            ui.display_error(f"Failed to connect to GitGuardian: {e}")
-            ctx.exit(ExitCode.UNEXPECTED_ERROR)
+            catalog_error = f"Failed to connect to GitGuardian: {e}"
+
+        if catalog_error is not None:
+            # Degrade rather than abort when there's still other work to do:
+            # GitHub-release plugins and the non-updatable footer don't depend
+            # on the GitGuardian catalog, so a single failed fetch shouldn't
+            # kill an otherwise valid ``update --all`` invocation. When the
+            # platform fetch is the only thing we were asked to do (no
+            # github_release plugins in the batch), surface as a hard error
+            # so the exit code still reflects the failure.
+            if github_release_plugins:
+                ui.display_warning(
+                    f"Skipping GitGuardian-hosted plugin updates: {catalog_error}"
+                )
+            else:
+                ui.display_error(catalog_error)
+                ctx.exit(ExitCode.UNEXPECTED_ERROR)
+                return
 
     # Check GitHub releases for updates
     for name in github_release_plugins:
@@ -226,6 +272,18 @@ def update_cmd(
                         }
                     )
 
+    # When the platform check was skipped (catalog fetch failed but we
+    # still have github_release work to do), the "everything is up to
+    # date" wording lies about half the picture. Surface that the
+    # platform side wasn't actually checked so the user knows their
+    # GitGuardian-hosted plugins are unverified, not confirmed-current.
+    no_updates_msg = (
+        "No updates found for checked plugins; "
+        "GitGuardian-hosted plugins were skipped."
+        if gitguardian_plugins and not platform_checked
+        else "All updatable plugins are up to date."
+    )
+
     # Check-only mode
     if check_only:
         if updates_available:
@@ -233,7 +291,7 @@ def update_cmd(
             for update in updates_available:
                 source_label = (
                     "GitGuardian"
-                    if update["source_type"] == PluginSourceType.GITGUARDIAN_API
+                    if update["source_type"] == PluginSourceType.PLATFORM
                     else "GitHub"
                 )
                 ui.display_info(
@@ -243,16 +301,9 @@ def update_cmd(
             ui.display_info("")
             ui.display_info("Run 'ggshield plugin update --all' to update.")
         else:
-            ui.display_info("All updatable plugins are up to date.")
+            ui.display_info(no_updates_msg)
 
-        if non_updatable_plugins:
-            ui.display_info("")
-            ui.display_heading("Cannot Auto-Update")
-            for plugin in non_updatable_plugins:
-                ui.display_info(
-                    f"  {plugin['name']}: installed from {plugin['source_type']}"
-                )
-            ui.display_info("Re-install these plugins manually to update.")
+        _display_non_updatable_footer(non_updatable_plugins)
         return
 
     # Update mode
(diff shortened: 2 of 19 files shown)


ggshield/core/plugin/client.py
```diff
@@ -2,24 +2,82 @@
...
+    """Return a wheel filename safe to use as a single path segment.
+
+    Strips any path components the server may have included and rejects
+    values that would resolve outside the plugin directory (``..``, empty
+    segment, embedded NUL, trailing ``.whl`` missing).
+    """
+    name = PurePosixPath(raw).name
+    if not name or name in {".", ".."} or "\x00" in name or "\\" in name:
+        raise PluginAPIError(f"Server returned unsafe filename: {raw!r}")
+    return name
```
paulpetit-gg-ext: - This doesn't check the ".whl" suffix.
- There is almost the same logic twice in `downloader.py`, we should probably make this a util.

ggshield/core/plugin/client.py
```diff
@@ -133,118 +215,201 @@ def get_available_plugins(self) -> PluginCatalog:
...
                     "platform": platform_info.os,
                     "arch": platform_info.arch,
                 },
-                headers=self._get_headers(),
             )
+            if response.status_code == 404:
+                raise PluginsNotEnabledError()
             response.raise_for_status()
+        except PluginsNotEnabledError:
+            raise
```
paulpetit-gg-ext: I don't get the purpose of these two lines :/

ggshield/core/plugin/downloader.py
```diff
@@ -28,13 +28,85 @@
...
+    PEP 427 wheel filenames are
+    ``{distribution}-{version}(-{build})?-{python}-{abi}-{platform}.whl``.
+    The distribution segment uses ``_`` in place of ``-`` from the
+    canonical name, so the inverse normalisation
+    (``lower()`` + ``_ -> -``) recovers the PEP 503 form. We only need
+    the distribution segment, so a simple split on ``-`` is enough.
+    """
+    if not filename.endswith(".whl"):
+        raise DownloadError(f"Not a wheel filename: {filename!r}")
+    stem = filename[: -len(".whl")]
```
paulpetit-gg-ext: nit: `stem = filename.removesuffix(".whl")` exists :)

ggshield/core/plugin/client.py
```diff
@@ -133,118 +215,201 @@ def get_available_plugins(self) -> PluginCatalog:
...
+                    supported_platforms=[],
+                    reason=p.get("reason"),
                 )
-                for p in data.get("plugins", [])
+                for p in plugins_data
             ],
         )
 
-    def get_download_info(
+    @contextmanager
```
paulpetit-gg-ext, line 298: The API feels awkward: The PluginAPIClient class is only used when downloading from the platform and `PluginDownloader.download_and_install()` is only used with this context manager. Given how the install/update command works, I suppose we could simply have something like `PluginDownloader.download_from_platform()` that instantiates a `PluginAPIClient` to work.

ggshield/core/plugin/downloader.py
```diff
@@ -100,79 +172,155 @@ def __init__(self) -> None:
     def download_and_install(
```
paulpetit-gg-ext, line 165: There are inconsistencies within the method names: this one is called "download_and_install", there is "install_from_wheel" (makes sense because no download), but "download_from_github_artifact" and "download_from_github_release" also install, whereas their name suggests they only download.

Also, as stated in the review of `client.py`, it is really a "download_from_platform" in disguise, because all other methods don't reuse it with bytes downloaded from various sources.

ggshield/core/plugin/downloader.py
```diff
@@ -476,20 +629,24 @@ def download_from_github_artifact(
             except Exception as e:
                 raise GitHubArtifactError(f"Failed to extract artifact: {e}") from e
 
-            # Find wheel file in extracted contents
-            wheel_files = list(extract_dir.glob("**/*.whl"))
+            # Sort to ensure deterministic wheel selection when multiple
+            # wheels are shipped in the same artifact — without this, the
+            # order depends on filesystem traversal and an attacker-shaped
```
paulpetit-gg-ext: nit: not clear what "attacker-shaped" means here. If we suppose the upstream repository can be compromised or that we downloaded an adversarial wheel, this is most likely game over already.
However, handling the multiple-wheel-per-artifact case is good regardless.

paulpetit-gg-ext (commented):
A few comments here and there for the actual PR.

More broadly, my general impression (not restricted to this PR) is that the plugin management's logic is hard to follow. There is a lot of duplicated work and not very clear boundaries/separation of concerns, including:

- The `PluginDownloader` is more a "PluginManager" as it also handles installing/uninstalling the plugins and do a lot of the security checks...
- ... except for some plugin management logic which is still in the commands
- the different download methods repeat almost identical logic for checksum computation, signature verification and registration/cleanup. There should be a common flow that does all that and only delegate fetching the wheel file to a specific method per possible source type.
- install and update commands repeat the same error management and logging for every kind of download

This PR could be merged (after resolving a few comments) as it doesn't really make things worse in that regard, but I think there is a strong opportunity to refactor this module.

ggshield/core/plugin/client.py
benjaminrigaud-gg: Fixed across fb4a26c1 + 5606d658 + 44861745 — extracted to `wheel_utils.sanitize_wheel_filename` with the `.whl`-suffix check; `client.py` and the URL-download filename fallback in `downloader.download_from_url` now use it (the latter via `try/except InvalidWheelError` so the `"plugin.whl"` fallback still kicks in for query-string URLs).

ggshield/core/plugin/client.py
benjaminrigaud-gg: Fixed in 5606d658 — confirmed dead code. `PluginsNotEnabledError(Exception)` does not inherit from `RequestException`, so the bare `except RequestException` below would never have caught it. Removed.

ggshield/core/plugin/downloader.py
benjaminrigaud-gg: Fixed in 44861745.

ggshield/core/plugin/client.py
benjaminrigaud-gg: Agreed; deferring to a follow-up PR. The plan there is to collapse `PluginAPIClient` into `PluginDownloader.download_from_platform()` so we have one entry point per source type and a single shared post-download flow. Keeping this PR review-fix-scoped to keep the diff small.

ggshield/core/plugin/downloader.py
benjaminrigaud-gg: Agreed; deferring to the same follow-up PR as #4. The intent is to make `download_from_*` only download and `install_*` only install, with a single post-download flow (verify → register → cleanup) reused across all sources.

ggshield/core/plugin/downloader.py
benjaminrigaud-gg: Fixed in 44861745 — rewrote the comment to spell out the multi-wheel-artifact reasoning. You're right that game-over is game-over once an attacker controls the upstream; the comment now says we just don't want to add ordering non-determinism on top.

benjaminrigaud-gg:
Inline comments resolved as of 44861745. The structural feedback (boundaries between `PluginAPIClient` / `PluginDownloader`, duplicated post-download flow, repeated command-side error handling) is the right call — addressing it as a separate PR (Tranche 2) so this one stays narrow. Thanks for the careful review.

ggshield/cmd/plugin/install.py
```diff
@@ -186,30 +189,32 @@ def _install_from_gitguardian(
...
-            download_info, plugin_name, signature_mode=signature_mode
-        )
-
-        # Enable in config
-        enterprise_config.enable_plugin(plugin_name, version=download_info.version)
+        with plugin_api_client.download_plugin(
+            plugin_name, platform_info=platform_info, version=version
+        ) as (info, chunks):
+            bundle_bytes = (
+                plugin_api_client.download_signature_bundle(info.signature_url)
```
clement-tourriere: `download_signature_bundle` raises `PluginAPIError` on any bundle fetch failure (network, oversize, foreign origin, 5xx). With `--allow-unsigned` (WARN mode) the user has explicitly opted out of signature checks, but a broken `X-Plugin-Signature-URL` here still falls through to the generic `except Exception` below and aborts the install. Could we catch `PluginAPIError` in WARN mode, log a warning, and continue with `bundle_bytes=None`? STRICT should keep failing fast. Same pattern in `update.py:307`.

No test currently covers `--allow-unsigned + broken signature URL`, worth adding one.

ggshield/core/plugin/downloader.py
```diff
@@ -663,11 +817,11 @@ def get_installed_signature_label(self, plugin_name: str) -> Optional[str]:
             return None
 
         trusted_unsigned = False
-        wheel_path = self.get_wheel_path(plugin_name)
-        if wheel_path is not None:
+        stored_sha256 = manifest.get("sha256")
```
clement-tourriere: This passes the manifest-stored SHA into `is_trusted`, but `PluginTrustStore.is_trusted` is documented as comparing against the *exact wheel hash on disk*. Feeding it `manifest["sha256"]` makes the comparison a tautology: the trust record was written from the same value, so it always matches as long as the manifest is intact.

The previous implementation on `main` computed the SHA from `get_wheel_path()` — could we restore that? Otherwise a wheel tampered with after install (manifest left untouched) keeps rendering as `unsigned (trusted)` in `plugin list` / `plugin status`.

ggshield/core/plugin/wheel_utils.py
```diff
@@ -120,3 +120,22 @@ def validate_wheel_file(wheel_path: Path) -> bool:
...
+    Strips path components the server may have included and rejects
+    values that would resolve outside the plugin directory (``..``,
+    empty segment, embedded NUL, backslash). Requires a ``.whl``
+    suffix.
+
+    Raises:
+        InvalidWheelError: when the input doesn't satisfy the rules.
+    """
+    name = PurePosixPath(raw).name
+    if not name or name in {".", ".."} or "\x00" in name or "\\" in name:
```
clement-tourriere: `PurePosixPath(raw).name` doesn't strip Windows drive prefixes, so a server returning e.g. `filename="D:evil-1.0-py3-none-any.whl"` passes this check. Downstream, `Path(plugins_dir) / "D:evil.whl"` on Windows is drive-relative and discards the intended base path — the temp wheel, the sigstore bundle and the manifest tmp-rename all land outside `plugins_dir`.

Could we also reject `:` (or use `ntpath.splitdrive` to catch drive-qualified / `\\?\C:\...` shapes) and add parametrised cases for these in `TestSanitizeWheelFilename`?

ggshield/core/plugin/downloader.py
```diff
@@ -204,28 +345,26 @@ def install_from_wheel(
...
 
-        # Create plugin directory
+        # Verify on the caller-provided wheel path — the bundle (if any)
+        # lives alongside the source wheel. Copying first would leave a
+        # rejected wheel under plugins/<name>/ on STRICT failure.
+        sig_info = verify_wheel_signature(wheel_path, signature_mode)
+
+        sha256 = compute_file_sha256(wheel_path)
+
         plugin_dir = self.plugins_dir / plugin_name
```
clement-tourriere, line 356: `install_from_wheel` uses the raw METADATA `Name:` field for the directory, but `download_and_install` canonicalises via `_wheel_distribution_name` (PEP 503: `lower()` + `_ -> -`). A wheel with `Name: Foo_Bar` ends up under `plugins_dir/Foo_Bar/` for the local-file flow and `plugins_dir/foo-bar/` for the platform flow — two parallel directories for the same package, which is the case the PR description says we're trying to avoid.

Could we canonicalise the local name too, e.g. reuse `_wheel_distribution_name(wheel_path.name)` so both code paths share one helper (or `packaging.utils.canonicalize_name` on `metadata.name`)?

clement-tourriere (changes requested):
Overall direction looks good and the convergence story is the right move. CI is green, coverage is solid. Four things I'd want fixed before merge — left as inline comments. Also a handful of smaller non-blocking notes:

- `download_signature_bundle` uses `stream=True` but never `.close()`s the response on error.
- Same-filename reinstall can lose the previous wheel if `_sync_trust_record` / `_write_manifest` fail after `temp_path.replace(wheel_path)` — the except branch then deletes the new wheel too.
- `plugin update --all` returns early when there are no updates and skips the "cannot be auto-updated" footer at the bottom of the function.
- Source labels are humanized in `plugin list` (`local_file` -> `local file`) but `update.py` still shows the raw enum value.
- The `except PluginAPIError: raise` lines in `download_plugin` / `download_signature_bundle` are no-ops (no broader handler below catches them) — fine to drop.

Thanks for the careful work on the verify-before-swap flow, that part reads really nicely.

ggshield/cmd/plugin/install.py
benjaminrigaud-gg: Done in e1a3c0a4. Wrapped the `download_signature_bundle` call in both `install.py` and `update.py` so STRICT keeps re-raising and WARN catches `PluginAPIError`, surfaces it via `ui.display_warning`, and continues with `bundle_bytes=None`. Added two tests in `test_install.py`: `--allow-unsigned` + broken bundle URL → install succeeds with `bundle_bytes=None`; default mode + same broken URL → install aborts before `download_and_install`.

ggshield/core/plugin/downloader.py
benjaminrigaud-gg: Good catch. Restored the pre-rewrite behaviour in e1a3c0a4: `get_installed_signature_label` now hashes the on-disk wheel via `compute_file_sha256(get_wheel_path(...))` and passes that to `is_trusted`. Added `test_signature_label_drops_trusted_when_wheel_tampered` — install + trust, then overwrite the wheel bytes on disk → label drops from `unsigned (trusted)` back to `missing`.

ggshield/core/plugin/wheel_utils.py
benjaminrigaud-gg: Done in eea666c4. `sanitize_wheel_filename` now rejects `:` alongside backslash/NUL/empty/dot — catches `D:evil.whl`, `\\?\C:\…`, and `\\share\…`. Cases added to the parametrised `test_rejects_unsafe`.

ggshield/core/plugin/downloader.py
benjaminrigaud-gg: Done in 8a25a96c. `install_from_wheel` now derives `plugin_name` via `_wheel_distribution_name(wheel_path.name)` instead of the raw METADATA `Name`, so a wheel whose `Name:` is `Foo_Bar` lands under `plugins_dir/foo-bar/` — the same directory `download_and_install` writes for the catalog flow. New `test_install_from_wheel_canonicalises_distribution_name` exercises the non-canonical case.

benjaminrigaud-gg:
@clement-tourriere — thanks, all five non-blocking notes addressed in this push too:

- `download_signature_bundle` now wraps the streamed response in `with self.client.session.get(...) as response:`, so the connection closes on every error path. (e1a3c0a4)
- Post-swap `_sync_trust_record`/`_write_manifest` failures no longer trigger `_cleanup_failed_install`: the verified wheel stays on disk and a retry rewrites manifest/trust. Existing test re-purposed to assert the wheel is preserved. (8b596dd3)
- `update --all`'s no-updates-available branch now calls the same `_display_non_updatable_footer` helper as `--check`, so the user sees their non-auto-updatable installs even when everything else is current. (1c36e3a6)
- Source labels in `update.py` go through `.value.replace("_", " ")` like `plugin list` — `local_file` → `local file` consistently. (1c36e3a6)
- `except PluginAPIError: raise` dropped from `download_plugin` and `download_signature_bundle`. (e1a3c0a4)

Branch has also been rebased onto main; 14/14 QA scenarios still passing.

ggshield/cmd/plugin/update.py
```diff
@@ -187,10 +188,16 @@ def update_cmd(
...
+                            "source_type": PluginSourceType.PLATFORM,
                         }
                     )
 
+        except PluginsNotEnabledError:
+            ui.display_error(
+                "Plugin system is not available on this workspace. "
+                "Contact your administrator."
+            )
+            ctx.exit(ExitCode.UNEXPECTED_ERROR)
```
6d7a: This also aborts all updates of other non-platform plugins (or those of unknown source).

ggshield/cmd/plugin/install.py
```diff
@@ -186,30 +189,48 @@ def _install_from_gitguardian(
...
+            downloader.download_and_install(
+                info,
+                chunks,
+                plugin_name,
+                signature_mode=signature_mode,
+                bundle_bytes=bundle_bytes,
+            )
 
-        # Save config
+        enterprise_config.enable_plugin(plugin_name, version=info.version)
```
6d7a: We don't need to address this in this PR, but currently we enable by default, regardless of whether the plugin was disabled before. We should either fix this or just drop the `enable` toggle completely and just rely on install/uninstall.

ggshield/core/plugin/client.py
```diff
@@ -133,118 +196,198 @@ def get_available_plugins(self) -> PluginCatalog:
...
+        plugins_data = response.json()
 
         return PluginCatalog(
-            plan=account_data.get("plan", data.get("plan", "unknown")),
-            features=account_data.get("features", data.get("features", {})),
             plugins=[
                 PluginInfo(
-                    name=p.get("name", "unknown"),
-                    display_name=p.get("display_name", p.get("name", "Unknown")),
+                    name=p["reference"],
```
6d7a, line 282: Did you remove the unknown fallback we had before for a particular reason? As far as I see, a KeyError will not be caught here.

ggshield/core/plugin/client.py
```diff
@@ -133,118 +196,198 @@ def get_available_plugins(self) -> PluginCatalog:
...
                     description=p.get("description", ""),
-                    available=self._is_plugin_available(p, current_platform),
-                    latest_version=p.get("latest_version"),
-                    supported_platforms=p.get("supported_platforms", []),
-                    reason=self._get_unavailable_reason(p, current_platform),
+                    available=p.get("available", False),
+                    latest_version=(
+                        p["releases"][0]["version"] if p.get("releases") else None
+                    ),
+                    supported_platforms=[],
```
6d7a: Do we still get data about platform support from the server? If not, we should remove this field.

ggshield/core/plugin/client.py
```diff
@@ -133,118 +196,198 @@ def get_available_plugins(self) -> PluginCatalog:
...
-                f"{self.base_url}/{self.API_VERSION}/plugins/{plugin_name}/download",
+                f"{self.base_url}/{self.API_VERSION}/endpoints/plugins/{reference}/download",
                 params=params,
-                headers=self._get_headers(),
+                stream=True,
+                timeout=HTTP_TIMEOUT_SECONDS,
             )
+            assert_all_https(response, exc_factory=PluginAPIError)
+            if response.status_code in (403, 404):
+                # Surface the server's `detail` (e.g. "No active release
```
6d7a: nit: this comment is rather self-explanatory

ggshield/core/plugin/client.py
```diff
@@ -133,118 +196,198 @@ def get_available_plugins(self) -> PluginCatalog:
...
+
+                size_bytes = int(response.headers.get("Content-Length", 0))
+                if size_bytes > MAX_BUNDLE_SIZE_BYTES:
+                    raise PluginAPIError(
+                        f"Signature bundle size {size_bytes} exceeds maximum "
+                        f"of {MAX_BUNDLE_SIZE_BYTES} bytes"
+                    )
+
+                buffer = bytearray()
+                for chunk in response.iter_content(chunk_size=65536):
```
6d7a: This duplicates `_iter_with_size_cap`

ggshield/cmd/plugin/update.py
benjaminrigaud-gg: Fixed in fa53de07. Restructured the three `except` arms into a single `catalog_error` variable; when there are any `github_release_plugins` in the batch we surface as a warning and continue to the GitHub-releases loop instead of `ctx.exit`-ing. When the platform fetch is the only thing in scope (single-plugin update of a platform plugin, or `--all` with no github_release plugins installed) we keep the hard error so the exit code still reflects the failure.

Covered by `test_update_catalog_failure_does_not_abort_github_release_check` (existing single-plugin error tests still pass unchanged).

ggshield/cmd/plugin/install.py
benjaminrigaud-gg: Agreed and tracked as [END-243](https://linear.app/gitguardian/issue/END-243/ggshield-plugin-install-enable-is-implicit-on-every-install) so it doesn't get lost. Leaning toward option (2) — drop the `enable`/`disable` toggle entirely and make `install` / `uninstall` the single source of truth — which pairs naturally with Tranche 2's `PluginDownloader` / `PluginAPIClient` reshuffle that's already deferred in [END-30](https://linear.app/gitguardian/issue/END-30).

ggshield/core/plugin/client.py
benjaminrigaud-gg: Good catch — confirmed. The PoC had `p.get("name", "unknown")`; the rewrite tightened to `p["reference"]` and the surrounding `except RequestException` doesn't catch `KeyError`, so a malformed row would leak as a raw traceback. Fixed in cea46249: pulled `response.json()` + the comprehension into a `try` with `(KeyError, TypeError, ValueError)` re-raised as `PluginAPIError("Malformed plugin catalog response: …")` — same envelope as every other catalog failure mode. Two new tests in `test_client.py` cover the missing-field and non-JSON-body cases.

ggshield/core/plugin/client.py
benjaminrigaud-gg: Confirmed — the backend's `PluginListItemOutputSchema` returns `reference / display_name / available / reason / releases`, no top-level `supported_platforms`. We were hardcoding it to `[]` everywhere, which made `is_platform_supported` a constant `True` function with no live production caller (only `test_client.py` exercised the method). Platform filtering already happens server-side via the `?platform=` query param.

Removed in 7affc788: dropped the `supported_platforms` field, the `is_platform_supported` method, the three dead unit tests, and the now-unused `field` import.

ggshield/core/plugin/client.py
benjaminrigaud-gg: Agreed — gone in 7affc788. The block collapsed to:

```python
detail = _extract_server_detail(response)
if not detail and response.status_code == 404:
    detail = "Plugin or version not found"
raise PluginNotAvailableError(reference, detail)
```

`_extract_server_detail`'s name carries the intent.

ggshield/core/plugin/client.py
benjaminrigaud-gg: You're right — the bundle path was open-coding the same size-cap loop. Fixed in ad58cca0:

```python
buffer = bytearray()
for chunk in _iter_with_size_cap(
    response.iter_content(chunk_size=65536), MAX_BUNDLE_SIZE_BYTES
):
    buffer.extend(chunk)
return bytes(buffer)
```

The pre-flight `Content-Length` check keeps the bundle-specific phrasing; the streamed-cap overflow now reads "Response body exceeded maximum size of N bytes" (matches what the wheel proxy raises) which I think is fine — the pre-check already names the bundle when the server is honest about the size, and the streamed-overflow case is the same class of error in both paths.

benjaminrigaud-gg:
@6d7a — thanks, all 6 comments addressed in this push (`3e00d7fa..ad58cca0`):

| # | File | Fix | Commit |
|---|---|---|---|
| 1 | `update.py` | github_release updates now survive a platform catalog failure | fa53de07 |
| 2 | `install.py` (`--enable` implicit) | Deferred — tracked as [END-243](https://linear.app/gitguardian/issue/END-243/ggshield-plugin-install-enable-is-implicit-on-every-install), leaning toward dropping the toggle entirely in Tranche 2 | — |
| 3 | `client.py:216` (KeyError on `reference`) | Catalog parse now re-raises as `PluginAPIError("Malformed plugin catalog response: …")` | cea46249 |
| 4 | `client.py:223` (`supported_platforms`) | Dead field + method + tests removed | 7affc788 |
| 5 | `client.py:263` (long comment) | Comment trimmed, block simplified | 7affc788 |
| 6 | `client.py:361` (`_iter_with_size_cap` dup) | Bundle path reuses the helper | ad58cca0 |

395/395 plugin tests pass locally.

ggshield/core/plugin/client.py
```diff
@@ -133,118 +186,188 @@ def get_available_plugins(self) -> PluginCatalog:
 
         try:
             response = self.client.session.get(
-                f"{self.base_url}/{self.API_VERSION}/plugins",
+                f"{self.base_url}/{self.API_VERSION}/endpoints/plugins",
```
clement-tourriere, line 260: This is an authenticated request, but the platform client never pre-validates that `base_url` is HTTPS before sending the token-bearing request. `download_plugin()` / `download_signature_bundle()` call `assert_all_https()` only after the request has already been sent, and `get_available_plugins()` / `report_installation()` don't assert the redirect chain at all. Since `GGClient` accepts both `http://` and `https://` base URIs, a misconfigured instance URL can leak the API token in cleartext. Please validate the target/base URL before every authenticated platform request (preserving the explicit loopback bypass) and still assert the response history after the request.

ggshield/core/plugin/client.py
```diff
@@ -133,118 +186,188 @@ def get_available_plugins(self) -> PluginCatalog:
...
+                    f"Server response missing X-Plugin-SHA256 header for {reference}"
+                )
+            resolved_version = response.headers.get("X-Plugin-Version")
+            if not resolved_version:
+                raise PluginAPIError(
+                    f"Server response missing X-Plugin-Version header for {reference}"
                 )
 
-            response.raise_for_status()
+            size_bytes = int(response.headers.get("Content-Length", 0))
```
clement-tourriere: A malformed `Content-Length` header will raise a raw `ValueError` here (same in `download_signature_bundle()`), bypassing the typed `PluginAPIError` path the CLI handles. Could you parse this through a small helper that catches `TypeError`/`ValueError` and raises `PluginAPIError("Malformed Content-Length ...")`, with tests for both wheel and bundle responses?

ggshield/core/plugin/client.py
```diff
@@ -133,118 +186,188 @@ def get_available_plugins(self) -> PluginCatalog:
...
-    def _get_headers(self) -> Dict[str, str]:
-        return {
-            "Authorization": f"Token {self.client.api_key}",
-            "Content-Type": "application/json",
-        }
+        Never raises — a failure here must not fail the install.
+        """
+        try:
+            response = self.client.session.post(
+                f"{self.base_url}/{self.API_VERSION}/endpoints/plugins/{reference}/installed",
```
clement-tourriere, line 448: `report_installation()` is documented as best-effort, but this POST has no explicit timeout. In `update.py` it is called before `enterprise_config.enable_plugin()` and before the final `save()`, so a stalled `/installed` endpoint can leave the command hanging after the wheel was installed but before config state is persisted. Please pass `timeout=HTTP_TIMEOUT_SECONDS` here, and consider moving the update reporting after the config has been updated/saved.

ggshield/cmd/plugin/install.py
```diff
@@ -252,7 +271,6 @@ def _install_from_local_wheel(
             wheel_path, signature_mode=signature_mode
         )
 
-        # Enable in config
         enterprise_config.enable_plugin(plugin_name, version=version)
```
clement-tourriere: This enables the name returned by the downloader (now the wheel distribution name for local wheels). However `PluginLoader.discover_plugins()` keys wheel-installed plugins by their `ggshield.plugins` entry-point name when present, and checks enablement with that key. For wheels whose distribution/package name differs from the entry-point name, install will report success but `plugin list` will show the plugin disabled and it won't load. Please derive the config key from the installed wheel's entry-point name when present (falling back to the distribution name), and apply the same rule to URL/GitHub-release/artifact installs.

ggshield/core/plugin/downloader.py
```diff
@@ -842,21 +964,65 @@ def _cleanup_failed_install(self, wheel_path: Path) -> None:
...
+          finished writing), and
+        - it has a ``manifest.json`` (so it really is a plugin
+          directory, not some unrelated path).
+        """
+        if not self._is_valid_plugin_name(catalog_reference):
+            return
+        legacy_dir = self.plugins_dir / catalog_reference
+        if not legacy_dir.exists() or legacy_dir.resolve() == current_dir.resolve():
+            return
+        if not (legacy_dir / "manifest.json").exists():
```
clement-tourriere: Preserving a bare `plugins_dir/<catalog_reference>/` directory here can still break the migration path: `_resolve_plugin_dir()` returns any direct path before trying the entry-point scan, and `_get_manifest_path()` then gives up when this directory has no manifest. That means `status`/`uninstall <catalog_reference>` can miss the new install under the wheel distribution directory. Either remove/ignore such stale non-plugin dirs during cleanup, or make `_resolve_plugin_dir()` fall back to `_find_plugin_dir_by_entry_point()` when the direct path lacks a manifest.

ggshield/core/plugin/downloader.py
```diff
@@ -842,21 +964,65 @@ def _cleanup_failed_install(self, wheel_path: Path) -> None:
...
+            return
+        try:
+            shutil.rmtree(legacy_dir)
```
clement-tourriere, line 1092: This cleanup deletes any distinct directory named after the catalog reference as long as it has a `manifest.json`. That can delete a real user-installed local/URL/GitHub plugin that happens to use the same directory name when the platform wheel distribution differs from the catalog reference. Please read the manifest and only remove directories that are clearly previous platform/legacy GitGuardian installs (for example source missing/`platform`/legacy `gitguardian_api`, ideally with matching entry point), and add a regression proving `local_file` / `url` / GitHub sources are preserved.

ggshield/cmd/plugin/update.py
```diff
@@ -187,16 +189,36 @@ def update_cmd(
...
+        if catalog_error is not None:
+            # Degrade rather than abort when there's still other work to do:
+            # GitHub-release plugins and the non-updatable footer don't depend
+            # on the GitGuardian catalog, so a single failed fetch shouldn't
+            # kill an otherwise valid ``update --all`` invocation. When the
+            # platform fetch is the only thing we were asked to do (no
+            # github_release plugins in the batch), surface as a hard error
+            # so the exit code still reflects the failure.
+            if github_release_plugins:
+                ui.display_warning(
```
clement-tourriere, line 239: When we skip GitGuardian-hosted updates here and GitHub-release checks find no updates, the later `--check` branch still prints `All updatable plugins are up to date.` That is misleading because platform plugins were not checked. Please track that the platform check was skipped and adjust the empty-result message (for example `No updates found for checked plugins; GitGuardian-hosted plugins were skipped`) with a no-GitHub-update regression.

clement-tourriere (changes requested):
Thanks for the follow-up fixes. I re-reviewed the current head and ran the focused plugin unit suite locally (`285 passed`). I still see a few issues that should be addressed before merge, especially around authenticated HTTP validation and legacy cleanup potentially deleting user installs. Details inline; the remaining lower-severity observations are included here for completeness.

Additional non-inline notes:
- `download_and_install()`'s docstring says the catalog reference is stored in the manifest, but the current call writes the wheel distribution name as `plugin_name`; please either store the catalog reference explicitly (for example in `source.reference`) or update the contract.
- URL/GitHub-artifact installs still use `METADATA Name` for the install directory while local/platform now use the wheel filename's PEP 503 distribution name, so the same wheel can still land in different directories depending on source. It would be worth normalizing those paths too, or tracking as a follow-up if deliberately out of scope.

6d7a:
@benjaminrigaud-gg can you also squash or rewrite your commit history on this branch before you merge? Thanks!

clement-tourriere (approved):

benjaminrigaud-gg:
@clement-tourriere — thanks, all 7 R3 comments addressed in `6fe73708`. A subsequent full-PR review surfaced three follow-ups (one of which was a miss in my own R3 work — the update flow had the same config-key bug as install); those are folded into the same commit.

| # | File | Fix | Where |
|---|---|---|---|
| 1 | `client.py` | `_assert_base_url_https` (loopback bypass via `GITGUARDIAN_ALLOW_INSECURE_LOOPBACK`) on all 4 authenticated endpoints + `assert_all_https` now wired on `get_available_plugins` and `report_installation` too. `_origin()` helper normalises default ports so `https://host` ↔ `https://host:443` compare equal in `download_signature_bundle`. | `client.py` |
| 2 | `client.py` | `_parse_content_length` raises `PluginAPIError` on malformed values (was raw `ValueError`); used by wheel + bundle paths. | `client.py` |
| 3 | `client.py` / `update.py` | `report_installation` POST passes `timeout=HTTP_TIMEOUT_SECONDS`; in `update.py` the report calls are buffered and fired after `enterprise_config.save()` so a hang can't desync wheel-on-disk from config. | `client.py`, `update.py` |
| 4 | `install.py` / `update.py` | Hoisted to `loader.py:resolve_config_key`. Applied to all 5 install paths (platform / local-wheel / url / github-release / github-artifact) AND both update paths (PLATFORM / GITHUB_RELEASE) so the row written by `enable_plugin` matches the loader's enablement key. | `loader.py`, `install.py`, `update.py` |
| 5 | `downloader.py` | `_resolve_plugin_dir` now returns the direct path only when `manifest.json` is present; otherwise falls through to entry-point scan. | `downloader.py` |
| 6 | `downloader.py` | `_cleanup_legacy_install_dir` parses the candidate manifest and only removes when `source.type ∈ {platform, gitguardian_api}`. `local_file` / `url` / `github_*` installs whose name collides with the catalog reference are preserved. | `downloader.py` |
| 7 | `update.py` | `platform_checked` flag drives the empty-results message: `No updates found for checked plugins; GitGuardian-hosted plugins were skipped.` when the catalog fetch failed and we degraded to github-only. | `update.py` |
| bonus | `downloader.py` | New `_remove_stale_wheels()` sweep after every install path; upgrades no longer leave the previous `*.whl` + `.sigstore` sidecar in the plugin dir. | `downloader.py` |

**Validation**: 427/427 plugin unit tests pass (+10 new); 17/17 QA scenarios pass against a live backend (added 3 new — config-key, legacy-cleanup preservation, legacy-cleanup still removes platform).

Resolving the 7 threads since the fixes are now on disk; let me know if you'd prefer any reopened.

---

**Note**: PR commits were restructured pre-merge into a clean 2-commit shape (refactor / feat, with the ``.worktrees/`` chore folded into feat) so the merged history reads as logical units rather than chronological build order. Branch tip: `6fe73708`. R3 fixes are now folded into the `feat(plugin): install plugins from the GitGuardian platform` commit.

PR #1236, opened 2026-05-07
fix(plugins): skip TUF update on --allow-unsigned

6d7a:
## Context

`ggshield plugin install --allow-unsigned` and `ggshield plugin update --allow-unsigned` currently fail if TUF metadata cannot be successfully updated. This blocks plugin installation for the user entirely.

## What has been done

When passing `--allow-unsigned`, the verifier is now built with `offline=True`, which makes `sigstore-python` use the embedded/cached trust root and skip the TUF metadata refresh. `STRICT` mode is unchanged: it continues to refresh TUF and fail closed on network errors.

## Validation

- block egress to `tuf-repo-cdn.sigstore.dev` (e.g. `127.0.0.1 tuf-repo-cdn.sigstore.dev` in `/etc/hosts`)
- `uv run ggshield plugin install <wheel> --allow-unsigned` should now succeed (previously failed with a TUF refresh error)
- `uv run ggshield plugin install <wheel>` (no flag, STRICT) should still fail with a TUF refresh error — verifying the offline opt-in is gated on `--allow-unsigned`
- `uv run pytest tests/unit/core/plugin/test_signature.py`

## PR check list


benjaminrigaud-gg (approved):
LGTM

PR #1218, opened 2026-04-27
fix(client): retry transient connection errors and 5xx responses

benjaminrigaud-gg:
## Context

Customer report tracked in [END-176](https://linear.app/gitguardian/issue/END-176/ggshield-connection-aborted-error-message): a `ggshield secret scan` on a 25k-file repository against the SaaS API fails with:

```
Scanning failed: ('Connection aborted.', ConnectionResetError(54, 'Connection reset by peer'))
```

Quota was untouched, so it's not a 429. The error surfaces in `ggshield/core/errors.py:217` (`status_code=None`) and exits with code 128.

## Root cause

`ggshield/core/client.py` mounts an `HTTPAdapter` with `pool_maxsize=100` (added in [`ba63b2ff`](https://github.com/GitGuardian/ggshield/commit/ba63b2ff) / #1141, shipped in v1.46.0) but **no retry policy**. A single transient TCP reset on one of the many concurrent POST `multi_content_scan` batches therefore aborts the whole scan; there is no recovery for connection errors or 5xx responses.

The 100-slot pool added in #1141 lets ggshield run a lot more in-flight POST batches than before, which raises the probability that a customer scanning a very large repo hits a transient reset on at least one batch — exactly what's happening here.

A draft adapter with retry config did exist on the never-merged debug branch [`agateau/debug-555-2`](https://github.com/GitGuardian/ggshield/commits/agateau/debug-555-2) ([4d60dd1e](https://github.com/GitGuardian/ggshield/commit/4d60dd1e), April 2024, related to open issue #555), but it was never landed on `main`.

## What has been done

Add a retry policy to the existing pool-tuned adapter:

```python
retries = urllib3.Retry(
    total=5,
    backoff_factor=0.2,
    status_forcelist=[502, 503, 504],
    allowed_methods=frozenset({"HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE", "POST"}),
)
adapter = HTTPAdapter(pool_maxsize=100, max_retries=retries)
```

- `total=5` with `connect=None` makes urllib3 retry connection errors (`ConnectionResetError` and friends) too, not just response statuses.
- `backoff_factor=0.2` gives 0.2 / 0.4 / 0.8 / 1.6 / 3.2 s waits — bounded, gentle on the SaaS edge.
- `status_forcelist=[502, 503, 504]` retries transient gateway errors but lets real failures (4xx, 500) propagate.
[...]

PR #1237, opened 2026-05-07
fix(client): scope retry policy by command

benjaminrigaud-gg:
## Context

Follow-up to #1218 ([END-176](https://linear.app/gitguardian/issue/END-176/ggshield-connection-aborted-error-message)). The retry policy restored in #1218 (`total=5, backoff_factor=0.2`, ~6 s worst case, no jitter) is too short to outlast a typical SaaS edge incident on large multi-batch scans, and customers kept reporting `Connection aborted / ConnectionResetError` failures after it shipped.

An earlier draft of this PR bumped the policy to a single, session-wide `total=8, backoff_factor=0.3, backoff_jitter=0.5, backoff_max=20` (~58 s worst case). Code review caught that **GitHub Enterprise Server enforces a fixed 5 s timeout shared across all pre-receive hooks** ([docs](https://docs.github.com/en/enterprise-server@3.17/admin/enforcing-policies/enforcing-policy-with-pre-receive-hooks/about-pre-receive-hooks)), so any retry budget that adds measurable wall clock would break `secret scan pre-receive` on GHES — exactly the customers most reliant on it.

## What has been done

Introduce two named retry profiles and select per command via a keyword-only argument threaded through `create_session` / `create_client` / `create_client_from_config`. The retry policy lives in one place (`ggshield/core/client.py`) and the only command that opts out of the default is `prereceive.py`.

```python
class RetryProfile(Enum):
    DEFAULT = "default"           # ~15 s budget, jittered
    PRE_RECEIVE = "pre_receive"   # 1 immediate retry, no backoff
```

- **`DEFAULT`** — `total=5, backoff_factor=0.5, backoff_max=8, backoff_jitter=0.5`. Sleep schedule before each retry: `0, 1, 2, 4, 8 s`, each (after the first) jittered by up to 0.5 s; the 8 s sleep is capped by `backoff_max`. Worst case wall clock ≈ 15 s (≈ 16.5 s with maximum jitter). Long enough to absorb a typical SaaS edge blip; jitter de-synchronizes the up-to-`pool_maxsize=100` batches that can fail together when a load balancer aborts many in-flight POSTs at once. Used by every command **except** pre-receive.
- **`PRE_RECEIVE`** — `total=1, backoff_factor=0, backoff_jitter=0`. One immediate retry, no backoff, ~0 s added wall clock. Recovers the instant `ConnectionResetError` that motivated the original ticket without eating into GHES's 5 s budget. Used by `secret scan pre-receive` only.
- `status_forcelist=[502, 503, 504]` and `allowed_methods` (with POST, since scan endpoints are POST-based) are unchanged from #1218.
- `pool_maxsize=100` is unchanged.

`urllib3 >= 2.2.2` (already a hard dependency in `pyproject.toml`) exposes `backoff_max` and `backoff_jitter`, so no dependency change is needed.

## Why scope per command instead of a single policy

A single ~15 s session-wide policy would still risk exceeding GHES's 5 s pre-receive timeout in the worst case (initial attempt + first retry + connection time). A single ~0 s policy regresses #1218 for all other commands. The two-profile split keeps the policy decision local to where it's set (one line in `prereceive.py`), with `DEFAULT` covering every other call site (including `pre-commit`, `pre-push`, `ai_hook`, and the entire CLI surface for `auth`, `quota`, `status`, etc., which run on developer machines with no hard server-side timeout).

## Validation

- `tests/unit/core/test_client.py` — replaced the single retry-config test with three: `test_create_session_default_retry_profile`, `test_create_session_pre_receive_retry_profile`, `test_create_client_threads_retry_profile`. All pass.
- `tests/unit/cmd/scan/test_prereceive.py` (10 tests) — all pass; pre-receive command still behaves correctly with the minimal profile.
- Wider sweep across `tests/unit/cmd/scan/` and `tests/unit/core/test_client.py` shows 153 passes (8 pre-existing pyfakefs teardown E's, unrelated and reproduced on the unchanged baseline).

## PR check list

### DIFF
--- a/changelog.d/20260507_140000_brigaud_tune_retry_backoff.md
+++ b/changelog.d/20260507_140000_brigaud_tune_retry_backoff.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Scans no longer fail on a single transient network glitch. ggshield retries connection errors (e.g. `ConnectionResetError`) and 502/503/504 responses with bounded exponential backoff (~15 s budget with jitter). `ggshield secret scan pre-receive` uses a minimal retry policy instead so it stays inside GitHub Enterprise Server's fixed 5 s pre-receive hook timeout.
--- a/ggshield/cmd/secret/scan/prereceive.py
+++ b/ggshield/cmd/secret/scan/prereceive.py
@@ -16,7 +16,7 @@ from ggshield.cmd.utils.common_decorators import fail_on_server_error_option
 from ggshield.cmd.utils.context_obj import ContextObj
 from ggshield.core import ui
 from ggshield.core.cache import ReadOnlyCache
-from ggshield.core.client import create_client_from_config
+from ggshield.core.client import RetryProfile, create_client_from_config
 from ggshield.core.config import Config
 from ggshield.core.errors import ExitCode, handle_exception
 from ggshield.core.git_hooks.prereceive import (
@@ -93,7 +93,11 @@ def prereceive_cmd(
     configuration.
     """
     ctx_obj = ContextObj.get(ctx)
-    ctx_obj.client = create_client_from_config(ctx_obj.config)
+    # GitHub Enterprise Server enforces a fixed 5 s timeout shared across all
+    # pre-receive hooks. Use a minimal retry profile to stay inside it.
+    ctx_obj.client = create_client_from_config(
+        ctx_obj.config, retry_profile=RetryProfile.PRE_RECEIVE
+    )
     config = ctx_obj.config
     output_handler = create_output_handler(ctx)
     if os.getenv("GL_PROTOCOL") == "web":
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from enum import Enum
 from typing import Optional
 
 import requests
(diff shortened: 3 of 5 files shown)


6d7a (changes requested):
Thanks for the MR. I don't think this is a sustainable solution for the issue though. If the user having [OOM issues](https://linear.app/gitguardian/issue/END-176/ggshield-connection-aborted-error-message) is lucky, the jitter will prevent waves of concurrent retry requests and maybe even the OOM, but I doubt that, since the config the user shared is below what we recommend [here](https://docs.gitguardian.com/self-hosting/management/infrastructure-management/scaling). Extending retries also impacts `ggshield` hooks, and ~60s of retry time vs ~6s before can lead to a significant UX impact. Adding jitter to the existing retry policy still makes sense, though.

benjaminrigaud-gg:
@6d7a 

> Thanks for the MR. I don't think this is a sustainable solution for the issue though. If the user having [OOM issues]

They have both SaaS & On-Prem IIRC.

The Retries tweaks is more for SaaS mode: we don't want to retry too fast if any network point (Load Balancer / Proxy) Is having some transient issue.

They have to tweak their On-Prem Instance to fix the OOM Kill issues: I suggested that in the ticket.

**edited: handling now different retry policies depending on the command.**

6d7a (approved):

PR #1239, opened 2026-05-12
build(docker): use wolfi python as base

6d7a:
## Context

The current docker base image is vulnerable to CVE-2026-33845.

## What has been done

Use `ghcr.io/gitguardian/wolfi/python:3.10-dev` as a build image and `cgr.dev/chainguard/wolfi-base` as a runtime base.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,20 @@
-FROM python:3.10-slim AS build
+FROM ghcr.io/gitguardian/wolfi/python:3.10-dev AS build
+
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PYTHONFAULTHANDLER=1
+
+WORKDIR /app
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.10-slim AS build
```
clement-tourriere, line 1: I don't think we should use this image as base image since it's huge ! 

```
ghcr.io/gitguardian/wolfi/python:3.10-dev                                       9e5d28054a50        814MB             0B
python:3.10-slim                                                                57d9edee8ad2        147MB             0B
```

We should instead find a better base image ! Let me do some tests

clement-tourriere: I pushed a new image, I'm not 100% sure it will perfectly work for all envs, but tested it locally and it's working

benjaminrigaud-gg:
The  Wolfi images provided by chainguard are going to be charged. I'll keep you updated :zoi:

PR #1240, opened 2026-05-12
Split VSCode/Copilot CLI and fix hooks

paulpetit-gg-ext:
## Context

Some customers notified us that the hooks weren't working for their developpers.

After some investigation, it turned out that they were using "Copilot CLI", which is a distinct product than VSCode with Copilot Chat.

Although their configuration is mostly interchangeable, they behave differently (distinct payloads, not the same response to hook results, etc.). Most notably, Copilot CLI currently doesn't allow blocking a user prompt, only a tool use.

Some more work will be needed to properly support MCP servers and their differences between VSCode and Copilot CLI, which will be for a subsequent PR.

## What has been done

- the first commit is a small refactoring to have more custom-logic in `Agent`, and remove the inheritance between `Copilot` and `Claude`.
- then a new `VSCode` class replaces what was `Copilot`, which becomes a very light subclass.
  - because they can share the same configuration, from a UX perspective this is like simply having `vscode` and `copilot` being two aliases for the same hook installation
- then Copilot specifics are added so that it is properly recognized and hooks work
- while doing tests, I noticed that VSCode changed its behavior when files were "@-ed" in a prompt. This is fixed in another commit
- add changelog

## Validation

```
ggshield install -t copilot -m global
copilot -p "read the content of this-file-contains-secrets.txt"
```

## PR check list

### DIFF
--- a/changelog.d/20260430_000000_add_codex_ai_hooks.md
+++ b/changelog.d/20260430_000000_add_codex_ai_hooks.md
@@ -1,3 +1,3 @@
 ### Added
 
-- Add Codex support to `ggshield secret scan ai-hook` and `ggshield install -t codex`.
+- Add Codex support to `ggshield secret scan ai-hook` and `ggshield install -t codex`. (thanks to trickyfalcon)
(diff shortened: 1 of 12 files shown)


Blaxav (approved):
That's qualitative, I have no concern, thanks for the refacto !

xblanchot-gg (approved):

PR #1241, opened 2026-05-13
AI hook basic command parsing

paulpetit-gg-ext:
## Context

Some users raised the concern that we may not detect secrets leaked to AI agents in PostTool if the Agent uses bash command instead of a proper "Read" tool to access a file's content.

Our first stance was that trying to perfectly do that is probably a lost battle, and that agents prefer using a "Read" tool anyway.

After some testing with the recent Codex support, we noticed that, at least in Windows, it favors the Powershell "Get-Content" command over an agent-specific "Read" tool.

This PR adds basic support for detecting file reads in bash commands.

## What has been done

- In PreToolUse hooks, detect if the command is of the form `{read_command} {filename}`.
- Two "read_commands" supported for now: `Get-Content` and `cat`.

## Validation

- Explicitly making an agent use a bash command instead of simply prompting "use this file"
- Reading a file with Codex on a Windows computer.

## PR check list

### DIFF
--- a/changelog.d/20260513_155719_paul.petit.ext_HEAD.md
+++ b/changelog.d/20260513_155719_paul.petit.ext_HEAD.md
@@ -1,0 +1,3 @@
+### Changed
+
+- AI hooks naively try to detect file read by shell commands.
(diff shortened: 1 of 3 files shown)


xblanchot-gg (approved):
Thanks !

PR #1242, opened 2026-05-13
fix(ai): include command in notification

paulpetit-gg-ext:
# Context

GitGuardian's Claude Code hook notifies users when the assistant runs a command that exposes secrets. Today the notification reads:

"Claude Code got access to 3 secrets by running a command"

User feedback:

> Hey there,
Claude ran a command and got secrets from it. I got the notification Claude Code got access to 3 secrets by running a command. I'd assume this creates an incident somewhere but I see no actionable thing in either Claude or on the Dashboard.
Any insights on where I can find the detected leaked secrets? (I have them in my Claude transcript)"

## Expected behavior

- The notification should include the exact command that led to the secret access. This lets the user:
  - Identify what Claude Code ran
  - Judge whether the exposure is a concern
  - Take local action (rotate, redact transcript, etc.)

## Scope (for now)

- Notification message only.
- We are explicitly not creating incidents from these detections at this stage.

## PR check list

### DIFF
--- a/changelog.d/20260513_172255_paul.petit.ext_HEAD.md
+++ b/changelog.d/20260513_172255_paul.petit.ext_HEAD.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- (AI hooks): the command that leaked a secret is now shown in the notification message.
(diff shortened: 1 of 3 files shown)


xblanchot-gg (approved):
Nice improvement thanks !

PR #1243, opened 2026-05-15
fix: atomic writes for cache and config files

HelifeWasTaken:
## Summary

> **Note — possible `SIGBUS` under concurrency.** Torn writes on a cache file that another process has memory-mapped trigger `SIGBUS` (bus error, core dumped) on the reader the moment it touches a page that's no longer backed by the file. This has been observed in practice when 5–10 `ggshield secret scan ai-hook` processes were running in parallel from AI-agent `PreToolUse` hooks: ggshield exits with `bus error (core dumped)` and the failure is silent in the surrounding `except` blocks. Native extensions in the dependency tree (`cryptography`, `_cffi_backend`, `pydantic_core`, the mypyc-compiled module, …) and downstream consumers of these caches are exactly the kind of readers that can hit this. Eliminating torn writes — never letting a reader see a half-written / truncated cache file — closes that class of crash.

Several ggshield cache/config writes were unsafe under concurrent invocations:

- `ai_discovery.json` (`ggshield/verticals/ai/cache.py`), `auth_config.yaml` / `update_check.yaml` (`ggshield/core/config/utils.py`), the secret-scan cache (`ggshield/core/cache.py`) and the AI fallback user-id (`ggshield/verticals/ai/user.py`) used **truncate-in-place** writes (`open("w")` / `O_TRUNC`). A second process reading mid-write would see a zero-length or partial file.
- The docker-image ID cache (`ggshield/core/scan/id_cache.py`) and the plugin manifest (`ggshield/core/plugin/downloader.py`) already used a tmp+rename pattern, but with a **fixed `.tmp` filename**, so two concurrent writers would race on the same tempfile.

Both classes of bug surface when several `ggshield secret scan ai-hook` invocations run in parallel from AI-agent `PreToolUse` hooks. Failures are silently swallowed by the surrounding `except OSError: pass` / `except Exception` blocks, so when nothing else goes wrong the visible symptom is just degraded caching — but as called out in the note above, a torn write on a file another process has mmap'd surfaces as `SIGBUS` and kills the reader outright.

## Changes

- Introduce `atomic_write_text` / `atomic_write_bytes` in `ggshield/utils/files.py`. They use `tempfile.mkstemp` (per-process unique name in the destination directory) followed by `os.replace`. POSIX guarantees the rename is atomic — readers see either the previous content or the new content, never a partial write.
- Switch every affected call site to use the helper:
  - `ggshield/verticals/ai/cache.py` (`save_discovery_cache`)
  - `ggshield/verticals/ai/user.py` (UUID fallback)
  - `ggshield/core/config/utils.py` (`save_yaml_dict`, covers `auth_config.yaml` and `update_check.yaml`)
  - `ggshield/core/cache.py` (`Cache.save`)
  - `ggshield/core/scan/id_cache.py` (`IDCache._save`)
  - `ggshield/core/plugin/downloader.py` (`_write_manifest`)
- Add 7 unit tests under `tests/unit/utils/test_files.py`, including a `ThreadPoolExecutor`-based concurrent-writers test that asserts the final file is always one of the inputs (never a torn mix) and no temp files leak.
- Update `tests/unit/core/plugin/test_downloader.py::test_tmp_file_is_cleaned_up_on_rename_failure` to patch `os.replace` (now used by the helper) instead of `pathlib.Path.replace`, and assert no leftover temp files in the directory.

## Test plan

- [x] `pytest tests/unit` — full unit suite (1919 passed)
- [x] `pytest tests/unit/utils/test_files.py` — new atomic-write tests (21 passed, incl. concurrent-writers)
- [x] `pytest tests/unit/core/test_cache.py tests/unit/core/config tests/unit/core/scan tests/unit/core/plugin tests/unit/verticals/ai tests/unit/core/test_check_updates.py` — modules with touched call sites
- [ ] CI on this PR

### DIFF
--- a/ggshield/core/cache.py
+++ b/ggshield/core/cache.py
@@ -1,5 +1,4 @@
 import json
-import os
 from pathlib import Path
 from typing import Any, Dict, List
 
@@ -7,6 +6,7 @@ from ggshield.core import ui
 from ggshield.core.constants import CACHE_PATH
 from ggshield.core.errors import UnexpectedError
 from ggshield.core.types import IgnoredMatch
+from ggshield.utils.files import atomic_write_text
 from ggshield.utils.git_shell import gitignore, is_gitignored
 
 
@@ -61,20 +61,16 @@ class Cache:
             # if there are no found secrets, don't modify the cache file
             return
         try:
-            fd = os.open(
-                str(self.cache_path), os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600
+            text = json.dumps(self.to_dict())
+        except Exception as e:
+            raise UnexpectedError(
+                f"Failed to save cache in {self.cache_path}:\n{str(e)}"
             )
-            f = os.fdopen(fd, "w")
+        try:
+            atomic_write_text(self.cache_path, text, mode=0o600)
         except OSError:
             # Hotfix: for the time being we skip cache handling if permission denied
             return
-        with f:
-            try:
-                json.dump(self.to_dict(), f)
-            except Exception as e:
-                raise UnexpectedError(
-                    f"Failed to save cache in {self.cache_path}:\n{str(e)}"
-                )
 
     def purge(self) -> None:
         self.last_found_secrets = []
--- a/ggshield/utils/files.py
+++ b/ggshield/utils/files.py
@@ -1,3 +1,5 @@
+import os
+import time
 from enum import Enum, auto
 from pathlib import Path, PurePath, PurePosixPath
 from typing import List, Pattern, Set, Tuple, Union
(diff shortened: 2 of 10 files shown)


ggshield/utils/files.py
```diff
@@ -145,3 +147,98 @@ def url_for_path(path: PurePath) -> str:
...
+        os.chmod(new_path, mode)
+        with os.fdopen(fd, "wb") as f:
+            f.write(data)
+        _replace_with_retry(new_path, path)
+    except BaseException:
+        try:
+            os.unlink(new_path)
+        except OSError:
+            pass
+        raise
```
paulpetit-gg-ext, line 244: This is, barring the fdopen flags, an exact duplication of `atomic_write_text`. We should probably have a single implementation. If we want to keep the two *_text and *_bytes methods (which is fine), have them be thin wrappers around a common implementation.

tests/unit/utils/test_files.py
```diff
@@ -160,3 +165,86 @@ def test_get_gitignored_files(tmp_path):
...
+
+    def test_bytes_round_trip(self, tmp_path: Path) -> None:
+        target = tmp_path / "blob.bin"
+        payload = b"\x00\x01\x02\xff"
+
+        atomic_write_bytes(target, payload)
+
+        assert target.read_bytes() == payload
+
+    def test_concurrent_writers_produce_valid_file(self, tmp_path: Path) -> None:
```
paulpetit-gg-ext, line 231: This test does not justify the change, as this naive (and non-atomic) implementation of `atomic_write_text()` also passes it:

```python
def atomic_write_text(path, data, mode, encoding):    
    path.parent.mkdir(parents=True, exist_ok=True)
    path.write_text(data, encoding=encoding)
    path.chmod(mode)
```

To be honest, I tried to make this test exhibit the error with the naive implementation but couldn't, even with larger and more convoluted payloads.

paulpetit-gg-ext (changes requested):
I'm curious about the context: is this an error you or someone actually encountered? Under which OS?

The PR description looks AI generated and based on a code analysis, but mentions that the phenomenon has been "observed in practice".

The main idea seems sound, and I see how concurrent writes could corrupt a file like `ai_discovery.json`, so this kind of change is likely welcomed. Still, I'm very surprised this would ever trigger a SIGBUS, which sounds more like a "Python" problem than a "ggshield" problem. I would have expected a simple deserialization error (still undesirable, but which would have been handled gracefully in the code).

Moreover, the most important test, which is supposed to exhibit that the new code prevents the corruption, passes consistently with a naive implementation, so there is no guarantee that we actually fixed an issue here.

paulpetit-gg-ext:
Two additional comments:
- it looks like #1216 would benefit from this PR, as it implements something similar on its own :)
- this PR avoids having corrupted files (the latest process to write wins), but it doesn't prevent "cache stampede": multiple `ggshield` processes won't know there is already another one accessing/updating the cache. This is something a lib like [filelock](https://py-filelock.readthedocs.io/en/latest/index.html) would provide. We can also probably use it for the use case addressed by this PR as well.

Maybe we want to have utils for both needs, but the "cache stampede" avoidance is definitely useful: we will soon need to debounce hook calls as some agents call multiple times the same hook with the same payload if it is configured in multiple places.

ggshield/utils/files.py
mattisdalleau-gg: Hi the actual problem was verified:
```
❯ ggshield auth login
[1]    368219 bus error (core dumped)  ggshield auth login
```
`SIGBUS` is sound due to data corruption opening the file with `O_TRUNC` and re-reading it conccurently (around 5~6 agents using the `ggshield secret scan ai-hook`) which is probably a timeframe of `<1-2ms`  but very existant actually

Once the `SIGBUS` occur and the agents are actively calling it becomes persistent as `mmap` (probably?) cannot sanitize the `inode` of the file fast enough when truncating

ggshield/core/cache.py
```diff
@@ -61,20 +61,16 @@ def save(self) -> None:
             # if there are no found secrets, don't modify the cache file
             return
         try:
-            fd = os.open(
-                str(self.cache_path), os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600
+            text = json.dumps(self.to_dict())
+        except Exception as e:
```
paulpetit-gg-ext, line 65: the except should be after the next try

PR #1245, opened 2026-05-18
feat(git): harden git commands by using --end-of-options separator

6d7a:
## Context

Snyk flags a couple of medium command injection issues that we should resolve.

## What has been done

git 2.24+ supports the --end-of-options flag to separate command options from command arguments. This commit adds the separator to harden against command injections.

## PR check list

### DIFF
--- a/ggshield/core/git_hooks/ci/commit_range.py
+++ b/ggshield/core/git_hooks/ci/commit_range.py
@@ -52,7 +52,7 @@ def _fetch_target_branch(remote: str, branch: str) -> None:
     """
     try:
         ui.display_verbose(f"\tFetching {branch} from {remote}")
-        git(["fetch", remote, branch])
+        git(["fetch", "--end-of-options", remote, branch])
     except (subprocess.CalledProcessError, GitCommandTimeoutExpired):
         logger.warning("Failed to fetch %s from %s", branch, remote)
 
--- a/ggshield/core/scan/commit_information.py
+++ b/ggshield/core/scan/commit_information.py
@@ -75,5 +75,5 @@ class CommitInformation:
 
     @staticmethod
     def from_sha(sha: str, cwd: Optional[Path] = None) -> "CommitInformation":
-        header = git(["show", sha] + HEADER_COMMON_ARGS, cwd=cwd)
+        header = git(["show"] + HEADER_COMMON_ARGS + ["--end-of-options", sha], cwd=cwd)
         return CommitInformation.from_patch_header(header)
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -299,7 +299,7 @@ def is_valid_git_commit_ref(ref: str, wd: Optional[Union[str, Path]] = None) -> 
         wd = Path.cwd()
 
     ref += "^{commit}"
-    cmd = ["cat-file", "-e", ref]
+    cmd = ["cat-file", "-e", "--end-of-options", ref]
 
     try:
         git(cmd, cwd=wd)
@@ -428,7 +428,8 @@ def get_filepaths_from_ref(
     check_git_ref(ref, wd)
 
     filepaths = git(
-        ["ls-tree", "--name-only", "--full-name", "-r", ref], cwd=wd
+        ["ls-tree", "--name-only", "--full-name", "-r", "--end-of-options", ref],
+        cwd=wd,
     ).splitlines()
     return [Path(path_str) for path_str in filepaths]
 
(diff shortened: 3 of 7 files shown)


ggshield/utils/git_shell.py
```diff
@@ -299,7 +299,7 @@ def is_valid_git_commit_ref(ref: str, wd: Optional[Union[str, Path]] = None) ->
         wd = Path.cwd()
 
     ref += "^{commit}"
-    cmd = ["cat-file", "-e", ref]
+    cmd = ["cat-file", "-e", "--end-of-options", ref]
```
clement-tourriere, line 302: On git < 2.24 this fails with `error: unknown option` (exit 129). That is caught as `CalledProcessError` below, so this returns `False` for every ref, even `HEAD`. Then `check_git_ref()` rejects valid refs and scanning stops working.

Proposal: one helper in this file, used everywhere instead of the raw string.

```python
GIT_MIN_VERSION = (2, 24)  # --end-of-options
_RX_GIT_VERSION = re.compile(r"git version (\d+)\.(\d+)")


@lru_cache(None)
def _supports_end_of_options() -> bool:
    try:
        match = _RX_GIT_VERSION.match(git(["--version"], log_stderr=False).strip())
    except (subprocess.CalledProcessError, GitError):
        return True
    if match is None:
        return True
    return (int(match.group(1)), int(match.group(2))) >= GIT_MIN_VERSION


def end_of_options() -> List[str]:
    return ["--end-of-options"] if _supports_end_of_options() else []
```

Then here:

```python
cmd = ["cat-file", "-e", *end_of_options(), ref]
```

Two details I checked:
- It has to return `True` when the output cannot be parsed, because `tests/unit/utils/test_git_shell.py:669` mocks `subprocess.run` with `stdout=b""`.
- The regex needs to accept vendor suffixes. I tested `git version 2.50.1 (Apple Git-155)`, `2.45.1.windows.1`, `2.24.0.rc1` and `1.8.3.1`, all give the right answer.

ggshield/core/git_hooks/ci/commit_range.py
```diff
@@ -52,7 +52,7 @@ def _fetch_target_branch(remote: str, branch: str) -> None:
     """
     try:
         ui.display_verbose(f"\tFetching {branch} from {remote}")
-        git(["fetch", remote, branch])
+        git(["fetch", "--end-of-options", remote, branch])
```
clement-tourriere, line 55: This call site is the reason I would not just document the git 2.24 requirement.

On old git the error is caught below and only logged, so the fetch is skipped and we keep using a stale `origin/<branch>`. That gives a wrong commit range, which is the exact bug this function was added to prevent, and it happens with no visible error. A scanner that quietly scans less is worse than one that stops.

With `end_of_options()` this cannot happen.

ggshield/core/scan/commit_information.py
```diff
@@ -75,5 +75,5 @@ def from_staged(cwd: Optional[Path] = None) -> "CommitInformation":
 
     @staticmethod
     def from_sha(sha: str, cwd: Optional[Path] = None) -> "CommitInformation":
-        header = git(["show", sha] + HEADER_COMMON_ARGS, cwd=cwd)
+        header = git(["show"] + HEADER_COMMON_ARGS + ["--end-of-options", sha], cwd=cwd)
```
clement-tourriere, line 78: `Commit.from_sha` runs `git show` with this same sha a second time, in `ggshield/core/scan/commit.py:74`, and that call is not hardened. Would be good to do both:

```python
cmd = ["show", *PATCH_COMMON_ARGS, *end_of_options(), sha, "--"]
```

I checked this form works.

clement-tourriere (commented):
Nice fix, the approach is right. I ran all 9 new calls on git 2.50 and they all behave the same as before, and the test suite passes.

One thing to solve before merge: `--end-of-options` needs git 2.24 (Nov 2019), and nothing declares that. Our deb/rpm packages depend on plain `git` with no version bound, and they ship their own Python, so a user on an old distro can install ggshield and then hit broken scans.

My proposal is in the first comment: detect the git version once, and drop the flag when git is too old. Old git then keeps working exactly like today.

Two smaller notes:
- `get_list_commit_SHA()` in `git_shell.py` is still not hardened, and it gets its value straight from CI env vars (`CI_COMMIT_BEFORE_SHA`, `CIRCLE_RANGE`, `--commit-range`). It also does `.split()`, so one env var can become several args. It is not a one liner because `verticals/secret/repo.py:56` calls it with `"--all"`, so maybe a follow up ticket.
- No test pins the new flag. A grep of `tests/` finds no reference to these args, so this can regress silently later.

PR #1249, opened 2026-05-21
Ctourriere/fix root plugin installation

clement-tourriere:
## Context

 Fix plugin loading reliability for machine-scan/root-admin scenarios.

 - Enable installed wheel plugins by their ggshield.plugins entry point name instead of the wheel package name.
 - Keep legacy package-name config as an alias so existing installs continue to load.
 - Extract plugin wheels into the active runtime cache instead of beside the installed wheel, avoiding root/admin-owned extraction dirs breaking user runs.
 - Prune stale extracted wheel caches on load and remove plugin extraction cache on uninstall.

## PR check list

### DIFF
--- a/ggshield/cmd/plugin/install.py
+++ b/ggshield/cmd/plugin/install.py
@@ -26,7 +26,7 @@ from ggshield.core.plugin.downloader import (
     InsecureSourceError,
     PluginDownloader,
 )
-from ggshield.core.plugin.loader import resolve_config_key
+from ggshield.core.plugin.loader import enable_installed_plugin
 from ggshield.core.plugin.platform import get_platform_info
 from ggshield.core.plugin.signature import (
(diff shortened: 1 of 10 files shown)


benjaminrigaud-gg:
Did some QA, it looks ok. Thanks for All the ~~Fish~~ Fixes!

benjaminrigaud-gg (approved):

ggshield/cmd/plugin/install.py
```diff
@@ -148,6 +148,27 @@ def install_cmd(
         _install_from_gitguardian(ctx, plugin_source, version, signature_mode)
 
 
+def _enable_installed_plugin(
```
6d7a: Don't we also need to use this in `update.py` to correctly migrate during updates?

PR #1251, opened 2026-05-25
feat(api-status): show account id of the token

GG-HH:
## Summary
- `ggshield api-status` now displays the account ID (workspace ID) of the currently authenticated token, shown directly under `API URL` in the text output and as `account_id` in the JSON output.
- The new field is sourced from `workspace_id` returned by the `/api_tokens/self` endpoint and is gracefully omitted when the call fails (same behavior as `Token scopes`).
- The JSON schema (`doc/schemas/api-status.json`) and tests are updated accordingly.

## Test plan
- [x] `python -m pytest tests/unit/cmd/test_status.py`
- [x] `ggshield api-status` prints `Account ID:` right below `API URL:`
- [x] `ggshield api-status --json` includes the new `account_id` integer field

### DIFF
--- a/changelog.d/20260525_account_id_api_status.md
+++ b/changelog.d/20260525_account_id_api_status.md
@@ -1,0 +1,3 @@
+### Added
+
+- `ggshield api-status` now displays the account ID (workspace ID) associated with the current token, in both text and JSON output.
--- a/changelog.d/20260525_account_id_api_status.md
+++ b/changelog.d/20260525_account_id_api_status.md
@@ -1,3 +1,0 @@
-### Added
-
-- `ggshield api-status` now displays the account ID (workspace ID) associated with the current token, in both text and JSON output.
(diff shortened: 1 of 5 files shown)


GG-HH:
CI is partially failing because of the release in progress (tag 1.50.5 exists but no release yet)

xblanchot-gg (approved):
LGTM thanks !

6d7a (commented):
Thank you for tackling this so quickly. The changes look good but we should keep the wording consistent and use `workspace_id` here. We talk about workspaces in the [public docs](https://docs.gitguardian.com/platform/collaboration-and-sharing/users#access-levels) and also the `auth_config.yaml` associates tokens to a `workspace_id`.

GG-HH:
Unfortunately, it is not that clear, we actually use both (I checked beforehand) :( But you are correct, let's stick with something closer to the API.

6d7a (approved):
Nice, thanks for the changes!

PR #1252, opened 2026-05-26
Debounce AI hooks

paulpetit-gg-ext:
## Context

Some agents can detect multiple hook configurations (for instance VSCode can load Claude's hooks in addition to its own).

In these cases, the same payload is usually called once for each hook, which create unnecessary load to GIM and latency to the user.

## What has been done

- Remember the hash of the latest payload in a cache file and use it to debounce hook payloads.
- To avoid a race condition where the two hook processes are spawned quickly enough for the cache to be stale, file locking is done.
- New dependency to the [filelock](https://py-filelock.readthedocs.io/en/latest/index.html) library that makes this locking cross-platform.


ggshield/verticals/ai/hooks.py
```diff
@@ -32,6 +34,31 @@
...
+
+def debounce(content: str) -> bool:
+    """Return True if the payload is identical to the most recent hook call.
+
+    Some agents install hooks from multiple assistants, which can invoke ggshield
+    twice with the same payload.
+    """
+    payload_hash = hashlib.sha256(content.encode()).hexdigest()
+    debounce_path = get_cache_dir() / AI_HOOK_DEBOUNCE_FILENAME
+    debounce_path.parent.mkdir(parents=True, exist_ok=True)
```
xblanchot-gg: Is it ok for ggshield to write files this way ? Could it happen that ggshield does not have the rights to write the file or mkdir ?

ggshield/verticals/ai/hooks.py
```diff
@@ -231,6 +258,8 @@ def __init__(self, scanner: ScannerProtocol):
 
     def scan(self, content: str) -> int:
         """Scan the content, print the result and return the exit code."""
+        if content.strip() and debounce(content):
```
xblanchot-gg: Maybe it's standard, but I do not find the name "debounce" clear, I would go for a "has_already_been_seen" or seomthing like that

xblanchot-gg (commented):
A few small comments, overall looks good !

ggshield/verticals/ai/hooks.py
paulpetit-gg-ext: I think we already count on the cache being writeable elsewhere in ggshield, but you are right, it doesn't hurt being able to proceed in case of error. Will fix.

ggshield/verticals/ai/hooks.py
paulpetit-gg-ext: [It is](https://developer.mozilla.org/en-US/docs/Glossary/Debounce), but sure, I'll go with `has_already_been_seen`.

xblanchot-gg: Then you can keep as is !

paulpetit-gg-ext: woops, too late. :P

xblanchot-gg (approved):

PR #1253, opened 2026-05-26
Detect MCP servers' display_name

paulpetit-gg-ext:
## Context

Most MCP servers installed via extensions/plugins have a proper "human-readable" name. They can be used as display names in our platform to show data in a more user-friendly way.

## What has been done

This PR scans for these names whenever possible, and fill the `display_name` field of MCPServers (trusting the first non-empty value found). As an implementation detail, this meant adding a `display_name` field to the MCPConfiguration class as well, in order to transiently store our findings. This field is only used locally, and won't change the format of what is sent to the platform.

### DIFF
--- a/ggshield/verticals/ai/agents/vscode.py
+++ b/ggshield/verticals/ai/agents/vscode.py
@@ -8,7 +8,7 @@ from pygitguardian.models import AIDiscovery, MCPActivityRequest
 
 from ggshield.core.dirs import get_user_home_dir
 
-from ..models import Agent, EventType, HookPayload, HookResult
+from ..models import Agent, EventType, HookPayload, HookResult, MCPConfiguration
 
 
 class VSCode(Agent):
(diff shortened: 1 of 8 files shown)


ggshield/verticals/ai/agents/vscode.py
```diff
@@ -70,6 +70,28 @@ def discover_project_directories(self) -> Iterator[Path]:
                 if path.is_dir():
                     yield path.resolve()
 
+    def _get_user_mcp_configurations(self) -> Iterator[MCPConfiguration]:
+        confs = list(Agent._get_user_mcp_configurations(self))
```
xblanchot-gg: nit, not usual to explicitly call the class directly rather than using `super`, is there a good reason ?

ggshield/verticals/ai/models.py
```diff
@@ -7,17 +7,21 @@
...
+from pygitguardian.models import AIDiscovery, MCPActivityRequest
+from pygitguardian.models import MCPConfiguration as BaseMCPConfiguration
+from pygitguardian.models import MCPServer
 
 from ggshield.core.scan import File, Scannable, StringScannable
 from ggshield.utils.files import is_path_binary
 
 
+@dataclass
+class MCPConfiguration(BaseMCPConfiguration):
```
xblanchot-gg, line 19: Does it really deserves a new class ? Optional with default will do the work directly in the base class no ?

xblanchot-gg (commented):
A few marginal comment, but looks good to me overall ! Thanks !

ggshield/verticals/ai/agents/vscode.py
paulpetit-gg-ext: good catch

ggshield/verticals/ai/models.py
paulpetit-gg-ext: The base class is in `py-gitguardian`, and the new field is not sent to GIM, so doesn't really make sense in the interface schema :/

xblanchot-gg: Ah I did not have in mind this was in py-gitguardian ok !

xblanchot-gg (approved):

PR #1244, opened 2026-05-18
Pbeslin/parse mcp history

pbeslin-gg:
## Context

New option in `ggshield ai discover`: `--history` allows parsing and sending MCP calls found in agent transcripts, in order to obtain insight on calls made before installing the hook.  
:warning: Requires py-gitguardian changes: https://github.com/GitGuardian/py-gitguardian/pull/173

image

## What has been done

- `ggshield ai discover` takes a new optional flag option: `--history`. Not passing the flag results in the same behavior as before.  
- Passing the flag runs the history parsing in addition to the default behavior
- Transcript files are parsed for all supported agents: Claude Code, Copilot, Cursor
  - Claude Code and VSCode Chat store them in .jsonl files
  - Cursor uses a database
- Each agent implements its own way of iterating over the past events
- All activities are sent to the GitGuardian API in batches (500). The GitGuardian API handles idempotency.

## Validation

1. Wire py-gitguardian correct version (https://github.com/GitGuardian/py-gitguardian/pull/173)
2. Wire the local GitGuardian API with the new route (if not deployed yet)
3. Run `ggshield ai discover --history`
4. Go to the Agentic AI dashboard

## PR check list

- [x] As much as possible, the changes include tests (unit and/or functional)
- [x] If the changes affect the end user (new feature, behavior change, bug fix) then the PR has a changelog entry (see doc/dev/getting-started.md). If the changes do not affect the end user, then the `skip-changelog` label has been added to the PR.

PR #1259, opened 2026-06-01
feat(ai): collect agent activity

amascia-gg:
## Context

Builds on the MCP-history framework (#1244 / #1257). Today `ggshield ai discover --history` ships only MCP `PreToolUse` events (~5% of agent activity). This adds collection of **per-event usage metadata** for the whole session, to power agentic-AI usage dashboards.

Design: [AI Agent Activity Pipeline](https://www.notion.so/gitguardian/AI-Agent-Activity-Pipeline-36da119c1d9f8047aaaef9bd6737b5a7).

## What has been done

- **Per-agent activity sources** (Claude Code, Codex, Cursor) walk the on-disk transcripts/SQLite and emit one event per record via `client.send_agent_activity(...)`.
- **Metadata-only, fail-closed.** Each source's `serialize()` builds a NEW dict of allow-listed safe fields (`event_type`, tool name, model, timestamps, …) and drops every free-text field (prompts, command strings, tool inputs/outputs, file contents). The base `serialize()` **raises** by default, so a source can never accidentally ship raw content — **no secret or PII ever leaves the machine** (no server-side redaction needed).
- **PII/size hardening:** home paths anonymised (`/Users/x` → `~`); per-record size cap + byte-batching.
- Renamed the package `raw_history` → `agent_activity` (it's metadata, not raw).
- **Review fixes** to the original framework: `GGClient` typing + `Detail` error handling in the orchestrator; home-path leak fix in `source_path_for`.
- Bumps `pygitguardian` to the commit adding `send_agent_activity()` (GitGuardian/py-gitguardian#175).

## Validation

- `pytest tests/unit/verticals/ai/` → 387 passing.
- Verified end-to-end against a local GIM (`ward-runs-app`): `ggshield ai discover --history` → `POST /v1/nhi/ai/activity` → ClickHouse `agent_activity`. Stored content is metadata only (e.g. `{"type":"assistant","model":"…","tools":["Bash"]}`); a table-wide sweep confirmed prompts/commands/outputs are absent.

## Related / depends on

- SDK: GitGuardian/py-gitguardian#175 (`send_agent_activity`). Re-pin to the merged commit before landing.
- Server: GIM `ward-runs-app` MR (endpoint + `agent_activity` ClickHouse table).

## PR check list

- [ ] Changelog entry — TODO before un-drafting.

> Draft. Based on `pbeslin/collect-ai-agent-history` (#1257); registering the agent sources turns on real data shipping, so this is the "sources + metadata" layer on top of the framework.


ggshield/verticals/ai/agent_activity/orchestrator.py
```diff
@@ -0,0 +1,95 @@
...
+def collect_agent_activity(
+    client: GGClient, agents: "Iterable[Agent]"
+) -> AgentActivityReport:
+    """Walk each agent's raw sources and ship records in BATCH_SIZE-event batches."""
+    report = AgentActivityReport()
+    buffer: List[AgentActivityEvent] = []
+    buffer_bytes = 0
+
+    def flush() -> None:
+        nonlocal buffer_bytes
```
xblanchot-gg: Unused here

ggshield/verticals/ai/agent_activity/orchestrator.py
```diff
@@ -0,0 +1,95 @@
...
+        buffer.clear()
+        buffer_bytes = 0
+
+    for agent in agents:
+        for event in agent.iter_agent_activity_events():
+            buffer.append(event)
+            buffer_bytes += len(event.content.encode("utf-8"))
+            report.parsed += 1
+            if len(buffer) >= BATCH_SIZE or buffer_bytes >= MAX_BATCH_BYTES:
+                flush()
```
xblanchot-gg: flush performs an API call, I think we can find a better name here

ggshield/verticals/ai/agent_activity/readers.py
```diff
@@ -0,0 +1,60 @@
...
+
+def iter_jsonl(path: Path) -> Iterator[str]:
+    """Yield each non-blank line of a JSONL file as a raw string.
+
+    Returns an empty iterator on I/O failure.
+    """
+    if not path.is_file():
+        return
+    try:
+        with path.open("r", encoding="utf-8", errors="replace") as fp:
```
xblanchot-gg: Avoid two letter var names such as `fp`

ggshield/verticals/ai/agent_activity/orchestrator.py
xblanchot-gg: We should wrap this loop in a try/except so that an unexpected error for a given agent do not break the entire collection process for subsequent agents

ggshield/verticals/ai/agents/copilot.py
```diff
@@ -22,12 +23,30 @@
...
 
+class CopilotActivitySource(JSONLActivitySource):
+    """Every Copilot CLI session line, shipped raw.
+
+    Copilot CLI appends one JSON object per line to
+    ~/.copilot/session-state/<uuid>/events.jsonl. The line is shipped
+    verbatim; GitGuardian scans and strips secrets server-side before storing it.
+    """
+
+    kind = "session_events"
```
xblanchot-gg: I don't understand the point of those `kind` things. They are not really helpful themselves, and are just a way to say "event" in the jargon of a given AI coding tool. The name of the agent allow to perform the same discrimination between all events, and is more reliable, as we will not recall the difference between `session_events` and `session_transcripts` in the future

xblanchot-gg (commented):
First batch of comments as I've not finished the review yet, as the PR is quite long to read. Overall looks good, but I think there is a bug somewhere, when testing on my machine, only 10k events over 63k are actually sent to GitGuardian.

tests/unit/verticals/ai/agent_activity/test_models.py
```diff
@@ -0,0 +1,25 @@
+import json
+
+from ggshield.verticals.ai.agent_activity.models import AgentActivityEvent
+
+
+def test_agent_activity_event_to_dict_is_json_serializable() -> None:
```
xblanchot-gg: Unnecessary test, this tests dataclass library, not our code

tests/unit/verticals/ai/agent_activity/test_orchestrator.py
```diff
@@ -0,0 +1,170 @@
...
+    agent = _make_agent("a", events)
+    client = MagicMock()
+    client.send_agent_activity.return_value = MagicMock(
+        success=True, ingested=BATCH_SIZE, duplicates=0
+    )
+
+    report = collect_agent_activity(client, [agent])
+
+    assert client.send_agent_activity.call_count == 2
+    assert report.parsed == BATCH_SIZE + 3
```
xblanchot-gg: I'm not super fan of those tests that instanciate "batch_size" elements, where batch_size can be several thousands. Can we patch batch_size to be something like 5 ?

xblanchot-gg (commented):
Second batch, a few other comments, nothing really blocking on my side

ggshield/verticals/ai/agents/copilot.py
pbeslin-gg: We will want to collect data from several types of history files, which may not have the same structure. The `kind` attribute in addition to the agent name allows processing them case by case on the server.  
For instance, Claude Code has different files for direct transcripts and subagents transcripts

ggshield/verticals/ai/agent_activity/orchestrator.py
amascia-gg: Refactored the send buffer into a small `_Batch` value object, so there's no `nonlocal` here at all anymore — `send_batch()` just mutates the batch.

ggshield/verticals/ai/agent_activity/orchestrator.py
amascia-gg: Renamed `flush` → `send_batch`, since it does perform the API call.

ggshield/verticals/ai/agent_activity/readers.py
amascia-gg: Renamed `fp` → `handle`.

ggshield/verticals/ai/agent_activity/orchestrator.py
amascia-gg: Done — each agent's history walk is now wrapped in a try/except that logs and moves on, so one agent failing can't abort collection for the rest.

ggshield/verticals/ai/agents/copilot.py
amascia-gg: Keeping `kind` (the per-source `source_kind`): as Paul said, a single agent can have several history-file types (e.g. Claude Code's direct vs subagent transcripts) that the server needs to process case-by-case, and the agent name alone can't distinguish them. I'll keep the values descriptive.

tests/unit/verticals/ai/agent_activity/test_models.py
amascia-gg: Agreed — removed the test (and the file, as it was the only one). `to_dict()` is just `dataclasses.asdict`, so it was testing the stdlib.

tests/unit/verticals/ai/agent_activity/test_orchestrator.py
amascia-gg: Done — the batching tests now `monkeypatch` `BATCH_SIZE` (and `MAX_BATCH_BYTES`) to small values instead of allocating thousands of events.

paulpetit-gg-ext (approved):

xblanchot-gg (approved):

PR #1260, opened 2026-06-02
feat(honeytoken): add honeytoken plant command

jCOTINEAU:


ggshield/cmd/honeytoken/plant.py
```diff
@@ -0,0 +1,313 @@
...
+            path, section = resolve_placement(item.method, item.config, target.home)
+            expected = item.token.access_token_id if item.token else None
+            result = remove_aws_profile(path, section, expected)
+            if result is RemoveOutcome.FOREIGN_KEPT:
+                click.echo(
+                    f"[{target.username}] deployment {item.id}: profile holds a "
+                    "different key, left untouched",
+                    err=True,
+                )
+            _confirm(client, item, ConfirmStatus.REMOVED, target)
```
clement-tourriere, line 227: Small ownership concern here: on delete, `remove_aws_profile()` may rewrite the AWS file through a temp file + `os.replace()`. When ggshield runs as root and the file still has other profiles, the resulting credentials/config file can become root-owned with `0600`, so the target user may lose access to their own AWS file. Could we re-apply `apply_perms_and_owner()` after a successful removal when the file still exists?

ggshield/verticals/honeytoken/targets.py
```diff
@@ -0,0 +1,169 @@
...
+        "username": username,
+        "hostname": _get_hostname(),
+    }
+
+
+def resolve_targets(user: Optional[str], user_dir: Optional[Path]) -> List[Target]:
+    """Resolve the set of users to plant for. Raises ``LookupError`` if an explicit
+    ``--user`` can't be resolved."""
+    if user_dir is not None:
+        return [Target(username=user or _get_username(), home=user_dir, uid=None)]
```
clement-tourriere: If both `--user` and `--user-dir` are passed, we keep the right username but lose the uid here. In a root run this means the file/dir created under the overridden home will stay root-owned. Maybe we should still resolve the passwd entry for `user` in this branch and keep its uid?

ggshield/verticals/honeytoken/endpoint_deployments.py
```diff
@@ -0,0 +1,231 @@
...
+                headers=self._headers,
+                timeout=SYNC_TIMEOUT_SECS,
+                **kwargs,
+            )
+        except requests.RequestException as exc:
+            raise EndpointDeploymentsError(
+                f"endpoint-deployments API request error: {exc}"
+            )
+
+        if response.status_code in (200, 201):
```
clement-tourriere: Tiny API-compat thought: confirm-style PATCH endpoints often return `204 No Content`. If this endpoint does that, we would log every successful confirm as an API error. Would it be safer to accept any 2xx here (or at least include 204)?

clement-tourriere (approved):

PR #1264, opened 2026-06-10
Jeremycotineau/hon 1419 harden honeytoken plant aws file handling comments read only

jCOTINEAU:
## Context

Ggshield honeytoken plant was not preserving file comments nor file permissions

## What has been done

Moved to the `configupdater` Library to benefit from INI file handling safety

### E2E results — `ggshield honeytoken plant` (local instance)

| # | Scenario | Before | After | Result |
|---|----------|--------|-------|--------|
| A | Commented file, mode 0600 | comments + `[default]`/`[work]` | comments intact, `[prod-backup]` added, mode 0600 | ✅ comments preserved |
| B | Custom mode 0640 | `-rw-r-----` | `-rw-r-----` (`[prod-backup]` added) | ✅ mode preserved (no 0600 reset) |
| C | Read-only 0444 | `-r--r--r--` | modified with no error, `[prod-backup]` added, mode 0444 kept | ✅ trust-the-system (os.replace writes) |
| D | No trailing newline | `[default]` at EOF | `[prod-backup]` is a real section, file ends with `\n` | ✅ trailing-newline guard |
| E | Idempotence | — | run 1 = `2 written`; run 2 = `0 written, 2 skipped` | ✅ no-op on rerun |
| F | Collision (`[prod-backup]` holds other creds) | foreign key present | no `--force` → refused, key kept; `--force` → overwrites, comment kept | ✅ `ForceRefusal` intact |

Tests added (7)
Commit 1 — comments / configupdater

**test_write_preserves_existing_comments**, **test_write_leaves_other_sections_byte_for_byte** → A
**test_write_appends_cleanly_to_file_without_trailing_newline** → D
**test_write_does_not_mix_newlines_on_crlf_file** (CRLF), **test_remove_preserves_comments_on_other_profiles**
Commit 2 — permissions

**test_write_preserves_existing_file_mode** → B, C
**test_apply_perms_non_root_keeps_dir_private_and_leaves_file_mode** (rewritten: dir 0700, no file chmod)
E & F use pre-existing tests (**test_write_is_idempotent_for_same_key**, **test_write_refuses_*_without_force**).

## PR check list

- [X] As much as possible, the changes include tests (unit and/or functional)

### DIFF
--- a/ggshield/verticals/honeytoken/aws_profile.py
+++ b/ggshield/verticals/honeytoken/aws_profile.py
@@ -15,9 +15,12 @@ from __future__ import annotations
 import configparser
 import enum
 import os
+import stat
 import tempfile
 from pathlib import Path
 from typing import Optional, Tuple
+
+from configupdater import ConfigUpdater
 
 from ggshield.verticals.honeytoken.endpoint_deployments import (
     DeploymentMethod,
@@ -87,31 +90,62 @@ def resolve_placement(
     raise PlacementError("unsupported deployment method — client may be out of date")
 
 
-def _load(path: Path) -> configparser.ConfigParser:
-    # interpolation=None: AWS secret keys can contain ``%``, which the default
-    # interpolation would choke on.
-    parser = configparser.ConfigParser(interpolation=None)
+def _load(path: Path) -> ConfigUpdater:
+    # configupdater edits in place and keeps comments (configparser wouldn't).
+    parser = ConfigUpdater()
     if path.exists():
         try:
-            parser.read(path, encoding="utf-8")
+            text = path.read_text(encoding="utf-8")
+            # Else configupdater glues our new section onto the last line.
+            if text and not text.endswith("\n"):
+                text += "\n"
+            parser.read_string(text)
         except (configparser.Error, UnicodeDecodeError) as exc:
-            # A malformed/garbage file is an expected condition, not a crash: surface it
-            # as a PlacementError so the caller reports a clean per-deployment failure
-            # and leaves the file untouched.
+            # configupdater parse errors subclass configparser.Error.
             raise PlacementError(
                 f"could not parse {path}: not a valid AWS credentials/INI file ({exc})"
             )
     return parser
 
 
-def _atomic_write(path: Path, parser: configparser.ConfigParser) -> None:
+def _get_value(parser: ConfigUpdater, section: str, key: str) -> Optional[str]:
+    sec = parser[section]
+    if sec.has_option(key):
+        return sec[key].value
+    return None
+
+
+def _reject_symlinked_target(path: Path) -> None:
+    """Refuse a symlinked ``.aws`` dir or credentials file: as root in the fan-out,
+    following it would redirect the write/chmod/chown outside the user's home."""
+    parent = path.parent
+    if parent.is_symlink():
+        raise PlacementError(
+            f"refusing to use {parent}: '.aws' is a symlink "
+            "(possible redirect outside the home directory)"
+        )
+    if path.is_symlink():
+        raise PlacementError(
+            f"refusing to write through symlinked credentials file {path}"
+        )
+
+
+def _atomic_write(path: Path, parser: ConfigUpdater) -> None:
+    _reject_symlinked_target(path)
     path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
+    # Preserve the file's existing mode (new file → 0600).
+    try:
+        mode = stat.S_IMODE(path.stat().st_mode)
+    except FileNotFoundError:
+        mode = 0o600
     fd, tmp = tempfile.mkstemp(prefix=".plant.", suffix=".tmp", dir=str(path.parent))
     try:
-        os.chmod(tmp, 0o600)
         with os.fdopen(fd, "w", encoding="utf-8") as handle:
             parser.write(handle)
+        # Swap first (temp stays at mkstemp's 0600 while holding the secret), then
+        # restore the final mode — chmod'ing the temp pre-swap would briefly widen it.
         os.replace(tmp, path)
+        os.chmod(path, mode)
     except BaseException:
         try:
             os.unlink(tmp)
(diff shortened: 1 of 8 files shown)


clement-tourriere (changes requested):
nop

clement-tourriere (commented):
Adding the same review as inline comments so the requested changes are visible on the relevant diff lines.

ggshield/verticals/honeytoken/aws_profile.py
```diff
@@ -87,30 +90,44 @@ def resolve_placement(
...
-def _atomic_write(path: Path, parser: configparser.ConfigParser) -> None:
+def _get_value(parser: ConfigUpdater, section: str, key: str) -> Optional[str]:
+    sec = parser[section]
+    if sec.has_option(key):
+        return sec[key].value
+    return None
+
+
+def _atomic_write(path: Path, parser: ConfigUpdater) -> None:
     path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
```
clement-tourriere: This follows `home/.aws` if it is a symlink, and the subsequent `mkstemp`, `chmod`, and root `chown` path follows that target too. In a root fan-out run, a user can redirect the write plus directory permission/ownership changes outside their home. Please reject symlinked `.aws` directories here (or use `openat`/`dir_fd` + no-follow checks and verify the real directory stays under the target home) before writing/chowning.

ggshield/verticals/honeytoken/aws_profile.py
```diff
@@ -87,30 +90,44 @@ def resolve_placement(
...
+        mode = stat.S_IMODE(path.stat().st_mode)
+    except FileNotFoundError:
+        mode = 0o600
     fd, tmp = tempfile.mkstemp(prefix=".plant.", suffix=".tmp", dir=str(path.parent))
     try:
-        os.chmod(tmp, 0o600)
+        # chmod after writing: the secret stays in a private 0600 temp until the swap.
         with os.fdopen(fd, "w", encoding="utf-8") as handle:
             parser.write(handle)
+        os.chmod(tmp, mode)
```
clement-tourriere: `os.chmod(tmp, mode)` runs before `os.replace()`, so an existing permissive credentials file (`0644`/`0640`) makes `.plant.*.tmp` readable while it already contains the planted honeytoken; a failure after this point can also leave a readable temp file behind. Keep the temp file `0600` until after the replace, then restore the final file mode after ensuring the parent directory is private.

ggshield/verticals/honeytoken/aws_profile.py
```diff
@@ -166,7 +183,7 @@ def remove_aws_profile(
         return RemoveOutcome.ALREADY_ABSENT
```
clement-tourriere: In this remove path, the code below still uses `parser.sections()` to decide whether to unlink the file. With `ConfigUpdater`, a file containing comments plus only this honeytoken section has no sections after `remove_section()`, but still has preserved comments, so `path.unlink()` deletes them. Please update the logic around the `parser.remove_section()`/`if parser.sections()` block to write back non-empty rendered content and unlink only when the remaining document is truly empty; add a regression test for comments + only the honeytoken section.

clement-tourriere (changes requested):
Thanks for the follow-up fixes. I still see one issue that needs changes.

The symlink checks are still time-of-check/time-of-use prone. `_reject_symlinked_target()` verifies that `.aws` / `credentials` are not symlinks (`ggshield/verticals/honeytoken/aws_profile.py:118-130`), but the later write/remove operations use the paths again (`mkdir`, `mkstemp`, `os.replace`, `chmod`, and `unlink`; `aws_profile.py:133-148`, `aws_profile.py:213-217`). A user controlling `~/.aws` could swap the path after the check, so a root fan-out run can still write/chmod/remove outside the intended home.

The same pattern exists in `apply_perms_and_owner()`: the symlink check is before later path-based `chmod` / `chown` calls (`ggshield/verticals/honeytoken/targets.py:95-109`).

Could you switch these privileged operations to no-follow / fd-based operations instead of relying on path pre-checks? For example: open `.aws` with `O_DIRECTORY|O_NOFOLLOW`, operate relative to that dir fd, use `dir_fd` / `follow_symlinks=False` / `fchmod` / `fchown` where appropriate, and add a regression test that swaps `.aws` or `credentials` between the check and the operation.

Minor: `ggshield/cmd/honeytoken/plant.py:223-225` still says the rewritten file is `root-owned 0600`, but the new code preserves file mode and `apply_perms_and_owner()` no longer changes it.

ggshield/verticals/honeytoken/aws_profile.py
```diff
@@ -87,31 +90,62 @@ def resolve_placement(
...
+        )
+    if path.is_symlink():
+        raise PlacementError(
+            f"refusing to write through symlinked credentials file {path}"
+        )
+
+
+def _atomic_write(path: Path, parser: ConfigUpdater) -> None:
+    _reject_symlinked_target(path)
     path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
```
clement-tourriere: This still looks TOCTOU-prone: `_reject_symlinked_target()` can pass, then `.aws` or `credentials` can be swapped before the later path-based `mkstemp` / `replace` / `chmod` / `unlink` operations. Could we make the write/remove path no-follow/fd-based instead (for example operate relative to an `O_DIRECTORY|O_NOFOLLOW` dir fd) and add a regression test that swaps the path between the check and the operation?

ggshield/verticals/honeytoken/targets.py
```diff
@@ -87,15 +87,15 @@ def resolve_targets(user: Optional[str], user_dir: Optional[Path]) -> List[Targe
...
     if os.name != "posix":
         return
-    try:
-        os.chmod(path, 0o600)
-    except OSError:
-        pass
     parent = path.parent
+    # Don't chmod/chown through a symlink: as root that could redirect onto e.g. /etc
+    # (write/remove already reject symlinks; this closes the chown leg + the TOCTOU).
+    if parent.is_symlink() or path.is_symlink():
```
clement-tourriere: Same TOCTOU concern here: the path can be swapped after this check and before the path-based `chmod`/`chown` below. Could we avoid following paths for the privileged ownership/perms operations (e.g. fd-based operations or `follow_symlinks=False` where applicable) and cover the swap case in tests?

clement-tourriere (changes requested):

ggshield/verticals/honeytoken/aws_profile.py
jCOTINEAU: Thanks, i tried added a specific guardrail for that, but its not as easy, there are some difference with macos and linux capabilities, the current approach seems to solve the problem

clement-tourriere (approved):

PR #1268, opened 2026-06-11
feat(auth): surface credential-store failures and add `ggshield auth status`

### LINKED ISSUE #1267: Keyring migration fails silently when Keychain entry ACL doesn't match current binary (e.g. mise installs)
## Environment

- ggshield version: 1.51.0
- Operating system (Linux, macOS, Windows): macOS (Apple Silicon)
- Operating system version: Darwin 25.5.0
- Python version: 3.14 (bundled with ggshield, installed via mise as part of a toolchain)

## Describe the bug

When ggshield is installed via a tool version manager like mise (or asdf, pyenv, pipx, etc.), the binary path can change between installs or reshims. On macOS, Keychain entries have an ACL tied to the creating application path. If the path changes, the new binary gets error `-25244` (`errSecInvalidOwnerEdit`) when trying to overwrite the existing entry.

The result is that:

- `is_available()` returns `True` (the probe with `__ggshield_probe__` always writes fine as a fresh entry)
- `store_token()` silently fails on the real instance URL key due to the ACL mismatch
- The token stays in cleartext in `auth_config.yaml` with no visible feedback to the user

This is not specific to Homebrew upgrades — it affects any installation method where the binary path is managed externally (mise, asdf, pipx, Docker volumes, CI runners, etc.).

Steps to reproduce:

1. Install ggshield via mise and authenticate (`ggshield auth login`)
2. Reinstall or reshim ggshield via mise (e.g. `mise install`, toolchain update)
3. Run any ggshield command
4. Check `~/.config/ggshield/auth_config.yaml`

Actual result:

The token is still in cleartext in `auth_config.yaml`, and no warning or error is shown anywhere.

Expected result:

The token is migrated to the Keychain, or — when that is impossible — the user is told the token stays in cleartext and how to fix it.

## Root cause

Two issues compound each other:

1. `is_available()` probes with a dummy key `__ggshield_probe__`, which always succeeds since it's a fresh entry with no ACL conflict. It never detects that the real instance URL key is locked to a different binary path.

2. `_persist_to_keyring()` catches the exception but only logs it at `logger.warning` level, invisible to the end user:

```python
except Exception:
    logger.warning(
        "Failed to store token in keyring for %s, storing in config file",
        inst.url,
        exc_info=True,
    )
    fallback_urls.add(inst.url)
```

## Proposed fixes

1. **Surface the error visibly with an actionable fix** — warn on stderr that the token stays in plaintext, including the commands to recover:

```
security delete-generic-password -s ggshield -a '<instance url>'
ggshield auth login
```

2. **Detect the conflict where it matters** — the dummy-key probe cannot catch per-entry ACL conflicts; the diagnosis should exercise the real instance URL keys.

3. **Add a diagnostic command** reporting the credential-store backend, whether it is reachable, and where each instance's token actually lives (credential store vs cleartext config file), with a copy-pasteable fix when a token unexpectedly stays in plaintext.

4. **Notify the user after migration attempts** — confirmation on success, warning on failure.

AGallouin:
## Context

When the ggshield binary path changes (mise/asdf reshim, pyenv/pipx reinstall, Homebrew upgrade…), the macOS Keychain entry created by the previous binary can no longer be overwritten: writes fail with `-25244` (`errSecInvalidOwnerEdit`) because the entry's ACL is tied to the old path. ggshield silently fell back to storing the token in cleartext in `auth_config.yaml`, with only a debug-level log. The availability probe could not catch this, since it writes a fresh key with no ACL conflict.

Fixes #1267.

## What has been done

- **Failed credential-store writes are now surfaced to the user**: `AuthConfig.save()` prints a warning with a humanized error (the `-25244` case is translated to plain language), states that the token will stay in cleartext, and gives the copy-pasteable recovery commands (`security delete-generic-password …` + `ggshield auth login`). The fix advice is gated on the *active keyring backend* (not the OS), so a custom backend on macOS doesn't get Keychain-specific commands.
- **Successful cleartext → credential-store migrations are confirmed** with an info message. Routine re-saves and fresh logins stay silent to avoid noise; the prior on-disk state is used to tell the two apart.
- **New `ggshield auth status` command** diagnosing token storage, in text and `--json`:
  - reports the credential-store backend (macOS Keychain, Windows Credential Locker, Linux Secret Service, KWallet…), whether it is reachable, and where each instance's token actually lives (`ok` / `failed` / `plaintext` / `disabled` / `skipped`), with fix commands when actionable;
  - is strictly **read-only**: it never writes to the credential store (no token migration as a side effect, and no write-probe — a new `is_reachable()` read-only check is used instead of `is_available()`);
  - never touches the credential store at all when `GGSHIELD_NO_KEYRING` is set (`reachable` is reported as `null`);
  - the diagnosis is read-centric: a token that reads back fine is `ok` even if overwriting its entry would fail, because reads are what every command needs at runtime — write failures are surfaced at save time instead.
  - JSON output has a stable shape (`credential_store` + per-instance `{instance, status, message, fix}`, with `null` for non-applicable fields) intended to be parseable by scripts; tests freeze the contract.

Differences from the issue's proposed fixes: instead of making `is_available()` probe real instance keys (which would put writes on every command's startup path), the conflict is detected where it matters — the save path warns on the actual failed write, and `auth status` verifies each real instance key with reads only. The diagnostic command lives under `ggshield auth status` rather than `ggshield config keyring-status`, and its output is ASCII-only for legacy Windows terminals.

## Validation

Unit tests cover the new behavior (failure warning, migration message, silence on re-saves, read-only guarantee, JSON contract). To validate manually on macOS:

- Authenticate: `ggshield auth login`, then run `ggshield auth status` — the instance should report `token_storage: ok` with the macOS Keychain location.
- Simulate the ACL conflict: re-create the Keychain entry from another binary path (or reinstall ggshield via mise/brew so its path changes), then re-trigger a save (`ggshield auth login`) — a warning should explain the failure and print the `security delete-generic-password` fix; `auth status` should report `plaintext` with the same fix.
- Apply the fix commands — `auth status` should report `ok` again, and the save should print the migration confirmation.
- `GGSHIELD_NO_KEYRING=1 ggshield auth status --json` should report `"reachable": null` and not prompt for Keychain access.

## PR check list


ggshield/cmd/auth/__init__.py
```diff
@@ -6,9 +6,16 @@
...
 from .logout import logout_cmd
+from .status import auth_status_cmd
 
 
-@click.group(commands={"login": login_cmd, "logout": logout_cmd})
+@click.group(
+    commands={
+        "login": login_cmd,
+        "logout": logout_cmd,
+        "status": auth_status_cmd,
```
GG-HH: There is already a command `ggshield api-status` that i believe has the intended behavior. What I would suggest is to add an alias with ggshield auth status. WDYT ?

ggshield/core/config/auth_config.py
```diff
@@ -117,6 +121,30 @@ def prepare_auth_config_dict_for_save(data: Dict[str, Any]) -> Dict[str, Any]:
     return data
 
 
+def read_stored_tokens(config_path: Path) -> Dict[str, str]:
```
GG-HH: Should we rename this to read_config_tokens ? because the tokens from the keyring are also stored

GG-HH (commented):
Thanks for this,
While reviewing it, I wondered if we should have the keyring path depend on the installation path if it can only be accessed by that path ?

WDYT ?

AGallouin:
> Thanks for this, While reviewing it, I wondered if we should have the keyring path depend on the installation path if it can only be accessed by that path ?
> 
> WDYT ?

You mean that instead of having (for example in mac):
`keyring.set_password(KEYRING_SERVICE, instance_url, token)`
we would have something like:
`keyring.set_password(f"{installation_path}_{KEYRING_SERVICE}", instance_url, token)`
Scoping it only to MacOS ?

I'm not a big fan for many reasons:
- if you have many installs of ggshield on your computer (brew + mise) both would be able to have different passwords, different behaviors => not really clear for an end user, hard to debug / understand
- usually, upgrading the software changes its path (for brew, it lives in /opt/homebrew/Cellar/ggshield/1.51.0), so upgrading it to 1.52 would mean: loosing your password (need to re auth) and leaving an orphaned valid token in the keyring. Not great from a security PoV. You would also erase all the ACLs (always allow behaviors) set by user.

In short, if I understood your point correctly, I'd lean toward keeping a single canonical entry, with the failure surfaced loudly and a one-time fix (what this PR does). But open to discussing it!
@GG-HH

ggshield/cmd/auth/__init__.py
AGallouin: From what i understood, api-status is more "can i talk to GG, is my token valid ?" it does network calls and check the token no matter its origin (plain text, keyring on mac or other password managers).
auth status  runs only locally, checks if we have a token and if it is correctly stored. It would be able to be enriched with other OS specific diagnostics.

I don't thing they server the same purpose, so I think that 2 commands is the right call. In a world full of LLMs, I would keep it really straightforward, one command, one scoped mission, simple actionnable output. If you think it will be more manually ("humanly") called than programmatically called, having an "overall dashboard" can make sense, it's more of a product call.

PR #1269, opened 2026-06-11
chore(release): split build/ staging from dist/ output

nriviere-gitguardian:
The release pipeline wrote OS packages (.deb/.rpm/.pkg/.msi/...) to a top-level packages/ directory that also held intermediate staging. We want that name free for source packages in the monorepo, and we want a clean convention: build/ for intermediates, dist/ for shippable output.

- Final artifacts (the files uploaded to releases/cloudsmith/chocolatey) now go to dist/: nfpm --target, the tar.gz/pkg/zip/msi paths, and the chocolatey --outdir.
- Intermediates go to build/: the PyInstaller bundle (PYINSTALLER_OUTPUT_DIR, via --distpath), its analysis cache (--workpath build/.pyinstaller, so it no longer collides with the bundle's ggshield/ subdir), the per-target staging tree, and the macOS tar restructuring scratch dir.

Release workflows are unchanged: every artifact glob was already dist/ggshield-*.<ext>, which now matches only the final output. build/ is already gitignored.

### DIFF
--- a/.github/workflows/build_release_assets.yml
+++ b/.github/workflows/build_release_assets.yml
@@ -281,13 +281,13 @@ jobs:
         with:
           name: os-packages-${{ matrix.os }}
           path: |
-            packages/ggshield-*.gz
-            packages/ggshield-*.pkg
-            packages/ggshield-*.zip
-            packages/ggshield-*.rpm
-            packages/ggshield_*.deb
-            packages/ggshield.*.nupkg
-            packages/ggshield-*.msi
+            dist/ggshield-*.gz
+            dist/ggshield-*.pkg
+            dist/ggshield-*.zip
+            dist/ggshield-*.rpm
+            dist/ggshield_*.deb
+            dist/ggshield.*.nupkg
+            dist/ggshield-*.msi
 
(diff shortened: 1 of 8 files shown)


GG-HH:
I think you missed a few spots:
- linux package smoke tests, in the setup, it seems to read from the packages directory
- documentation (os-packages.md), it mention packages/....
Otherwise sounds good to me!

clement-tourriere (approved):
Approved but two small comments:  
 - Low/latent bug — .github/workflows/build_release_assets.yml:337
   The Clear Linux smoke-test branch still uses pkg_dir=$PWD/packages after artifacts are downloaded to dist at line 318. If Clear Linux is re-enabled in the matrix, this branch will fail to find the tarball.
   Should be pkg_dir=$PWD/dist. Current matrix does not include Clear Linux, so CI won’t catch this.
 - Low/docs — doc/dev/os-packages.md:17-20
   The OS packaging docs still describe PyInstaller output as dist/ggshield and staging as packages/.... With this PR those should be build/ggshield and build/..., with final artifacts in dist/.

PR #1273, opened 2026-06-11
fix(config): run keyring probe in subprocess to survive segfaults

GG-HH:
## Summary

On some Linux machines, native keyring backends (libsecret, KWallet) segfault
during the availability probe, crashing the ggshield process entirely instead
of gracefully falling back to file-based token storage.

## Changes

The `is_available` probe now runs in a child process via `subprocess.run`. A
non-zero exit code — including the -11/139 from SIGSEGV — is treated as
unavailable and triggers the file store fallback. The subprocess uses distinct
exit codes: 0 (available), 2 (fail backend detected), 3 (round-trip mismatch),
anything else (crash or unexpected error). Tests are updated to patch
`subprocess.run` and a new test case covers the segfault scenario (exit 139).

### DIFF
--- a/ggshield/core/config/token_store.py
+++ b/ggshield/core/config/token_store.py
@@ -27,13 +27,15 @@ def _keyring_probe() -> None:
     kr = keyring.get_keyring()
     if isinstance(kr, keyring.backends.fail.Keyring):
         sys.exit(2)
-    keyring.set_password(KEYRING_SERVICE, "__ggshield_probe__", "test")
+
+    PROBE_VALUE = "test"
+    keyring.set_password(KEYRING_SERVICE, "__ggshield_probe__", PROBE_VALUE)
     val = keyring.get_password(KEYRING_SERVICE, "__ggshield_probe__")
     try:
         keyring.delete_password(KEYRING_SERVICE, "__ggshield_probe__")
     except Exception:
         pass
-    sys.exit(0 if val == "test" else 3)
+    sys.exit(0 if val == PROBE_VALUE else 3)
 
 
 class TokenStore(ABC):
--- a/ggshield/core/config/token_store.py
+++ b/ggshield/core/config/token_store.py
@@ -1,4 +1,6 @@
 import logging
+import subprocess
+import sys
 from abc import ABC, abstractmethod
 from typing import Optional
 
@@ -58,22 +60,34 @@ class KeyringTokenStore(TokenStore):
             logger.debug("No keyring entry to delete for instance %s", instance_url)
 
     def is_available(self) -> bool:
-        """Check if keyring is usable by probing with a test key."""
+        """Check if keyring is usable by probing with a test key.
+
+        The probe runs in a subprocess so that a segfault in a native backend
+        (e.g. libsecret or KWallet) doesn't crash the main process.
+        """
+        probe = (
+            "import keyring, keyring.backends.fail;"
+            f"kr = keyring.get_keyring();"
+            f"exit(2) if isinstance(kr, keyring.backends.fail.Keyring) else None;"
+            f"keyring.set_password({KEYRING_SERVICE!r}, '__ggshield_probe__', 'test');"
+            f"val = keyring.get_password({KEYRING_SERVICE!r}, '__ggshield_probe__');"
+            f"(keyring.delete_password({KEYRING_SERVICE!r}, '__ggshield_probe__'));"
+            f"exit(0 if val == 'test' else 3)"
+        )
         try:
-            kr = keyring.get_keyring()
-            if isinstance(kr, keyring.backends.fail.Keyring):
-                return False
-            # Probe the backend to verify it actually works (e.g. a
-            # ChainerBackend may pass the isinstance check but still fail).
-            probe_key = "__ggshield_probe__"
-            keyring.set_password(KEYRING_SERVICE, probe_key, "test")
-            val = keyring.get_password(KEYRING_SERVICE, probe_key)
-            try:
-                keyring.delete_password(KEYRING_SERVICE, probe_key)
-            except Exception:
-                logger.debug("Failed to clean up keyring probe key")
-            return val == "test"
+            result = subprocess.run(
+                [sys.executable, "-c", probe],
+                timeout=10,
+                capture_output=True,
+            )
+            if result.returncode != 0:
+                logger.debug(
+                    "Keyring probe subprocess exited with code %d",
+                    result.returncode,
+                )
+            return result.returncode == 0
         except Exception:
+            logger.debug("Keyring probe subprocess failed", exc_info=True)
             return False
 
 
--- a/ggshield/core/config/token_store.py
+++ b/ggshield/core/config/token_store.py
@@ -1,5 +1,5 @@
 import logging
-import subprocess
+import multiprocessing
 import sys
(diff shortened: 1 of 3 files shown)


ggshield/core/config/token_store.py
```diff
@@ -58,22 +60,34 @@ def delete_token(self, instance_url: str) -> None:
...
-            probe_key = "__ggshield_probe__"
-            keyring.set_password(KEYRING_SERVICE, probe_key, "test")
-            val = keyring.get_password(KEYRING_SERVICE, probe_key)
-            try:
-                keyring.delete_password(KEYRING_SERVICE, probe_key)
-            except Exception:
-                logger.debug("Failed to clean up keyring probe key")
-            return val == "test"
+            result = subprocess.run(
+                [sys.executable, "-c", probe],
```
clement-tourriere: **Blocker — this breaks standalone (PyInstaller) builds.**

This assumes `sys.executable` is a Python interpreter. That's true for pip/pipx installs, but not for our PyInstaller standalone bundles (`scripts/build-os-packages/build-os-packages` runs `pyinstaller ggshield/__main__.py --name ggshield`). In a frozen build `sys.executable` is the `ggshield` binary, so this actually runs:

```
ggshield -c "import keyring, keyring.backends.fail;..."
```

`-c` is `--config-path` (`__main__.py:104`), typed `RealPath(exists=True)` and eager, so Click rejects the probe string as a non-existent path and exits with code 2 → `returncode != 0` → `is_available()` returns `False`. Net effect: keyring is reported unavailable for *every* standalone user, and tokens silently fall back to the YAML config file instead of the OS credential store.

Since `multiprocessing.freeze_support()` is already called in `__main__.py:288`, a `multiprocessing.Process` probe would be frozen-safe (or a hidden probe sub-command handled before Click parses args). Either way it needs a smoke check against an actual frozen build — the in-process unit tests can't catch this.

ggshield/core/config/token_store.py
```diff
@@ -58,22 +60,34 @@ def delete_token(self, instance_url: str) -> None:
...
-            # ChainerBackend may pass the isinstance check but still fail).
-            probe_key = "__ggshield_probe__"
-            keyring.set_password(KEYRING_SERVICE, probe_key, "test")
-            val = keyring.get_password(KEYRING_SERVICE, probe_key)
-            try:
-                keyring.delete_password(KEYRING_SERVICE, probe_key)
-            except Exception:
-                logger.debug("Failed to clean up keyring probe key")
-            return val == "test"
+            result = subprocess.run(
```
clement-tourriere: Worth weighing the cost on the happy path too: `AuthConfig.load()` calls `get_token_store()` on most commands unless `GITGUARDIAN_API_KEY` is set (`auth_config.py:213`), so this now spawns a fresh interpreter + `import keyring` (backend discovery) on nearly every invocation. It's cached per-process, but that's roughly 150–400ms of cold start added to e.g. every `secret scan`. Might be worth measuring, or only running the probe when we actually have keyring-backed tokens to read.

ggshield/core/config/token_store.py
```diff
@@ -58,22 +60,34 @@ def delete_token(self, instance_url: str) -> None:
...
+        The probe runs in a subprocess so that a segfault in a native backend
+        (e.g. libsecret or KWallet) doesn't crash the main process.
+        """
+        probe = (
+            "import keyring, keyring.backends.fail;"
+            f"kr = keyring.get_keyring();"
+            f"exit(2) if isinstance(kr, keyring.backends.fail.Keyring) else None;"
+            f"keyring.set_password({KEYRING_SERVICE!r}, '__ggshield_probe__', 'test');"
+            f"val = keyring.get_password({KEYRING_SERVICE!r}, '__ggshield_probe__');"
+            f"(keyring.delete_password({KEYRING_SERVICE!r}, '__ggshield_probe__'));"
```
clement-tourriere: `delete_password` is unguarded, so if cleanup raises *after* a successful set/get round-trip, the subprocess exits non-zero and we disable a backend that actually works. We already know delete can legitimately raise — `delete_token()` catches `PasswordDeleteError` at line 57, and the previous `is_available()` wrapped this exact call in `try/except`. Suggest making the delete best-effort and exiting on `val == 'test'`, e.g. wrap it in a `try/except` inside the probe.

ggshield/core/config/token_store.py
```diff
@@ -58,22 +60,34 @@ def delete_token(self, instance_url: str) -> None:
...
+            f"exit(0 if val == 'test' else 3)"
```
clement-tourriere: Two small things:

1. The `0/2/3/other` exit-code scheme doesn't change behaviour — everything non-zero collapses to `False` in `is_available`, and code 2 collides with Click's own usage-error exit (see the `sys.executable` thread). So the distinct codes mostly add complexity without a payoff.
2. `exit()` is injected by the `site` module; under `-S` it would raise `NameError`. `sys.exit(...)` / `raise SystemExit(...)` is more robust. (And the parens around the `delete_password` call on the line above are a no-op.)

tests/unit/core/config/test_token_store.py
```diff
@@ -55,41 +56,34 @@ def test_delete_token_not_found(self):
...
         store = KeyringTokenStore()
-        mock_keyring = MagicMock()
-        with (
-            patch("keyring.get_keyring", return_value=mock_keyring),
-            patch("keyring.set_password") as mock_set,
-            patch("keyring.get_password", return_value="test") as mock_get,
-            patch("keyring.delete_password") as mock_delete,
-        ):
+        ok = subprocess.CompletedProcess(args=[], returncode=0)
+        with patch("subprocess.run", return_value=ok):
```
clement-tourriere: Now that every test patches `subprocess.run` and asserts only on `returncode`, the probe string itself is never executed — a syntax error in it would still pass CI, and the `sys.executable`/frozen problem is invisible to the suite. The 2 / 3 / 139 cases have also collapsed into identical "non-zero → False" assertions. Could we keep at least one test that runs the real probe against a fake backend, plus one asserting the command we spawn, so this path stays covered?

clement-tourriere (changes requested):
Thanks for tackling this — isolating a crash-prone native backend in a child process so a SIGSEGV becomes a graceful fallback is the right instinct.

My main concern is the choice of child process: `sys.executable -c <python>` assumes `sys.executable` is a Python interpreter, which holds for pip/pipx installs but not for our PyInstaller standalone bundles — and those are a primary distribution channel. In a frozen build this ends up running `ggshield -c "<probe>"`, which Click rejects, so keyring reports as unavailable and **every standalone user silently falls back to file-based token storage**. That trades a rare Linux crash for a guaranteed loss of OS-keyring storage for the majority of installs, so I'd hold the merge on it. Details inline, plus a couple of smaller points on cleanup handling and test coverage.

Good news: `multiprocessing.freeze_support()` is already wired in `__main__.py`, so a `multiprocessing.Process`-based probe would be frozen-safe. Happy to pair on it.

ggshield/core/config/token_store.py
GG-HH: I tested on my machine, it indeed seems to add ~100ms. (to be compared to ~1s to run api-status). I agree it is not great.

GG-HH:
Closing as we saw that the problem was very unlikely and had other causes than initially thought.

PR #1275, opened 2026-06-12
fix(ai): make the AI hook resilient to auth failures

amascia-gg:
## Context

On macOS, `ggshield secret scan ai-hook` could crash with `Error: No token is saved …` on **every** agent tool call, which agents misread as a block — so the hook looked like it was working when in fact it never scanned. 

### Root cause (refined during investigation)

The installed hook command was the **bare** string `ggshield secret scan ai-hook`. The hook runs as an agent-spawned process whose `PATH` is **not** the user's shell PATH and is **not stable across launch contexts** (a terminal-launched agent and a GUI/Dock-launched one see different PATHs). On a machine with more than one ggshield install (e.g. Homebrew **and** an MDM-managed copy), the bare command could resolve to a *different* binary than the one the user authenticated with. macOS Keychain read permission is tied to the exact calling binary (for the ad-hoc-signed Homebrew interpreter, to its code hash — so it is also invalidated by every `brew upgrade`), so the token read returned nothing → "No token is saved". Because that error was raised *outside* the hook's try/except, it crashed instead of degrading.

## Changes

Three independent, complementary fixes:

1. **Fail open with an agent-visible warning** (`ai_hook.py`). Any auth / network / unexpected failure no longer crashes the hook: it emits a well-formed *allow* response carrying a warning in the agent's own format (`systemMessage` for Claude Code / VS Code / Codex, `user_message` for Cursor) and exits 0. The agent stays usable and the user is told the action was **NOT** scanned, with remediation. Falls back to exit 1 only when the agent can't be identified.

2. **Install-time auth preflight** (`install.py`). After installing the hook, `ggshield install -t <agent>` verifies it can authenticate + reach GitGuardian, and prints either "ready to scan" or a clear warning. On macOS this also triggers the Keychain authorization prompt while the user is still interactive (rather than invisibly inside a hook).

3. **Pin the hook to the absolute ggshield path** (`installation.py`). `install` now writes the absolute path of the ggshield that ran it instead of a bare `ggshield`, so the hook always runs the same binary the user authenticated with. The stable launcher path is used (symlinks are **not** resolved) so it survives version upgrades; falls back to the bare command when the path can't be determined (e.g. `python -m ggshield`).

## Cross-platform

The path logic uses only portable primitives (`os.sep`/`os.altsep`/`os.name`, `shlex.quote`, `shutil.which`); macOS is only referenced in comments. POSIX and Windows branches (incl. space-quoting for paths like `C:\Users\John Doe\…`) are unit-tested. Verified live on macOS, including that invoking through a symlink preserves the stable launcher path rather than the version-pinned target. I could not execute on real Linux/Windows — a CI run on those platforms is the final confirmation.

## Tests

- New `TestBuildHookCommand` (7 cases): absolute argv0, PATH-resolved bare, `python -m` fallback, missing-file fallback, Windows quoted/unquoted, POSIX spaces.
- New fail-open command tests (auth failure, scan failure, unknown-agent) and `allow_with_warning` output tests for all four agent formats.
- New install preflight tests (success + failure warning).
- Full AI vertical + install suites pass locally (360 tests).

### DIFF
--- a/ggshield/cmd/install.py
+++ b/ggshield/cmd/install.py
@@ -1,12 +1,18 @@
 import os
 import subprocess
+import sys
 from pathlib import Path
 from typing import Any, Literal, Optional
 
 import click
 from click import UsageError
+from pygitguardian.models import HealthCheckResponse
 
 from ggshield.cmd.utils.common_options import add_common_options
+from ggshield.cmd.utils.context_obj import ContextObj
+from ggshield.core import ui
+from ggshield.core.client import create_client_from_config
+from ggshield.core.config import Config
 from ggshield.core.dirs import get_data_dir
 from ggshield.core.errors import UnexpectedError
 from ggshield.utils.git_shell import check_git_dir, git
@@ -46,7 +52,9 @@ fi
 @click.option("--force", "-f", is_flag=True, help="Overwrite any existing hook script.")
 @click.option("--append", "-a", is_flag=True, help="Append to existing script.")
 @add_common_options()
+@click.pass_context
 def install_cmd(
+    ctx: click.Context,
     mode: Literal["local", "global"],
     hook_type: str,
     force: bool,
@@ -62,7 +70,10 @@ def install_cmd(
     """
 
     if hook_type in AGENTS:
-        return install_hooks(name=hook_type, mode=mode, force=force)
+        returncode = install_hooks(name=hook_type, mode=mode, force=force)
+        if returncode == 0:
+            check_ai_hook_authentication(ContextObj.get(ctx).config)
+        return returncode
 
     return_code = (
         install_global(hook_type=hook_type, force=force, append=append)
@@ -70,6 +81,41 @@ def install_cmd(
         else install_local(hook_type=hook_type, force=force, append=append)
     )
     return return_code
+
+
+def _is_interactive() -> bool:
+    """Whether install is running with a user present at a terminal."""
+    return sys.stdout.isatty()
+
+
+def check_ai_hook_authentication(config: Config) -> None:
+    """Verify the freshly installed AI hook will be able to authenticate.
+
+    The hook runs as an agent-spawned, non-interactive process, where an auth
+    failure would otherwise only show up as a warning on every tool call.
+    Checking now also makes the OS credentials store ask for access (e.g. the
+    macOS Keychain authorization prompt) while the user can still answer it.
+
+    Skip this in non-interactive installs (CI, automated fleet/MDM
+    provisioning): there is no one to answer a credential-store popup or read
+    the result, and triggering the prompt there would be disruptive.
+    """
+    if not _is_interactive():
+        return
+    try:
+        client = create_client_from_config(config)
+        response = client.health_check()
+        if not isinstance(response, HealthCheckResponse) or response.status_code != 200:
+            raise UnexpectedError(str(getattr(response, "detail", response)))
+    except Exception as exc:
+        ui.display_warning(
+            f"The hook is installed but ggshield cannot reach GitGuardian: {exc}\n"
+            "The hook will NOT scan anything until this is fixed. Run "
+            "'ggshield auth login' to authenticate, then 'ggshield api-status' "
+            "to check."
+        )
+    else:
+        click.echo("ggshield successfully authenticated: the hook is ready to scan.")
 
 
 def install_global(hook_type: str, force: bool, append: bool) -> int:
--- a/ggshield/verticals/ai/installation.py
+++ b/ggshield/verticals/ai/installation.py
@@ -1,4 +1,7 @@
 import json
+import os
+import shlex
+import sys
 from copy import deepcopy
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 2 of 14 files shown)


ggshield/cmd/install.py
```diff
@@ -72,6 +82,30 @@ def install_cmd(
     return return_code
 
 
+def check_ai_hook_authentication(config: Config) -> None:
```
xblanchot-gg, line 91: Should we do this at install time, or at SessionStart in hooks ? 
Because with fleet MDM deployment, having a random popup for ggshield accessing keyring might be weird. No strong opinion here I don't know what's the best way. 

Something I'm not sure about: the machine scan already requires ggshield to access the token, so I don't get why ggshield here would not have access to the auth ? That looks like something that should be handled at a higher level in ggshield no ?

ggshield/verticals/ai/installation.py
```diff
@@ -12,12 +16,70 @@
...
+    (e.g. ``python -m ggshield``), so behavior is never worse than before.
+    """
+    executable = _find_ggshield_executable()
+    if executable is None:
+        return BARE_HOOK_COMMAND
+    return f"{_quote_executable(executable)} secret scan ai-hook"
+
+
+def _find_ggshield_executable() -> Optional[str]:
+    """Return the absolute path of the running ggshield executable, or None."""
```
xblanchot-gg: I think I would rather take everything until gghsield in the command, few examples to test `ggshield`, `/usr/local/bin/ggshield`, `uvx run ggshield`

xblanchot-gg (changes requested):
Thanks, a few comments on the scope to discuss

ggshield/verticals/ai/installation.py
```diff
@@ -12,12 +16,70 @@
...
+        # (unlike the hook's) is the user's real shell PATH.
+        path = shutil.which(candidate) if candidate else None
+        path = path or shutil.which("ggshield")
+
+    if not path or not os.path.isfile(path):
+        return None
+    # Guard against launchers where argv[0] is not the ggshield executable,
+    # such as `python -m ggshield` (argv[0] is a .py file or the interpreter).
+    if not os.path.basename(path).lower().startswith("ggshield"):
+        return None
```
xblanchot-gg: Following my comment above, if this is the command used to install the hooks, we should not try to add intelligence here, this is a lost battle IMO, we should trust the command used whatever it is

ggshield/cmd/install.py
amascia-gg: Made the check interactive-only, it only runs for a user installing at a terminal.
To answer the second part: in our case we run ggshield from an agent session, not the terminal. It means that that the ggshield binary used can be different from the one that run `auth login` which is not the case for `machine scan`

ggshield/verticals/ai/installation.py
amascia-gg: sys.argv[0] is already the resolved ggshield path.
I tried with:
  ▎ - ggshield → resolved via PATH → /opt/homebrew/bin/ggshield …
  ▎ - /usr/local/bin/ggshield → used as-is
  ▎ - uv run ggshield → …/.venv/bin/ggshield …

ggshield/verticals/ai/installation.py
amascia-gg: Tested that even a bare ggshield (resolved via PATH) gives an absolute sys.argv[0], so the which() fallback was dead code

xblanchot-gg (approved):
Thanks for the fix !

ggshield/verticals/ai/installation.py
```diff
@@ -18,6 +21,36 @@ class InstallationStats:
...
+    Homebrew plus an MDM-managed copy on macOS), a bare command can resolve to a
+    *different* binary than the one the user authenticated with, which then fails
+    to read the stored token.
+
+    We trust whatever executable ran ``install`` (``sys.argv[0]``); we only make
+    it absolute and shell-quote it. ``abspath`` does not resolve symlinks, so a
+    package manager's stable launcher (e.g. ``/opt/homebrew/bin/ggshield``) is
+    kept rather than a version-pinned path that would break on upgrade.
+    """
+    executable = os.path.abspath(sys.argv[0])
```
clement-tourriere, line 41: Blocking: `abspath(sys.argv[0])` does not implement the fallback behavior described in the PR. For `python -m ggshield`, `sys.argv[0]` points to the package `__main__.py`, so this installs something like `/.../ggshield/__main__.py secret scan ai-hook` instead of invoking Python. If `argv[0]` is bare, `abspath('ggshield')` also turns into `$PWD/ggshield` rather than resolving through PATH. Please distinguish an executable launcher from module/script invocation, use `shutil.which()` for bare launchers, and fall back to a quoted `sys.executable -m ggshield secret scan ai-hook` form when needed.

ggshield/verticals/ai/installation.py
```diff
@@ -39,7 +72,7 @@ def install_hooks(
     base_dir = get_user_home_dir() if mode == "global" else Path(".")
     settings_path = base_dir / agent.settings_path(mode)
 
-    command = "ggshield secret scan ai-hook"
+    command = build_hook_command()
```
clement-tourriere, line 75: Blocking: this new command is only used when a `<COMMAND>` placeholder is inserted. Existing ggshield-owned AI hooks such as `ggshield secret scan ai-hook` are located and counted as already present by `_fill_dict()`, but they are left untouched unless the user passes `--force`. That means the main broken legacy install keeps the bare command after users rerun `ggshield install -t ...`. We should migrate recognized ggshield AI-hook commands to this freshly built command by default, and add a regression test with a pre-existing bare command.

ggshield/cmd/install.py
```diff
@@ -72,6 +83,41 @@ def install_cmd(
...
+
+    Skip this in non-interactive installs (CI, automated fleet/MDM
+    provisioning): there is no one to answer a credential-store popup or read
+    the result, and triggering the prompt there would be disruptive.
+    """
+    if not _is_interactive():
+        return
+    try:
+        client = create_client_from_config(config)
+        response = client.health_check()
```
clement-tourriere, line 107: Important: `health_check()` only proves the server is reachable and that reading config/keychain did not throw; it does not prove that the token is accepted by the API. This can print “the hook is ready to scan” with an invalid token. Please reuse the same auth validation as `api-status` / `check_client_api_key(..., required_scopes=set())` (or an equivalent hook self-check) before reporting readiness.

clement-tourriere (changes requested):
Thanks for digging into this — fail-open plus install-time awareness is the right direction. I’m requesting changes because the current path fix does not reliably update the broken installed command for existing users, the argv0 handling does not cover the fallback cases described in the PR, and the preflight can report that the hook is ready even when the installed hook will still run a different/broken command.

PR #1276, opened 2026-06-12
fix(secret): censor ignored secrets appearing in context lines

GG-HH:
## Context

When a patch contains two adjacent secrets and the first one is on a context or deletion line (`DiffKind.CONTEXT` / `DiffKind.DELETION`), it is filtered out as `NOT_INTRODUCED` and `Result.censor()` never censors it. Since `Line` objects are shared by reference across all matches built from the same content, the ignored secret leaks in plaintext through the `lines_before_secret` / `lines_after_secret` context of the adjacent displayed secret, even with `show_secrets=False`.

Fixes [END-523](https://linear.app/gitguardian/issue/END-523/ggshield-ignored-secrets-leak-in-plaintext-as-context-lines-in-prci).

## What has been done

- Added a `filtered_out_matches` field to `Result`: in `from_scan_result()`, policy breaks that are dropped (when `--all-secrets` is off) now also get their matches converted to `ExtendedMatch` objects and kept there.
- `Result.censor()` now censors those matches alongside the displayed ones. `ExtendedMatch.censor()` modifies the shared `Line` objects in place with a length-preserving replacement, so the spans of displayed secrets are unaffected.
- Added a regression test reproducing the leak, plus a changelog entry.

With `--all-secrets`, ignored breaks become regular `Secret`s and were already censored, so behavior there is unchanged. With `--show-secrets`, nothing is censored, as before.

## Validation

- Run `pytest tests/unit/verticals/secret/output/test_ignored_secret_context_leak.py`: it fails on `main` (the context line shows the secret in plaintext) and passes on this branch (the context line renders as `SMTP_PASSWORD = "exi*****************123"`).
- Full unit suite passes (2106 tests).

## PR check list

### DIFF
--- a/.gitguardian.yaml
+++ b/.gitguardian.yaml
@@ -24,6 +24,10 @@ secret:
       name: Test secret mentioned in commit 6282ceb9035c1bfcceec5f6372f69063cdd588f7
     - match: e90eaa00d17c03309b660e1879c70b815509422c2d4cf0980b2e1221dffa0381
       name: Test secret mentioned in commit 6282ceb9035c1bfcceec5f6372f69063cdd588f7
+    - match: 796c064d6cfd48964052633c64979c19bf232e9fe7da667b75f1a0485d62fb8d
+      name: Fake secret in test_ignored_secret_context_leak.py
+    - match: 5af590dc986688e893c7cb22638a4175d8e3b3268a9499c4fb007d5f43cf3b8b
+      name: Fake secret in test_ignored_secret_context_leak.py
 
   ignored_paths:
     - '**/README.md'
--- a/changelog.d/20260612_henri.hubert_ignored_secret_context_leak.md
+++ b/changelog.d/20260612_henri.hubert_ignored_secret_context_leak.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Ignored secrets (for example secrets on context or deleted lines of a patch) no longer appear in plaintext when they show up in the context lines of another displayed secret.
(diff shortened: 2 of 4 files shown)


clement-tourriere (approved):

benjaminrigaud-gg (approved):
LGTM.

Claudes found that:

>   MEDIUM-1 — Default scan path now feeds every ignored match through assertion-laden from_match (my original MEDIUM ≡ revbot Finding 1)
  secret_scan_collection.py:226-230. The new comprehension materializes ExtendedMatch.from_match(...) for all filtered-out breaks in the default
  (--all-secrets off) path, where main only counted and discarded them. from_match/MatchSpan.from_match can raise ValueError/AssertionError,
  and from_scan_result (secret_scanner.py:230) is unguarded, so a non-mappable ignored match would turn a previously-clean scan into an
  UnexpectedError.
  Verified: DELETION/CONTEXT/CRLF cases all censor cleanly with no crash; no crash reproduced with valid patches. Real but low-likelihood (same
  path already live under --all-secrets).
  Action: add DELETION + multiline regression tests; optionally a narrow guard around the comprehension. Non-blocking.

 > MEDIUM-2 — test_show_secrets_reveals_both under-asserts and is misnamed (I rated LOW; revbot rated MEDIUM — I'll defer to MEDIUM since it's
  the only guard on the show_secrets=True path)
  tests/unit/.../test_ignored_secret_context_leak.py:127-135. Name/docstring say "both values appear" but it only asserts _NEW_SECRET in output;
  the ignored _EXISTING_SECRET is never checked, so it can't catch an over-censoring regression.
  Verified: with show_secrets=True both secrets render deterministically.
  Action: add assert _EXISTING_SECRET in output (drop the "may appear" hedge), or rename the test.

PR #1279, opened 2026-06-15
feat(scripts): add ggshield install/uninstall scripts

benjaminrigaud-gg:
## What

One-line install/uninstall scripts for ggshield, aimed at developers and non-developers:

- **Linux / macOS** (`install.sh`): `curl --proto '=https' --tlsv1.2 -sSfL …/install.sh | bash`
- **Windows** (`install.ps1`): `irm …/install.ps1 | iex`

The installer detects OS/arch and picks the best method available:

| Platform | Method |
| --- | --- |
| macOS | Homebrew, else standalone tarball |
| Linux + apt/dnf/yum/zypper + sudo | Cloudsmith deb/rpm repo |
| Linux without sudo (x86_64 / arm64 glibc) | standalone tarball in `~/.local/bin` |
| Linux musl | pipx |
| Windows | MSI (elevated) → zip (per-user) → Chocolatey (last resort) |

## Highlights

- **Authentication** after install: `--instance` targets EU / self-hosted; `GITGUARDIAN_API_KEY` does a non-interactive token login.
- **Plugins are opt-in**: `--plugin <name>` (repeatable).
- **Download integrity**: standalone downloads are checksum-verified against GitHub's per-asset digest, plus build-provenance attestation when `gh` is available.
- **Uninstall** (`uninstall.sh` / `uninstall.ps1`): confirms each step, removes plugins, logs out, deletes config/cache/data, and removes every install method it detects (system package, brew, pipx, uv, pip, mise, macOS pkg, choco, msi, tarball/zip).

## Notes

- Scripts self-reference `raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/…` — those resolve once merged to `main`.
- arm64 Linux standalone builds are recent; on releases that predate them the installer falls back to pipx.

Refs: END-511

### DIFF
--- a/changelog.d/20260615_143744_benjaminrigaud_install_scripts.md
+++ b/changelog.d/20260615_143744_benjaminrigaud_install_scripts.md
@@ -1,0 +1,3 @@
+### Added
+
+- Install and uninstall scripts under `scripts/install/`: a one-line `curl | bash` (Linux/macOS) and `irm | iex` (Windows) installer that detects the platform, picks the best install method, authenticates, and optionally installs plugins, plus a matching uninstaller that cleanly removes ggshield and its data.
--- a/scripts/install/README.md
+++ b/scripts/install/README.md
@@ -1,0 +1,149 @@
+# ggshield install scripts
+
+One-line scripts to install ggshield on your machine (for developers and
+non-developers), authenticate, optionally install plugins (`--plugin
+<name>`), and cleanly uninstall it later.
+
+## Install
+
+Linux / macOS:
+
+```sh
+curl --proto '=https' --tlsv1.2 -sSfL \
+  https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.sh | bash
+```
+
+Windows (PowerShell):
+
+```powershell
+irm https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.ps1 | iex
+```
+
+Prefer inspecting before running:
+
+```sh
+curl --proto '=https' --tlsv1.2 -sSfL -o install.sh \
+  https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.sh
+less install.sh
+bash install.sh
+```
+
+The script detects the OS and architecture (including Rosetta 2 and musl) and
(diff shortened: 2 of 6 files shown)


GG-HH:
Tested on my machine(linux, ubuntu):
- both install and uninstall worked correctly
- would be nice to have detection of servers so that the authentication uses oob method as I first tried to do it on a virtual machine and it failed because of port forwarding
- I was surprised that the uninstall suggested to remove the services (ie: handled by the MDM), but that seems correct.

clement-tourriere (commented):
1. Default installer should be standalone user-local only
     - Linux/macOS: download the release tarball, verify checksum, install under ~/.local/share/ggshield-standalone, and symlink to ~/.local/bin.
     - Windows: default to per-user ZIP install.
     - MSI should be explicit only.

 2. Document other install methods instead of auto-selecting them
     - brew install ggshield
     - Cloudsmith apt/rpm setup
     - pipx install ggshield
     - Chocolatey/MSI/ZIP on Windows

 3. Auth and plugin setup should be best-effort post-install convenience
     - Installing/verifying the binary should be the only hard-fail part.
     - After install, it’s fine to try setup by default when appropriate:
         - if GITGUARDIAN_API_KEY is set, try token auth;
         - if an interactive TTY exists, try browser auth;
         - if non-interactive and no API key, skip auth with a warning.
     - If auth fails or the user cancels, warn but do not fail the installer.
     - If plugin installation fails, warn per plugin but do not fail the installer.
     - Always print retry commands at the end, for example:
         - ggshield auth login --instance ...
         - ggshield plugin install ...

 4. Uninstaller should only undo script-owned installs by default
     - Do not automatically remove Homebrew/pipx/uv/pip/system-package installs.
     - Add explicit flags for destructive cleanup:
         - --purge for config/cache/data
         - --clean-pre-commit-cache
         - --prune-uv-cache
         - --remove-system-scheduling

 5. Fail closed on verification
     - If the GitHub digest/checksum cannot be retrieved, abort.
     - --yes should not mean “continue without checksum verification”.

benjaminrigaud-gg:
Moving to a simpler version on https://github.com/GitGuardian/ggshield/pull/1287

PR #1287, opened 2026-06-16
feat(scripts): add ggshield install/uninstall scripts

benjaminrigaud-gg:
Supersedes #1279 — the simplified, standalone-only version requested in review.

## What changed vs #1279
One install method per OS, everything else documented rather than auto-selected:

- **Linux / macOS** → standalone tarball → `~/.local/share/ggshield-standalone` + `~/.local/bin` symlink (no sudo). No more brew/apt-rpm/pipx auto-routing.
- **Windows** → per-user **zip** by default; **`-Method msi`** explicit (admin). Chocolatey dropped from the script.
- **Other methods** (brew, Cloudsmith apt/rpm, pipx, choco, Docker) are now documented as manual alternatives.

## Behaviour
- **Best-effort auth & plugins**: installing the verified binary is the only hard-fail step. Auth/plugin failures warn and print retry commands; the instance is shown, with an `--method oob` hint for headless sessions.
- **Fail-closed verification**: a download whose sha256 digest can't be retrieved is never installed (no `-y` or interactive opt-out), plus attestation when `gh` is present.
- **Alpine/musl out of scope** (glibc-only build): clean refusal pointing to the Docker image / pipx.
- **Uninstall removes only the script-owned install by default**; config/cache (`--purge`), pre-commit cache, uv cache and systemd scheduling are opt-in flags. No auto-removal of brew/pipx/uv/pip/system installs.

## QA
Unix verified (Docker + qemu): x86_64 and **arm64** tarball installs on 1.52.0, best-effort auth (bad key → exit 0), musl refusal, default uninstall preserves foreign installs + config, `--purge`. Windows `.ps1` parse-checked; VM round-trip pending. macOS path is identical to Linux's (different target triple) but has no CI.

Refs: END-511

### DIFF
--- a/changelog.d/20260615_143744_benjaminrigaud_install_scripts.md
+++ b/changelog.d/20260615_143744_benjaminrigaud_install_scripts.md
@@ -1,0 +1,3 @@
+### Added
+
+- Install and uninstall scripts under `scripts/install/`: a one-line `curl | bash` (Linux/macOS) and `irm | iex` (Windows) installer for the standalone ggshield build that authenticates and optionally installs plugins, plus a matching uninstaller that removes the install it created.
--- a/scripts/install/README.md
+++ b/scripts/install/README.md
@@ -1,0 +1,154 @@
+# ggshield install scripts
+
+One-line scripts to install the standalone ggshield build on your machine (for
+developers and non-developers), authenticate, optionally install plugins, and
+cleanly uninstall it later. No admin/sudo required.
+
+## Install
+
+Linux / macOS:
+
+```sh
+curl --proto '=https' --tlsv1.2 -sSfL \
+  https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.sh | bash
+```
+
+Windows (PowerShell):
+
+```powershell
+irm https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.ps1 | iex
+```
+
+The `irm | iex` form runs the script content directly, so it is not subject to
+the execution policy. If you instead download and run the `.ps1` file, Windows
+requires `powershell -ExecutionPolicy Bypass -File .\install.ps1`.
+
+Prefer inspecting before running:
+
+```sh
+curl --proto '=https' --tlsv1.2 -sSfL -o install.sh \
+  https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.sh
+less install.sh
+bash install.sh
+```
+
+The script installs the standalone build (detecting OS/arch, including Rosetta 2) to:
+
+- **Linux / macOS** → `~/.local/share/ggshield-standalone`, symlinked into
+  `~/.local/bin` (no sudo).
+- **Windows** → `%LOCALAPPDATA%\Programs\ggshield`, added to your user PATH (no
+  admin). `-Method msi` installs the system-wide MSI instead (requires admin).
+
+The download is **checksum-verified** against the digest GitHub publishes for
+each asset (the install aborts if it cannot be verified), and — when `gh` is
+available — against the release's [build provenance attestation](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds).
+
+Alpine/musl is not supported (the standalone build is glibc-only) — use the
+Docker image `gitguardian/ggshield` or `pipx install ggshield`.
+
+### Options
+
+```text
+-y, --yes           never prompt (for CI)
+    --instance URL  GitGuardian instance (default: https://dashboard.gitguardian.com)
+    --version X.Y.Z ggshield version to install (default: latest)
+    --install-only  install ggshield only, skip auth and plugins
+    --plugin NAME   install this ggshield plugin (repeatable)
+```
+
+Pass options through the pipe with `bash -s --`:
+
+```sh
+curl --proto '=https' --tlsv1.2 -sSfL \
+  https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.sh |
+  bash -s -- --plugin <plugin_name>
+```
+
+`install.ps1` takes the equivalent options (`-Yes`, `-Instance URL`,
+`-Version X.Y.Z`, `-Method zip|msi`, `-InstallOnly`, `-Plugin name[,name]`):
+
+```powershell
+& ([scriptblock]::Create((irm https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.ps1))) `
+  -Plugin <plugin_name> -Yes
+```
+
+Authentication and plugin installation are **best-effort**: if they fail (wrong
+instance, cancelled login, no network), the script warns and prints the retry
+commands, but the install itself still succeeds.
+
+### Instance
+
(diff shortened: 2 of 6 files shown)


clement-tourriere:
1. Support GITGUARDIAN_INSTANCE
    ggshield already supports GITGUARDIAN_INSTANCE, so the scripts should probably honor it too, especially for EU/custom/self-hosted instances.
     - Unix: use GITGUARDIAN_INSTANCE
     - PowerShell: same idea, prefer $env:GITGUARDIAN_INSTANCE.

 2. Keep uninstall focused on the script-owned install
    I’d remove the extra cleanup flags/code for:
     - --clean-pre-commit-cache
     - --prune-uv-cache
     - --remove-system-scheduling

    --purge for ggshield config/cache/data is fine, but pre-commit/uv/systemd cleanup feels broader than what this standalone installer should own.

 3. Guard against unsafe uninstall paths
    Before doing rm -rf "$OPT_DIR", refuse unsafe values such as:
     - empty path
     - /
     - $HOME
     - $BIN_DIR

    This protects users who accidentally set GGSHIELD_OPT_DIR incorrectly.

 4. Do not overwrite an existing non-symlink binary
    Before ln -sf "$bin" "$BIN_DIR/ggshield", if $BIN_DIR/ggshield exists and is not a symlink, abort with a clear message. This avoids replacing a manually installed binary.

benjaminrigaud-gg:
Thanks @clement-tourriere — all four addressed in the latest push (force-pushed; history is now 3 commits: unix / windows / docs):

1. **`GITGUARDIAN_INSTANCE`** — honored. `install.sh` defaults `--instance` to `$GITGUARDIAN_INSTANCE`; `install.ps1`'s `-Instance` defaults to `$env:GITGUARDIAN_INSTANCE` (it was wrongly reading `GGSHIELD_INSTANCE`). Added to the env-var table.
2. **Extra cleanup flags removed** — dropped `--clean-pre-commit-cache`, `--prune-uv-cache`, `--remove-system-scheduling` (and the Windows equivalents) and their code. Only `--purge` (config/cache/data) remains.
3. **Unsafe uninstall paths guarded** — `uninstall.sh` refuses empty / `/` / `$HOME` / `$BIN_DIR` before `rm -rf`; `uninstall.ps1` refuses empty / drive-root / `%USERPROFILE%` / `%LOCALAPPDATA%`.
4. **Won't clobber a non-symlink binary** — install aborts if `$BIN_DIR/ggshield` exists and is not a symlink.

### QA (Docker + qemu-arm64 + a Windows VM, against the Sandbox workspace)
Each fix verified, plus a real auth + `machine_scan` plugin round-trip on **both Linux and Windows** (`machine_scan v0.58.0`, signed; `machine` command tree available; clean uninstall).

The real-key run also caught and fixed a bug: `run_gg` guarded the `/dev/tty` reconnect with `[ -e /dev/tty ]`, but the device can exist yet be unopenable with no controlling terminal — plugin install died with `/dev/tty: No such device or address` in containers/CI. It now tests that `/dev/tty` actually opens, so non-interactive `curl | bash` plugin installs work.

benjaminrigaud-gg:
Pushed fixes for both points you hit:

**Shadowing / wrong version reported.** Root cause: the script ran whatever `ggshield` resolved to on PATH, and Windows searches the **Machine PATH before the user PATH** where the per-user zip lands — so your pre-existing 1.51.0 shadowed the fresh 1.52.0 (hence "Installing 1.52.0" → "Installed: 1.51.0", and the old binary printing the update notice). `install.ps1` now:
- reports/authenticates/installs-plugins using the **exact binary it just installed** (the zip's `ggshield.exe`, or the MSI's `InstallLocation`), so the version is always correct;
- **warns when another ggshield on PATH shadows it**, naming both paths.

This matches what `install.sh` already did on Linux.

Reproduced your case on a Windows VM — MSI **1.51.0** on the machine PATH, then the **1.52.0** zip under it:
```
==> Installed: ggshield, version 1.52.0
warning: another ggshield on your PATH (C:\Program Files\GitGuardian\ggshield\ggshield.exe) shadows the one just installed (...\Programs\ggshield\ggshield-1.52.0-x86_64-pc-windows-msvc\ggshield.exe)
warning: remove it, fix your PATH order, or open a new terminal
```

**ExecutionPolicy.** Expected when running the saved `.ps1` file (Windows blocks unsigned script files). The documented `irm … | iex` runs the content directly with no bypass; for the file form use `powershell -ExecutionPolicy Bypass -File .\install.ps1`. Added a README note.

(The `codecov/project` red is unrelated — this PR has no Python changes; coverage is unchanged.)

benjaminrigaud-gg:
Thanks for testing on macOS — good catch. That's the **bash 3.2** quirk: macOS ships bash 3.2.57, where expanding an *empty* array as `"${arr[@]}"` under `set -u` errors (only fixed in bash 4.4+). With no `--plugin`, the plugin loop tripped it right at the end — after a successful install + auth. Fixed with the `${PLUGINS[@]+"${PLUGINS[@]}"}` guard, and verified under the real `bash:3.2` image (old form reproduces your `PLUGINS[@]: unbound variable`; new form is clean for empty and populated arrays).

Everything else in your run worked as intended — the `arm64-apple-darwin` tarball + attestation, replacing the old uv symlink, and the shadow warning correctly flagging `/usr/local/bin/ggshield`. Re-pull and it'll complete cleanly. 🙏

clement-tourriere (commented):
Small remark
```zsh ❯ ./install.sh --plugin machine_scan
==> Platform: darwin/arm64 — installing the standalone build
==> Resolving latest ggshield version
==> Downloading ggshield-1.52.1-arm64-apple-darwin.tar.gz
==> Verifying sha256 checksum
==> Verifying build provenance attestation
==> Installing to /Users/ctourriere/.local/share/ggshield-standalone/ggshield-1.52.1-arm64-apple-darwin
==> Installed: ggshield, version 1.52.1
warning: another ggshield at /usr/local/bin/ggshield shadows the one just installed (/Users/ctourriere/.local/bin/ggshield); fix your PATH order
==> Authenticating against https://dashboard.gitguardian.com
ggshield is already authenticated without an expiry date
==> Installing the machine_scan plugin
Installing machine_scan...
Installed machine_scan v0.58.0
==> ggshield is installed. To finish setup later:
    ggshield auth login --instance https://dashboard.gitguardian.com
    # EU workspace: ggshield auth login --instance https://dashboard.eu1.gitguardian.com
    ggshield plugin install machine_scan
```


It's strange to have the "setup later" message even if everything worked well before

benjaminrigaud-gg:
Fixed — thanks, that "To finish setup later" line is now only printed when something is actually left to do.

- **Full success** (auth + any requested plugins both OK, or auth OK with no `--plugin`) → the script just prints `==> ggshield is ready.`
- **Something pending** → `To finish setup:` followed by *only* the missing step(s): the `auth login` hint if auth failed/was skipped, the `plugin install` hint if a plugin failed. `--install-only` shows both (by design).

So your run (install + auth + `machine_scan` all succeeding) now ends with `ggshield is ready.` instead of re-listing the plugin you just installed.

Verified by exercising the real `post_install` across all five branches (success / auth-fail / plugin-fail / install-only / no-plugins), on both modern bash and bash 3.2. Pushed.

clement-tourriere (approved):

PR #1293, opened 2026-06-17
fix(macos): sign ggshield launcher with allow-jit entitlement

### LINKED ISSUE #1292: macOS: ggshield machine scan 6-7x slower (hardened launcher missing allow-jit entitlement)
## Summary

On macOS, `ggshield machine scan` runs ~6–7× slower than it should. The machine-scan plugin matches secrets with **PCRE2**, whose JIT is several times faster than its interpreter on regex-heavy content. The standalone `ggshield` launcher is code-signed with the **hardened runtime** but **no entitlements**, so the kernel denies JIT (`MAP_JIT`) memory: `pcre2_jit_compile` fails and PCRE2 silently falls back to the interpreter.

## Impact

- Machine scans on macOS run at a fraction of full speed (≈4–5 MB/s vs ≈30 MB/s on regex-dense files) while pegging every core.
- Version-independent — every hardened build is affected (observed on 1.52.0 and 1.52.1).

## Root cause

`scripts/build-os-packages/macos-functions.bash` signs the launcher with `rcodesign sign --code-signature-flags runtime` and **no** `--entitlements-xml-path`, so it ships hardened with zero entitlements:

```
$ codesign -dv --entitlements - .../ggshield
flags=0x10000(runtime)    # hardened runtime
(no entitlements)         # no allow-jit
```

JIT permission on macOS is a property of the **main executable**, so the `ggshield` launcher itself must carry `com.apple.security.cs.allow-jit`.

## Evidence

Same engine, same files, only the code signature changed:

| signature | CPU (user) | throughput |
|---|---|---|
| adhoc (no hardened runtime) | 73 s | 24 MB/s |
| hardened, no `allow-jit` | 528 s | 2 MB/s |
| hardened + `allow-jit` | 73–76 s | 18–22 MB/s |

The same engine runs at full speed under a non-hardened interpreter (`uvx ggshield`), confirming the launcher signature — not the engine — is the cause.

## Fix

Attach `com.apple.security.cs.allow-jit` to the launcher when signing (see linked PR). Library validation is intentionally left intact: first-party plugins are same-team signed and continue to load.

clement-tourriere:
## What

Sign the macOS `ggshield` launcher with the `com.apple.security.cs.allow-jit` entitlement.

## Why

The launcher is signed with the hardened runtime but no entitlements, so the kernel denies JIT (`MAP_JIT`) memory. The machine-scan plugin matches with PCRE2, whose JIT is several times faster than its interpreter on regex-heavy content; without the entitlement `pcre2_jit_compile` fails and PCRE2 silently falls back to the interpreter, making `ggshield machine scan` ~6–7× slower on macOS.

JIT permission is a property of the process's main executable, so the entitlement is attached to the `ggshield` launcher only (the bundled libraries inherit the process permission). Library validation is intentionally left intact — first-party plugins are same-team signed and continue to load.

Closes #1292.

## Changes

- Add `scripts/build-os-packages/macos-entitlements.plist` (just `allow-jit`).
- `macos_sign_file`: pass `--entitlements-xml-path …/macos-entitlements.plist` when signing the `ggshield` launcher.

## Test plan

Reproduced and verified by signing a binary three ways and scanning the same files (only the signature changed):

| signature | CPU (user) | throughput |
|---|---|---|
| adhoc (no hardened runtime) | 73 s | 24 MB/s |
| hardened, no `allow-jit` | 528 s | 2 MB/s |
| hardened + `allow-jit` | 73–76 s | 18–22 MB/s |

On a signed release artifact, confirm `codesign -d --entitlements - …/ggshield` shows `com.apple.security.cs.allow-jit`, and that `ggshield machine scan` runs at full speed.

### DIFF
--- a/changelog.d/20260617_091316_clement.tourriere_macos_allow_jit_entitlement.md
+++ b/changelog.d/20260617_091316_clement.tourriere_macos_allow_jit_entitlement.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- macOS: `ggshield machine scan` is no longer several times slower than on other platforms. The signed launcher now carries the `com.apple.security.cs.allow-jit` entitlement, so the scanner's PCRE2 JIT works under the hardened runtime instead of silently falling back to the interpreter.
--- a/scripts/build-os-packages/macos-entitlements.plist
+++ b/scripts/build-os-packages/macos-entitlements.plist
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <!--
+      The machine-scan plugin (satori) matches with PCRE2, whose JIT is ~6-7x
+      faster than its interpreter on regex-heavy files. JIT needs executable
+      (MAP_JIT) memory, which the hardened runtime denies unless the process's
+      MAIN executable carries this entitlement. Without it PCRE2 silently falls
+      back to the interpreter and `ggshield machine scan` is dramatically
+      slower. Attached to the `ggshield` launcher in macos_sign_file().
+    -->
+    <key>com.apple.security.cs.allow-jit</key>
+    <true/>
+</dict>
+</plist>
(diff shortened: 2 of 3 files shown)


benjaminrigaud-gg (approved):
It could work 🤷

PR #1301, opened 2026-06-17
feat: support Python 3.13

benjaminrigaud-gg:
## Summary

ggshield already *installed* on Python 3.13 (`requires-python` is uncapped) but 3.13 was neither tested nor advertised, and the dev/test suite crashed there because the locked `pyfakefs` 5.5.0 is not 3.13-compatible.

- Widen the `pyfakefs` constraint to `>=5.7.0` and re-lock. uv forks it to **6.2.0** on Python ≥3.10 and **5.10.2** on 3.9 (pyfakefs 6.x dropped 3.9 support).
- Add `'3.13'` to the CI build/test matrix in `.github/workflows/ci.yml`.
- Add the `Programming Language :: Python :: 3.13` trove classifier.

`requires-python` is **left uncapped on purpose** — capping it would make pip/uv silently pin 3.13-only users to a stale ggshield. No runtime code changes are needed: ggshield imports none of the stdlib modules removed in 3.13 (PEP 594). Track 2 (a 3.13-based frozen binary) is intentionally out of scope — the release binary is built on Python 3.10 and is self-contained.

## Test plan

- [x] Full unit suite passes on **Python 3.13** (2146 passed, run serially as CI does)
- [x] Full unit suite passes on **Python 3.12** — no regression from the pyfakefs 6.2.0 bump (2146 passed)
- [x] `uv sync --locked --group dev --group tests` consistent on 3.13
- [x] `pip install` of the built wheel into a clean 3.13-only venv works; `ggshield --version` / `--help` run; compiled deps (`cryptography`, `cffi`, `charset-normalizer`) import — all from cp313 wheels (nothing built from sdist)
- [x] No GitLab mirror leak in `uv.lock`; pre-commit hooks pass

Closes END-561

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04, macos-15-intel, windows-2022]
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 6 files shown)


GG-HH:
Good idea! Actually, in homebrew, we even use python3.14!

clement-tourriere (approved):

PR #1302, opened 2026-06-18
Install script: trim redundant curl --proto and fix Windows uninstall prompt

benjaminrigaud-gg:
Two small install-script changes (both on END-511).

## 1. Trim redundant curl flags from the install commands (`docs`)

Every documented install/uninstall command targets a literal `https://` URL on GitHub, which serves only TLS 1.2+. So both `--proto '=https'` (curl uses the scheme from the URL) and `--tlsv1.2` (the handshake negotiates ≥1.2 regardless) are redundant. The commands keep `-sSfL` — in particular `-f`, so an HTTP error page is never piped into `bash`.

Scope: the top-level `README.md`, `scripts/install/README.md`, and the `install.sh` / `uninstall.sh` documented one-liners. The scripts' own download helpers (`fetch`, `asset_http_status`) keep `--proto '=https' --tlsv1.2` as cheap defense-in-depth.

## 2. Fix the missing path in the Windows uninstall prompt (`fix`)

`uninstall.ps1` prompted `Remove the standalone ggshield at $ZipDir?`. In PowerShell `?` is a valid variable-name character (it's why the automatic `$?` exists), so the parser read the name as `ZipDir?` — undefined, expands to empty. The prompt rendered as `Remove the standalone ggshield at  [y/N]`, with no path and no `?`. Bracing the name (`${ZipDir}?`) keeps `ZipDir` as the variable and `?` literal.

Verified in a real PowerShell container (7.4.2); the `?`-as-name-char grammar is identical on Windows PowerShell 5.1. AST confirms `$ZipDir?` parses to `VariablePath = ZipDir?`.

Refs: [END-511](https://linear.app/gitguardian/issue/END-511/simple-install-script-to-test-ggshield)

### DIFF
--- a/scripts/install/README.md
+++ b/scripts/install/README.md
@@ -9,7 +9,7 @@ cleanly uninstall it later. No admin/sudo required.
 Linux / macOS:
 
 ```sh
-curl --proto '=https' --tlsv1.2 -sSfL \
+curl -sSfL \
   https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.sh | bash
 ```
 
@@ -26,7 +26,7 @@ requires `powershell -ExecutionPolicy Bypass -File .\install.ps1`.
 Prefer inspecting before running:
 
 ```sh
-curl --proto '=https' --tlsv1.2 -sSfL -o install.sh \
+curl -sSfL -o install.sh \
   https://raw.githubusercontent.com/GitGuardian/ggshield/main/scripts/install/install.sh
 less install.sh
 bash install.sh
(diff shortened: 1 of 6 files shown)


GG-HH (approved):

scripts/install/README.md
```diff
@@ -26,7 +26,7 @@ requires `powershell -ExecutionPolicy Bypass -File .\install.ps1`.
 Prefer inspecting before running:
 
 ```sh
-curl --proto '=https' --tlsv1.2 -sSfL -o install.sh \
+curl --tlsv1.2 -sSfL -o install.sh \
```
GG-HH: Do we need the tlsv1.2 ?

PR #1303, opened 2026-06-18
feat(ai): send hook installation status

paulpetit-gg-ext:
## Context

We want to be able to show in the dashboard which agent has its hooks globally installed.

## What has been done

- reuse and complete the installation logic to be able to surface whether a hook configuration is already correct or should be installed/modified.
- send the result during an AI discovery

## Validation

```
ggshield ai discover
```

## PR check list

### DIFF
--- a/ggshield/verticals/ai/installation.py
+++ b/ggshield/verticals/ai/installation.py
@@ -5,20 +5,29 @@ import sys
 from copy import deepcopy
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Literal, Optional
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple
 
 import click
 
 from ggshield.core.dirs import get_user_home_dir
 from ggshield.core.errors import UnexpectedError
 
-from .agents import AGENTS
+from .agents import AGENTS, Agent
 
 
 @dataclass
 class InstallationStats:
-    added: int
-    already_present: int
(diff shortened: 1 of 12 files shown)


xblanchot-gg (approved):
LGTM thanks ! I think we just need to test the install command manually in many different situations

ggshield/verticals/ai/installation.py
```diff
@@ -182,3 +219,13 @@ def _fill_dict(
                 stats.added += 1
 
     return stats
+
+
+def are_hooks_installed_globally(agent_name: str) -> Tuple[bool, Optional[str]]:
+    """Whether the ggshield AI hooks are installed in this agent's global settings file."""
+    result = build_hook_config(agent_name, "global", force=False)
+    # We are conservative: returns False if any hook should change
```
amascia-gg: I think this comment is wrong?

> It calls build_hook_config(..., force=False) → _fill_dict(overwrite=False). With overwrite=False, an outdated ggshield command (e.g. the embedded absolute path changed after a pip→Homebrew move) is located, not counted as added, and reported as hooks_installed=True with the stale command. To actually be "conservative" (flag anything that would change), it'd need force=True. So the code reports "installed" for configs that install would in fact rewrite. Either fix the call or fix the comment.

paulpetit-gg-ext: indeed, comment removed.

PR #1308, opened 2026-06-22
feat(auth): add --only-scopes flag and auth.login_scopes config

benjaminrigaud-gg:
## What

Adds `--only-scopes` to `ggshield auth login` (requests an **exact** scope set, replacing the defaults) plus an `auth.login_scopes` config field to pin the base set in `.gitguardian.yaml`.

```bash
# self-hosted: avoid the never-grantable endpoints:send default
ggshield auth login --only-scopes "scan nhi:send-inventory"
```

## Why

Since v1.52.0, login requests `scan honeytokens:check endpoints:send` by default. On self-hosted, `endpoints:send` is never grantable (ClickHouse-backed; the backend raises `IS_ON_PREM`) and token creation is all-or-nothing, so login fails outright (END-581). The existing `--scopes` flag only **appends**, so it cannot drop a default scope.

## Precedence (base scope set; first match wins)

| Source (highest first) | Base scopes | `--scopes` extras applied? |
| -- | -- | -- |
| `--only-scopes` (CLI) | exactly those | ❌ ignored (stderr note) |
| `auth.login_scopes` (config) | the config list | ✅ appended |
| neither | `DEFAULT_SCOPES` | ✅ appended |

`--only-scopes` is rejected for `--method token`; an empty value errors. `_warn_missing_scopes` now compares granted scopes against the **effective requested set** instead of the hardcoded defaults.

## Config

```yaml
# .gitguardian.yaml
auth:
  login_scopes:
    - scan
    - nhi:send-inventory
```

## Testing

- 10 new unit tests (CLI replace/override/append, token-method rejection, empty-value error, false-warning suppression, config parse + base resolution).
- 272 passing across `tests/unit/cmd/auth/`, `verticals/auth/test_oauth.py`, `tests/unit/core/config/`.
- black / isort / flake8 / pyright clean on changed files.

Refs: END-583, END-581

### DIFF
--- a/changelog.d/20260622_091042_benjamin.rigaud_auth_login_only_scopes.md
+++ b/changelog.d/20260622_091042_benjamin.rigaud_auth_login_only_scopes.md
@@ -1,0 +1,3 @@
+### Added
+
+- `ggshield auth login` gains a `--only-scopes` option that requests an exact set of scopes, replacing the defaults (unlike `--scopes`, which appends). This unblocks login on self-hosted instances where a default scope such as `endpoints:send` is unavailable, e.g. `ggshield auth login --only-scopes "scan nhi:send-inventory"`. The base scope set can also be pinned in the configuration file via `auth.login_scopes`.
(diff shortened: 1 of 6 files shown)


benjaminrigaud-gg:
`uv run ggshield --instance=https://on-prem.preprod.gitguardian.tech auth login --only-scopes "scan nhi:send-inventory"`

image

Same with 
image

PR #1305, opened 2026-06-19
fix(auth): surface clear error on non-JSON token-claim response

benjaminrigaud-gg:
## Context

While testing the upcoming self-hosted release, `ggshield auth login` crashed with an opaque `AssertionError` and a raw `socketserver` traceback when the token-claim step received a non-JSON HTTP error response (the server returned HTTP 405).

Root cause: `_claim_token` raises `UnexpectedError` for a non-JSON response, but the callback handler `do_GET` only caught `OAuthError`. The `UnexpectedError` escaped the request-handler thread (swallowed by `socketserver`), so `_request_error_message` was never set, `_wait_for_callback` returned as if login had succeeded, and `_print_login_success` then tripped on `assert self.instance_config.account is not None`.

Refs: END-579

## What has been done

- `do_GET` now catches `UnexpectedError` in addition to `OAuthError` and routes it through `_handle_error`, so the error is recorded and the main thread raises it cleanly instead of mistaking the callback for a success.
- `_print_login_success` raises a clean `UnexpectedError` instead of a bare `assert` when no account was saved (defense-in-depth; also survives `python -O`).
- Added regression tests that exercise the real `RequestHandler.do_GET`. The existing web-flow harness bypasses it (its `FakeOAuthClient` calls `process_callback` directly in the main thread), which is why the existing garbage-response test passed while production crashed.

## Validation

- `ggshield auth login` against a server whose `/v1/oauth/token` returns a non-JSON error (e.g. an nginx 405) now prints a clear `Error: Server response is not JSON (HTTP code: 405).` and exits, instead of an `AssertionError` + traceback.
- `make unittest` (full suite, serial): all pass. `black` / `isort` / `flake8` / `pyright` clean.

## Out of scope / follow-up

This makes the failure legible but does not fix the underlying 405. For that instance the cause is that `dashboard_to_api_url` classifies hosts under `*.gitguardian.tech` as SaaS and drops the `/exposed` API prefix — tracked as a separate follow-up under END-579.

## PR check list

- [x] As much as possible, the changes include tests (unit and/or functional)
- [x] If the changes affect the end user (new feature, behavior change, bug fix) then the PR has a changelog entry (see doc/dev/getting-started.md). If the changes do not affect the end user, then the `skip-changelog` label has been added to the PR.

PR #1312, opened 2026-06-22
fix(core): surface clear errors on non-JSON GitGuardian API responses

benjaminrigaud-gg:
## Summary

`ggshield api-status` (and other commands that read token scopes) crashed with a raw `requests.exceptions.JSONDecodeError` traceback when the GitGuardian API returned a **2xx with a non-JSON body** — e.g. the dashboard SPA's HTML, served when the instance URL is wrong. The same unguarded `response.json()` pattern existed in the `auth login` token-claim / token-validate paths.

Root cause is in pygitguardian (`GGClient.api_tokens()` calls `resp.json()` unconditionally on `resp.ok`), a pinned dependency — so this PR hardens the **ggshield** side. The complementary root-cause fix is in GitGuardian/py-gitguardian#184.

### Changes

**`api_tokens()` callers**
- New `safe_api_tokens()` (`core/client.py`): converts a non-JSON `JSONDecodeError` into a clean `UnexpectedError` naming the instance URL (+ a body snippet); passes `APITokensResponse` / `Detail` through unchanged.
- Routed through it: `api-status` (`cmd/status.py`), `check_client_api_key` (`core/client.py`), and the post-login `_warn_missing_scopes` (`cmd/auth/login.py`).

**Raw `response.json()` in the auth-login flow**
- New `safe_response_json()` (shares the `_non_json_error()` message builder with `safe_api_tokens`).
- Routed through it: `token_login()` (`auth login --method token`) and `OAuthClient._validate_access_token()`.

**`_warn_missing_scopes()` hardening**
- Swallows any `RequestException` — a post-login network blip must not crash an already-successful login.
- Skips the warning on an error `Detail` instead of reporting every default scope as "missing".

**Browser / localhost OAuth callback handler**
- `RequestHandler.do_GET()` gains a catch-all: any exception escaping the callback handler thread (not just `UnexpectedError`) is routed through `_handle_error()` instead of being swallowed by `socketserver` and misread as a login success.

## Test plan
- [x] `safe_api_tokens` / `safe_response_json` unit tests: non-JSON → clean error; valid response / error `Detail` pass through
- [x] `api-status`: non-JSON 2xx → exit 128 with an "instance URL" message, no raw traceback
- [x] `check_client_api_key`: non-JSON → clean `UnexpectedError`
- [x] `auth login --method token`: non-JSON 2xx → clean error, no traceback
- [x] `_warn_missing_scopes`: swallows non-JSON / `ConnectionError`; silent on an error `Detail`
- [x] `RequestHandler.do_GET`: an arbitrary callback exception is recorded (no swallowed false-success)
- [x] Existing scopes-omitted-on-error test still passes (normal error `Detail` tolerated)
- [x] Full unit suite: **2171 passed**; `black` / `isort` / `flake8` / `pyright` clean

### DIFF
--- a/ggshield/cmd/auth/login.py
+++ b/ggshield/cmd/auth/login.py
@@ -8,7 +8,11 @@ from pygitguardian.models import APITokensResponse
 
 from ggshield.cmd.utils.common_options import add_common_options
 from ggshield.cmd.utils.context_obj import ContextObj
-from ggshield.core.client import create_client, create_client_from_config
+from ggshield.core.client import (
+    create_client,
+    create_client_from_config,
+    safe_api_tokens,
+)
 from ggshield.core.config import Config
 from ggshield.core.constants import DEFAULT_INSTANCE_URL
 from ggshield.core.errors import UnexpectedError
@@ -17,7 +21,12 @@ from ggshield.verticals.auth import DEFAULT_SCOPES, OAuthClient
 
 
 def _warn_missing_scopes(client: GGClient) -> None:
-    token_info = client.api_tokens()
+    try:
+        token_info = safe_api_tokens(client)
+    except UnexpectedError:
+        # Best-effort warning: don't fail a successful login if the token-info
+        # fetch hits a non-JSON response.
+        return
     granted = (
         token_info.scopes
         if isinstance(token_info, APITokensResponse) and token_info.scopes
--- a/ggshield/cmd/auth/login.py
+++ b/ggshield/cmd/auth/login.py
@@ -12,6 +12,7 @@ from ggshield.core.client import (
     create_client,
     create_client_from_config,
     safe_api_tokens,
+    safe_response_json,
 )
 from ggshield.core.config import Config
 from ggshield.core.constants import DEFAULT_INSTANCE_URL
@@ -23,15 +24,15 @@ from ggshield.verticals.auth import DEFAULT_SCOPES, OAuthClient
 def _warn_missing_scopes(client: GGClient) -> None:
     try:
         token_info = safe_api_tokens(client)
-    except UnexpectedError:
-        # Best-effort warning: don't fail a successful login if the token-info
-        # fetch hits a non-JSON response.
+    except (UnexpectedError, requests.exceptions.RequestException):
+        # Best-effort warning: never fail an already-successful login if the
+        # token-info fetch errors out (non-JSON body, network blip, timeout...).
         return
-    granted = (
-        token_info.scopes
-        if isinstance(token_info, APITokensResponse) and token_info.scopes
-        else []
-    )
+    if not isinstance(token_info, APITokensResponse):
+        # An error Detail (e.g. 401/500) tells us nothing about scopes; don't
+        # warn that scopes are "missing" when we never managed to read them.
+        return
+    granted = token_info.scopes or []
     missing = [s for s in DEFAULT_SCOPES if s not in granted]
     if missing:
         click.echo(
@@ -253,7 +254,7 @@ def token_login(config: Config, instance: Optional[str]) -> None:
     if not response.ok:
         raise UnexpectedError("Authentication failed with token.")
 
-    api_token_data = response.json()
+    api_token_data = safe_response_json(response)
     scopes = api_token_data["scope"]
     if "scan" not in scopes:
         raise UnexpectedError("This token does not have the scan scope.")
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -1,7 +1,7 @@
 import logging
 import os
 from enum import Enum
-from typing import Optional
+from typing import Optional, Union
 
 import requests
 import urllib3
@@ -167,6 +167,28 @@ def create_session(
     return session
 
 
+def safe_api_tokens(client: GGClient) -> Union[APITokensResponse, Detail]:
+    """Call GGClient.api_tokens() but turn a non-JSON 2xx body into a clean error
+    instead of letting a raw JSONDecodeError escape.
+
+    pygitguardian's api_tokens() calls resp.json() unconditionally when resp.ok;
+    a wrong instance URL can answer a 2xx with the dashboard SPA's HTML, which
+    otherwise crashed callers with an opaque traceback. Mirrors the non-JSON
+    hardening already applied to the auth-login token-claim path.
+    """
+    try:
+        return client.api_tokens()
+    except requests.exceptions.JSONDecodeError as exc:
+        snippet = " ".join((getattr(exc, "doc", "") or "").split())[:100]
+        message = (
+            f"Unexpected non-JSON response from {client.base_uri}. "
+            "Check your instance URL settings."
+        )
+        if snippet:
+            message += f"\nResponse body: {snippet}"
+        raise UnexpectedError(message) from exc
+
+
 def check_client_api_key(client: GGClient, required_scopes: set[TokenScope]) -> None:
     """
     Raises APIKeyCheckError if the API key configured for the client is not usable
@@ -231,7 +253,7 @@ def check_client_api_key(client: GGClient, required_scopes: set[TokenScope]) -> 
     # Check token scopes if required_scopes is not empty
     if required_scopes:
         try:
-            response = client.api_tokens()
+            response = safe_api_tokens(client)
         except requests.exceptions.ConnectionError as e:
             raise ServiceUnavailableError(
                 message="Failed to connect to GitGuardian server. Check your"
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -1,7 +1,7 @@
 import logging
 import os
 from enum import Enum
-from typing import Optional, Union
+from typing import Any, Optional, Union
 
 import requests
 import urllib3
@@ -167,26 +167,41 @@ def create_session(
     return session
 
 
+def _non_json_error(
+    url: Optional[str], exc: requests.exceptions.JSONDecodeError
+) -> UnexpectedError:
+    """Build a clear error for a non-JSON body, with a snippet of what came back."""
+    snippet = " ".join((getattr(exc, "doc", "") or "").split())[:100]
+    message = (
+        f"Unexpected non-JSON response from {url}. Check your instance URL settings."
+    )
+    if snippet:
+        message += f"\nResponse body: {snippet}"
+    return UnexpectedError(message)
+
+
 def safe_api_tokens(client: GGClient) -> Union[APITokensResponse, Detail]:
     """Call GGClient.api_tokens() but turn a non-JSON 2xx body into a clean error
     instead of letting a raw JSONDecodeError escape.
 
     pygitguardian's api_tokens() calls resp.json() unconditionally when resp.ok;
     a wrong instance URL can answer a 2xx with the dashboard SPA's HTML, which
-    otherwise crashed callers with an opaque traceback. Mirrors the non-JSON
-    hardening already applied to the auth-login token-claim path.
+    otherwise crashed callers with an opaque traceback.
     """
     try:
         return client.api_tokens()
     except requests.exceptions.JSONDecodeError as exc:
-        snippet = " ".join((getattr(exc, "doc", "") or "").split())[:100]
-        message = (
-            f"Unexpected non-JSON response from {client.base_uri}. "
-            "Check your instance URL settings."
-        )
-        if snippet:
-            message += f"\nResponse body: {snippet}"
-        raise UnexpectedError(message) from exc
+        raise _non_json_error(client.base_uri, exc) from exc
+
+
+def safe_response_json(response: requests.Response) -> Any:
+    """Parse a JSON response body, raising a clean UnexpectedError that names the
+    instance URL instead of letting a raw JSONDecodeError escape when the body is
+    not JSON (e.g. the dashboard SPA's HTML served on a wrong instance URL)."""
+    try:
+        return response.json()
+    except requests.exceptions.JSONDecodeError as exc:
+        raise _non_json_error(response.url, exc) from exc
 
 
 def check_client_api_key(client: GGClient, required_scopes: set[TokenScope]) -> None:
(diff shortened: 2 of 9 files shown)


ggshield/cmd/auth/login.py
```diff
@@ -17,7 +21,12 @@
 
 
 def _warn_missing_scopes(client: GGClient) -> None:
-    token_info = client.api_tokens()
+    try:
+        token_info = safe_api_tokens(client)
+    except UnexpectedError:
```
clement-tourriere: **[minor] "Best-effort" vs. the catch scope.** The comment says *"don't fail a successful login"*, but only `UnexpectedError` is caught. `safe_api_tokens()` still lets `requests.exceptions.ConnectionError` (and other `RequestException`s / timeouts) propagate, so a transient network blip here — *after* the token is already saved and "Authentication was successful" has been printed — would still crash the command with a traceback. Pre-existing, but since you're explicitly making this best-effort now, consider widening to `except (UnexpectedError, requests.exceptions.RequestException):`.

ggshield/cmd/auth/login.py
```diff
@@ -17,7 +21,12 @@
...
+        # Best-effort warning: don't fail a successful login if the token-info
+        # fetch hits a non-JSON response.
+        return
     granted = (
         token_info.scopes
         if isinstance(token_info, APITokensResponse) and token_info.scopes
```
clement-tourriere: **[minor, pre-existing] Misleading warning on an error `Detail`.** If `safe_api_tokens()` returns a `Detail` (e.g. a 401/500 from `/v1/api_tokens/self`), it isn't an `APITokensResponse`, so `granted` falls through to `[]` and the block below warns that *every* default scope is missing — misleading, since the real problem is an API error, not missing scopes. Consider early-returning unless `isinstance(token_info, APITokensResponse)`, and adding a regression test for the `Detail` path (the new tests only cover the `JSONDecodeError` path). Independently flagged by Codex too.

ggshield/core/client.py
```diff
@@ -167,6 +167,28 @@ def create_session(
...
+    instead of letting a raw JSONDecodeError escape.
+
+    pygitguardian's api_tokens() calls resp.json() unconditionally when resp.ok;
+    a wrong instance URL can answer a 2xx with the dashboard SPA's HTML, which
+    otherwise crashed callers with an opaque traceback. Mirrors the non-JSON
+    hardening already applied to the auth-login token-claim path.
+    """
+    try:
+        return client.api_tokens()
+    except requests.exceptions.JSONDecodeError as exc:
```
clement-tourriere, line 193: **[nit] Exception class is narrower than the "mirror" it cites.** The docstring says this mirrors the `oauth.py` hardening, but that path catches `json.decoder.JSONDecodeError` (`oauth.py:337`) whereas this catches `requests.exceptions.JSONDecodeError`. It works today because the latter subclasses the former — just flagging the inconsistency; either is fine functionally.

Separately: nice that this message is actually *better* than `oauth.py`'s (it names the instance URL and includes a body snippet). Worth back-porting that wording to `oauth.py` so the two non-JSON paths read consistently.

clement-tourriere (commented):
## Review

Solid, well-scoped fix. `safe_api_tokens()` centralizes the `JSONDecodeError` → `UnexpectedError` conversion and all three `api_tokens()` call sites (`status.py`, `check_client_api_key`, `_warn_missing_scopes`) are routed through it — no site left behind.

**Verified locally on the PR branch:**
- ✅ `tests/unit/core/test_client.py`, `tests/unit/cmd/test_status.py`, `tests/unit/cmd/auth/test_login.py` → **189 passed**
- ✅ `black` / `isort` / `flake8` clean; `pyright` clean (with the venv interpreter)
- ✅ Confirmed `requests.exceptions.JSONDecodeError` subclasses stdlib `json.JSONDecodeError`, and the tests' 3-arg constructor sets `.doc` as the snippet logic expects.

---

### ⚠️ One substantive gap (not inline — it's outside the diff)

`token_login()` at `ggshield/cmd/auth/login.py:256` still does `api_token_data = response.json()` **without** non-JSON hardening:

```python
response = client.get(endpoint="token")   # GGClient.get() -> raw requests.Response
...
if not response.ok:
    raise UnexpectedError("Authentication failed with token.")
api_token_data = response.json()           # <-- raw JSONDecodeError on a 2xx non-JSON body
```

This is the **exact** wrong-instance-URL → SPA-HTML scenario the PR targets, in the **same file**, and it crashes *before* `_warn_missing_scopes()` is ever reached — reproducible via `ggshield auth login --method token` against a misconfigured instance URL. The PR description says it "mirrors the non-JSON hardening of the auth-login token-claim path," but only the *OAuth* token-claim path is hardened (`oauth.py:337`); this *manual-token* claim path is not.

Recommend hardening line 256 here too (a shared `safe_json(response)`-style helper would cover both this and the `oauth.py` path), **or** explicitly calling out in the PR that it's deferred to the follow-up py-gitguardian fix. Not strictly a regression, but it leaves the stated "no more raw `JSONDecodeError` on a wrong instance URL" promise half-kept.

The inline comments below are minor / nit.

<sub>Reviewed with assistance from Codex (GPT‑5).</sub>

ggshield/cmd/auth/login.py
benjaminrigaud-gg: Widened to `except (UnexpectedError, requests.exceptions.RequestException)` in 3cc30177 — a post-login `ConnectionError`/timeout (after the token is already saved and "Authentication was successful" printed) no longer crashes the command. Added `test_warn_missing_scopes_swallows_connection_error`.

ggshield/cmd/auth/login.py
benjaminrigaud-gg: Fixed in 3cc30177: `_warn_missing_scopes()` now early-returns unless `token_info` is an `APITokensResponse`, so an error `Detail` (401/500) no longer makes the block report every default scope as "missing". Added `test_warn_missing_scopes_silent_on_error_detail` covering the `Detail` path.

ggshield/core/client.py
benjaminrigaud-gg: Dropped the inaccurate "mirrors the auth-login token-claim path" line from the docstring. For the message back-port: extracted a shared `safe_response_json(response)` (same `_non_json_error()` builder as `safe_api_tokens`) and routed the raw `response.json()` sites through it, so they now emit the instance-URL + body-snippet message consistently. Applied to `token_login()` and the previously-unguarded `oauth.py:366` (`_validate_access_token`).

Left `oauth.py:336` as-is on purpose: it parses `dct` *before* the `resp.ok` check to pull `detail` out of error bodies, and it's already covered by the existing non-JSON test — didn't want to change that tested behavior just for message wording.

benjaminrigaud-gg:
Addressed the review in 3cc30177 (kept as a separate commit so the delta is reviewable).

**The token_login gap** — hardened rather than deferred. `token_login()` (`auth login --method token`) now parses via the new `safe_response_json()`, so a wrong instance URL serving the SPA's HTML on a `2xx` produces the clean instance-URL error instead of a raw `JSONDecodeError`. While there, I caught the same unguarded pattern in `OAuthClient._validate_access_token()` (`oauth.py:366`). New test: `TestAuthLoginToken::test_auth_login_token_non_json_response`.

**Inline nits** — all three addressed (replies on each thread): widened the `_warn_missing_scopes` catch to `RequestException`, skip the warning on an error `Detail`, and consolidated the non-JSON message into one shared helper.

Verification on the branch: full unit suite **2157 passed**; `black` / `isort` / `flake8` / `pyright` clean.

A complementary root-cause fix (so the client returns a `Detail` instead of raising at all) is up at GitGuardian/py-gitguardian#184.

ggshield/verticals/auth/oauth.py
```diff
@@ -363,7 +364,7 @@ def _validate_access_token(self) -> Dict[str, Any]:
         ).get(endpoint="token")
         if not response.ok:
             raise OAuthError("The created token is invalid.")
-        return response.json()
+        return safe_response_json(response)
```
clement-tourriere, line 367: **[medium, pre-existing] This clean error doesn't reach the user in the default browser flow.**

In the localhost/browser login, `_validate_access_token()` runs inside `process_callback()` → `RequestHandler.do_GET()`, and `do_GET` (oauth.py:537) only catches `OAuthError`:

```python
try:
    self.oauth_client.process_callback(callback_url)
except OAuthError as error:
    self._handle_error(error.message)
else:
    self._redirect(...authenticated...)
```

The `UnexpectedError` raised by `safe_response_json()` here is **not** an `OAuthError`, so it escapes `do_GET`. `HTTPServer` then dumps a raw traceback to stderr and swallows it; `_request_error_message` is never set; and since `do_GET` already set `_request_finished = True` (line 528) *before* the call, `_wait_for_callback()` exits its loop, sees no error message, and returns as **success** — so `oauth_process` goes on to `_print_login_success()` even though `_save_token()` was never reached. I reproduced this against a real `HTTPServer.handle_request()`.

Not a regression (the old `response.json()` raised a raw `JSONDecodeError` that escaped `do_GET` the same way, and `_claim_token` at :339/:350 already raises `UnexpectedError` inside the callback), but the PR's changelog now claims `ggshield auth login` reports a clear error on non-JSON bodies — true for `oob`/`token`, not for the default browser method.

**Suggested fix:** in `do_GET`, also catch `UnexpectedError` (or a bare `except Exception`) and route it through `self._handle_error(str(error))`, and add a `RequestHandler`-level test for a non-JSON callback. That also covers the pre-existing `_claim_token` case. Alternatively, scope the changelog to exclude the browser callback path and defer it.

clement-tourriere (commented):
## Review — round 2

Thanks for the thorough follow-up in `3cc30177` — it resolves all three round-1 points cleanly: shared `_non_json_error()` + `safe_response_json()` helper, both raw `response.json()` sites (`token_login` and `_validate_access_token`) routed through it, `_warn_missing_scopes` now swallows any `RequestException` and skips the warning on an error `Detail`, plus 3 new regression tests.

**Re-verified locally on `3cc30177`:** 192 tests pass across the affected files; `black` / `isort` / `flake8` clean; `pyright` clean (venv interpreter).

I ran a second Codex pass focused on the new code. It surfaced **one Medium finding** worth your attention (details inline on `oauth.py`). Short version:

### ⚠️ Medium — the default (browser) `auth login` flow doesn't actually surface this clean error

`_validate_access_token()` now raises `UnexpectedError` on a non-JSON body, but in the **localhost/browser** flow it runs inside `RequestHandler.do_GET()` (`oauth.py:537`), which only catches `OAuthError`. So the `UnexpectedError` escapes the handler → Python's `HTTPServer` dumps a **raw traceback to stderr** and swallows it → `_request_error_message` is never set → `_wait_for_callback()` sees `_request_finished=True` with no error and returns as **success** → ggshield prints the login-success message **without having saved a token**.

I confirmed this empirically (driving a real `HTTPServer.handle_request()`): the exception is swallowed by `socketserver`, `_request_error_message` stays `None`, and control returns as if successful.

**Important caveats (so this is judged fairly):**
- This is **pre-existing**, *not* a regression from this PR — pre-`3cc30177`, `_validate_access_token` did `return response.json()`, whose raw `JSONDecodeError` is *also* not an `OAuthError` and escaped `do_GET` identically. `_claim_token()` (`oauth.py:339`/`:350`) already raises `UnexpectedError` inside `process_callback` too, so in the realistic wrong-instance-URL case the claim step trips this boundary first.
- The reason I'm still raising it: this PR's changelog now advertises that **`ggshield auth login`** reports a clear error on non-JSON bodies. That holds for `--method oob` and `--method token`, but **not** for the default browser method — there you still get a traceback + a misleading "success". So the fix is one boundary short of its stated scope.

**Options:** either fix the boundary (catch `UnexpectedError` — or all exceptions — in `do_GET` and route through `_handle_error(str(error))`, with a `RequestHandler`-level test for a non-JSON callback), or narrow the changelog wording to make clear the browser-callback path isn't covered here and is deferred.

Everything else looks good — this is the only item above nit level. Not flagging it as a merge blocker given it's pre-existing.

<sub>Reviewed with assistance from Codex (GPT‑5); the control-flow claim was independently reproduced.</sub>

ggshield/verticals/auth/oauth.py
benjaminrigaud-gg: Root cause was a **stale branch**, not a missing fix. This branch forked at `2c09dff8` — 8 commits before END-579 (#1305) merged — so on `3cc30177` (what you reproduced against) `do_GET` still only caught `OAuthError`. END-579 already landed the `except (OAuthError, UnexpectedError)` catch *and* the `_print_login_success` no-account guard on `main`.

Rebased onto current `main` (now `ac5ce83f`), so the `UnexpectedError` raised by `safe_response_json()` in `_validate_access_token()` is caught by `do_GET` → `_handle_error()` → surfaced cleanly. No traceback, no false success.

I also closed the **general** class (your bare-`except` option), since the narrow catch only covers the `UnexpectedError` subset: added a catch-all `except Exception: self._handle_error(str(error))` after the specific clause, so any callback exception that's neither `OAuthError` nor `UnexpectedError` (a stray `ConnectionError`, a `KeyError` on a malformed-but-JSON body) is recorded instead of swallowed into a misleading success. New RequestHandler-level test `test_request_handler_records_arbitrary_exception` covers it; END-579's `test_request_handler_records_unexpected_error` already covers the `UnexpectedError` path.

Full unit suite: **2171 passed**; `black` / `isort` / `flake8` / `pyright` clean.

benjaminrigaud-gg:
Pushed `ac5ce83f` (force-push — **rebased onto current `main`** + one new commit).

The round-2 Medium was a stale-branch artifact: END-579's handler-boundary fix (`do_GET` catching `UnexpectedError` + the `_print_login_success` no-account guard) is already on `main`, and this branch forked just before #1305 merged. Rebasing brings it in, so the browser flow now surfaces the non-JSON error cleanly.

Went one step further: a catch-all in `do_GET` closes the swallow→false-success class for *any* escaping exception, not just non-JSON. Details + the new RequestHandler test in the thread above.

clement-tourriere (approved):

PR #1322, opened 2026-06-23
fix(ai): use the documented catch-all matcher for AI hooks

amascia-gg:
## Context

A security engineer deploying AI hooks via MDM noted that ggshield writes `"matcher": ".*"` into the Claude hook config — non-standard and undocumented ([Slack thread](https://gitguardian.slack.com/archives/C0BAQGFKH4Y/p1781181048632349)). Per the hook spec, the match-all idiom is `"*"` (or omitted); `".*"` only works as an incidental catch-all regex.

This matcher lives in the **shared** `Agent.settings_template`, so it affects **Claude Code, Codex, Copilot and VS Code** (Copilot via VS Code). **Cursor is unaffected** — it overrides the template and has no `matcher`.

## What this PR does

- Switch the shared template's matcher from `".*"` to `"*"`.
- Make `settings_locate` reuse an existing ggshield hook block even when its matcher differs from the template (e.g. a legacy `".*"`), so re-running `install` / `machine setup` over an old config **migrates it in place** instead of appending a duplicate hook block.

## Validation

- New unit tests: template uses `*`; a legacy `.*` block is reused by `settings_locate`; re-installing over a legacy `.*` config does not duplicate the hook.
- 340 AI-vertical tests pass; black / isort / flake8 / pyright clean.

## Note for reviewers

`"*"` is the documented match-all for Claude Code. Codex / Copilot / VS Code share the same template (and hook schema), so it applies to them too — please confirm against their hook specs if you know of any divergence.

## PR check list

### DIFF
--- a/changelog.d/20260623_150000_achille.mascia_ai_hook_matcher.md
+++ b/changelog.d/20260623_150000_achille.mascia_ai_hook_matcher.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- The AI hook configuration written for Claude Code, Codex, Copilot and VS Code now uses the documented catch-all matcher `"*"` instead of `".*"`. Re-running the install over an existing `".*"` configuration migrates it in place instead of adding a duplicate hook entry.
--- a/ggshield/verticals/ai/models.py
+++ b/ggshield/verticals/ai/models.py
@@ -166,7 +166,7 @@ class Agent(ABC):
             "hooks": {
                 "PreToolUse": [
                     {
-                        "matcher": ".*",
+                        "matcher": "*",
                         "hooks": [
                             {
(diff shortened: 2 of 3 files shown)


paulpetit-gg-ext (commented):
The PR looks fine, and I tested with a few agents that the new version of the hooks' file still works.

However, I would argue that the initial reason this PR was created is unfounded: `.*` *is* documented and officially supported in the doc, as a valid regular expression (because it contains a dot). See https://code.claude.com/docs/en/hooks#matcher-patterns

Thus, there is no reason to think it would break at some point, and as people say, _if it ain't broke, don't fix it_.

So merge it or not :), I don't think it makes a difference.

PR #1323, opened 2026-06-23
test: deflake the unit suite under pytest -n auto

benjaminrigaud-gg:
## What

Fixes two independent causes of unit-suite flakiness under `pytest -n auto` (each reproduced deterministically). These were the intermittent "Run unit tests" failures on `main`.

### 1. Global `logging.disable()` leaking between tests
`__main__.main()` calls `disable_logs()` (a process-wide `logging.disable()`) on every run, including when unit tests invoke it via `main(args=...)`. The per-test UI reset (`reset()`, run by an autouse fixture) removed the log handler but never reverted the disable — so logging stayed globally disabled for later tests in the same xdist worker, and `caplog`-based tests captured nothing and failed with `IndexError: list index out of range` (e.g. `test_load_last_check_time_swallows_errors`, the `test_id_cache` cannot-be-read/written tests).

→ Re-enable logging in `_reset_log_handler()` so every test starts with logging on.

### 2. Shared `.cache_ggshield` corrupted by concurrent workers
`Cache()` writes `CACHE_PATH` (`"./.cache_ggshield"`, relative to the CWD) non-atomically with no per-test isolation. Under `-n auto` all workers share the repo-root CWD, so they concurrently read/write the same file → interleaved/corrupted JSON → `UnexpectedError` in ~30 tests (`test_secret_scanner`, `test_cmd_ai`, `test_install`). The corrupt file is even written into the working tree.

→ Autouse fixture redirecting `ggshield.core.cache.CACHE_PATH` into each test's `tmp_path`; the `cache` fixture becomes function-scoped so it is built after that redirect. `test_cache.py` opts out (it exercises the real relative path via `isolated_fs`/`cd`).

## Verification
- Each cause reproduced deterministically, then confirmed fixed.
- Full unit suite under `pytest -n auto` run repeatedly (11×+ across the change): **2178 passed, 0 failed, 0 errors** every run (previously ~10–40% of runs failed).
- No `.cache_ggshield` left in the repo root; lint clean.

Test-only change; no production behaviour affected (`reset()` is unit-test-only).

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -723,8 +723,11 @@ def no_api_key(monkeypatch):
     monkeypatch.delenv("GITGUARDIAN_API_KEY", raising=False)
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture()
 def cache() -> Cache:
+    # Function-scoped so it is built after _isolate_cache_path has redirected
+    # CACHE_PATH into the test's tmp_path (a session cache would write the
+    # shared .cache_ggshield into the CWD).
     c = Cache()
     c.purge()
     return c
@@ -824,6 +827,18 @@ def clear_cache():
     _get_git_path.cache_clear()
     _git_rev_parse_absolute.cache_clear()
     read_git_file.cache_clear()
+
+
+@pytest.fixture(autouse=True)
+def _isolate_cache_path(monkeypatch, tmp_path):
+    """Point the secrets cache into the test's tmp_path.
+
+    Without this, ``Cache()`` writes ``.cache_ggshield`` relative to the CWD,
+    which all xdist workers share, leading to interleaved/corrupted JSON.
+    """
+    monkeypatch.setattr(
+        "ggshield.core.cache.CACHE_PATH", str(tmp_path / ".cache_ggshield")
+    )
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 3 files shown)


tests/unit/conftest.py
```diff
@@ -826,6 +829,18 @@ def clear_cache():
     read_git_file.cache_clear()
 
 
+@pytest.fixture(autouse=True)
```
clement-tourriere, line 832: Don't we want instead of this `autouse` fixture to handle xdist different cache directories inside the cache() fixture directly using https://pytest-xdist.readthedocs.io/en/stable/how-to.html#identifying-the-worker-process-during-a-test ?

benjaminrigaud-gg: Having a look :zoi:

benjaminrigaud-gg:
```
  The blocker: most failing tests don't go through the cache() fixture

  - test_secret_scanner.py: 5 tests use the cache fixture, but 6 construct cache=Cache() directly.
  - test_cmd_ai.py and test_install.py: zero cache-fixture references. They're CLI tests — Cache() is created inside production code at
  ggshield/__main__.py:137 (ctx_obj.cache = Cache()).

  So handling xdist cache dirs inside the cache() fixture would only cover those 5 tests and miss the bulk of the ~30 failures (the
  direct-Cache() ones and every CLI test). A fixture-scoped solution fundamentally can't reach a Cache() built inside main(). The only lever
  that covers all of them is patching the global CACHE_PATH that Cache reads — which is what the autouse fixture does.
```

clement-tourriere (approved):

PR #1324, opened 2026-06-23
fix(scripts): give shell-specific PATH guidance after install

clement-tourriere:
## Context

Fixes [END-598](https://linear.app/gitguardian/issue/END-598/ggshield-not-in-path-after-install).

The `install.sh` script symlinks `ggshield` into `~/.local/bin`, which is **not on `PATH` everywhere**, so users sometimes can't run `ggshield` right after installing:

- **macOS** never adds `~/.local/bin` to `PATH` by default.
- **Debian/Ubuntu** add it from `~/.profile` *only if the directory already exists at login* — a fresh install creates it afterwards, so it isn't picked up until the shell is restarted (the most common cause of the report).

I checked whether we could just install into a directory that is *always* on `PATH`: there is **no sudo-less directory guaranteed on `PATH` across every Linux distro and macOS** (`/usr/local/bin` needs `sudo`, and on Apple Silicon Homebrew uses `/opt/homebrew/bin`, itself not on the stock macOS `PATH`). So we keep `~/.local/bin` (the no-sudo XDG/systemd convention) and make the guidance actionable instead — the approach other CLIs take. Windows (`install.ps1`) already edits the user `PATH` automatically and is unchanged.

## What has been done

- When `BIN_DIR` is not on `PATH`, the installer now prints **shell-specific** instructions at the **end** of the run (instead of a generic note buried before the auth/plugin output), detecting the login shell via `$SHELL`:
  - **zsh** → `~/.zshrc`
  - **bash** → `~/.bashrc` on Linux; on macOS the first existing of `~/.bash_profile` / `~/.bash_login` / `~/.profile` (avoids creating `~/.bash_profile` and silently shadowing an active `~/.profile`)
  - **fish** → `fish_add_path` (persistent, no restart needed)
  - **other / POSIX sh** → `~/.profile`, using `.` rather than the non-POSIX `source`
  - …each with a reminder to **restart the terminal** (plus a `source` one-liner to apply it immediately).
- Emitted commands quote the rc path so paths containing spaces copy-paste correctly.
- Added an early guard that fails with a clear message if `HOME` is unset (every install path derives from it).
- Documented the `PATH` behaviour in `scripts/install/README.md`.

The logic is flag-based (`PATH_NEEDS_SETUP`): set in `install_tarball`, surfaced at the end of `post_install`, listed before the auth/plugin hints (which assume `ggshield` is already callable), and folded into the "ggshield is ready." gate.

## Validation

- `bash -n scripts/install/install.sh` passes.
- Exercised the hint function across `SHELL` = zsh/bash/fish/dash and `OS` = linux/darwin, including the macOS rc-file selection (only `.profile` present → `.profile`; `.bash_profile` present → `.bash_profile`; none present → `.bash_profile`) and a path with spaces (quotes correctly).
- Verified the `HOME` guard: unset/empty `HOME` → exit 1 with a clear error; normal `--help` → exit 0.
- `pre-commit run --files …` passes for all changed files (prettier, trailing-whitespace, end-of-file-fixer).

To try it manually:
```sh
PATH="/usr/bin:/bin" GGSHIELD_BIN_DIR="$HOME/.local/bin" bash scripts/install/install.sh --install-only
# end of output shows the shell-specific "add to PATH, then restart" hint
```

## PR check list

- [x] As much as possible, the changes include tests (unit and/or functional) — n/a: pure shell installer change; validated manually as above (no shell test harness exists for these scripts).
- [x] If the changes affect the end user (new feature, behavior change, bug fix) then the PR has a changelog entry

### DIFF
--- a/scripts/install/install.sh
+++ b/scripts/install/install.sh
@@ -332,6 +332,9 @@ emit_plugin_hint() {
 # Uses $SHELL (the login shell) rather than $0: this script always runs under
 # bash (curl | bash), which says nothing about the shell the user reopens.
 emit_path_hint() {
+    # The headline goes through warn() (bold yellow, stderr) so it stands out;
+    # as an indented "# ..." comment it blended into the command block below and
+    # users missed it. The commands stay on stdout, unprefixed, to copy-paste cleanly.
     # reload defaults to `source` (zsh/bash); `.` is the POSIX form for the
     # generic-sh fallback. Paths are double-quoted in the emitted commands so a
     # BIN_DIR/rc with spaces still copy-pastes correctly.
@@ -339,7 +342,7 @@ emit_path_hint() {
     case "$(basename "${SHELL:-sh}")" in
     fish)
         # fish persists PATH via universal variables — no file edit, no restart.
-        printf '    # %s is not on your PATH; add it permanently with:\n' "$BIN_DIR"
+        warn "$BIN_DIR is not on your PATH. Add it permanently with:"
         printf '    fish_add_path -- "%s"\n' "$BIN_DIR"
         return 0
         ;;
@@ -363,7 +366,7 @@ emit_path_hint() {
         reload="."
         ;;
     esac
-    printf '    # %s is not on your PATH; add it, then restart your terminal:\n' "$BIN_DIR"
+    warn "$BIN_DIR is not on your PATH. Add it, then restart your terminal:"
     printf '    echo '\''export PATH="%s:$PATH"'\'' >> "%s"\n' "$BIN_DIR" "$rc"
     printf '    # (or apply it to the current shell now: %s "%s")\n' "$reload" "$rc"
     return 0
--- a/scripts/install/install.sh
+++ b/scripts/install/install.sh
@@ -13,6 +13,11 @@
 
 set -euo pipefail
 
+# Every install path (and every PATH hint) is derived from $HOME; fail early
+# with a clear message rather than building paths like /.local/bin under set -u.
+# die() is not defined yet, so emit raw.
+[ -n "${HOME:-}" ] || { printf '\033[1;31merror:\033[0m HOME is not set\n' >&2; exit 1; }
+
 GITHUB_REPO="GitGuardian/ggshield"
 DEFAULT_INSTANCE="https://dashboard.gitguardian.com"
 EU_INSTANCE="https://dashboard.eu1.gitguardian.com"
@@ -25,6 +30,9 @@ STATE_FILE="$STATE_DIR/state"
 
 ASSUME_YES=0
 INSTALL_ONLY=0
+# Set when BIN_DIR is not on PATH, so the final summary can tell the user how to
+# expose it (see emit_path_hint).
+PATH_NEEDS_SETUP=0
 PLUGINS=()
 # honor GITGUARDIAN_INSTANCE like ggshield does; --instance overrides it
 INSTANCE="${GITGUARDIAN_INSTANCE:-}"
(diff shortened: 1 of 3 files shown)


scripts/install/install.sh
```diff
@@ -276,6 +283,51 @@ emit_plugin_hint() {
...
+            # Linux: interactive terminals read ~/.bashrc.
+            rc="$HOME/.bashrc"
+        fi
+        ;;
+    *)
+        rc="$HOME/.profile"
+        reload="."
+        ;;
+    esac
+    printf '    # %s is not on your PATH; add it, then restart your terminal:\n' "$BIN_DIR"
```
benjaminrigaud-gg: Can't we have a warn on this line (same for fish & co)?

It isn't that visible in the paté!

Image

benjaminrigaud-gg: It was more visible before IMO 🤷 

Image

benjaminrigaud-gg: What does it look like on macos & **main** for you?

clement-tourriere: Good catch — fixed. The headline now goes through `warn()` (bold-yellow `warning:`, on stderr) for **every** shell branch (zsh/bash, the generic-sh fallback, **and** fish), so it stands out instead of blending into the indented command block. The copy-paste commands stay on stdout, unprefixed, so they paste cleanly. This restores the visibility the pre-rework `warn` had, while keeping the new shell-specific guidance.

While dogfooding the installer I also hit a related gap that this thread made me look at: the **more common** END-598 symptom is actually that `~/.local/bin` *is* on `PATH`, but an older `ggshield` (brew/pipx/manual) sits **ahead** of it and shadows the fresh install. The old code only said the dead-end `fix your PATH order`, and the new hint never fired (because `BIN_DIR` was technically present). That now routes through the same `emit_path_hint`, naming the offending binary and giving the exact prepend command:

```
warning: an older ggshield at /usr/local/bin/ggshield shadows the new one; put /Users/you/.local/bin first on your PATH, then restart your terminal:
    echo 'export PATH="/Users/you/.local/bin:$PATH"' >> "/Users/you/.zshrc"
    # (or apply it to the current shell now: source "/Users/you/.zshrc")
```

(fish uses `fish_add_path --move` to pull it to the front.)

> What does it look like on macos & main for you?

On `main`, macOS printed two generic `warning:` lines **mid-run** (before the auth/plugin output): `warning: $BIN_DIR is not in your PATH. Add this to your shell profile:` / `warning:   export PATH=...`. So it was visible but generic and buried mid-install. With these two commits it's visible **and** shell-specific **and** at the very end. Verified across all 6 cases (ready / shadowed×{zsh,bash,fish} / absent×{zsh,fish}).

benjaminrigaud-gg: better!

```
➜  projects curl -sSfL https://raw.githubusercontent.com/GitGuardian/ggshield/6724be4c196e624e476371e75f1b9320ec11b80c/scripts/install/install.sh | PATH="/usr/bin:/bin" GGSHIELD_BIN_DIR="$HOME/.local/bin" bash
==> Platform: linux/x86_64 (gnu) — installing the standalone build
==> Resolving latest ggshield version
==> Downloading ggshield-1.52.2-x86_64-unknown-linux-gnu.tar.gz
==> Verifying sha256 checksum
==> Skipping build provenance check (set GGSHIELD_REQUIRE_ATTESTATION=1 to require it). To verify a downloaded asset yourself:
    gh attestation verify <asset> --repo GitGuardian/ggshield
==> Installing to /home/benjaminrigaud/.local/share/ggshield-standalone/ggshield-1.52.2-x86_64-unknown-linux-gnu
==> Installed: ggshield, version 1.52.2
==> Authenticating against https://dashboard.gitguardian.com
ggshield is already authenticated without an expiry date
==> To finish setup:
warning: /home/benjaminrigaud/.local/bin is not on your PATH. Add it, then restart your terminal:
    echo 'export PATH="/home/benjaminrigaud/.local/bin:$PATH"' >> "/home/benjaminrigaud/.zshrc"
    # (or apply it to the current shell now: source "/home/benjaminrigaud/.zshrc")
```

benjaminrigaud-gg (approved):
LGTM

PR #1327, opened 2026-06-24
feat(machine): install git hooks machine-wide when setup runs as root

amascia-gg:
## Context

A security engineer deploying ggshield via MDM flagged that `machine setup`'s git-hook step is per-user. MDM runs as **root**, and `git config --global` writes to *root's* `~/.gitconfig` — so the hooks protect nobody who actually commits. (The honeytoken step already fans out across user homes when root; the git-hook step did not.)

## What this PR does

- `ggshield machine setup` now installs git hooks **machine-wide** when running **as root** or with the new **`--system`** flag, instead of only for the invoking user.
  - **As root / MDM** → sets git's **`--system` `core.hooksPath`** (one `/etc/gitconfig` entry) → every user on the machine is covered by a single install, for all their repos, now and in the future. No per-repo files, no repo scanning — it's git's one hooks-path pointer.
  - **Normal user, no flag** → unchanged: per-user (`--global`), affects only that user.
  - Selection is `--system or os.geteuid() == 0`.
- The shared system hooks dir and scripts are created **world-readable/executable (0755)** — the hook runs as each committing user — vs the per-user `0700`.

## Implementation

- `core/dirs.py`: `get_system_data_dir()` — site-wide data dir (`platformdirs.site_data_dir`: `/Library/Application Support/ggshield` on macOS, `/usr/local/share/ggshield` on Linux), with a `GG_SYSTEM_DATA_DIR` override for tests.
- `cmd/install.py`: `install_system()`, `get_system_hook_dir_path()`, `get_default_system_hook_dir_path()`; `create_hook(..., world_readable=False)` controls the 0755 vs 0700 mode of the dir + script.
- `cmd/machine/setup.py`: `--system` flag, `_is_root()`, and `_setup_git_hooks(system)` selecting system vs global scope. The idempotency/foreign-hook handling is unchanged across both scopes.

## Notes

- `--system` writing `/etc/gitconfig` needs root; as a normal user it fails unless redirected (that's how the tests exercise it — `GIT_CONFIG_SYSTEM` + `GG_SYSTEM_DATA_DIR`).
- System scope is a machine-wide **default**: git precedence is local > global > system, so a user/repo with its own `core.hooksPath` (e.g. Husky) still wins.

## Validation

```
# system scope without needing root (redirected):
GIT_CONFIG_SYSTEM=$TMP/sys GG_SYSTEM_DATA_DIR=$TMP/data \
  ggshield machine setup --system --no-ai-hooks --no-honeytokens
git config --system --get core.hooksPath   # -> $TMP/data/git-hooks ; scripts are 0755
```

- New unit tests: system-scope via `--system`, root-implies-system, flag-forwarding, per-user path pinned non-root (CI-deterministic), world-readable perms, `install_system` end-to-end, and `get_system_data_dir` override/default.
- black / isort / flake8 / pyright clean; import-linter both contracts KEPT.

## PR check list

### DIFF
--- a/changelog.d/20260624_160000_achille.mascia_machine_setup_system_hooks.md
+++ b/changelog.d/20260624_160000_achille.mascia_machine_setup_system_hooks.md
@@ -1,0 +1,3 @@
+### Added
+
+- `ggshield machine setup` now installs git hooks **machine-wide** (for every user) when run as root or with the new `--system` flag, instead of only for the invoking user. This makes MDM/fleet deployments work as expected: a single root-run `machine setup` sets git's system `core.hooksPath`, so every user on the machine is covered. Without root (and without `--system`) it keeps installing per-user as before.
--- a/ggshield/cmd/install.py
+++ b/ggshield/cmd/install.py
@@ -9,7 +9,7 @@ from click import UsageError
 from ggshield.cmd.utils.common_options import add_common_options
 from ggshield.cmd.utils.context_obj import ContextObj
 from ggshield.core import ui
-from ggshield.core.dirs import get_data_dir
+from ggshield.core.dirs import get_data_dir, get_system_data_dir
 from ggshield.core.errors import UnexpectedError
 from ggshield.utils.git_shell import check_git_dir, git
 from ggshield.verticals.ai.installation import (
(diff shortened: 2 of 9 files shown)


xblanchot-gg (commented):
LGTM, just the question of is_root implem that might already exist somewhere else, and should be shared across ggshield, not tied to hooks

xblanchot-gg (approved):

PR #1329, opened 2026-06-24
fix(cli): fail fast with a clear message on Python < 3.9

clement-tourriere:
`requires-python` is already `>=3.9`, but `uv tool install` / pipx can still place ggshield on an old *system* interpreter (e.g. Python 3.6 on RHEL 8 / CentOS 8 / openSUSE Leap 15), where it dies with a cryptic `ModuleNotFoundError: No module named 'dataclasses'` deep in startup.

This adds an import-time guard in `ggshield/__init__.py` that exits with a clear, actionable message (pointing at both `uv` and `pipx`) before any 3.7+ import is reached (kept free of 3.7+ syntax so it runs on the old interpreter itself).

Tests: `tests/unit/test_python_version_guard.py` (rejects 3.6/3.7/3.8 with the message; accepts 3.9+). Python 2.x is intentionally not covered: its annotations are 3.x-only syntax, so the module fails at parse time and the guard is never reached. Reviewed via Codex.

Internal ref: END-618 (surfaced by the satori install matrix).

### DIFF
--- a/ggshield/__init__.py
+++ b/ggshield/__init__.py
@@ -1,1 +1,26 @@
+import sys as _sys
+
+
 __version__ = "1.52.2"
+
+_MIN_PYTHON = (3, 9)
+
+
+# Runs on the (possibly pre-3.9) interpreter itself, so it must avoid 3.7+ syntax
+# (f-strings, evaluated annotations) or it crashes before explaining why.
+def _ensure_supported_python(
+    version_info: "tuple[int, ...]" = _sys.version_info[:2],
+) -> None:
+    if tuple(version_info[:2]) < _MIN_PYTHON:
+        _sys.stderr.write(
+            "ggshield requires Python %d.%d or newer, but it is running on "
+            "Python %d.%d.\n"
+            "Reinstall it on a supported interpreter, for example:\n"
+            "    uv tool install --python 3.12 ggshield\n"
+            "    pipx install --python 3.12 ggshield\n"
+            % (_MIN_PYTHON[0], _MIN_PYTHON[1], version_info[0], version_info[1])
+        )
+        raise SystemExit(1)
+
+
+_ensure_supported_python()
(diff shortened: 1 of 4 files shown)


benjaminrigaud-gg (approved):

tests/unit/test_python_version_guard.py
```diff
@@ -0,0 +1,28 @@
+"""The import-time guard rejects pre-3.9 interpreters with a clear message."""
+
+from __future__ import annotations
+
+import pytest
+
+import ggshield
+
+
+@pytest.mark.parametrize("version", [(2, 7), (3, 6, 0), (3, 7, 5), (3, 8, 18)])
```
benjaminrigaud-gg: The `(2, 7)` case only exercises the version arithmetic — it can't reflect real 2.7 behavior. On an actual CPython 2.7 this module fails at *parse* time (function annotations like `version_info: "tuple[int, ...]"` / `-> None` are Python-3-only syntax), so `_ensure_supported_python` never runs and the friendly message is never printed. The guard does genuinely work for the 3.6–3.8 system interpreters this PR targets (annotations exist there). Suggest dropping the 2.7 row, or adding a comment that it only tests the comparison logic — otherwise it reads as "2.7 is handled gracefully," which isn't the case.

ggshield/__init__.py
```diff
@@ -1 +1,24 @@
...
+# (f-strings, evaluated annotations) or it crashes before explaining why.
+def _ensure_supported_python(
+    version_info: "tuple[int, ...]" = _sys.version_info[:2],
+) -> None:
+    if tuple(version_info[:2]) < _MIN_PYTHON:
+        _sys.stderr.write(
+            "ggshield requires Python %d.%d or newer, but it is running on "
+            "Python %d.%d.\n"
+            "Reinstall it on a supported interpreter, for example:\n"
+            "    uv tool install --python 3.12 ggshield\n"
```
benjaminrigaud-gg, line 19: The remediation hint is uv-only, but the PR description (and the bug) also call out pipx — a pipx user hitting this gets uv-specific advice. Consider covering both, e.g. add a `pipx install --python 3.12 ggshield` line, or make it more installer-agnostic ("reinstall on Python ≥ 3.9, e.g. ...").

PR #1331, opened 2026-06-24
feat(machine): provision a machine-wide service-account token in setup

amascia-gg:
## Context

The first MDM request: fleet auth. The system-wide git hooks (#1327) are **fail-closed** — every account that commits needs a working token — but there was no way to provision auth machine-wide. `ggshield auth login` is interactive (browser), and `GITGUARDIAN_API_KEY` is per-shell/per-user. This adds the auth counterpart to the system-wide git hooks: one **service-account token** an MDM provisions once, used by every account on the machine.

## What this PR does

`ggshield machine setup` can now provision a machine-wide service-account token:

```
GGSHIELD_SERVICE_ACCOUNT_TOKEN="$TOKEN" ggshield machine setup --system --instance https://dashboard.gitguardian.com
```

- **Token input**: `GGSHIELD_SERVICE_ACCOUNT_TOKEN` env var, or piped on stdin — **never a CLI argument**, so it can't leak into `ps`/shell history.
- **Validated before storing**: setup checks the token is reachable and has the `scan` scope (same check as `auth login --method token`); a bad token is rejected and nothing is written.
- **Stored machine-wide**: written to `get_system_config_dir()/auth.yaml` (e.g. `/Library/Application Support/ggshield` on macOS, `/etc/xdg/ggshield` on Linux), world-readable `0644` — by design, since it's one shared identity for every account on the machine (same trade-off as the system-wide git hooks). Writing requires root.
- **Read precedence**: config resolution gains a *system-config* source — **below** `GITGUARDIAN_API_KEY` and a user's own `auth login`, **above** the default. So a developer's personal token always wins; the machine-wide service-account token is the fallback every other account uses. The system token is only used for the instance it was provisioned for.
- Adds `--instance` to `machine setup`; provisioning is a no-op when no token is supplied (setup keeps working unchanged).

This completes the fleet story: `machine setup --system` (as root, via MDM) now configures **auth + git hooks + AI hooks + honeytoken** machine-wide in one command.

## Validation

Tested locally (real CLI):
- **Read fallback** — with the per-user config isolated, `api-status` resolves `api_key_source: SYSTEM_CONFIG` from the system file.
- **Write, bad token** — `machine setup` with a bad token prints "rejected", exits non-zero, writes nothing.
- **Write, real token** — validates and writes `auth.yaml` (`0644`).
- Unit: `system_auth` read/write + perms; config fallback (token + instance, env-wins, instance-mismatch ignored); setup provisioning (env/stdin token read, validate/reject/missing-scope/write-failure, `--instance`).


ggshield/core/config/system_auth.py
```diff
@@ -0,0 +1,80 @@
...
+        os.chmod(path.parent, 0o755)
+    except OSError:
+        pass
+    path.write_text(
+        yaml.safe_dump(
+            {"instance": instance, "token": token}, default_flow_style=False
+        ),
+        encoding="utf-8",
+    )
+    os.chmod(path, 0o644)
```
clement-tourriere, line 79: **[high] Long-lived bearer token stored cleartext + world-readable.** `0644` means any unprivileged local user or compromised process can `cat` this and use the SAT off-host; rotating the file can't revoke copies already taken. Per-user tokens here go to keyring or `0600` (`core/config/utils.py:71`) — this is strictly weaker for a *higher*-value credential.

"Shared across accounts" doesn't require world-readable: root-owned `0640` with a dedicated `ggshield` group (or `0600` if consumers run as root) gives every ggshield process read access without giving it to every process on the box. Please also document a rotation/revocation procedure.

ggshield/core/config/system_auth.py
```diff
@@ -0,0 +1,80 @@
...
+    """Write the machine-wide service-account auth and return its path.
+
+    The file and its directory are made world-readable (0644 / 0755) so every
+    account on the machine can use the token. Requires write access to the system
+    config dir (i.e. root); the OSError propagates otherwise.
+    """
+    path = get_system_auth_path()
+    path.parent.mkdir(parents=True, exist_ok=True)
+    try:
+        os.chmod(path.parent, 0o755)
```
clement-tourriere, line 70: **[med] Unconditionally widens the system config dir and swallows failures.** If an admin/MDM created this dir as `0750`, `chmod(..., 0o755)` silently overrides it; and `except OSError: pass` means a failed chmod leaves users unable to read the token with no signal. Create the dir with the intended mode, preserve stricter existing modes, and fail loudly on incompatible permissions.

clement-tourriere (commented):
## Design question before the line notes: is planting a *visible* SAT the right shape?

Two things worry me about the core approach, independent of the code details.

**1. We write a long-lived, high-value service-account token in cleartext, world-readable (`0644`), on every machine.** That is strictly weaker than what we give a *personal* token (keyring, or `0600` on the file fallback — see `core/config/utils.py:71`). One shared identity means one compromised unprivileged account on one machine yields a token that works fleet-wide, and rotating the file cannot claw back copies already exfiltrated. If we keep an on-disk file, it should at minimum be root-owned `0640` with a dedicated `ggshield` group (or `0600` if the consumers run as root), plus a documented rotation/revocation runbook — not world-readable.

**2. We will scan-detect our own planted secret.** `ggshield secret scan path` (and the `machine scan` plugin) over the system config tree will hit `auth.yaml`, recognize a GitGuardian token, and report it as a leak. The default exclusions in `secret_scan_common_options.py` do not cover this generated path, so the two features that ship together collide by default — and "GitGuardian left a GitGuardian token in cleartext on disk" is a bad thing to surface to a customer. We should either not store cleartext, or add an exact suppression for the generated path (with a test proving it doesn't hide arbitrary user `auth.yaml` files).

On **keyring**: it's the instinctive answer, but it genuinely doesn't fit here — the backends are per-user and need an unlocked session / D-Bus, so they can't back a *shared, headless, root-provisioned* fallback and would silently degrade to a file anyway. So a file is defensible; world-readable + unverified-token-type is not. Whatever we pick, let's put the rationale in the module docstring.

Also note this is a **fallback, not enforcement**: `GITGUARDIAN_API_KEY` / a user's own `auth login` still win, and the system hook runs plain `ggshield secret scan`, so a user on a managed machine can still route hooks to a personal token or a different instance. If the goal is fleet *enforcement*, that needs a locked mode; if it's convenience, let's say so.

**Smallest set to unblock, in priority order:**
1. Don't store world-readable — `0640` + dedicated group (or `0600` if consumers run as root).
2. Verify the token is actually an active *service-account* token (type/status/expiry), not any token that happens to carry `scan`.
3. Make stdin the documented entry point; mark the env var MDM-only.
4. Add the generated auth path to the scanner exclusions (or drop cleartext entirely, which removes items 1 and 4).

Inline notes below on the specifics. The top three are the ones I'd want resolved before merge.

ggshield/cmd/machine/setup.py
```diff
@@ -174,3 +211,92 @@ def _setup_honeytokens(ctx: click.Context) -> bool:
     """Plant a honeytoken on this machine (idempotent reconcile via `plant`)."""
     click.echo(click.style("Honeytoken", bold=True))
     return ctx.invoke(plant_cmd) == 0
+
+
+def _read_service_account_token() -> Optional[str]:
+    """Read the service-account token from the env var or piped stdin (never argv,
+    so it does not leak into `ps` / shell history). None if not provided."""
```
clement-tourriere, line 218: **[high] "does not leak into `ps` / shell history" only holds vs argv.** Once set, `GGSHIELD_SERVICE_ACCOUNT_TOKEN` is readable via `/proc/<pid>/environ` and `ps e`, inherited by every child process, and — if an operator types the literal token instead of `"$TOKEN"` — persisted to shell history. stdin is the safe path and it's already implemented below; please make it the *documented default* (`printf %s "$TOKEN" | ggshield machine setup --system --instance …`) and mark the env var MDM-only. Consider unsetting the env var after reading it.

ggshield/cmd/machine/setup.py
```diff
@@ -174,3 +211,92 @@ def _setup_honeytokens(ctx: click.Context) -> bool:
...
+        response = client.get(endpoint="token")
+    except Exception as exc:  # noqa: BLE001 - report any validation failure as a result
+        ui.display_warning(f"  could not validate the service-account token: {exc}")
+        return False
+
+    if not response.ok:
+        ui.display_warning(f"  the service-account token was rejected by {instance}.")
+        return False
+
+    scopes = set(safe_response_json(response).get("scope", []))
```
clement-tourriere, line 271: **[high] No verification that this is actually a service-account token.** Validation only checks `response.ok` + scopes. It never confirms the credential is a *service account*, active, non-expired, or bound to the expected workspace. A personal access token that happens to carry `scan` would pass here and get written machine-wide, world-readable, to every account — the worst case of the storage issue. Please parse `/token`, require the expected token type/status/expiry, and fail closed on personal/unknown tokens.

ggshield/cmd/machine/setup.py
```diff
@@ -89,12 +108,30 @@ def setup_cmd(
...
+
+    For fleet/MDM deployment, provide a service-account token via the
+    `GGSHIELD_SERVICE_ACCOUNT_TOKEN` environment variable (or piped on stdin) and
+    `--instance`: setup stores it machine-wide so every account on the machine
+    authenticates with it, without a per-user `ggshield auth login`. Requires root.
     """
     if agents and exclude_agents:
         raise click.UsageError("--agent and --exclude-agent cannot be used together.")
 
+    if instance:
```
clement-tourriere, line 120: **[low] `--instance` isn't enforced when a token is supplied.** The help text says provisioning uses `--instance`, but when it's omitted the SAT is written against whatever `config.instance_name` resolves to (env / user / default / system). During an MDM run that can silently target the wrong instance. When a token is supplied, require an explicit `--instance` and fail if absent — don't infer the target from mutable config.

ggshield/cmd/machine/setup.py
```diff
@@ -89,12 +108,30 @@ def setup_cmd(
...
+        ContextObj.get(ctx).config.cmdline_instance_name = instance
+
     failed = False
 
+    # Provision machine-wide auth first (if a service-account token is supplied) so
+    # the steps below authenticate with it.
+    if not _setup_service_account_auth(
```
clement-tourriere, line 127: **[med] No explicit root/`--system` gate, and failure doesn't halt setup.** (a) Provisioning runs purely because a token is present; it relies solely on the write `OSError` to fail on non-root. If `GG_SYSTEM_CONFIG_DIR` points somewhere user-writable, a non-root user can provision "machine-wide" auth. Gate on `is_root()`/`--system` *before* the network validation and write. (b) On failure this only sets `failed = True` but setup still installs the hooks + honeytoken with no working machine auth — decide whether a provisioning failure should abort.

ggshield/core/config/config.py
```diff
@@ -197,7 +206,19 @@ def get_api_key_and_source(self) -> Tuple[str, ConfigSource]:
...
+            except (MissingTokenError, UnknownInstanceError):
+                # No personal token for this instance — fall back to the machine-wide
+                # service-account token (fleet/MDM), but only if it is for this same
+                # instance, so we never use a token against the wrong instance.
+                system_auth = read_system_auth()
+                if (
+                    system_auth is not None
+                    and system_auth.instance == self.instance_name
+                ):
+                    return system_auth.token, ConfigSource.SYSTEM_CONFIG
```
clement-tourriere, line 220: **[med] This makes the SAT a fallback, not an enforced policy.** `GITGUARDIAN_API_KEY` and a user's own `auth login`/instance still win, and the system git hook runs plain `ggshield secret scan`. So on a managed fleet machine any user can route hooks to a personal token or a different instance — the machine token is only used when nothing else resolves. If the feature is meant to *enforce* machine auth, that needs a locked system mode; otherwise let's document it explicitly as non-enforcing fallback auth. (Instance-mismatch guarding here is good, by the way.)

PR #1332, opened 2026-06-24
fix(plugin): surface enabled plugins that fail to load (END-620)

clement-tourriere:
## Problem

The `machine_scan` endpoint plugin ships a native (PyO3) extension. When an **enabled** plugin fails to load in the field — e.g. a manylinux wheel installed on a musl/Alpine host raises an `ImportError` during import — the loader caught it, logged at `WARNING` (invisible at the default UI level), and moved on.

The user then runs `ggshield machine scan` and gets a bare:

```
Error: No such command 'machine'.
```

…while `ggshield plugin list` still shows the plugin as **enabled**. Confusing and unactionable — this was hit during endpoint-protection rollout.

## Fix

Make a failed load **visible and explained**:

- **`PluginRegistry`** records load failures (`record_load_failure` / `get_load_failures`).
- **`load_enabled_plugins`** records a failure on the version-incompatibility path, the generic exception path, and when the load returns `None`.
- **`_load_from_wheel`** now *propagates* import/load errors instead of swallowing them to `None`, so the real reason (e.g. native ABI mismatch) reaches the registry — this was the actual field scenario and the previous version still missed it.
- **`main()`** warns about each failed plugin **before Click dispatches**. The old code warned inside the group callback `cli()`, which Click never runs when the command is unknown — so the reported case showed no hint. Now the warning fires regardless.
- **`ggshield plugin list`** shows `enabled, FAILED TO LOAD: <reason>` instead of `enabled`.

A healthy plugin is unaffected; one plugin failing does not block the others.

### Example after this change

```
Plugin 'machine_scan' is enabled but failed to load: <reason>. Its commands are unavailable; run `ggshield plugin list` for status.
Error: No such command 'machine'.
```

## Tests

- `tests/unit/core/plugin/test_load_failures.py` (new): failure recorded + commands not registered; one failure doesn't block others (failing plugin ordered first); wheel import error records the real reason; `None` load recorded; disabled plugin is not a failure.
- `tests/unit/core/plugin/test_loader.py`: `test_load_from_wheel_propagates_import_error` (was `…_handles_exception`, now asserts propagation).
- `tests/unit/cmd/plugin/test_list.py`: `FAILED TO LOAD` shown for an enabled-but-failed plugin.
- `tests/unit/test_main_plugin_registration.py`: `_warn_about_failed_plugins` emits one warning per failure / silent when none; end-to-end test runs the real `cli.main` for an unknown plugin command and asserts the warning **and** "No such command" both appear.

Full unit suite green. Reviewed with Codex (two blockers found and fixed; both re-confirmed resolved).

## Follow-up (out of scope)

Top-level plugin-loading/register failures still route through `_deferred_warnings`, flushed only in the group callback — so a whole-registry load failure stays invisible on the unknown-command path. Pre-existing and separate from the per-plugin path fixed here; worth a follow-up ticket.

### DIFF
--- a/ggshield/__main__.py
+++ b/ggshield/__main__.py
@@ -221,6 +221,16 @@ def _register_plugin_commands() -> None:
         _deferred_warnings.append(f"Failed to register plugin commands: {e}")
 
 
+# Emitted before Click dispatches: a failed load leaves the command unregistered,
+# so Click rejects it as unknown and the group callback never runs to warn.
+def _warn_about_failed_plugins() -> None:
+    for name, reason in _load_plugins().get_load_failures().items():
+        ui.display_warning(
+            "Plugin '%s' is enabled but failed to load: %s. Its commands are "
+            "unavailable; run `ggshield plugin list` for status." % (name, reason)
+        )
+
+
 def _set_color(ctx: click.Context):
     """
     Helper function to override the default click default output color setting.
@@ -330,6 +340,11 @@ def main(args: Optional[List[str]] = None) -> Any:
 
     force_utf8_output()
     setup_truststore()
+
+    # cli.main() has not yet consumed _GGSHIELD_COMPLETE, so skip the warning
+    # during shell completion to avoid leaking it into completion output.
+    if "_GGSHIELD_COMPLETE" not in os.environ:
+        _warn_about_failed_plugins()
 
     show_crash_log = getenv_bool("GITGUARDIAN_CRASH_LOG")
     return cli.main(args, prog_name="ggshield", standalone_mode=not show_crash_log)
--- a/ggshield/core/plugin/loader.py
+++ b/ggshield/core/plugin/loader.py
@@ -251,14 +251,17 @@ class PluginLoader:
             try:
                 plugin = self._load_plugin(discovered)
                 if plugin is None:
+                    registry.record_load_failure(discovered.name, "could not be loaded")
                     continue
 
                 if not self._check_version_compatibility(plugin.metadata):
-                    logger.warning(
-                        "Plugin %s requires ggshield >= %s, skipping",
-                        plugin.metadata.name,
-                        plugin.metadata.min_ggshield_version,
+                    reason = (
+                        "requires ggshield >= %s" % plugin.metadata.min_ggshield_version
                     )
+                    logger.warning(
+                        "Plugin %s %s, skipping", plugin.metadata.name, reason
+                    )
+                    registry.record_load_failure(discovered.name, reason)
                     continue
 
                 plugin.on_load()
@@ -273,6 +276,7 @@ class PluginLoader:
 
             except Exception as e:
                 logger.warning("Failed to load plugin %s: %s", discovered.name, e)
+                registry.record_load_failure(discovered.name, str(e))
 
         return registry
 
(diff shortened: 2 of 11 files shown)


benjaminrigaud-gg (approved):
Two comments :o

ggshield/__main__.py
```diff
@@ -302,6 +312,8 @@ def main(args: Optional[List[str]] = None) -> Any:
     force_utf8_output()
     setup_truststore()
 
+    _warn_about_failed_plugins()
```
benjaminrigaud-gg: **The warning fires on every command invocation, not just the failed plugin's command.**

Because `_warn_about_failed_plugins()` runs unconditionally before `cli.main()` dispatches, *any* command emits it. Confirmed empirically — with `machine_scan` recorded as failed, `ggshield secret --help` prints `Warning: Plugin 'machine_scan' is enabled but failed to load…` before the help text.

This is the tradeoff that makes the unknown-command path work (you can't scope a warning to a command that was never registered), and it's on stderr so piped stdout (JSON/SARIF) is safe. But two consequences are worth a conscious call:

- **CI noise:** the field scenario is a host where `machine_scan` can't load — so every `ggshield secret scan` run there now prints this on *every* invocation.
- **Shell completion:** this runs before `cli.main()` reads `_GGSHIELD_COMPLETE`, so it also fires during tab-completion (the old `cli()`-callback warning didn't).

Not a blocker — but consider gating with `if "_GGSHIELD_COMPLETE" not in os.environ:` and/or confirming the every-invocation noise is acceptable for the CI-scan use case.

ggshield/core/plugin/loader.py
```diff
@@ -251,14 +251,17 @@ def load_enabled_plugins(self) -> PluginRegistry:
...
+                    reason = (
+                        "requires ggshield >= %s" % plugin.metadata.min_ggshield_version
+                    )
                     logger.warning(
-                        "Plugin %s requires ggshield >= %s, skipping",
-                        plugin.metadata.name,
-                        plugin.metadata.min_ggshield_version,
+                        "Plugin %s %s, skipping", plugin.metadata.name, reason
                     )
+                    registry.record_load_failure(discovered.name, reason)
```
benjaminrigaud-gg, line 264: **This version-incompatibility recording branch is untested.** The new `test_load_failures.py` covers the generic-exception, wheel-import, `None`-return, and disabled paths — but not this one. It's the one new failure-recording branch without a dedicated test. Cheap to add:

```python
def test_version_incompatible_plugin_is_recorded_as_failure():
    plugin = MagicMock()
    plugin.metadata.name = "machine_scan"
    plugin.metadata.min_ggshield_version = "99.0.0"
    loader = PluginLoader(EnterpriseConfig())
    with patch.object(PluginLoader, "discover_plugins", return_value=[_enabled()]), \
         patch.object(PluginLoader, "_load_plugin", return_value=plugin), \
         patch.object(PluginLoader, "_check_version_compatibility", return_value=False):
        registry = loader.load_enabled_plugins()
    assert "requires ggshield >= 99.0.0" in registry.get_load_failures()["machine_scan"]
```

benjaminrigaud-gg (commented):
Two notes from a review pass — one UX decision and one test-coverage gap. Neither blocks merge.

PR #1338, opened 2026-06-26
feat(machine): surface git core.hooksPath overrides that shadow ggshield

amascia-gg:
## Context

When `machine setup` installs git hooks machine-wide (system `core.hooksPath`, #1327), git's config precedence (**local > global > system**) means a repo that has its **own** `core.hooksPath` — Husky, lefthook, a corporate setup — silently **wins**, and ggshield's hook never runs there. Git uses exactly one hooks directory, so we cannot out-rank a more-specific override; there's no "system hook that also runs."

We can't *force* it (git precedence is a hard wall, and local hooks are bypassable anyway — the real backstop is GitGuardian's server-side scanning). So the fix is to **stop the silent gap**: surface it instead of giving a false sense of coverage.

## What this PR does

- **`machine doctor`** gains a **"Git hook precedence"** check. It resolves the *effective* `core.hooksPath` (`git config --get core.hooksPath`, respecting precedence) and, if a non-ggshield directory is in effect, reports ✗ — naming the shadowing path (e.g. `.husky/_`) and how to fix it (integrate ggshield into that hook manager, or unset the override; commits are still scanned server-side). Per-context: run it in a repo to check that repo. ✓ when nothing shadows ggshield.
- **`machine setup`** warns at install time when a `core.hooksPath` override takes precedence in the current context, so the operator knows its hook will be shadowed there.

Shared helpers in `cmd/install.py`: `get_configured_hook_dir_path()` (effective path), `is_ggshield_hook_dir()`, `get_shadowing_hooks_path()`.

## What it does NOT do

It does not (and cannot) make ggshield run in a repo that has its own `core.hooksPath` — git won't allow a lower-precedence hook to run alongside. That's a per-repo integration (add ggshield to the other tool) or a server-side concern. This PR makes the gap **visible**, not closed.

## Validation

Tested at runtime:
- doctor in a repo with `git config core.hooksPath .husky/_` → ✗ "core.hooksPath points to .husky/_ … ggshield's hook will NOT run"; in a clean repo → ✓.
- setup with such an override present → installs the hook, then warns "a core.hooksPath override (.husky/_) takes precedence here …".

Unit: `get_configured_hook_dir_path` / `is_ggshield_hook_dir` / `get_shadowing_hooks_path` (install), the doctor precedence check, and the setup warning. Existing `TestSetupGitHooks` pinned so they don't shell out to real `git config`. black/isort/flake8/pyright + import-linter clean.

## PR check list

### DIFF
--- a/ggshield/cmd/install.py
+++ b/ggshield/cmd/install.py
@@ -11,7 +11,7 @@ from ggshield.cmd.utils.context_obj import ContextObj
 from ggshield.core import ui
 from ggshield.core.dirs import get_data_dir, get_system_data_dir
 from ggshield.core.errors import UnexpectedError
-from ggshield.utils.git_shell import check_git_dir, git
+from ggshield.utils.git_shell import GitError, check_git_dir, get_git_root, git
 from ggshield.verticals.ai.installation import (
     AGENTS,
     check_ai_hook_authentication,
@@ -33,6 +33,10 @@ if [ -f "$_ggshield_local_hook" ]; then
     fi
 fi
 """
+
+# The line `create_hook` writes into every ggshield-managed hook script; its presence
+# is how we recognize a hook (or hooks dir) as ggshield's own.
+GGSHIELD_HOOK_MARKER = "ggshield secret scan"
 
 
 @click.command(context_settings={"ignore_unknown_options": True})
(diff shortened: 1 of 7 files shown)


ggshield/cmd/install.py
```diff
@@ -169,6 +173,80 @@ def get_system_hook_dir_path() -> Optional[Path]:
...
+
+def hook_invokes_ggshield(hook_path: Path) -> bool:
+    """Whether ``hook_path`` is an existing hook script that runs ggshield.
+
+    Single source of truth for recognizing a ggshield-managed hook by the marker line
+    ``create_hook`` writes.
+    """
+    return hook_path.is_file() and GGSHIELD_HOOK_MARKER in hook_path.read_text(
+        errors="ignore"
+    )
```
pbeslin-gg: might be worth some error handling (OSError?), since `errors="ignore"` seems to be only for encoding errors

ggshield/cmd/install.py
```diff
@@ -169,6 +173,80 @@ def get_system_hook_dir_path() -> Optional[Path]:
...
+def is_ggshield_hook_dir(hook_dir: Path) -> bool:
+    """Whether ``hook_dir`` holds (or, for Husky, sources) ggshield hook scripts.
+
+    For a Husky ``.husky/_`` directory the runnable hook lives in the parent
+    ``.husky/`` — that is where ``ggshield install --mode local`` writes its content
+    and what Husky's ``_`` wrappers source — so the parent is inspected too.
+    """
+    candidate_dirs = [hook_dir]
+    if is_husky_hooks_path(hook_dir):
+        candidate_dirs.append(hook_dir.parent)
```
pbeslin-gg, line 232: Seems like there are a few other special cases that need to be handled: for instance pre-commit.com and lefthook both seem to have their own config files, with `.git/hooks` only launching them

pbeslin-gg (commented):
Nice! Works well for me, code looks good

ggshield/cmd/install.py
amascia-gg: They're actually handled by the generic detection. In their default setup both pre-commit.com and lefthook install into `.git/hooks/`, which ggshield's global/system hook chains to (`LOCAL_HOOK_SNIPPET`), so they still run — correctly not flagged as a shadow. And if either sets its own `core.hooksPath`, `get_shadowing_hooks_path` flags any non-ggshield hooksPath, so it's caught with the right remediation.

The Husky parent-dir case is special only because ggshield's own `install --mode local` writes into `.husky/`'s parent (not the `_` dir git points at), so recognizing "this hooksPath is ggshield's" needs a parent lookup — not because Husky is allowlisted. pre-commit.com/lefthook have no such ggshield-side install indirection, so there's nothing analogous to special-case.

pbeslin-gg: Ok I thought the goal was to cover manual installations too, where the ggshield scan command would live in a `.pre-commit-config.yaml` for instance. Since `ggshield setup` will be the install method most of the time, I understand we don't need it :+1:

pbeslin-gg (approved):

gmendy1 (approved):

PR #1342, opened 2026-07-01
feat(scripts): offer to add the install dir to PATH instead of only printing it

benjaminrigaud-gg:
## Summary

- Detect the shell and offer to add `~/.local/bin` to `PATH` when it's missing, instead of only printing instructions: bash, zsh, fish, and — independent of `$SHELL`, since they're rarely the login shell — Nushell and PowerShell on Linux when installed.
- Prompted with a default of yes, silent under `-y`; a decline, an unreachable prompt, or the new `--no-modify-path` falls back to the existing printed instructions.
- `uninstall.sh` reverses the edit using the files `install.sh` recorded in its state file, not shell re-detection (`$SHELL` may differ by then).
- Wire `shellcheck` into pre-commit, scoped to `scripts/install/*.sh` (the repo's other shell scripts have pre-existing findings, one a real bug, that are out of scope here).
- Add a bats functional test suite (9 cases) that sources the scripts and drives fish/nu/pwsh with fake stub binaries — no real shell install needed beyond bats — wired into a new path-filtered GitHub Actions job (ubuntu + macos).

Refs: END-653

## Test plan

- [x] `shellcheck` clean on `scripts/install/install.sh` and `uninstall.sh`
- [x] `bats scripts/install/tests/` passes locally (9/9)
- [x] Manually verified: bash/zsh rc-file edits, idempotency on rerun, `--no-modify-path`, fish `fish_add_path`, nu/pwsh opportunistic config, pwsh gated to Linux only, decline-falls-back-to-print, empty-Enter defaults to yes (via a real pty), install→uninstall round trip leaves unrelated rc-file lines untouched
- [x] CI green on the new `install-scripts.yml` workflow

### DIFF
--- a/scripts/install/install.sh
+++ b/scripts/install/install.sh
@@ -13,29 +13,33 @@
 
 set -euo pipefail
 
-# Every install path (and every PATH hint) is derived from $HOME; fail early
-# with a clear message rather than building paths like /.local/bin under set -u.
-# die() is not defined yet, so emit raw.
+# All paths derive from $HOME; fail early with a clear message (die() isn't defined yet).
 [ -n "${HOME:-}" ] || { printf '\033[1;31merror:\033[0m HOME is not set\n' >&2; exit 1; }
 
 GITHUB_REPO="GitGuardian/ggshield"
 DEFAULT_INSTANCE="https://dashboard.gitguardian.com"
 EU_INSTANCE="https://dashboard.eu1.gitguardian.com"
 
+# Trailing slash stripped so the PATH-membership test and the fish canonical
+# form (builtin realpath -s) match what a user-supplied dir with a slash yields.
 BIN_DIR="${GGSHIELD_BIN_DIR:-$HOME/.local/bin}"
+BIN_DIR="${BIN_DIR%/}"
 # NOT ~/.local/share/ggshield: that is ggshield's own data dir (plugins…)
 OPT_DIR="${GGSHIELD_OPT_DIR:-$HOME/.local/share/ggshield-standalone}"
+OPT_DIR="${OPT_DIR%/}"
 STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/ggshield-install"
 STATE_FILE="$STATE_DIR/state"
 
+# Trailing marker on each rc source line so uninstall strips exactly the lines
+# we added, regardless of the OPT_DIR they point at. Kept identical in uninstall.sh.
+PATH_SENTINEL="# ggshield-install PATH"
+
 ASSUME_YES=0
 INSTALL_ONLY=0
-# Set when BIN_DIR is not on PATH, so the final summary can tell the user how to
-# expose it (see emit_path_hint).
+NO_MODIFY_PATH=0
+# Set when BIN_DIR is not on PATH; drives the final emit_path_hint summary.
 PATH_NEEDS_SETUP=0
-# Set to the path of an older ggshield that resolves ahead of the fresh one
-# while BIN_DIR *is* on PATH; emit_path_hint then says "put BIN_DIR first"
-# rather than "add it".
+# Path of an older ggshield shadowing the fresh one when BIN_DIR is already on PATH.
 SHADOWED_BY=""
 PLUGINS=()
 # honor GITGUARDIAN_INSTANCE like ggshield does; --instance overrides it
@@ -58,6 +62,9 @@ Options:
                       also via GGSHIELD_VERSION env var)
       --install-only  install ggshield only, skip auth and plugins
       --plugin NAME   install this ggshield plugin (repeatable)
+      --no-modify-path
+                      don't offer to add the install dir to PATH; only print
+                      instructions
   -h, --help          show this help
 
 Environment:
@@ -85,8 +92,22 @@ need() {
     command -v "$1" >/dev/null 2>&1 || die "required command not found: $1"
 }
 
-# A remote/headless session has no browser for the OAuth redirect; ggshield's
-# --method oob is the way out (reported by users behind SSH / VM port-forwarding).
+# BIN_DIR/OPT_DIR are interpolated into generated shell/nu/ps1/fish snippets and
+# sourced from an arbitrary cwd, so they must be absolute and free of chars that
+# could break out of those string contexts. Both arms die, so `case` (which is 0
+# on no match) never trips set -e as a trailing `&& die` would.
+assert_safe_path_value() {
+    case "$2" in
+    '' | [!/]*)
+        die "$1='$2' must be an absolute path (start with /)"
+        ;;
+    *[\"\'\`\$\;\\]* | *$'\n'*)
+        die "$1='$2' contains a quote, backtick, \$, ;, backslash, or newline, which cannot safely appear in a generated PATH snippet; use a plain path"
+        ;;
+    esac
+}
+
+# No browser for OAuth in a remote/headless session; ggshield's --method oob is the way out.
 is_headless() {
     [ -n "${SSH_CONNECTION:-}${SSH_TTY:-}" ] && return 0
     [ "$OS" = linux ] && [ -z "${DISPLAY:-}${WAYLAND_DISPLAY:-}" ] && return 0
@@ -151,8 +172,7 @@ resolve_version() {
     [ -n "$VERSION" ] || die "could not resolve the latest version from the GitHub API"
 }
 
-# GitHub computes a sha256 digest for every release asset; pair each "name"
-# with the "digest" that follows it in the release JSON.
+# Pair each asset "name" with the "digest" that follows it in the release JSON.
 asset_digest() {
     local asset="$1"
     fetch "https://api.github.com/repos/$GITHUB_REPO/releases/tags/v$VERSION" |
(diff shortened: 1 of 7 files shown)


scripts/install/uninstall.sh
```diff
@@ -90,6 +140,7 @@ remove_standalone() {
         say "Removing $BIN_DIR/ggshield"
         rm -f "$BIN_DIR/ggshield"
     fi
+    remove_path_lines
     say "Removing $OPT_DIR"
     rm -rf "$OPT_DIR"
```
clement-tourriere: **Bug: uninstall can brick a Nushell config.** `remove_path_lines` is best-effort, but this `rm -rf "$OPT_DIR"` runs unconditionally after it. If stripping the `source "$OPT_DIR/env.nu"` line fails — read-only `config.nu` (the failed `cp` above is `|| true`'d), or `XDG_STATE_HOME` set at install time but not at uninstall time (state file never found, while `found=1` still comes from `OPT_DIR` existing) — the line survives pointing at a deleted file. Nushell resolves `source` at **parse time**, so every subsequent `nu` launch fails to load the entire config. bash/zsh only degrade to a per-shell "No such file" message.

Consider skipping the `OPT_DIR` deletion (or at least keeping `env*`) when the strip step didn't verifiably succeed.

scripts/install/install.sh
```diff
@@ -292,12 +291,24 @@ other install methods in scripts/install/README.md" ;;
 
 write_state() {
     mkdir -p "$STATE_DIR"
-    cat >"$STATE_FILE" <<EOF
+    # Preserve any path_* lines from a prior run: a reinstall only re-records them
+    # when PATH still needs setup, so otherwise an upgrade would orphan uninstall's target.
+    local existing_path_lines=""
+    if [ -f "$STATE_FILE" ]; then
+        existing_path_lines=$(grep '^path_' "$STATE_FILE" 2>/dev/null || true)
```
clement-tourriere, line 311: **Bug: preserved `path_*` lines detach from the `opt_dir` they were written with.** This preserves old `path_rc_file` lines, but the heredoc below overwrites `opt_dir` with the *current* value. Install with the default `OPT_DIR`, then reinstall with `GGSHIELD_OPT_DIR=$HOME/gg-opt`: uninstall's `remove_path_lines` greps rc files only for `$HOME/gg-opt/`, so the old `. ".../ggshield-standalone/env"` line is never removed — and becomes a dangling source once that dir is deleted.

Recording the opt_dir alongside each path record (e.g. `path_rc_file=<rc>:<opt_dir>`, or one `path_opt_dir=` line preserved together with the group) would keep them consistent.

scripts/install/uninstall.sh
```diff
@@ -67,8 +67,57 @@ assert_safe_optdir() {
...
+            rm -f "$value.ggshield-tmp"
+            ;;
+        path_shell)
+            # BIN_DIR travels via an env var, not string interpolation (see install.sh's
+            # configure_fish). Filters with a literal loop, not `string match` (a glob
+            # by default), so a `*`/`?` in fish_bin_dir can't strip unrelated entries.
+            if [ "$value" = fish ] && command -v fish >/dev/null 2>&1; then
+                # shellcheck disable=SC2016 # intentional: fish expands this, not bash
+                GGSHIELD_FISH_BIN_DIR="$fish_bin_dir" fish -c \
+                    'set -U fish_user_paths (for p in $fish_user_paths; test "$p" != "$GGSHIELD_FISH_BIN_DIR"; and echo $p; end)' \
```
clement-tourriere: **Bug: exact-match removal misses fish's normalized path** (verified empirically with fish 4.7.1). `fish_add_path` normalizes before storing in `fish_user_paths` (trailing slash stripped, `..`/`//` collapsed), but install records `path_shell_bin_dir=$BIN_DIR` verbatim and this `test "$p" != …` compares exactly. Install with `GGSHIELD_BIN_DIR="$HOME/.local/bin/"` → fish stores `…/.local/bin`, uninstall looks for `…/.local/bin/`, never matches, and the entry stays in `fish_user_paths` forever.

Normalizing `BIN_DIR` at install time before recording (or comparing with `path normalize` here) closes it. Default value is already normalized, so this only bites user-supplied `GGSHIELD_BIN_DIR`.

scripts/install/install.sh
```diff
@@ -509,6 +673,9 @@ main() {
         shift
     done
 
+    assert_safe_path_value GGSHIELD_BIN_DIR "$BIN_DIR"
```
clement-tourriere, line 705: **Bug: relative `GGSHIELD_BIN_DIR`/`GGSHIELD_OPT_DIR` pass this guard and get written verbatim into profiles.** `assert_safe_path_value` rejects metacharacters but not relative paths: `GGSHIELD_OPT_DIR=opt GGSHIELD_BIN_DIR=bin install.sh -y` from `/tmp` puts `. "opt/env"` in the rc file, resolved against whatever cwd each future shell starts in — missing-file noise at startup and a ggshield that's still not on PATH.

Since the guard already exists, adding an absolute-path check (`[ "${2#/}" != "$2" ] && die …`) here covers it for all four generated languages at once.

scripts/install/install.sh
```diff
@@ -328,70 +338,232 @@ emit_plugin_hint() {
...
+configure_extra_shells() {
+    command -v nu >/dev/null 2>&1 && { configure_nu || true; }
+    [ "$OS" = linux ] && command -v pwsh >/dev/null 2>&1 && { configure_pwsh || true; }
     return 0
 }
 
+# Records what this run touched so uninstall.sh can reverse it without re-detecting the shell.
+record_path_state() {
+    mkdir -p "$STATE_DIR"
+    printf '%s\n' "$1" >>"$STATE_FILE"
```
clement-tourriere: **Minor bug: duplicate `path_*` lines accumulate across same-session reruns.** `write_state` preserves prior `path_` lines, and this blind append re-adds the same key when `emit_path_hint` fires again (rerun before the terminal restart, BIN_DIR still off PATH). Harmless at uninstall (removal is idempotent) but the state file grows on every rerun. A `grep -qxF` guard mirroring `append_line_once` fixes it in one line.

scripts/install/install.sh
```diff
@@ -328,70 +338,232 @@ emit_plugin_hint() {
...
+# ggshield shell setup
+\$env.PATH = (\$env.PATH | prepend "$BIN_DIR" | uniq)
+EOF
+}
+
+write_env_ps1() {
+    mkdir -p "$OPT_DIR" 2>/dev/null || return 1
+    cat 2>/dev/null >"$OPT_DIR/env.ps1" <<EOF || return 1
+# ggshield shell setup
+if ((\$env:PATH -split [IO.Path]::PathSeparator) -notcontains "$BIN_DIR") {
```
clement-tourriere: **Nit-level bug: `-notcontains` is case-insensitive in PowerShell.** On a case-sensitive Linux filesystem, a PATH entry like `/home/a/.Local/bin` suppresses adding the real `/home/a/.local/bin`, leaving ggshield unavailable in `pwsh`. `-cnotcontains` is the case-sensitive form.

scripts/install/uninstall.sh
benjaminrigaud-gg: Fixed. `remove_path_lines` now sets `PATH_STRIP_VERIFIED=0` whenever a strip can't be confirmed — it re-greps the rc file after the write, and a missing state file also counts as unverified. When unverified, `remove_standalone` keeps `env`/`env.nu`/`env.ps1` as neutralized no-op stubs and removes only the rest of `OPT_DIR` (state kept for a retry), so a surviving `source` still resolves. Confirmed with nushell 0.113.1 that the dangling `source` aborts the whole config at parse time, and that keeping the stub prevents it.

scripts/install/install.sh
benjaminrigaud-gg: Fixed. Stripping is no longer keyed on the recorded `opt_dir`. `install.sh` appends an anchored trailing sentinel (`# ggshield-install PATH`) to every rc source line, and uninstall strips by that marker — so a line pointing at a prior/different `OPT_DIR` is removed too. I went with the sentinel over per-line `opt_dir` pairing because it also composes with the dedup fix and removes the pre-existing over-broad `$opt_dir/` substring match.

scripts/install/uninstall.sh
benjaminrigaud-gg: Fixed. `install.sh` records the canonical form via `builtin realpath -s` (the exact primitive `fish_add_path` uses to store the entry, available since 3.2.0), and uninstall re-canonicalizes the recorded value the same way before comparing — so a trailing slash (or `..`/`//`) no longer defeats removal. Verified with fish 4.7.1: install with a trailing slash, then uninstall, removes it correctly. Also switched removal to `set -a keep` + plain `set` (not `set -U`), so entries with newlines/flag-like values aren't mangled and a globally-managed `fish_user_paths` isn't forced universal.

scripts/install/install.sh
benjaminrigaud-gg: Fixed. `assert_safe_path_value` now also rejects non-absolute values (`'' | [!/]*`), applied in both scripts, and trailing slashes are stripped at startup. One note: the suggested `[ "${2#/}" != "$2" ]` is inverted — it rejects absolute paths and accepts relative ones — so the guard uses the `=` form, and it's folded into the existing `case` rather than a trailing `[ … ] && die` (that returns 1 for a valid absolute path and would trip `set -e` at the top-level call).

scripts/install/install.sh
benjaminrigaud-gg: Fixed. `record_path_state` now guards with `grep -qxF` before appending. I also collapsed the fish record from two lines (`path_shell_bin_dir` + `path_shell`) to a single self-contained `path_fish_bin_dir` — the two-line form had the same accumulation/mismatch problem when `BIN_DIR` changed between installs (the dedup guard would keep the new value line but drop the trigger line).

scripts/install/install.sh
benjaminrigaud-gg: Fixed. `env.ps1` now uses `-cnotcontains`. Confirmed on pwsh 7.6.3: with `/x/.Local/bin` on `PATH`, `-notcontains` skips adding the real `/x/.local/bin`, while `-cnotcontains` adds it.

clement-tourriere (approved):

PR #1348, opened 2026-07-07
fix(machine-id): share cache path with satori, drop dead ~/.satori shim

benjaminrigaud-gg:
## Summary

Part of the cross-repo fix for machine-identity divergence between ggshield and satori (**END-699**). Sibling MR (satori): _added below_.

- The reconciliation shim read `~/.satori/machine_id`, but satori writes `~/.ggshield/machine_id` two renames ago — so the two tools never actually shared the random-UUID fallback. Removed the dead shim.
- `_get_machine_id` now derives the hardware id **first**, then falls back to the **shared** `~/.ggshield/machine_id` cache (read + write), matching satori's derivation order. New `_machine_id_cache_path()` helper.
- Net effect: on the common Windows machine ggshield already reported the SMBIOS UUID and is unchanged; satori's companion MR makes satori report the *same* SMBIOS UUID, so the `Endpoint.machine_id` join finally matches.

Identity contract (algorithm, per-platform sources, normalization, cache path, re-keying decision) is documented in the satori MR.

## Test plan
- [x] `tests/unit/core/test_machine_id.py` — 32 passed (added: shared-cache fallback, system-id-precedence, dead-shim-not-consulted; updated write-path assertions to `~/.ggshield`)
- [x] Parity: `_parse_wmic_uuid` output byte-identical to satori's Rust `parse_wmic_uuid` on shared SMBIOS vectors
- [x] `black`, Import Linter, `ty`, commitizen — all green (pre-commit)

Refs END-699

### DIFF
--- a/ggshield/core/machine_id.py
+++ b/ggshield/core/machine_id.py
@@ -219,15 +219,29 @@ def _machine_id_cache_path() -> Path:
 _CACHE_READ_MAX = 65536
 
 
+def _is_trusted_cache_owner(file_uid: int, euid: int, home_uid: Optional[int]) -> bool:
+    """Whether ``file_uid`` may own the fallback-id cache.
+
+    Identity is client-asserted end to end (the server stores whatever
+    machine_id a client reports), so an own-files-only rule buys no integrity.
+    The requirement is that no OTHER unprivileged user can plant the file:
+    trust the current user, root, and the owner of the home the cache lives
+    in — so a ``sudo -E`` run and a plain user run converge on one id instead
+    of diverging.
+    """
+    return file_uid == euid or file_uid == 0 or file_uid == home_uid
+
+
 def _read_trusted_cache_line(path: Path) -> Optional[str]:
     """First non-empty line of the shared fallback-id cache, if trustworthy.
 
-    On POSIX, refuse a group/world-writable file or one not owned by the
-    current effective user, so a local attacker can't pre-seed the machine id
-    for a higher-privilege run (mirrors satori's ``paths::open_trusted_cache_file``).
-    Opens with ``O_NOFOLLOW`` and validates ownership/mode on the very
-    descriptor it reads: a stat-then-open by path could be raced, follows
-    planted symlinks, and blocks forever on a FIFO.
+    On POSIX, refuse a group/world-writable file or one owned by an unrelated
+    user (see ``_is_trusted_cache_owner``), so a local attacker can't pre-seed
+    the machine id for another account's run (mirrors satori's
+    ``paths::open_trusted_cache_file``). Opens with ``O_NOFOLLOW`` and
+    validates ownership/mode on the very descriptor it reads: a stat-then-open
+    by path could be raced, follows planted symlinks, and blocks forever on a
+    FIFO.
     """
     flags = os.O_RDONLY
     if os.name == "posix":
@@ -240,8 +254,15 @@ def _read_trusted_cache_line(path: Path) -> Optional[str]:
         st = os.fstat(fd)
         if not stat.S_ISREG(st.st_mode):
             return None
-        if os.name == "posix" and (st.st_mode & 0o022 or st.st_uid != os.geteuid()):
-            return None
+        if os.name == "posix":
+            if st.st_mode & 0o022:
+                return None
+            try:
+                home_uid: Optional[int] = os.stat(path.parent.parent).st_uid
+            except OSError:
+                home_uid = None
+            if not _is_trusted_cache_owner(st.st_uid, os.geteuid(), home_uid):
+                return None
         data = os.read(fd, _CACHE_READ_MAX)
     except OSError:
         return None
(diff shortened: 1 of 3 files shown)


ggshield/core/machine_id.py
```diff
@@ -24,6 +25,11 @@
 
 _MAC_IOREG_UUID_RE = re.compile(r'"IOPlatformUUID"\s*=\s*"([^"]+)"')
 
+# Bound the system-id subprocesses: a hung WMI service must not freeze scan
+# startup (and thus git hooks) indefinitely — on timeout we fall through to the
```
6d7a, line 29: I see this was not introduced in this MR, but why do we need to send the machine ID unconditionally for all types of scans? At least for linux the machine ID is considered confidential, so if we send it at all we should only attach it to those requests that need it.

ggshield/core/machine_id.py
```diff
@@ -90,25 +97,77 @@ def _get_macos_system_id() -> Optional[str]:
...
 
 
+# The character class is ASCII-only on purpose (no \d / \w, which also match
+# fullwidth Unicode digits).
+_UUID_HEX_RE = re.compile(r"[0-9a-fA-F]{32}")
+
+# Placeholder SMBIOS UUIDs reported by unconfigured or cloned firmware (blank
+# boards, golden images, AMI default). They identify an image, not a machine:
+# adopting one collapses every such endpoint in an account onto one machine_id.
+_SENTINEL_UUID_HEX = frozenset(
```
6d7a, line 108: It seems there are [more](https://github.com/osquery/osquery/issues/8887) [of them](https://github.com/osquery/osquery/blob/99239193e9ffbbf7b1931622d158692967880d62/osquery/core/system.cpp#L90).

ggshield/core/machine_id.py
```diff
@@ -120,38 +179,114 @@ def _get_windows_system_id() -> Optional[str]:
...
+        except OSError:
+            st = None
+        if st is not None:
+            if os.name == "posix" and st.st_uid != os.geteuid():
+                return
+            os.unlink(path)
+        flags = os.O_WRONLY | os.O_CREAT | os.O_EXCL
+        if os.name == "posix":
+            flags |= os.O_NOFOLLOW
+        fd = os.open(path, flags, 0o600)
```
6d7a: Isn't this going to lead to two different machine IDs (and thus partition IDs of the user's findings) between a scan that is initiated by `root` and a scan initiated by a user?

ggshield/core/machine_id.py
benjaminrigaud-gg: I don't know :pokerface:
@pbeslin-gg do you have more detail about https://github.com/GitGuardian/ggshield/pull/1333 ?

pbeslin-gg: The intent is to allow users to see on the GitGuardian dashboard which secrets were detected by scans on their machine. All scan types are concerned in the end, but only scans triggered by the AI agent hooks currently need it.

ggshield/core/machine_id.py
benjaminrigaud-gg: I loosened the guard and created a follow-up.

ggshield/core/machine_id.py
benjaminrigaud-gg: updated!

6d7a (approved):

clement-tourriere (approved):

ggshield/core/machine_id.py
benjaminrigaud-gg: Agreed. The header comes from #1333 (unreleased), and the flows that genuinely need endpoint attribution (AI discovery, honeytoken endpoint lookups) already carry the id in their payloads — the header exists for scans that need attribution, like the AI-hook flows. Plan: gate it to those scan modes and drop it from generic/CI secret scans. @pbeslin-gg could you confirm which scan modes the server-side attribution actually consumes, so we gate on the right set? On "should the raw id be sent at all": that's the HMAC/keyed-hash decision tracked separately — it needs client+server lockstep and should be settled before First-Class Partitions freezes the id into partition keys.

ggshield/core/machine_id.py
benjaminrigaud-gg: Done in 08b1320c — list aligned with osquery's `kPlaceholderHardwareUUIDList` plus the `FEFEFEFE…` Windows placeholder from osquery#8887, mirrored in the satori MR.

ggshield/core/machine_id.py
benjaminrigaud-gg: Partially — scope matters. On any machine with a derivable hardware id (all physical machines, SMBIOS on Windows), root and user derive the same id; divergence only exists on fallback machines (no hardware id: mostly containers and placeholder-firmware clones). Two changes address it: in this PR, the trust gate now lets runs sharing a `$HOME` adopt each other's id (trusted owners: current user, root, home owner); and END-767 tracks a machine-wide fallback file (`/var/lib/ggshield/machine_id`, root-owned) so root services and users converge too — scheduled ahead of First-Class Partitions for exactly the partition-key reason you raise. One extra wrinkle from the satori review: satori strips `$HOME` under elevation, so ggshield will align by resolving the cache home via `getpwuid` when elevated (proposal pending on the satori MR).

PR #1349, opened 2026-07-09
[AI] Improve server format support

paulpetit-gg-ext:
## Context

In some cases, (see https://cursor.com/docs/reference/plugins#optional-fields), the `mcpServers` field can be a string or a list. Support these cases.


.claude/settings.json
amascia-gg, line 1: do we want to include these files?

paulpetit-gg-ext: woopsie, bad rebase :facepalm:

pbeslin-gg:
Changed to support list/path config formats for all agents, since it's not specific to Cursor.

amascia-gg (approved):

pbeslin-gg:
@amascia-gg Small changes if you want to re-check, otherwise I'll squash the fixup and merge later today https://github.com/GitGuardian/ggshield/pull/1349/changes/94894036a9bc09e0e7072d3231099419434aad6e
- Avoid crashes on files we can't read
- Remove support for nested lists in config, which are documented for none of the agents.  Simplifies a bit the process, avoids questions about 1000-nested lists possibly crashing ggshield

PR #1355, opened 2026-07-22
fix(ai-hooks): use native osascript for macOS secret notifications

amascia-gg:
Fixes [NHI-1681](https://linear.app/gitguardian/issue/NHI-1681).

On macOS the AI hook's PostToolUse notification crashed the **whole hook**, so the "Secrets detected" block was never emitted. notifypy's bundled notifier is stripped on Homebrew and Intel-only on Apple Silicon, and the `Notify()` constructor wasn't guarded.

**Changes:**
- Guard the entire notification step so a notifier failure can never crash the hook.
- Deliver via native `osascript` on macOS (no bundled binary, arm64-native, injection/unicode-safe argv, timeout-bounded). Keep notifypy for Linux/Windows.


pbeslin-gg (approved):
Still works on my Linux

PR #1356, opened 2026-07-22
fix(ai-hooks): resolve ggshield path for bare-name install (NHI-1842)

amascia-gg:
## Summary

`ggshield install` for AI hooks pinned the hook command to a non-existent binary path when ggshield was invoked by bare name (`ggshield`, the normal pip/pipx/Homebrew case). `sys.argv[0]` is **not** rewritten to an absolute path by the shell, so `os.path.abspath` resolved it against the current directory — the installed hooks then silently never ran.

Fixes NHI-1842. Resolve the running ggshield against the install-time PATH (or `sys.executable` for the frozen standalone packages) so the hook command always points at a real binary.

## Behavior

| Invocation | `sys.frozen` | Before | After |
|---|---|---|---|
| Bare `ggshield` (pip/pipx/Homebrew) | F | `$CWD/ggshield` — dead path | `which()` → `/opt/homebrew/bin/ggshield` |
| Explicit `/opt/homebrew/bin/ggshield` | F | absolutized | unchanged |
| `./ggshield` / `uv run` venv path | F | absolutized | unchanged |
| Standalone `.pkg`/`.deb`/`.rpm` | T | `abspath(argv0)` (fragile) | `sys.executable` |
| Bare name, not on PATH | F | `$CWD/ggshield` | falls back to `abspath` (no regression) |

Preserves the two deliberate properties of the original pinning: symlinks are not resolved (the stable launcher path survives upgrades), and frozen mode never consults PATH (so it can't grab a different install).

## Tests

Rewrote the test that encoded the false assumption (*"a bare `ggshield` is resolved to its absolute path by the shell"* — the root cause that let this ship green) and added coverage for bare-name-via-PATH, PATH-miss fallback, and frozen-uses-`sys.executable`. Full AI-vertical suite (403 tests) passes.

### DIFF
--- a/changelog.d/20260722_120000_achille.mascia_fix_ai_hook_bare_name_path.md
+++ b/changelog.d/20260722_120000_achille.mascia_fix_ai_hook_bare_name_path.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- `ggshield install` for AI hooks no longer pins the hook to a non-existent
+  binary path. When ggshield was invoked by bare name (`ggshield`, the normal
+  pip/pipx/Homebrew case), `sys.argv[0]` is not rewritten by the shell, so the
+  hook command was resolved against the current directory and pointed nowhere —
+  the installed hooks then silently never ran. The running ggshield is now
+  resolved against the install-time `PATH` (or `sys.executable` for the frozen
+  standalone packages), so the hook command always points at a real binary.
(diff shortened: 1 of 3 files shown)


benjaminrigaud-gg:
Could we have shorter comments? :bow:

benjaminrigaud-gg (approved):

PR #1357, opened 2026-07-22
fix(ai-hooks): scan MCP tool inputs for secrets (NHI-1845)

amascia-gg:
## Summary

On a `PreToolUse` hook event, ggshield picks the content to secret-scan by tool: `Bash` → the command, `Read` → the file. For an **MCP tool call** (`tool_name` like `mcp__github__create_issue`) neither branch matched, so the scanned content stayed empty and the payload was allowed without any secret scan. A secret passed as an MCP tool argument was sent to the (potentially external) MCP server unscanned.

Fixes NHI-1845 (finding B9 in Eric's AI hooks test report, ggshield 1.52.2).

## Fix

In `parse_hook_input`, when the tool is `Tool.MCP`, set `content = json.dumps(tool_input)` — mirroring the existing `PostToolUse` `tool_output` serialization — so the arguments are scanned. If a secret is found, the call is blocked (`deny`) before it fires; otherwise MCP-activity governance runs as before.

## Scope

Deliberately limited to MCP. `Tool.OTHER`/`Write`/`Edit` inputs remain unscanned on `PreToolUse` — broadening would add an API scan on every tool call (latency) and risk new false-positive blocks. That's a separate decision.

## Tests

Added `test_claude_pre_tool_use_mcp_scans_tool_input` (payload is non-empty and the secret-bearing argument reaches the scannable). Full AI-vertical suite (397 tests) passes; black + flake8 clean.

### DIFF
--- a/ggshield/verticals/ai/hooks.py
+++ b/ggshield/verticals/ai/hooks.py
@@ -153,6 +153,12 @@ def parse_hook_input(raw_content: str) -> list[HookPayload]:
         elif tool == Tool.READ:
             # We only need to deal with the identifier, the content will be read by the Scannable
             identifier = lookup(tool_input, ["file_path", "filePath", "path"], "")
+        elif tool_input:
+            # MCP and unrecognized tool arguments can carry secrets bound for
+            # potentially external servers. Scan them, like tool_output below.
+            # Also covers agents whose MCP tools are only identified later, in
+            # post_process_payload (e.g. Copilot).
+            content = json.dumps(tool_input)
 
     elif event_type == EventType.POST_TOOL_USE:
         tool = _parse_tool(data)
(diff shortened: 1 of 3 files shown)


ggshield/verticals/ai/hooks.py
```diff
@@ -153,6 +153,11 @@ def parse_hook_input(raw_content: str) -> list[HookPayload]:
         elif tool == Tool.READ:
             # We only need to deal with the identifier, the content will be read by the Scannable
             identifier = lookup(tool_input, ["file_path", "filePath", "path"], "")
+        elif tool == Tool.MCP:
```
pbeslin-gg: do you think this should be a "else" (= we pass the whole tool_input for unknown tools)? The only issue I see is having inputs too large, which could impact the scan

pbeslin-gg (dismissed):

pbeslin-gg (dismissed):

pbeslin-gg (approved):

ggshield/verticals/ai/hooks.py
amascia-gg: agreed, let's try that 👍

PR #1358, opened 2026-07-23
fix(auth): serialize keyring reads to survive concurrent hook runs

amascia-gg:
## Problem

The AI hook fails intermittently with `ggshield could not authenticate to GitGuardian — this action was NOT scanned for secrets`, even when `ggshield api-status` is healthy in a terminal. It shows up during **parallel** agent activity: sub-agent fan-outs, parallel tool calls, or several agents at once.

## Root cause

Every hook process reads the GitGuardian token from the OS keyring. On macOS, securityd rejects most **concurrent** reads of the login Keychain, so under a fan-out the majority of reads fail with an auth error, and each hook then fails open and skips scanning. It's not the sandbox and not a missing Keychain grant, it reproduces on a machine where the grant is in place:

```
serial (10)        -> 10 OK,  0 fail
5 concurrent       ->  2 OK,  3 fail
15 concurrent      ->  6 OK,  9 fail
30 concurrent      -> 12 OK, 18 fail   (brew build, no fix)
```

Because the hook fails **open and silent**, affected tool calls are simply not scanned and nobody sees an error, which is the real risk at scale.

## Fix

Serialize keyring access across ggshield processes with an advisory file lock (`fcntl.flock`), applied to `get_token`, `store_token`, `delete_token`, and the availability probe in `KeyringTokenStore`. The lock is held only for the keychain call itself, never the network scan, so throughput is unaffected. It falls back to running unlocked after a timeout (so a stuck holder can't hang a hook), is a no-op on Windows (no `flock`), and never engages on the file-based fallback (no keyring in use).

## Verification

Same 30-way burst with the fixed build:

```
30 concurrent (fixed) -> 30 OK, 0 fail   (same wall-clock as before)
15 concurrent (fixed) -> 15 OK, 0 fail
```

- New regression test `TestKeyringConcurrency::test_get_token_is_serialized` asserts concurrent reads are serialized (fails without the lock: 20 overlapping reads; passes with it).
- Full `tests/unit/core/config/test_token_store.py` suite: 25 passing.

## Notes on alternatives

- Retry-with-jitter alone was only probabilistic (4/30 still failed after 4 retries) and adds more latency under load than the lock.
- Caching the token to a file would put a plaintext credential on disk, undoing the 1.50 keyring migration.

### DIFF
--- a/ggshield/core/config/token_store.py
+++ b/ggshield/core/config/token_store.py
@@ -1,11 +1,14 @@
+import contextlib
 import logging
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Iterator, Optional
 
+import filelock
 import keyring
 import keyring.backends.fail
 import keyring.errors
 
+from ggshield.core.dirs import get_cache_dir
 from ggshield.utils.os import getenv_bool
 
 
@@ -13,6 +16,35 @@ logger = logging.getLogger(__name__)
 
 KEYRING_SERVICE = "ggshield"
 KEYRING_SENTINEL = "__KEYRING__"
+
+# Concurrent reads of the OS credential store can fail (macOS securityd rejects
+# most simultaneous reads of the login Keychain), so fetching the token in
+# parallel loses reads to spurious auth errors. Serialize keyring access with a
+# file lock, held only for the keychain call, never the network scan.
+_KEYRING_LOCK_TIMEOUT = 10.0
+
+
+@contextlib.contextmanager
+def _keyring_lock() -> Iterator[None]:
+    """Serialize keyring access across ggshield processes.
+
+    Proceeds unlocked if the lock can't be acquired within the timeout, so a
+    stuck holder can never hang a caller.
+    """
+    cache_dir = get_cache_dir()
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    lock = filelock.FileLock(
+        str(cache_dir / "keyring.lock"), timeout=_KEYRING_LOCK_TIMEOUT
+    )
+    try:
+        lock.acquire()
+    except filelock.Timeout:
+        yield
+        return
+    try:
+        yield
+    finally:
+        lock.release()
 
 
 class TokenStore(ABC):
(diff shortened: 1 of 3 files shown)


ggshield/core/config/token_store.py
```diff
@@ -1,19 +1,72 @@
...
 
 KEYRING_SERVICE = "ggshield"
 KEYRING_SENTINEL = "__KEYRING__"
 
+# macOS securityd rejects most *concurrent* reads of the login Keychain, so a
+# fan-out of hook processes (parallel agents/sub-agents) loses the majority of
+# its token reads to spurious auth errors and silently skips scanning. An
+# advisory file lock serializes keyring access across our processes; each holder
+# only holds it for the keychain call itself (never the network scan), so the
+# queue drains quickly.
```
pbeslin-gg: nit: probably should be made shorter and more generic (parallel agents is a case where the problem occurs, but the fix applies to anything that fetch tokens in parallel)

ggshield/core/config/token_store.py
```diff
@@ -1,19 +1,72 @@
...
+                break
+            except OSError:
+                if time.monotonic() >= deadline:
+                    break
+                time.sleep(0.05)
+        try:
+            yield
+        finally:
+            if acquired:
+                fcntl.flock(handle, fcntl.LOCK_UN)
```
pbeslin-gg: The project already uses `filelock`, should we use it instead?

ggshield/core/config/token_store.py
amascia-gg: Good call, reworded to be generic about parallel token fetches rather than agent-specific.

ggshield/core/config/token_store.py
amascia-gg: Done, switched to `filelock.FileLock` (matching its use in hooks.py). Bonus: it's cross-platform, so the lock now also covers Windows and I dropped the fcntl no-op and the Windows-only test skip. Re-verified: 30/30 concurrent hooks pass, same wall time.

ggshield/core/config/token_store.py
```diff
@@ -14,6 +17,35 @@
...
+def _keyring_lock() -> Iterator[None]:
+    """Serialize keyring access across ggshield processes.
+
+    Proceeds unlocked if the lock can't be acquired within the timeout, so a
+    stuck holder can never hang a caller.
+    """
+    cache_dir = get_cache_dir()
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    lock = filelock.FileLock(
+        str(cache_dir / "keyring.lock"), timeout=_KEYRING_LOCK_TIMEOUT
```
pbeslin-gg, line 37: nit: FileLock seems to accept a Path as arg, so str() shouldn't be necessary

pbeslin-gg (approved):

gmendy1 (approved):

PR #1359, opened 2026-07-23
Pbeslin/ai discover scope checks

pbeslin-gg:
## What has been done

1. Add `ai-discover:send` to `ggshield machine doctor` checks.

## PR check list

### DIFF
--- a/changelog.d/20260723_000000_paul.beslin.ext_ai_discover_scope_checks.md
+++ b/changelog.d/20260723_000000_paul.beslin.ext_ai_discover_scope_checks.md
@@ -1,0 +1,5 @@
+### Changed
+
+- `ggshield machine doctor` now also verifies the token carries the
+  `ai-discover:send` scope, so a passing doctor run guarantees AI agent
+  discovery telemetry can be uploaded.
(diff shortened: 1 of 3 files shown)


ggshield/cmd/ai/discover.py
```diff
@@ -83,24 +87,39 @@ def discover_cmd(
...
-        config = submit_ai_discovery(client, config)
-        save_discovery_cache(config)
-        if scan_history:
-            backfill_report = backfill_mcp_history(client, config)
-        if scan_activity:
-            activity_report = collect_agent_activity(
-                client, config.user, AGENTS.values()
+    if scope_warning is not None:
+        ui.display_warning(f"Skipping upload of AI discovery: {scope_warning}")
+    else:
```
amascia-gg: This means that it's the client which decides to not send the AI discovery to the server. I would instead drop the hardcoded pre-check and let the server's 403 be the authority (it already reports the exact scope it wants, nhi:send-inventory or ai-discover:send). That works against every backend version with zero coordination, and it barely changes 403 volume since a token that has the right scope never 403s.

pbeslin-gg: ok right, I reverted that part

gmendy1 (approved):
LGTM

PR #1372, opened 2026-07-30
fix: prevent install prompt on git availability check on macos

6d7a:
## Context

`ggshield` checks if git is available even when running commands that do not require git so that it can load environment variables from a .env file at a repository root. On MacOS, this check triggered a prompt to install Apple's Command Line Tools if git was not installed. This was because Apple inserts a shim in `/usr/bin/git`, that is linked to `libxcselect`, which will run the prompt if the respective tools aren't installed.

## What has been done

This commit adds a check for that shim on darwin systems. The check is conservative, so that in the worst case, a user without a git install will encounter the install prompt, but no user with a valid git install will be erroneously blocked.

## PR check list

### DIFF
--- a/ggshield/utils/git_shell.py
+++ b/ggshield/utils/git_shell.py
@@ -3,6 +3,7 @@ import os
 import platform
 import re
 import subprocess
+import sys
 from enum import Enum
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


ggshield/utils/git_shell.py
```diff
@@ -68,10 +69,62 @@ def is_git_available() -> bool:
...
+def _is_macos_clt_stub(git_path: str) -> bool:
+    """True if git_path is the xcrun stub and no Command Line Tools are behind it.
+
+    Running such a stub opens a GUI prompt asking to install them.
+    """
+    if sys.platform != "darwin" or _macos_developer_git_exists():
+        return False
+    try:
+        with open(git_path, "rb") as f:
+            return b"libxcselect" in f.read(256 * 1024)
```
GG-HH, line 106: Can you add a comment here to explain the rational of this ? It surprised me at first.

ggshield/utils/git_shell.py
```diff
@@ -68,10 +69,62 @@ def is_git_available() -> bool:
...
+    except OSError:
+        return False
+
+
 @lru_cache(None)
 def _get_git_path() -> str:
     git_path = which("git")
 
+    if git_path is not None and _is_macos_clt_stub(git_path):
+        stub_dir = os.path.dirname(os.path.abspath(git_path))
```
GG-HH, line 119: Same here, a small comment would help

ggshield/utils/git_shell.py
```diff
@@ -68,10 +69,62 @@ def is_git_available() -> bool:
...
+        path = os.pathsep.join(
+            entry
+            for entry in os.environ.get("PATH", "").split(os.pathsep)
+            if entry and os.path.abspath(entry) != stub_dir
+        )
+        git_path = which("git", path=path) if path else None
+        if git_path is None:
+            raise GitExecutableNotFound(
+                "git is the Xcode Command Line Tools stub and the tools are not "
+                "installed. Install them by running `xcode-select --install`."
```
GG-HH: We only care about git, correct ? Should we prompt the user to install git only ? (I know it is not possible with xcode-select)

GG-HH (approved):
Comment on comments, otherwise, it seems good to me!

ggshield/utils/git_shell.py
6d7a: Yes, good point, especially since this is somewhat of a hack. Added one.

ggshield/utils/git_shell.py
6d7a: Yes, updated the message.

PR #1376, opened 2026-08-03
perf(ai-hook): cache clean scan verdicts to skip duplicate API calls

amascia-gg:
## Context

The ai-hook pays the ~380 ms API round trip twice for identical bytes. For a Read, **both** `PreToolUse` and `PostToolUse` scan the same file from disk: `hooks.py` sets `identifier = file_path` at both events, and `HookPayload.scannable` turns that into `File(path=...)`, so `PostToolUse` discards its own `tool_response` and re-reads the file. The bytes are provably identical, so caching the verdict is a correct short-circuit rather than a heuristic.

The existing debounce (`has_already_been_seen`) doesn't catch this pair — it hashes payload `content`, which is `""` for Pre-Read and the JSON `tool_response` for Post-Read.

## What has been done

New `verticals/ai/cache.py`: after an unambiguously clean scan, store `sha256(scanned_bytes) → timestamp`; a fresh hit returns allow without calling the API. Keyed on the bytes actually sent (`payload.scannable.content`), 15 min TTL, 500 entries max.

Hardening, since a poisoned cache would wave secret-bearing content through:
- file created `0600`, parent dir `0700`
- refuses to read a group/other-writable file, and `os.fstat` on the open fd + `S_ISREG` check (no TOCTOU window, rejects symlinks)
- every failure path fails **closed** — scan anyway, never allow

`ResultProtocol` gains `ignored_secrets_count_by_kind` so we only cache when the API itself found nothing — not when ggshield filtered secrets out locally, which would otherwise survive an ignore-rule change as a stale clean verdict.

## Validation

- `tests/unit` → **2525 passed, 4 skipped**; new `test_cache.py` covers hit/miss/expiry, corrupt JSON, world-writable file ignored, and "never cache a non-clean scan"
- Benchmark, API leg simulated at 380 ms identically on both arms, 20 iterations, fresh file each: **772.6 ms → 395.1 ms** per Read Pre+Post pair, API legs **40 → 20**

Honest scope: this removes only the API leg. The ~440 ms startup and ~135 ms config cost are untouched and still paid twice, so a tool call goes from ~2 s to ~1.6 s, not ~1 s. Write is deliberately not helped — its Pre and Post payloads genuinely differ. No skip table and no minimum-size gate: both were prototyped and rejected (the skip table would have skipped Post-Write content that Pre never saw).

## PR check list

- [x] Tests included
- [x] Changelog entry (`changelog.d/`)

## Follow-up: the config is part of the key

Review feedback was that a verdict goes stale when the config changes. The literal case (removing an entry from the ignore list) is already safe — a verdict is only stored when `ignored_secrets_count_by_kind` is empty, so a clean answer that depended on a local ignore rule is never cached. The general point holds though, so the key now also covers the *effective* secret config.

In the key: `filename_only` (rewrites the filename sent to the API, while the key held the local one), `all_secrets` (moves locally-ignored breaks into `secrets` instead of the counter that gates caching) and `source_uuid` (switches the endpoint to `scan_and_create_incidents`). Out of the key: presentation-only settings (`show_secrets`, verbosity, output format) and the ignore settings, which cannot make a stored verdict stale and would only cost misses.

Behaviour: the key derivation changes, so entries written by earlier builds become misses. A miss means "scan", so this is fail-safe and needs no migration.

**#1381 (Rust port) must mirror this**, since it deliberately shares the same cache file and key derivation. The new derivation is:

```
sha256(instance \0 api_key \0 config \0 filename \0 content)
```

- separator: a single NUL byte (`\0`), field order exactly as above
- `config` = `all_secrets=<0|1>;filename_only=<0|1>;source_uuid=<uuid or empty string>`, in that order, `;` separated, `=` between name and value, booleans as `0`/`1`, `source_uuid` as its lowercase canonical hyphenated form or the empty string when unset
- the whole joined string is UTF-8 encoded with `surrogatepass` (WTF-8: lone surrogates encoded as-is, not replaced), then SHA-256, then lowercase hex
- the config is inlined as plain text, **not** pre-hashed

### DIFF
--- a/changelog.d/20260803_120000_achille.mascia_ai_hook_verdict_cache.md
+++ b/changelog.d/20260803_120000_achille.mascia_ai_hook_verdict_cache.md
@@ -1,0 +1,6 @@
+### Changed
+
+- AI hooks no longer call the API twice for the same document. An unambiguously clean
+  scan result is cached locally for 15 minutes, keyed on the exact document sent and on
+  the instance and token it was sent with, so a file read costs one API round trip
+  instead of two (`PreToolUse` and `PostToolUse` scan the same file).
--- a/ggshield/core/scan/scanner.py
+++ b/ggshield/core/scan/scanner.py
@@ -3,8 +3,8 @@ Protocols for SecretScanner and its results,
 so that other verticals can use the scanner if they are provided one.
 """
 
-from collections.abc import Sequence
-from typing import Iterable, Optional, Protocol
+from collections.abc import Mapping, Sequence
+from typing import Any, Iterable, Optional, Protocol
 
 from pygitguardian import GGClient
 from pygitguardian.models import Match
@@ -39,6 +39,12 @@ class SecretProtocol(Protocol):
 class ResultProtocol(Protocol):
     @property
     def secrets(self) -> Sequence[SecretProtocol]: ...
+
+    @property
+    def ignored_secrets_count_by_kind(self) -> Mapping[Any, int]:
+        """Secrets the API reported but ggshield filtered out locally. Non-empty
+        means `secrets` reflects the local configuration, not just the API."""
+        ...
 
 
 class ResultsProtocol(Protocol):
--- a/ggshield/core/scan/scanner.py
+++ b/ggshield/core/scan/scanner.py
@@ -9,6 +9,7 @@ from typing import Any, Iterable, Optional, Protocol
 from pygitguardian import GGClient
 from pygitguardian.models import Match
 
+from ggshield.core.config.user_config import SecretConfig
 from ggshield.core.scanner_ui import ScannerUI
 
 from . import Scannable
@@ -57,6 +58,9 @@ class ScannerProtocol(Protocol):
 
     @property
     def client(self) -> GGClient: ...
+
+    @property
+    def secret_config(self) -> SecretConfig: ...
 
     def scan(
         self,
--- a/ggshield/verticals/ai/cache.py
+++ b/ggshield/verticals/ai/cache.py
@@ -9,6 +9,7 @@ from typing import Dict, Optional, Tuple
 from marshmallow.exceptions import ValidationError
 from pygitguardian.models import AIDiscovery, MCPConfiguration, MCPServer
 
+from ggshield.core.config.user_config import SecretConfig
 from ggshield.core.dirs import get_cache_dir
 
 from .models import Scope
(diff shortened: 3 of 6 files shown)


pbeslin-gg (approved):
Too many lengthy comments IMO, but the code itself looks good. 

One small note is that the verdict becomes stale if the config changes, for instance through .gitguardian.yaml. Removing a secret from the config won't work for 15mn if it was just scanned. I don't see any situation where that might be an issue, but invalidating the cache on .gitguardian.yaml change would be great if easy to do. The same issue exists for remote config (ignored incidents on the dashboard, things like that), but we can't do much about it.

amascia-gg:
Thanks! Both points addressed.

**Comments trimmed** — `18872be5`, −50/+28 lines of comments, code untouched. Kept only the non-obvious parts (why the key covers instance and token, why a broken cache reads as empty, why only an unfiltered single-`Result` scan is cached).

**On staleness — the `.gitguardian.yaml` case you describe is already handled, but you're right about the general class.**

Your specific scenario doesn't happen: a verdict is only stored when `ignored_secrets_count_by_kind` is empty, and `compute_ignore_reason()` increments that for *every* local filtering path. So content whose "clean" verdict depends on an ignore rule is never cached in the first place — removing that rule takes effect on the very next call, not 15 minutes later. Pinned by `test_locally_filtered_verdict_is_not_cached`.

Where you're right is the broader point. The key is:

```python
verdict_key(instance, api_key, filename, content)
```

No config in it. So any option that changes *what we send* or *how we read the answer*, without changing the content or filename, can serve a stale verdict for up to 15 min. The concrete one I found is **`secret.filename_only`** — it rewrites the filename sent to the API (`secret_scanner.py:102`) but the key uses the local `scannable.filename`, so flipping it reuses the other setting's verdict.

**Rather than watching `.gitguardian.yaml` for changes, I'd fold a hash of the effective secret config into the key.** It closes the whole class in one line instead of one instance, and it also covers CLI flags and user-level config, which a file watcher would miss. Your remote-config point (dashboard exclusions, known incidents) stays out of reach either way — that's what the 15 min TTL is for.

Happy either way: I can add the config hash here, or merge as-is and follow up. Slight preference for a follow-up so this PR stays a single concern, but say the word.

pbeslin-gg:
Ok for the config handling, I missed the part where the number of ignored secrets is checked

amascia-gg:
Pushed 39329e2d: the effective secret config is now part of the verdict cache key (`filename_only`, `all_secrets`, `source_uuid` — the settings that change what we send or how we read the answer; presentation-only ones stay out to avoid pointless misses).

I went with a config hash rather than watching `.gitguardian.yaml` because it closes the whole class in one go — CLI flags and the user-level config change the same things and a file watcher would miss them — and because your literal case is already covered: a verdict is only stored when `ignored_secrets_count_by_kind` is empty, so a clean answer that depended on a local ignore rule is never cached (`test_locally_filtered_verdict_is_not_cached`).

Note for #1381: it shares this cache file, so it needs to mirror the new derivation — spelled out byte-for-byte in the PR description.

PR #1377, opened 2026-08-03
perf(cli): import only the invoked command at startup

amascia-gg:
## Context

Every `ggshield` invocation imports all 12 command groups (`__main__.py:13-24`) and runs plugin discovery, before doing any work. On the shipped PyInstaller binary that's ~440 ms of the ~954 ms the ai-hook takes — and the hook fires twice per agent tool call. It also costs every `secret scan pre-commit`, which runs on every `git commit`.

Complements #1354 (`perf(plugin): defer sigstore imports off the startup path`) — the sigstore stack is already deferred; this removes the command-group imports and the per-invocation plugin discovery that remain.

## What has been done

- `cmd/utils/lazy_group.py`: click's canonical `LazyGroup` — a `name → "module:attr"` map imported on lookup. `list_commands()` does not import.
- Plugin discovery deferred to when it's actually needed (command-name miss, or `--help`); `ContextObj.plugin_registry` stays populated lazily, and the `_GGSHIELD_COMPLETE` guard that keeps plugin warnings out of shell completion is retained.
- **`scripts/build-os-packages`**: lazy imports are strings, so PyInstaller's static analysis can't see them — and for the ai-hook a non-zero exit reads as *allow*, i.e. a silent fail-open. So the build gate is the load-bearing part of this PR:
  - `--collect-submodules ggshield`, so everything under the package is bundled regardless of how it's imported
  - `smoke_check_bundle()` no longer runs a hardcoded 4-command list — it **generates** a walk of the click tree and resolves every leaf command from the built bundle, so a command added later is covered without anyone updating a list
  - plus one real end-to-end check: a canned Claude-Code payload piped into `secret scan ai-hook` from the bundle, asserting exit 0 and parseable JSON

## Validation

- `tests/unit` → **2502 passed, 4 skipped**; new `test_lazy_commands.py` asserts every entry in the lazy map resolves to a click command (catches `"module:attr"` typos without needing a build)
- Startup, source venv, n=25/side alternated, isolated `GG_*` dirs, no network: **551 ms → 209 ms**
- The new gate logic was exercised verbatim against a shim (44 leaf commands resolved, ai-hook exit 0 with valid JSON) and correctly **failed** when a lazy target was deliberately corrupted

✅ **Verified on CI** ([run 30892251277](https://github.com/GitGuardian/ggshield/actions/runs/30892251277), head `76a3119`): `build_os_packages` green on all five targets — windows-2022, macos-15-intel, macos-14, ubuntu-22.04 (rockylinux, x86_64), ubuntu-24.04-arm — and every one logged `Resolved 44 leaf commands from the bundle` then `Bundle smoke check OK`. `--collect-submodules` costs no build time: the PyInstaller step matches `main` (ubuntu 27s vs 27s, windows 38s vs 38s, macos-14 27s vs 26s); the +15–35s in the `Build` step is the new 44-command smoke walk itself. Packaged artifacts grow **+21–46 KB (≤0.05%)** per target vs the three latest `main` runs.

Venv numbers are not frozen-binary numbers: on the binary, expect roughly 954 ms → ~730 ms per invocation (~460 ms per tool call), since the API and config legs are untouched.

## PR check list

- [x] Tests included
- [x] Changelog entry (`changelog.d/`)

### DIFF
--- a/ggshield/cmd/utils/lazy_group.py
+++ b/ggshield/cmd/utils/lazy_group.py
@@ -1,0 +1,202 @@
+"""Lazy click groups: subcommands are imported only when actually resolved.
+
+Every ggshield invocation used to import all 12 top-level command groups (and
+run plugin discovery), so `ggshield secret scan pre-commit` paid for `hmsl`,
+`honeytoken`, `auth`, ... Declaring subcommands as "module:attr" strings defers
+each import to the moment Click resolves that name.
+
+The strings are invisible to PyInstaller's static analysis: the frozen binary
+relies on `--collect-submodules ggshield` plus the generated command-tree smoke
+check in scripts/build-os-packages/build-os-packages to guarantee every command
+still resolves from the bundle.
+"""
+
+import importlib
+import os
+from typing import Any, Dict, List, Optional
+
+import click
+
+from ggshield.core import ui
+
+
+class LazyGroup(click.Group):
+    """A click Group whose subcommands can be declared as "module:attr" strings
+    and are imported only when resolved (canonical click lazy-loading pattern).
+
+    `list_commands()` deliberately does not import anything, so listing names
+    (shell completion) is free. `--help` still shows the short help of every
+    subcommand, because Click's help formatter calls `get_command()` for each
+    listed name -- help is the one slow path, which is fine.
+
+    Use this class for *nested* groups (e.g. `secret scan`), and
+    PluginAwareLazyGroup for top-level ones. Plugins contribute top-level
+    commands only, so a nested group has nothing to merge; making it
+    plugin-aware would just pay for plugin discovery (including wheel signature
+    verification) on `--help` and on every unknown subcommand, and would let a
+    plugin group that happens to share a nested group's name leak its
+    subcommands in there.
+    """
+
+    def __init__(
+        self,
+        *args: Any,
+        lazy_commands: Optional[Dict[str, str]] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.lazy_commands: Dict[str, str] = lazy_commands or {}
+
+    def list_commands(self, ctx: click.Context) -> List[str]:
+        return sorted(set(self.lazy_commands) | set(self.commands))
+
+    def get_command(self, ctx: click.Context, cmd_name: str) -> Optional[click.Command]:
+        cmd = super().get_command(ctx, cmd_name)
+        if cmd is None and cmd_name in self.lazy_commands:
+            cmd = self.load_lazy_command(cmd_name)
+        return cmd
+
+    def load_lazy_command(self, name: str) -> click.Command:
+        module_name, attr = self.lazy_commands[name].split(":")
+        cmd = getattr(importlib.import_module(module_name), attr)
+        # Cache it, so later lookups (and plugin merges) see it in self.commands
+        self.add_command(cmd, name)
+        return cmd
+
+
+def _merge_subcommands(
+    target: click.Group, source: click.Group, warnings: List[str]
+) -> None:
+    """Merge `source`'s subcommands into `target`, never overriding an existing
+    one (a plugin must not be able to silently replace a built-in command).
+    Lazy subcommands count as existing even though they are not imported yet.
+
+    Conflicts warn instead of raising: `source` comes from a third-party plugin
+    wheel resolved at runtime, so a conflict is not necessarily a ggshield bug
+    we would catch during dev, and an installed plugin must never be able to
+    make the whole CLI unusable."""
+    lazy: Dict[str, str] = getattr(target, "lazy_commands", {})
+    for sub_name, sub_cmd in source.commands.items():
+        if target.commands.get(sub_name) is sub_cmd:
+            # Already merged: the root group and `target` itself both merge the
+            # same plugin group, so the second pass must stay silent instead of
+            # reporting the command it just added as a conflict.
+            continue
+        if sub_name in target.commands or sub_name in lazy:
+            warnings.append(
+                f"Skipping plugin subcommand '{source.name} {sub_name}' because it "
+                "conflicts with an existing command"
+            )
+            continue
+        target.add_command(sub_cmd, sub_name)
+
+
+def add_or_merge_plugin_command(
+    root: click.Group, command: click.Command, warnings: List[str]
+) -> None:
+    """Add a plugin command to the CLI, merging into a built-in group on conflict.
+
+    A plugin contributes a top-level command. If its name is free it is added
+    as-is. If it collides with a built-in we used to skip it entirely, which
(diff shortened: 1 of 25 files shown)


ggshield/cmd/utils/lazy_group.py
```diff
@@ -0,0 +1,179 @@
...
+def _merge_subcommands(
+    target: click.Group, source: click.Group, warnings: List[str]
+) -> None:
+    """Merge `source`'s subcommands into `target`, never overriding an existing
+    one (a plugin must not be able to silently replace a built-in command).
+    Lazy subcommands count as existing even though they are not imported yet."""
+    lazy: Dict[str, str] = getattr(target, "lazy_commands", {})
+    for sub_name, sub_cmd in source.commands.items():
+        if sub_name in target.commands or sub_name in lazy:
+            warnings.append(
```
GG-HH, line 91: Should we raise in that case, I guess we would catch it during dev ?

ggshield/cmd/utils/lazy_group.py
```diff
@@ -0,0 +1,179 @@
...
+            continue
+        target.add_command(sub_cmd, sub_name)
+
+
+def add_or_merge_plugin_command(
+    root: click.Group, command: click.Command, warnings: List[str]
+) -> None:
+    """Add a plugin command to the CLI, merging into a built-in group on conflict.
+
+    A plugin normally contributes a brand-new top-level command. When its name
```
GG-HH: As stated in your comment, this is false and there is not really a norm

ggshield/cmd/utils/lazy_group.py
```diff
@@ -0,0 +1,179 @@
...
+        _merge_subcommands(existing, command, warnings)
+        return
+
+    warnings.append(
+        f"Skipping plugin command '{name}' because it conflicts with an "
+        "existing command"
+    )
+
+
+class PluginAwareLazyGroup(LazyGroup):
```
GG-HH, line 133: Do we need to make this distinction ? Could we have all groups being PluginLazyGroups or would that make perf worse  ?

ggshield/cmd/utils/lazy_group.py
```diff
@@ -0,0 +1,179 @@
...
+        # triggers list_commands(), so skip the warnings during shell completion
+        # to avoid leaking them into completion output.
+        if "_GGSHIELD_COMPLETE" in os.environ:
+            return
+        # A failed load leaves the command unregistered, so Click rejects it as
+        # unknown; warn here or the user never learns why it is missing.
+        for name, reason in registry.get_load_failures().items():
+            ui.display_warning(
+                "Plugin '%s' is enabled but failed to load: %s. Its commands are "
+                "unavailable; run `ggshield plugin list` for status." % (name, reason)
```
GG-HH: Could we use f strings ?

ggshield/cmd/utils/lazy_group.py
```diff
@@ -0,0 +1,179 @@
...
+                warnings.append("Skipping unnamed plugin command")
+            elif self.plugin_scope is None:
+                add_or_merge_plugin_command(self, cmd, warnings)
+            elif cmd.name == self.plugin_scope and isinstance(cmd, click.Group):
+                _merge_subcommands(self, cmd, warnings)
+
+        # cli.main() has not yet consumed _GGSHIELD_COMPLETE when completion
+        # triggers list_commands(), so skip the warnings during shell completion
+        # to avoid leaking them into completion output.
+        if "_GGSHIELD_COMPLETE" in os.environ:
```
GG-HH, line 206: What is the purpose if this value and which process publishes it ?

GG-HH (commented):
Very nice improvement! It also has a huge effect on the pre-commit! (even bigger than what you stated). using the binaries produced by the CI, I have the following results:
  secret scan ai-hook (unauthenticated, isolated config dir — isolates process bootstrap + import + plugin loading from network noise):
  - main: ~0.85s median
  - PR #1377: ~0.72s median
  - -14%

  secret scan pre-commit:
  - main: ~1.65s median
  - PR #1377: ~1s median
  - -40%

It is a shame that the code is a bit complex

ggshield/cmd/utils/lazy_group.py
amascia-gg: Kept as a warning, and I added a docstring saying why: `source` is a third-party plugin wheel resolved at runtime, so a conflict is not necessarily a ggshield bug we would catch in dev — it can appear when a plugin ships a new subcommand against an older ggshield. Raising would make an installed plugin able to render the whole CLI unusable, which felt like the wrong failure mode for a security tool. The conflict is still surfaced loudly via `ui.display_warning`.

Related fix while I was here: `_merge_subcommands()` now skips a subcommand it has already merged. The root group and a `plugin_scope` group both merge the same plugin group into the same object, so the second pass was reporting a command it had just added as a conflict — a warning claiming `machine scan` was dropped when it had in fact been registered. Reproduced with the real `machine_scan` plugin.

ggshield/cmd/utils/lazy_group.py
amascia-gg: You are right, that was wrong — removed. The docstring now just says "A plugin contributes a top-level command" and describes the two cases (name free → added as-is; collides with a built-in group → merge subcommands) without claiming a norm that does not exist.

ggshield/cmd/utils/lazy_group.py
amascia-gg: Good question — I added the reasoning to the `LazyGroup` docstring rather than leave it implicit. Two reasons not to make everything plugin-aware:

1. **Perf**, which is the point of the PR: plugins only contribute *top-level* commands, so a nested group like `secret scan` has nothing to merge. Making it plugin-aware would trigger plugin discovery — including wheel signature verification — on `--help` and on every unknown subcommand, i.e. re-adding cost on a path that currently pays none.
2. **Correctness**: a plugin group that happened to share a nested group name could leak its subcommands in there.

So the distinction is load-bearing, but it was undocumented — that is fixed.

ggshield/cmd/utils/lazy_group.py
amascia-gg: Done, converted to an f-string.

ggshield/cmd/utils/lazy_group.py
amascia-gg: Expanded the comment. `_GGSHIELD_COMPLETE` is Click's completion protocol variable (`_<PROG_NAME>_COMPLETE`). The publisher is the user's shell: the completion function Click generates — installed with e.g. `eval "$(_GGSHIELD_COMPLETE=bash_source ggshield)"` — re-invokes ggshield with it set to ask for candidates. `cli.main()` has not consumed it yet at the point completion triggers `list_commands()`, which is why it is checked directly from `os.environ` here. Without the guard these warnings would be parsed by the shell as completion candidates.

GG-HH (approved):

pbeslin-gg (approved):

PR #1379, opened 2026-08-03
ci: scan CI-built packages for vulnerabilities with Grype

GG-HH:
## Summary
- Scans the packages `build_release_assets.yml` builds with [Grype](https://github.com/anchore/grype), reporting directly in the `Scan built packages for vulnerabilities` job — both its log and its summary. Since `ci.yml` already calls `build_release_assets.yml` on every PR, this runs continuously with no extra scheduling or GitHub-Releases querying needed.
- `scripts/vulnscan/scan_packages.py`: one stdlib-only script does extract + scan + report for a directory of downloaded build artifacts. Lists every individual finding (package, component, installed/fixed version, CVE, severity), deduped by (package, component, CVE) — grype's binary classifier flags the same component through more than one file (e.g. Windows ships both `python3.dll` and `python310.dll`), which isn't meaningful to a reader of this report. Pure reporting, no severity gating.
- `.grype.yaml`: ignore-rule config, seeded with the 12 CVEs already present in the current release build (bundled Python 3.14.6 + Windows' bundled SQLite 3.50.4.0) so the scanner starts clean and only flags genuinely new findings going forward.
- `doc/dev/vulnerability-scanning.md`: short doc on how to add an ignore rule.

## Test plan
- [x] Verified end-to-end against real artifacts downloaded from this PR's own completed CI runs.
- [x] Verified the `.grype.yaml` ignore mechanism against real CVEs: grype respects it, `--show-suppressed` confirms matches are actively suppressed (not coincidentally absent), and the report goes from listing all 12 baseline CVEs to "No vulnerabilities found."
- [x] black/isort/ty/pyright/flake8/prettier/GitHub-workflow-schema all pass.
- [x] CI green on this PR, including the new scan job.

Refs: END-846


doc/dev/vulnerability-scanning.md
```diff
@@ -0,0 +1,31 @@
...
+
+Rules can also match on package name/version/type instead of (or in addition
+to) a CVE ID — see `grype config` for the full list of fields. Grype has no
+built-in "reason" field, so leave a YAML comment above the entry explaining
+why it's ignored (a false positive, an accepted risk, a fix already in
+flight, ...):
+
+```yaml
+ignore:
+  # False positive: we don't use the vulnerable code path. See END-XXX.
```
6d7a: nit: remove the internal issue prefix
 
```suggestion
  # False positive: we don't use the vulnerable code path.
```

.grype.yaml
```diff
@@ -0,0 +1,19 @@
+# Vulnerability ignore rules for the Grype scan run by build_release_assets.yml.
+# See doc/dev/vulnerability-scanning.md for how to add an entry here.
+ignore:
+  # Baseline: vulnerabilities already present in the bundled Python 3.14.6
+  # interpreter and Windows' bundled SQLite 3.50.4.0 as of END-846. None are
```
6d7a:
```suggestion
  # interpreter and Windows' bundled SQLite 3.50.4.0. None are
```

6d7a (approved):
Two nits only

PR #1381, opened 2026-08-04
native (Rust) ai-hook — ~320 ms → ~20 ms, no daemon

amascia-gg:
## Why

`ggshield secret scan ai-hook` fires **twice per agent tool call**, and most of each call is Python startup + config/keychain work — roughly **320 ms before a byte is scanned**. Detection is 100% server-side, so this reimplements only the transport in Rust: **~20 ms startup instead of ~320 ms**, with no background daemon.

## What

A Cargo workspace under `ggshield-hook-rs/` (crates `common`, `config`, `hook`) that reads the hook event, resolves config + token (including the OS keychain), calls `/v1/multiscan`, and prints the allow/deny JSON.

Covers all five agents (Claude Code, Cursor, Copilot, Codex, VSCode), `.gitguardian.yaml` (incl. legacy v1), read-range slicing (#1382), and the shared clean-verdict cache (#1376). Document-size / batch limits are read from `/v1/metadata`, matching the Python client. No changes to the Python source.

## Kept honest

`tests/equivalence.py` runs the Rust and Python hooks against the same inputs and diffs stdout, exit code, and the `/v1/multiscan` request body — currently **byte-identical**. CI enforces it two ways:

- `ai_hook_equivalence` fails the build on any drift (the ignore rules now exist in two languages).
- a cross-OS `rust` job builds, lints (`clippy -D warnings`) and tests the workspace on Linux, macOS and Windows.

## Not included

MCP policy and AI discovery (#1383/#1384) — this is the hook path only. Distribution (making `ggshield` use this binary) is handled in #1392, stacked on top.


rust/INTEGRATION.md
6d7a, line 1: Can we remove this entirely? It's progress documentation for an agent and contains some internals.

rust/common/src/python_json.rs
```diff
@@ -0,0 +1,123 @@
+//! `json.dumps()`-compatible serialization.
+//!
+//! Not cosmetic. Two places need Python's exact bytes:
```
6d7a: Do we really need this? Regarding 1. IMO we should rather address this in the equivalence tests, not in the serialization. Regarding 2 I may lack some context. I guess we use the `tool_input` hash as a cache key, but I'd say we can assume that the client (rust/python) will be stable over one session.

rust/config/src/config.rs
```diff
@@ -0,0 +1,673 @@
...
+}
+
+/// `get_config_dir()`. platformdirs' `user_config_dir("ggshield", "GitGuardian")`
+/// agrees with `dirs::config_dir()/ggshield` on macOS and Linux (both ignore the
+/// appauthor). It does *not* on Windows: platformdirs' `user_config_dir` is
+/// `user_data_dir`, which reads `CSIDL_LOCAL_APPDATA` and appends appauthor then
+/// appname (platformdirs/windows.py: `Windows.user_data_dir`, `_append_parts`),
+/// giving `%LOCALAPPDATA%\GitGuardian\ggshield` — Local, not the `%APPDATA%`
+/// Roaming that `dirs::config_dir()` returns. Reasoned from the source, not run
+/// (no Windows host here); see INTEGRATION.md.
```
6d7a: Can we refactor this comment. It persists model reasoning.

rust/config/src/config.rs
```diff
@@ -0,0 +1,673 @@
...
+    #[cfg(not(windows))]
+    let base = dirs::config_dir().map(|d| d.join("ggshield"));
+    base
+}
+
+/// `get_cache_dir()`. Same Windows caveat, plus platformdirs' cache is
+/// "opinionated": `user_cache_dir` appends a `Cache` subfolder
+/// (platformdirs/windows.py: `Windows.user_cache_dir`, `opinion_value="Cache"`),
+/// so the cache is `%LOCALAPPDATA%\GitGuardian\ggshield\Cache`. macOS/Linux keep
+/// `dirs::cache_dir()/ggshield`, which already matches platformdirs there.
```
6d7a: Same here. Reading this as a maintainer is very noisy. It's important to know what the directories resolve to, not how the model got to this solution.

rust/config/src/config.rs
```diff
@@ -0,0 +1,673 @@
...
+        candidates.push(cwd.join(".env"));
+        candidates.push(user_config::project_root().join(".env"));
+    }
+
+    for path in candidates {
+        let Ok(content) = std::fs::read_to_string(&path) else {
+            continue;
+        };
+        let mentions_gitguardian = content.lines().map(str::trim_start).any(|line| {
+            let line = line.strip_prefix("export ").unwrap_or(line);
```
6d7a: [We should](https://github.com/motdotla/dotenv/blob/2fc7eac8ad77cbd5f7814355c7fa352dbcf3c358/lib/main.js#L12) `trim` here.

rust/config/src/config.rs
```diff
@@ -0,0 +1,673 @@
...
+                path.display()
+            )));
+        }
+        // Python stops at the first .env it finds.
+        return Ok(());
+    }
+    Ok(())
+}
+
+/// `is_saas_netloc()`.
```
6d7a:
```suggestion
```

rust/common/src/error.rs
```diff
@@ -0,0 +1,98 @@
...
+        Error::Fail(format!(
+            "ggshield could not scan ({first_line}) — this action was NOT scanned \
+             for secrets. Run 'ggshield api-status' to diagnose."
+        ))
+    }
+
+    /// A config-loading failure. See the `Fatal` variant docs: these do not
+    /// fail open, because in Python they happen before the command runs.
+    pub fn fatal(message: impl Into<String>) -> Self {
+        Error::Fatal(message.into())
```
6d7a: Some upstream errors like `serde_yaml_ng`'s are pretty opaque to the user. We should wrap these like the other ones with a warning, that the payload wasn't scanned.

rust/config/src/config.rs
```diff
@@ -0,0 +1,673 @@
...
+    std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_secs() as i64)
+        .unwrap_or(0)
+}
+
+/// Days-from-civil algorithm (Howard Hinnant), for `YYYY-MM-DDTHH:MM:SS...`.
+/// Any timezone suffix is ignored: tokens are minted with hour-plus lifetimes,
+/// so a sub-day offset cannot flip the comparison in a way that matters.
+fn unix_seconds(ts: &str) -> Option<i64> {
```
6d7a: We should not implement this by hand. `chrono` is very reliable and very light compared to other stuff we have in our dependency tree.

rust/config/src/config.rs
```diff
@@ -0,0 +1,673 @@
...
+    // never called, and no write probe is performed — a write probe is what was
+    // popping blocking Keychain modals in the Python implementation (PR #1378).
+    let raw = security_framework::passwords::get_generic_password(KEYRING_SERVICE, instance_url)
+        .map_err(|_| Error::auth())?;
+    String::from_utf8(raw).map_err(|_| Error::auth())
+}
+
+#[cfg(target_os = "linux")]
+fn keyring_token(instance_url: &str) -> Result<String, Error> {
+    use std::collections::HashMap;
```
6d7a: nit: can we have all imports at the top of the module for consistency? It doesn't make a runtime difference like in python.

rust/config/src/config.rs
```diff
@@ -0,0 +1,673 @@
...
+
+// The keyring service name every backend stores under
+// (`keyring.set_password("ggshield", instance_url, token)`, token_store.py).
+// Read by the macOS, Linux and Windows credential-store code; unused only on
+// other platforms, which fall back to `unimplemented`.
+#[cfg_attr(
+    not(any(target_os = "macos", target_os = "linux", windows)),
+    allow(dead_code)
+)]
+pub const KEYRING_SERVICE: &str = "ggshield";
```
6d7a, line 45: This is one of my main concerns with shipping rust as a quick performance fix. As far as I can reproduce locally (MacOS), the rust binary will not be able to access the keychain entry without a second prompt, even if keychain access has been `always allow`ed for the python ggshield binary. Also, if we do not sign the rust binary for mac, this prompt will return on every ggshield update, because the rust binary's keychain identifier will change.

rust/config/src/user_config.rs
```diff
@@ -0,0 +1,552 @@
+//! `.gitguardian.yaml` loading. Mirrors `UserConfig.load()` (user_config.py)
```
6d7a: IMO we should deviate from the Python implementation in one edge case that [we already considered a corner case when adding it](https://github.com/GitGuardian/ggshield/pull/875#issuecomment-2058462602). If we decide not to handle conflicts between `_` and `-` in the same file, we can just use serde on structs. It would make this a bit easier to read and maintain.

rust/hook/src/api.rs
```diff
@@ -0,0 +1,508 @@
...
+    pub matches: Vec<Match>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Match {
+    #[serde(rename = "match")]
+    pub value: String,
+    // Needed for the ignore sha, which hashes "<match>,<match_type>" pairs
+    // sorted by match_type. The wire field is `type`.
+    #[serde(rename = "type", default)]
```
6d7a: This should not be optional with an empty `default` fallback.

rust/hook/src/api.rs
```diff
@@ -0,0 +1,508 @@
...
+    pub match_type: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct RawPolicyBreak {
+    #[serde(default)]
```
6d7a: Same here, as far as I can see, `policy` is expected to be defined.

rust/hook/src/api.rs
```diff
@@ -0,0 +1,508 @@
...
+///
+/// The fetch is lazy: callers only reach here once a check has already decided
+/// the default cannot answer (a document past 1 MiB, or a batch past 20), so the
+/// common scan keeps its single round trip.
+fn limits(config: &Config) -> Limits {
+    *config.limits.get_or_init(|| fetch_limits(config))
+}
+
+fn fetch_limits(config: &Config) -> Limits {
+    let default = Limits {
```
6d7a: Are these only local defaults or can we make this an `impl Default for Limits`?

rust/hook/src/main.rs
```diff
@@ -0,0 +1,770 @@
...
+        return Ok(found);
+    }
+
+    let max_documents = api::max_documents_per_scan(config, pending.len());
+    // Read after the doc-count check above, so a fetch it triggered (or a large
+    // document earlier) is reflected; otherwise this is the default.
+    let max_payload_size = api::max_payload_size(config);
+    for chunk in chunks(&pending, max_documents, max_payload_size) {
+        let documents: Vec<&api::Document> = chunk.iter().map(|item| &item.document).collect();
+        let results = api::multiscan(config, payloads[chunk[0].index].agent, &documents)?;
```
6d7a: You can directly pass the iterator here, that way you don't have to collect the document references only to pass them.

```suggestion
        let results = api::multiscan(config, payloads[chunk[0].index].agent, chunk.iter().map(|item| &item.document))?;
```

rust/hook/src/api.rs
```diff
@@ -0,0 +1,508 @@
...
+                &doc.filename
+            };
+            let filename: String = filename
+                .chars()
+                .rev()
+                .take(API_PATH_MAX_LENGTH)
+                .collect::<Vec<_>>()
+                .into_iter()
+                .rev()
+                .collect();
```
6d7a, line 324:
```suggestion
            let filename: String = filename
                .chars()
                .skip(filename.len().saturating_sub(API_PATH_MAX_LENGTH))
                .collect();
```

rust/hook/src/api.rs
```diff
@@ -0,0 +1,508 @@
...
+    fn parse(json: &str) -> Vec<Vec<Secret>> {
+        parse_with(json, &SecretConfig::default())
+    }
+
+    const AWS: &str = r#"[{"policy_breaks": [
+        {"policy": "Secrets detection", "type": "AWS Keys", "known_secret": true,
+         "matches": [{"match": "AKIAsomething", "type": "client_id"},
+                     {"match": "s3cr3t", "type": "client_secret"}]}]}]"#;
+
+    #[test]
```
6d7a, line 756: Could you generate GIVEN-WHEN-THEN docstrings like we use in python?

rust/hook/src/main.rs
```diff
@@ -0,0 +1,770 @@
...
+
+fn run(stdin_content: &str) -> Result<i32, Error> {
+    // Config first, as in Python, where client creation precedes the debounce:
+    // a missing token must fail open even for a duplicate payload.
+    let config = config::resolve()?;
+    // Client creation is where Python emits the two mandatory insecure warnings
+    // (`create_session` in core/client.py, via `ui.display_warning` → stderr).
+    // Reproduce them verbatim, once per invocation, on stderr. They are not part
+    // of the hook's stdout contract, so the verdict is unaffected.
+    warn_if_insecure(&config);
```
6d7a: nit: can we remove these comments?

rust/hook/src/main.rs
```diff
@@ -0,0 +1,770 @@
...
+use ggshield_config::config;
+
+use std::io::Read;
+use std::path::Path;
+
+use error::Error;
+use output::HookResult;
+use payload::{EventType, Payload, Tool};
+
+/// stdin is capped exactly as in `ai_hook.py`.
```
6d7a: Can you remove the agent's _"Trust me, everything is really like it is in Python"_ re-assertions? 😅 

I guess mentioning the required parity once in the module header is good, but elsewhere it's going to be annoying  when we extract common logic from this.

rust/hook/src/main.rs
```diff
@@ -0,0 +1,770 @@
...
+        }
+        // Skip a document the instance itself would reject. 1 MiB is only the
+        // fetch trigger: the real ceiling is the instance's
+        // `maximum_document_size` (raised on some deployments), which is only
+        // fetched once a document is actually over 1 MiB. So a large
+        // secret-bearing file on a raised-limit instance is now scanned instead
+        // of being allowed unscanned — `scanner_ui.on_skipped(...)`, and
+        // skipping is not a block.
+        if content.len() > api::MAXIMUM_DOCUMENT_SIZE
+            && content.len() > api::max_document_size(config)
```
6d7a: This ignores any instance limit that is lower than `api::MAXIMUM_DOCUMENT_SIZE`.

rust/hook/src/api.rs
```diff
@@ -0,0 +1,508 @@
...
+    limits(config).maximum_document_size
+}
+
+/// This instance's `maximum_documents_per_scan`, for a batch of `needed`
+/// documents. The default handles every batch up to 20 without a fetch; only a
+/// larger batch — which the default would 400 on, failing open — pays the round
+/// trip.
+pub fn max_documents_per_scan(config: &Config, needed: usize) -> usize {
+    if needed <= DEFAULT_MAX_DOCUMENTS_PER_SCAN {
+        return DEFAULT_MAX_DOCUMENTS_PER_SCAN;
```
6d7a: This ignores any instance limit lower than `DEFAULT_MAX_DOCUMENTS_PER_SCAN`.

rust/hook/src/main.rs
```diff
@@ -0,0 +1,770 @@
...
+            start = position;
+            size = 0;
+        }
+        size += document_size;
+    }
+    if start < pending.len() {
+        chunks.push(&pending[start..]);
+    }
+    chunks
+}
```
6d7a: The logic is very difficult to trace here. Could you consider using iterators here?
```suggestion
fn chunks(
    pending: &mut impl Iterator<Item = Pending>,
    config: &Config,
) -> Vec<Pending> {
    let mut remaining_allowed_bytes = api::max_payload_size(config);

    pending
        .take(api::max_documents_per_scan(config, 0))
        .take_while(move |pending| {
            match remaining_allowed_bytes.checked_sub(pending.document.content.len()) {
                Some(remaining) => {
                    remaining_allowed_bytes = remaining;
                    true
                }
                None => false,
            }
        })
        .collect()
}
```

rust/hook/src/main.rs
```diff
@@ -0,0 +1,770 @@
...
+///
+/// One event routinely carries several payloads — a prompt mentioning N files
+/// produces N+1, a `cat` command 2 — and each request costs a round trip, which
+/// on this path is the whole remaining cost. Returns the secrets found per
+/// payload, in payload order.
+fn scan_contents(
+    config: &config::Config,
+    payloads: &[Payload],
+) -> Result<Vec<Vec<api::Secret>>, Error> {
+    let mut found: Vec<Vec<api::Secret>> = payloads.iter().map(|_| Vec::new()).collect();
```
6d7a: Why do we need to return a vector of vectors? We are only interested in the found secrets and index of the first contaminated payload.

rust/hook/src/main.rs
```diff
@@ -0,0 +1,770 @@
...
+
+    let mut pending: Vec<Pending> = Vec::new();
+    for (index, payload) in payloads.iter().enumerate() {
```
6d7a: You can take ownership here. That way you don't need to clone the content in `scannable`.
```suggestion
    for (index, payload) in payloads.into_iter().enumerate() {
```

rust/hook/src/api.rs
```diff
@@ -0,0 +1,508 @@
...
+
+/// A `PolicyBreak` reduced to the fields the hook message actually reads.
+#[derive(Debug, Deserialize)]
+pub struct Secret {
+    // The wire field is `type`; `break_type` is only pygitguardian's
+    // internal name for it (PolicyBreakSchema, data_key="type").
+    #[serde(rename = "type")]
+    pub detector_display_name: String,
+    #[serde(default)]
+    pub validity: Option<String>,
```
6d7a: Would be cool to have an `enum` for this, we'll probably re-use it in the future.

rust/hook/src/message.rs
```diff
@@ -0,0 +1,442 @@
...
+    } else {
+        plural
+            .map(str::to_string)
+            .unwrap_or_else(|| format!("{name}s"))
+    }
+}
+
+/// `translate_validity()`. An unknown id is returned as-is: its text is more
+/// useful than a generic "Unknown".
+fn translate_validity(validity: Option<&str>) -> String {
```
6d7a: The only live location calling this immediately lowercases the result.

6d7a (commented):
Not entirely done, publishing because I'm afraid of my local github extension overwriting my pending comments 😅

rust/hook/src/message.rs
```diff
@@ -0,0 +1,442 @@
...
+{secret_lines}
+
+The command was not executed and the {secrets} {were} not leaked.
+
+> How to remediate
+
+  Since the {secrets} {were} detected before the command was run:
+  1. consider moving the {secrets} to a secrets manager.
+
+{false_positive_instructions}";
```
6d7a: How about using `include_str!` to load the template from a separate file. That would make this module easier to read.
```suggestion
const PRE_BASH_TEMPLATE: &str = include_str!("templates/pre_bash.in");
```

rust/hook/src/message.rs
```diff
@@ -0,0 +1,442 @@
...
+            out.push(*c);
+        }
+    }
+    out
+}
+
+/// `markdown_hard_breaks()`: agents render the message as markdown, where a
+/// single newline collapses into a space. Two trailing spaces keep the break.
+/// A line is only padded when it is non-empty *and* the next line is non-empty.
+pub fn markdown_hard_breaks(text: &str) -> String {
```
6d7a: Can't we just adjust the templates without dynamically reformatting them?

rust/hook/src/message.rs
```diff
@@ -0,0 +1,442 @@
...
+                format!("{line}  ")
+            } else {
+                (*line).to_string()
+            }
+        })
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+fn secret_lines(secrets: &[Secret], escape_markdown: bool) -> String {
```
6d7a: `escape_markdown` is always `true`.

rust/hook/src/message.rs
```diff
@@ -0,0 +1,442 @@
...
+}
+
+fn false_positive_block(count: usize) -> String {
+    let this_is = pluralize(
+        "this is a false positive",
+        count,
+        Some("these are false positives"),
+    );
+    format!("> If {this_is}, run:\n\n    ggshield secret ignore --last-found")
+}
```
6d7a: Couldn't these be templates, too? It's takes some time to grasp what's happening here, and making changes to the messages becomes significantly harder when they are dynamically built.

rust/hook/src/message.rs
```diff
@@ -0,0 +1,442 @@
...
+            Some(Tool::Read) => POST_READ_TEMPLATE,
+            _ => POST_OTHER_TEMPLATE,
+        },
+        // PreToolUse, and EventType::Other, share the "not leaked yet" wording.
+        _ => match payload.tool {
+            Some(Tool::Bash) => PRE_BASH_TEMPLATE,
+            Some(Tool::Read) => PRE_READ_TEMPLATE,
+            _ => PRE_OTHER_TEMPLATE,
+        },
+    };
```
6d7a: nit: matching a tuple can prevent nested match statements and make it a bit easier to read.

```suggestion
   let template = match (payload.event_type, payload.tool) {
        (EventType::UserPrompt, Some(Tool::Read)) => USER_PROMPT_TEMPLATE,
        (EventType::PostToolUse, Some(Tool::Bash)) => POST_BASH_TEMPLATE,
        (EventType::PostToolUse, Some(Tool::Read)) => POST_READ_TEMPLATE,
        (EventType::PostToolUse, _) => POST_OTHER_TEMPLATE,
        (_, Some(Tool::Bash)) => PRE_BASH_TEMPLATE,
        (_, Some(Tool::Read)) => PRE_READ_TEMPLATE,
        (_, _) => PRE_OTHER_TEMPLATE,
    };
```

rust/hook/src/payload.rs
```diff
@@ -0,0 +1,970 @@
...
+/// (agents/__init__.py). Detection takes the FIRST match, so the order is part of
+/// the contract.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Agent {
+    Claude,
+    Codex,
+    Copilot,
+    Cursor,
+    VsCode,
+}
```
6d7a: These don't need `Eq`.

rust/common/src/python_json.rs
amascia-gg: Removed it, this implied a change from byte identical diffs to semantic comparison which works as well

rust/config/src/config.rs
amascia-gg: Should be better now, with the change in the python code which adds other binaries to the ACL, tell me what you think of it!

rust/config/src/user_config.rs
amascia-gg: Applied your suggestion!

rust/hook/src/main.rs
amascia-gg: I used `iter_mut()` + `mem::take` because scan still needs `payloads[index]` afterwards for the output contract. The clone your were targeting is gone, can you check the new implem?

rust/hook/src/verdict_cache.rs
```diff
@@ -0,0 +1,391 @@
+//! Clean-verdict cache — `ggshield/verticals/ai/cache.py` (PR #1376).
```
6d7a: This is prone to get out of sync.
```suggestion
//! Clean-verdict cache
```

rust/config/src/config.rs
```diff
@@ -0,0 +1,1223 @@
...
+        limits.maximum_payload_size,
+    ]
+    .iter()
+    .all(|value| *value > 0)
+    .then_some(limits)
+}
+
+/// Remember the limits fetched for (`api_url`, `token`). Best effort: a lost
+/// write costs the next invocation a round trip, nothing more.
+pub fn store_instance_limits(api_url: &str, token: &str, limits: Limits) {
```
6d7a, line 188: The file read/writing can be generalized and shared between limit and verdict caches.

6d7a (commented):
Some final things. The ACL solution looks alright to me.

ggshield/core/config/token_store.py
```diff
@@ -47,6 +49,121 @@ def _keyring_lock() -> Iterator[None]:
         lock.release()
 
 
+# Names of the sibling binaries that must be able to read the token without a
+# Keychain prompt: the Python CLI, the native dispatcher and the native hook.
+_MACOS_TRUSTED_BINARY_NAMES = ("ggshield", "ggshield-py", "ggshield-hook")
```
6d7a: Why does the dispatcher need to read the keychain?

rust/hook/src/message.rs
amascia-gg: Adopted the flattening with the first two arms swapped: `(UserPrompt, Some(Tool::Read))` must map to `PRE_READ_TEMPLATE`, because a UserPrompt event carries a Read payload per file mentioned in the prompt and the secret is in that file, not the prompt

rust/common/src/error.rs
```diff
@@ -0,0 +1,124 @@
...
+        ))
+    }
+
+    /// A capability this binary knowingly does not implement. Same shape as the
+    /// warnings above, but the remediation is "use the Python hook", because
+    /// running with the wrong rules is worse than not running.
+    pub fn unimplemented(what: impl AsRef<str>) -> Self {
+        Error::Fail(format!(
+            "ggshield-hook does not implement {} — this action was NOT scanned for \
+             secrets. Point the hook back at `ggshield secret scan ai-hook` for full \
```
6d7a: This is exactly the path that we intercept in the rust storefront binary, though.

amascia-gg: Thanks for the comment, this actually revealed some improvements that were needed: namely, .env file parsing in Rust's side as it ignored it until now, as well as the issue with using "scan and create incident" endpoint in AI hooks (we don't want to use that endpoint for now in AI hooks to not create too much noise in the dashboard until we can measure the impact of AI hooks with the data in CH so we will only use the `/multiscan` endpoint). Both fixed and modified wording here

ggshield/core/config/token_store.py
amascia-gg: Actually it's not the dispatcher but when we will merge the next PR, we will have the Rust binary called `ggshield` with all its crates that will need to read the keychain. In particular it will be the `ggshield-hook` sub crate which will but it's bundled under the global `ggshield` name. So only ggshield and ggshield-py needs to be in here, removed ggshield-hook which is irrelevant

6d7a (approved):

PR #1382, opened 2026-08-04
fix(ai-hook): scan the range an agent reads, not the whole file

amascia-gg:
## Context

`ggshield secret scan ai-hook` scanned the whole file on a `Read`, even when the agent only read a slice of it. `parse_hook_input` pulled `file_path` out of `tool_input` and dropped everything else.

The reason that matters most is not scope or latency, it is that **over-scanning can mean scanning nothing**. `SecretScanner._start_scans` does:

```python
if scannable.is_longer_than(maximum_document_size):
    scanner_ui.on_skipped(scannable, f"content is over {maximum_document_size:,} bytes")
    continue
```

`maximum_document_size` defaults to `DOCUMENT_SIZE_THRESHOLD_BYTES` (1 MiB). It is not a truncation and not an error: the document is dropped before any API call, no `Result` comes back, `_scan_content` sees no secrets and returns `allow`. So a `Read` of 300 lines from a 2 MB file was allowed having been scanned zero times, where those 300 lines would have scanned fine.

## What has been done

- New `Agent.read_range(tool_input)` returning `(first, last)`, 1-based and inclusive. Default: `None` — the whole file.
- `HookPayload.read_range` + `HookPayload.scannable` slices the file when a range is known.
- Both `PreToolUse` and `PostToolUse` read the range from the same `tool_input`, so one read produces one identical document at both events.

**Ranges implemented:**

| Agent | Parameters | Basis |
|---|---|---|
| Claude Code | `offset` / `limit` | `Read`'s documented parameters: `offset` is the line to start from, `limit` a line count |
| VS Code | `startLine` / `endLine` | The `read_file` PreToolUse fixture already in `test_hooks.py` (`startLine: 1, endLine: 200`), 1-based inclusive |

**Whole file — and the three are not the same case:**

- **Codex** genuinely only reads whole files, by construction. The only Codex read path ggshield recognises is `_parse_command`, which fires on a command starting with `cat ` or `Get-Content ` — both read the entire file. A partial read (`sed -n '10,20p'`, `head -50`) is not classified as a `Tool.READ` at all: it stays a Bash payload whose command string is scanned, and the file is never opened. There is nothing to slice here.
- **Cursor** and **Copilot CLI** almost certainly *do* support ranged reads — Cursor's `Read` mirrors Claude's, and Copilot's `view` is a text-editor-style tool of the kind that usually takes a view range. I have no payload showing the parameter names or their convention (0- vs 1-based, count vs end index), and the fixtures in `test_hooks.py` carry none. So they keep scanning the whole file: **unchanged from today, no regression, but no improvement either.** Guessing the names would silently under-scan if wrong, which is the one outcome worth avoiding.

Adding either is a small follow-up once someone captures one ranged `preToolUse` payload from each — a fixture and a `read_range` override, nothing more.

The same rule applies inside an implemented agent: a missing `offset` means "from the top", a missing `limit` means "to the end", never "nothing". `line_slice` also takes one line of slack on each side, so being one off on an agent's 0/1-based convention can never turn into unscanned content.

**Interaction with #1376:** its verdict cache is keyed on `(instance, token, filename, content)` and depends on Pre and Post hashing identically for a `Read`. They still do — both build the same `File` from the same `file_path` and apply the same range — and `filename` is unchanged (`url=self.identifier`). Test `test_pre_and_post_scan_the_same_bytes` pins exactly that property. No conflict with the diff; the two merge in either order.

## Validation

`tests/unit/verticals/ai/test_hooks.py::TestReadRange`, notably:

- `test_oversized_file_scans_once_only_a_slice_is_read` — a 1.4 MB file driven through the real `SecretScanner._start_scans`: whole file → `on_skipped`, nothing sent; 500-line slice → chunk sent, not skipped.
- `test_secret_inside_the_range_still_blocks_at_pre` / `test_secret_outside_the_range_is_not_reported` — asserting on the bytes handed to the scanner.
- `test_no_range_still_scans_the_whole_file` — previous behaviour preserved exactly.
- One test per agent, including the three that fall back.

`pytest tests/unit`: 2514 passed, 4 skipped. `black`, `flake8`, `isort` clean.

## PR check list

### DIFF
--- a/ggshield/verticals/ai/hooks.py
+++ b/ggshield/verticals/ai/hooks.py
@@ -141,6 +141,7 @@ def parse_hook_input(raw_content: str) -> list[HookPayload]:
     identifier = ""
     content = ""
     tool = None
+    read_range = None
 
     # Extract the identifier and content based on the event type
     if event_type == EventType.USER_PROMPT:
@@ -163,6 +164,7 @@ def parse_hook_input(raw_content: str) -> list[HookPayload]:
         elif tool == Tool.READ:
             # We only need to deal with the identifier, the content will be read by the Scannable
             identifier = lookup(tool_input, ["file_path", "filePath", "path"], "")
+            read_range = agent.read_range(tool_input)
         elif tool_input:
             # MCP and unrecognized tool arguments can carry secrets bound for
             # potentially external servers. Scan them, like tool_output below.
@@ -177,9 +179,12 @@ def parse_hook_input(raw_content: str) -> list[HookPayload]:
         if isinstance(content, (dict, list)):
             content = json.dumps(content)
         if tool == Tool.READ:
-            identifier = lookup(
-                data.get("tool_input", {}), ["file_path", "filePath", "path"], ""
-            )
+            # Same tool_input as PreToolUse, hence the same range: both events
+            # scan the exact same bytes for one read, which the verdict cache
+            # (keyed on the document) relies on.
+            tool_input = data.get("tool_input", {})
+            identifier = lookup(tool_input, ["file_path", "filePath", "path"], "")
+            read_range = agent.read_range(tool_input)
 
     # If identifier was not set, hash the content
     if not identifier:
@@ -194,6 +199,7 @@ def parse_hook_input(raw_content: str) -> list[HookPayload]:
             agent=agent,
             raw=data,
             timestamp=timestamp,
+            read_range=read_range,
         )
     )
 
--- a/ggshield/verticals/ai/models.py
+++ b/ggshield/verticals/ai/models.py
@@ -5,8 +5,9 @@ from collections.abc import Iterable, Iterator
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from enum import Enum, auto
+from functools import cached_property
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Literal, Optional
+from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Literal, Optional, Tuple
 
 
 if TYPE_CHECKING:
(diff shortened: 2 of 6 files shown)


amascia-gg:
Merged this locally with #1376 as a check: two `TestReadRange` tests failed with `AttributeError: Mock object has no attribute 'base_uri'`. `_scanner_finding` builds its client with `MagicMock(spec=GGClient)`, and `base_uri`/`api_key` are annotation-only on `GGClient`, so `spec=` never creates them — #1376's verdict cache key reads both. Fixed here in caa34d1f so this branch is merge-order-independent; #1376 already sets them on its own helper, so no change is needed there.

ggshield/verticals/ai/hooks.py
```diff
@@ -175,9 +177,13 @@ def parse_hook_input(raw_content: str) -> list[HookPayload]:
         if isinstance(content, (dict, list)):
             content = json.dumps(content)
         if tool == Tool.READ:
-            identifier = lookup(
-                data.get("tool_input", {}), ["file_path", "filePath", "path"], ""
-            )
+            # Same tool_input as the PreToolUse event, so the same range: Pre
+            # and Post scan the exact same bytes for one read. Pre is the only
+            # event that can block, and the verdict cache is keyed on the
+            # document, so the two must not drift apart.
```
pbeslin-gg: Is that comment relevant here, or a leakage from #1376 ?

ggshield/verticals/ai/models.py
```diff
@@ -111,7 +136,26 @@ def scannable(self) -> Scannable:
...
+                            # those are this hook's business. It is also the
+                            # difference between scanning a slice and scanning
+                            # nothing: a document over the API's
+                            # maximum_document_size is silently skipped by
+                            # SecretScanner, so over-scanning a large file ends
+                            # up allowing an unscanned read.
+                            return StringScannable(
+                                url=self.identifier,
+                                content=line_slice(file.content, *self.read_range),
+                            )
```
pbeslin-gg, line 152: Good point raised by Claude: this path now reads the file, and `scannable` happens to be read twice -> the file is read twice

> scannable is an uncached property and _scan_content reads it twice — payload.empty at hooks.py:588, then scanner.scan([payload.scannable]) at :592. On main this was free (File.is_longer_than(0) answers from the byte size without reading); the PR turns it into two full reads. functools.cached_property, or one local in _scan_content, halves every number in the first table for zero risk.

amascia-gg: Good catch and fixed, scannable is a cached_property now.

ggshield/verticals/ai/hooks.py
amascia-gg: If `PostToolUse` worked out the range differently from `PreToolUse`, the two events would send different `content` and the `Read` pair would silently stop hitting the cache so I would say that it's related to this PR.

 Recomputing from the same `tool_input` is what prevents that.

pbeslin-gg (approved):

PR #1383, opened 2026-08-04
perf(ai-hook): cut MCP PreToolUse latency from ~1.2s to ~0.4s

amascia-gg:
## Context

An MCP `PreToolUse` event is the slowest thing the AI hook does: it pays the
secret scan, plus a filesystem walk, plus up to two more API round trips, all
strictly serially. **p50 goes from 1183 ms to 387 ms** (p90 1210 -> 388 ms).
Other events are untouched.

## What has been done

- **Discovery walk cached for an hour.** It inventories assistant config files,
  which change on the order of days, but it ran on every MCP tool call. The
  existing cache file's mtime is the TTL marker; when the discovery is unchanged
  we `touch()` it rather than rewrite it. The "submit only if changed" logic is
  unchanged, it just runs at most once an hour.
- **Scan and `log_mcp_activity` now overlap** on a worker thread instead of
  running one after the other. Non-MCP payloads are detected before any thread is
  spawned and keep the plain serial path. The scan verdict still wins over the
  MCP policy verdict, so the decision the user sees is identical.

**Approved behaviour change** (agreed with the MCP feature owner): an MCP tool
call is now logged even when the secret scan blocks it. Before, a block skipped
the activity entirely.

Two things found on the way, **not changed here**:
- `has_changed_from` never converges, so discovery is re-submitted on *every*
  MCP call, not just when it changed: `discover_ai_configuration` builds
  `ggshield...MCPConfiguration` while the cache round trip yields
  `pygitguardian.models.MCPConfiguration`, and dataclass `__eq__` compares
  classes. Identical `to_dict()`, still `!=`. That third round trip is why the
  before-number is 1183 ms and not ~800 ms. The TTL hides it for an hour; it is
  worth fixing properly.
- `mcp.py` fails open on every exception. Left as is.

## Validation

Benchmarked in-process against a localhost mock API (fixed delay, identical on
both arms), measuring `AIHookScanner._scan_payloads` on the Claude MCP
`PreToolUse` fixture from `tests/unit/verticals/ai/test_hooks.py`. n=25 per arm,
arms alternated.

| arm (mock delay 380 ms = prod p50) | p50 | p90 |
| --- | --- | --- |
| MCP PreToolUse, before | 1183 ms | 1210 ms |
| MCP PreToolUse, TTL only | 773 ms | 775 ms |
| MCP PreToolUse, after | **387 ms** | **388 ms** |
| non-MCP, before | 385 ms | 388 ms |
| non-MCP, after | 386 ms | 387 ms |

With the mock at ~0 ms, to isolate local cost: MCP 26 -> 2 ms, non-MCP 1 -> 1 ms.

`pytest tests/unit` : 2520 passed, 4 skipped. `black --check`, `flake8`,
`isort --check-only` clean on touched files. The new tests were run against the
pre-change code and fail there (including the one asserting the two calls
actually overlap).

## PR check list

### DIFF
--- a/ggshield/verticals/ai/cache.py
+++ b/ggshield/verticals/ai/cache.py
@@ -23,12 +23,21 @@ VERDICT_CACHE_FILENAME = "ai_hook_verdicts.json"
 VERDICT_CACHE_TTL_SECONDS = 15 * 60
 VERDICT_CACHE_MAX_ENTRIES = 500
 
+# The walk inventories assistant configuration files, which change on the order of
+# days, but it runs on *every* MCP tool call. One hour keeps staleness well under a
+# coding session while making the walk negligible.
+DISCOVERY_CACHE_TTL_SECONDS = 3600
+
+
+def _discovery_cache_path() -> Path:
+    return get_cache_dir() / AI_DISCOVERY_CACHE_FILENAME
+
 
 def save_discovery_cache(config: AIDiscovery) -> None:
     """
     Save probe results to cache.
     """
-    cache_path = get_cache_dir() / AI_DISCOVERY_CACHE_FILENAME
+    cache_path = _discovery_cache_path()
     try:
         cache_path.parent.mkdir(parents=True, exist_ok=True)
         cache_path.write_text(json.dumps(config.to_dict(), indent=4))
@@ -41,7 +50,7 @@ def load_discovery_cache() -> Optional[AIDiscovery]:
 
     Returns None if the cache does not exist.
     """
-    cache_path = get_cache_dir() / AI_DISCOVERY_CACHE_FILENAME
+    cache_path = _discovery_cache_path()
     if not cache_path.exists():
         return None
     try:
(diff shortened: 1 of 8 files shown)


ggshield/verticals/ai/cache.py
```diff
@@ -23,12 +23,21 @@
...
 VERDICT_CACHE_MAX_ENTRIES = 500
 
+# The walk inventories assistant configuration files, which change on the order of
+# days, but it runs on *every* MCP tool call. One hour keeps staleness well under a
+# coding session while making the walk negligible.
+DISCOVERY_CACHE_TTL_SECONDS = 3600
+
+
+def _cache_path() -> Path:
+    return get_cache_dir() / AI_DISCOVERY_CACHE_FILENAME
```
pbeslin-gg: will need a more specific name since some other cache paths were added in other PRs

pbeslin-gg (approved):

PR #1385, opened 2026-08-04
perf(ai-hook): scan an event's payloads in one API call

amascia-gg:
## Context

A prompt mentioning three files costs **4 serial API round trips**: `_parse_user_prompt`
adds a `Tool.READ` payload per file, `_parse_command` one for `cat`/`Get-Content`, plus
the base payload — each scanned on its own. This is the most latency-visible event (the
prompt scan is the wait before every turn), and neither the verdict cache (#1376) nor
read-range slicing (#1382) helps it. `SecretScanner` already chunks and batches for
`secret scan path`; the ai-hook was the only caller passing it a single-element list.

## What has been done

All the payloads of an event that still need scanning go to `SecretScanner` in one call.
Measured against a local mock at the production p50 latency (380 ms), n=30, arms
alternated:

| event | before p50 | after p50 | before p90 | after p90 |
|---|---|---|---|---|
| `UserPromptSubmit` mentioning 3 files | 1561 ms | **395 ms** | 1564 ms | **398 ms** |
| single-payload prompt | 389 ms | 391 ms | 392 ms | 392 ms |

Behaviour:

- **Attribution**: results are matched to their document by url, not by position — a
  batch comes back neither in send order nor once per document. The lookup answers once
  per url and a round sends at most one entry per url, so no document inherits another's
  verdict.
- **Blocking**: verdicts are applied one payload at a time, in order.
- **Verdict cache**: still per document. A cached document is left out of the batch; a
  document is remembered as clean only if its own result came back with nothing filtered
  out locally.
- Empty payloads never reach the API; over 20 payloads are chunked by `SecretScanner`,
  none dropped.
- **Documents sent** — the one behaviour change: a blocking event now transmits every
  payload, not just up to the first block. On a prompt mentioning 25 files where one
  holds a secret, `main` sends 4 documents, this sends 26.

The batch cap is a global preference (default 20) served through `/v1/metadata`; the
ai-hook builds its `SecretScanner` with `check_api_key=True`, so
`client.secret_scan_preferences` carries the instance's real value before `_start_scans`
chunks on it — an on-prem install that lowered it still gets conforming batches. Backend
`multi_content_scan` appends exactly one result per document, in document order.

### Interaction with #1383

#1383 has merged; this branch is rebased on top and the textual conflict in
`_scan_payloads` is resolved so both wins survive. The loop submits the MCP activity
calls to the executor, runs the one batched scan alongside them, then applies verdicts
in payload order (scan verdict first). Non-MCP payloads get no activity call, and the
activity is still logged even when the scan blocks.

## Validation

End to end against a local ward instance (real server, not a mock), request counts from
the server's access log:

| payload | `main` | this PR |
|---|---|---|
| clean 3-file prompt | 4 requests x 1 doc | **1 request x 4 docs** |
| blocking 3-file prompt | 3 | **1** |
| 25-file prompt (26 docs) | 4 | **2** (20 + 6) |

- Attribution holds on the real server: with a GitHub token in a mentioned file and a
  Slack token in the prompt, the block reports the file's secret; a `cat` event names
  the file, not the command. Across two chunks completing out of order, the block still
  carries the right secret.
- The server answers `/v1/multiscan` with one result per document in request order,
  including two documents sharing a filename, and 400s at 21 documents. The client reads
  the cap from `/v1/metadata` (20) before every scan and chunks to it — no 400.
- Not verified: that a *lowered* on-prem cap is honoured.
- `pytest tests/unit` — 2610 passed, 11 new tests in `TestBatchedPayloadScan`. With
  `hooks.py` reverted, the new tests fail on the real signal (`Expected 'scan' to have
  been called once. Called 4 times.`, `assert 25 == 2`).

## PR check list

### DIFF
--- a/changelog.d/20260804_140000_achille.mascia_ai_hook_batch_payloads.md
+++ b/changelog.d/20260804_140000_achille.mascia_ai_hook_batch_payloads.md
@@ -1,0 +1,6 @@
+### Changed
+
+- AI hooks now scan all the payloads of an event in a single API call instead of one
+  call each. A prompt mentioning three files used to cost four round trips; measured
+  against a mock API at the production p50 latency, such a `UserPromptSubmit` goes from
+  1561 ms to 395 ms (p50). Single-payload events are unchanged.
--- a/changelog.d/20260805_100000_achille.mascia_ai_hook_prompt_file_message.md
+++ b/changelog.d/20260805_100000_achille.mascia_ai_hook_prompt_file_message.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- When a secret is found in a file mentioned in a prompt (`@path`), AI hooks now name
+  that file in the block message instead of reporting it as a secret "in your prompt".
+  The previous wording pointed at the wrong content and asked the user to edit a prompt
+  that did not contain the secret.
--- a/changelog.d/20260806_100000_achille.mascia_ai_hook_canonical_read_path.md
+++ b/changelog.d/20260806_100000_achille.mascia_ai_hook_canonical_read_path.md
@@ -1,0 +1,7 @@
+### Changed
+
+- AI hooks now resolve the file paths of `Read` tool calls and prompt `@`-mentions to an
+  absolute path against the event's working directory. A file mentioned in a prompt and the
+  same file later read by a tool now share one verdict-cache key, so it is scanned once
+  instead of twice. As a result, the file path reported in block messages (and sent for
+  scanning) is now absolute for prompt-mentioned files.
--- a/ggshield/core/scan/scanner.py
+++ b/ggshield/core/scan/scanner.py
@@ -40,9 +40,8 @@ class SecretProtocol(Protocol):
 class ResultProtocol(Protocol):
     @property
     def url(self) -> str:
-        """The url of the Scannable this result is about. A scan answers about
-        several documents at once, so this is what matches a result back to the
-        document it was produced for."""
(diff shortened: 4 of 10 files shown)


pbeslin-gg (approved):
Looks ok, one suggestion:

On some agents the file may be read by a tool invocation: I confirmed Copilot CLI does it (Claude Code doesn't, Codex is weird, didn't test Cursor and VSCode).
This means we scan the file twice: on UserPromptSubmit and PreToolUse.
But in PreToolUse we have an absolute path to the file, in UserPromptSubmit we have what the user wrote. File paths are different, resulting in two different verdict cache keys, the cache system is bypassed.
Could be great to make the filepaths uniform so that the cache is used in those cases (unless I missed something that already makes the paths absolute).

amascia-gg:
Good catch, your analysis is exactly right: `verdict_key` includes the filename, and the prompt gives a relative path (what the user typed) while the tool gives an absolute one, so the two produce different keys and the file is scanned twice (Copilot CLI does exactly this).

Fixed in 66c50b7: `Read` file paths are now canonicalized to an absolute path against the event's working directory (`cwd` for most agents, `workspace_roots` for Cursor), so the prompt mention and the tool read resolve to the same key and the second is served from the cache. It falls back to the raw path when no cwd is available, so the scan is never broken. One behavior change: the reported/sent filename is now absolute for prompt-mentioned files. Added a test that a `UserPromptSubmit @file` + `PreToolUse` Read of that file scans once (fails without the fix).

PR #1392, opened 2026-08-05
feat(packaging): one `ggshield` command — Rust dispatcher for the AI hook, Python for everything else

amascia-gg:
## What

Makes the bundled `ggshield` a small Rust dispatcher: it answers `secret scan ai-hook` natively (~20 ms) and hands every other command to the PyInstaller launcher, now renamed `ggshield-py`. Still one command on the `PATH`.

**Stacked on #1381 — do not merge first.**

```
/opt/gitguardian/ggshield-<version>/
├── _internal/   PyInstaller payload (unchanged)
├── ggshield     Rust dispatcher  ← what /usr/local/bin/ggshield points at
└── ggshield-py  the renamed Python launcher (internal, not on PATH)
```

- Dispatch is an **exact match** on `secret scan ai-hook`; anything else goes to Python. No option parsing here — the safe direction is always "hand it to Python".
- `exec` on Unix, spawn-and-wait on Windows; stdin/stdout/stderr and the exit code pass through unchanged. A missing/unresolvable `ggshield-py` fails loudly, never a silent success.
- Both binaries are signed/notarized (macOS and Windows); the macOS JIT entitlement is on `ggshield-py`.
- `build_hook_command()` writes the dispatcher path on a bundle install; pip / Homebrew / Docker (pure wheel) are untouched.

## Desktop notification

The banner raised once a secret has reached the agent now comes from a bundle we own: the GitGuardian icon instead of Script Editor's, `ggshield` as the app name, and a click that does nothing instead of opening Script Editor's file picker. `osacompile` builds it on the user's machine, so it is native on any architecture and nothing new needs signing or notarising.

Linux and Windows had no notifier in the port at all (`fn notify(_result) {}`), so they would have lost notifications entirely. They now get `notify-send` and a WinRT toast, both carrying the GitGuardian icon rather than the Python logo notifypy falls back to.

## Validation

- `cargo fmt` / `clippy -D warnings` clean; **127 Rust tests** including the delegation suite (argv / stdin-stdout / exit code / symlink / missing launcher) and the generated bundle's structure.
- Equivalence harness runs **through the dispatcher**: semantically equivalent to the Python hook (JSON compared as data).
- `test_installation.py` green; the release build exercised end-to-end with a stub PyInstaller (both binaries present, hook served through the bundle).

## PR check list

- [x] Tests (predicate, delegation, installation, notifier bundle)
- [x] Changelog entry


.github/workflows/build_release_assets.yml
```diff
@@ -134,6 +134,11 @@ jobs:
           tar xf nfpm.tar.gz nfpm
           cp nfpm /usr/local/bin
 
+          # Rust for the `ggshield` dispatcher is provisioned inside
+          # build-os-packages (build_dispatcher), where cargo actually runs, not
+          # here: doing it in this shared step perturbed the later `uv` step
+          # (uv: command not found).
+
```
6d7a: Please remove comments documenting the agent's reasoning.

rust/hook/Cargo.toml
```diff
@@ -5,9 +5,14 @@ edition.workspace = true
...
 
-[[bin]]
-name = "ggshield-hook"
-path = "src/main.rs"
+# A library: the dispatcher binary (../dispatcher) calls `run_hook()`. There is
+# deliberately no binary target here — the only `ggshield` on the PATH is the
+# dispatcher, and a second binary would double link time to produce an artifact
+# that must then be either signed and notarised or explicitly excluded from
+# those lists, for no user-visible gain. Tests and the equivalence harness
+# invoke the dispatcher with `secret scan ai-hook`, same as an agent does.
```
6d7a: Please remove model reasoning comments.

rust/hook/src/lib.rs
```diff
@@ -1,5 +1,874 @@
...
+//!
+//! Every verdict here has to mean the same thing as the Python one, so anything
+//! this binary cannot reproduce declines and fails open with a warning instead of
+//! guessing — see `config::check_no_dotenv_override()` and `user_config`.
+
 pub mod api;
 pub mod message;
 pub mod output;
 pub mod payload;
 pub mod verdict_cache;
```
6d7a: With `main.rs` gone, these do not need to be public.

doc/dev/os-packages.md
```diff
@@ -31,7 +31,26 @@ flowchart TD
...
+We use [PyInstaller](https://pyinstaller.org) to generate the standalone executable.
+
+### Two binaries, one command
+
+The `ggshield` the user runs is **not** the PyInstaller launcher. It is a small Rust dispatcher (`rust`), and the bundle contains both:
+
+| file          | what it is                                                                                                               |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------ |
+| `ggshield`    | the Rust dispatcher. The only entry point: what `/usr/local/bin/ggshield`, `/usr/bin/ggshield` and `%PATH%` point at.    |
+| `ggshield-py` | the PyInstaller launcher (what used to be called `ggshield`). Internal — never on `$PATH`, never run directly by a user. |
```
6d7a: When installing with `choco`, `ggshield-py` ends up on `$PATH`

rust/Cargo.toml
6d7a, line 9: We should keep this in sync in CI.

6d7a (commented):
This is a pet peeve of mine, but IMO the comments are going a bit overboard in this one. Some of them are just traces of model reasoning and many are redundant. What do you think of creating a dedicated document in `doc` that explains why we use the storefront binary and how it works plus brief summaries in the respective commits?

rust/hook/src/lib.rs
```diff
@@ -1,5 +1,874 @@
...
+/// Returns the secrets found and the index of the first payload holding them, or
+/// index 0 with no secrets — the caller only reads that payload for its output
+/// contract. One event routinely carries several payloads (a prompt mentioning N
+/// files produces N+1, a `cat` command 2) and each request costs a round trip.
+///
+/// `send_mcp_activity()`, which Python also calls per payload to ask the API
+/// whether an MCP tool call is allowed by policy, is not implemented. No warning
+/// either: it swallows every exception and returns "allowed" (mcp.py), so warning
+/// would be louder than Python is on its own failure path.
+fn scan_payloads(
```
clement-tourriere, line 218: The comment above is not exact: Python does not always fail open on MCP. `_scan_payloads` in hooks.py returns the MCP result when `mcp_result.block`, so when the API answers `allowed=false` Python **blocks** the MCP tool call. Here we skip `send_mcp_activity` completely, so on a bundle install the dashboard MCP policy is just not enforced (and the MCP telemetry disappears from the dashboard). Example: org denies `mcp__github__delete_repository`, the args have no secret, multiscan is clean, we allow. IMO this needs to be implemented, not skipped.

rust/hook/src/lib.rs
```diff
@@ -1,5 +1,874 @@
...
+/// The content is moved out of the payload rather than copied: a document can be
+/// a whole MiB, and nothing downstream reads it again.
+fn scannable(payload: &mut Payload) -> Option<(String, String)> {
+    if payload.tool == Some(Tool::Read) {
+        let path = Path::new(&payload.identifier);
+        if path.is_file() && !is_path_binary(path) {
+            // Python skips a file it cannot decode ("can't detect encoding")
+            // rather than scanning mojibake; a skip is not a block.
+            let content = std::fs::read(path).ok()?;
+            let content = String::from_utf8(content).ok()?;
```
clement-tourriere: `String::from_utf8` only accepts strict UTF-8, but Python decodes with charset_normalizer (UTF-16, latin-1, ... there are tests for UTF-32/CP1250). PowerShell `Out-File` writes UTF-16LE by default, so a very normal Windows secrets file fails here: `scannable()` returns None, the file is skipped, the read is allowed, no warning. The secret goes to the model unscanned. We need at least a BOM check + UTF-16 fallback.

rust/dispatcher/src/main.rs
```diff
@@ -0,0 +1,17 @@
...
+//! `dispatch` breaks every ggshield command, not just the hook. See
+//! `doc/dev/os-packages.md`.
+
+mod dispatch;
+
+fn main() {
+    // Dispatch before touching stdin: `exec()` hands the untouched descriptor to
+    // Python, and reading a byte here would eat the delegated command's input.
+    let args: Vec<std::ffi::OsString> = std::env::args_os().skip(1).collect();
+    if dispatch::is_native_hook(&args) {
```
clement-tourriere, line 13: When the native hook declines a config it can't handle (`.env` mentioning `GITGUARDIAN_*`, `secret.source_uuid`), it warns + allows, forever. But `ggshield-py` is right next to us and handles those cases fine — we just never delegate. So for those users this upgrade silently turns off the hook scanning, permanently. Worse, the error message tells them to point the hook back at `ggshield secret scan ai-hook`... which is exactly the command we intercept — circular advice. Suggestion: when the native hook returns the `unimplemented` decline, fall back to `dispatch::delegate()` instead of failing open.

rust/dispatcher/src/dispatch.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    // `exec()` only returns on failure.
+    let err = Command::new(launcher).args(args).exec();
+    broken_install(launcher, &err)
+}
+
+/// Windows has no `exec`, so spawn and wait.
+#[cfg(windows)]
+fn exec(launcher: &Path, args: &[OsString]) -> ! {
+    match Command::new(launcher).args(args).status() {
```
clement-tourriere, line 149: Ctrl-C problem: before, `ggshield.exe` WAS the Python process, so Ctrl-C was handled by Python (KeyboardInterrupt, cleanup, its exit code). Now CTRL_C_EVENT kills the dispatcher with the default handler while `ggshield-py` keeps running: the prompt comes back and the child still prints in the console, and a wrapping script sees the dispatcher's Ctrl-C code, not Python's. So "exit code propagated verbatim" is not true for interrupted commands. Need `SetConsoleCtrlHandler(NULL, TRUE)` (ignore in the parent, let the child handle it) like cmd.exe does with children.

scripts/build-os-packages/build-os-packages
```diff
@@ -221,9 +222,51 @@ step_build() {
...
+    # The Rocky 8 build container has neither rustup nor a C compiler (ring, the
+    # TLS backend, compiles C); macOS and Windows runners ship rustup. Not done
+    # in the CI's shared dependency step, because doing it there perturbed the
+    # later `uv` step (uv: command not found). --default-toolchain none because
+    # rust/rust-toolchain.toml pins the version.
+    if [ "$HUMAN_OS" = Linux ] ; then
+        command -v cc > /dev/null 2>&1 || yum install -y gcc
+    fi
+    if ! command -v cargo > /dev/null 2>&1 ; then
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
```
clement-tourriere: Supply chain: this downloads and runs an unpinned installer as root in the release container, before signing. Everything else in the release has pinned versions. Better: download a pinned `rustup-init` version + check its sha256. Also `cargo build` on line 254 should be `cargo build --locked` — we commit Cargo.lock but nothing enforces it in the release build.

clement-tourriere (commented):
Did a big review pass on this. Posting inline the most critical ones that live in THIS PR's diff. Big picture: every error in the native hook ends as "warn + allow", so each bug becomes a silent scan bypass.

Important: after the rebase, several criticals now live in #1381's diff, so I cannot anchor them here. Short list (can post them on #1381 if you want):

- **Windows absolute paths** (`payload.rs` `abs_read_path`): `C:\...` don't start with `/` so we build a garbage path, file never read → all pre-read scans dead on Windows. UNC paths too.
- **u64 overflow** (`payload.rs` read_range): `first + limit - 1` wraps in release with a huge `limit` (prompt-injectable) → empty slice → nothing scanned.
- **NUL bytes** (`api.rs`): pygitguardian replaces `\0` with `\x1a` before sending; we send verbatim → API 400 → whole event fails open.
- **No retry + dead 401 branch** (`api.rs`): one single POST, no retry (Python does 5 on 502/503/504) → one transient 503 = fail open. And ureq 3 makes every 4xx/5xx an `Err` by default so the 401→"auth login" message is dead code.
- **`filename_only` on Windows** (`api.rs`): basename split on `/` only → full local path sent to the API, exactly what the option must prevent.
- **Cleartext http** (`config.rs`): no scheme check, `http://` instance accepted → token + content in cleartext. Python refuses.
- **macOS keyring split-brain** (`token_store.py`): write goes direct to Apple Keychain, read/delete still use the configured backend → with `PYTHON_KEYRING_BACKEND` set, login "works" then nothing can authenticate anymore.
- **`api.create_cf` doesn't exist in keyring 24.x** (`token_store.py`): we allow `keyring>=24`, so the ACL silently never applies there (AttributeError eaten by the broad except).

rust/hook/src/lib.rs
amascia-gg: Right, only run_hook() is pub now

doc/dev/os-packages.md
amascia-gg: Good catch. We now drop a ggshield-py.exe.ignore marker next to it so Chocolatey skips shimming it. It was cosmetic rather than broken — Chocolatey's ggshield.exe shim resolves current_exe() back into the package directory, so the sibling lookup still works either way. Both bits are written up in doc/dev/os-packages.md.

rust/Cargo.toml
amascia-gg: Went a bit further than a CI check. hook/build.rs now reads `__version__` straight out of `ggshield/__init__.py` at build time, so there's nothing left to keep in sync

rust/hook/src/lib.rs
amascia-gg: Agreed, this needs implementing. Doing it in a separate PR stacked on this one, with a new ggshield-discovery crate.

rust/hook/src/lib.rs
amascia-gg: You're right, and it was worse than a skip: a skipped file is an allowed read, so the secret went to the model unscanned and the user was never told.

Now we honour the BOM — UTF-8, UTF-16 and UTF-32, both endians — and decode anything else lossily rather than giving up. Lossy is safe for this purpose since secrets are ASCII and ASCII survives byte for byte. What we still don't do is charset_normalizer's statistical guessing for BOM-less legacy encodings; those decode as UTF-8, which keeps their ASCII intact.

There's a test for the PowerShell Out-File case you described, plus one asserting a file of undecodable bytes keeps its ASCII instead of being skipped.

rust/dispatcher/src/main.rs
amascia-gg: Thanks for the comment, this issue was spotted as well on my agent's side so we implemented .env handling and the source_uuid issue. It should be inline with Python now

rust/dispatcher/src/dispatch.rs
amascia-gg: Good catch. Added SetConsoleCtrlHandler(NULL, TRUE) before spawning, so the child handles the event and we don't, same thing cmd.exe does.

scripts/build-os-packages/build-os-packages
amascia-gg: Both fixed, thanks!

changelog.d/20260805_170000_achille.mascia_ggshield_rust_dispatcher.md
```diff
@@ -0,0 +1,8 @@
+### Changed
+
+- In the standalone packages (`.pkg`, `.deb`, `.rpm`, `.msi`, Chocolatey, archives), `ggshield`
+  is now a small native dispatcher: it runs `ggshield secret scan ai-hook` — the AI
+  coding-assistant hook, which fires twice per agent tool call — in about 12 ms instead of
+  ~600 ms, and hands every other command to the bundled Python implementation. There is still
+  a single `ggshield` command, and every command behaves as before. pip, pipx and Homebrew
+  installs are unchanged.
```
6d7a: We will release the rust-based wheels in the same release, so we should omit the part about pip et al in this entry.
```suggestion
  a single `ggshield` command, and every command behaves as before.
```

changelog.d/20260805_170000_achille.mascia_ggshield_rust_dispatcher.md
```diff
@@ -0,0 +1,8 @@
+### Changed
+
+- In the standalone packages (`.pkg`, `.deb`, `.rpm`, `.msi`, Chocolatey, archives), `ggshield`
+  is now a small native dispatcher: it runs `ggshield secret scan ai-hook` — the AI
+  coding-assistant hook, which fires twice per agent tool call — in about 12 ms instead of
+  ~600 ms, and hands every other command to the bundled Python implementation. There is still
```
6d7a: We should keep numbers from local benchmarks out of the docs.

```suggestion
  is now a small native dispatcher: it runs `ggshield secret scan ai-hook` and hands every other command to the bundled Python implementation. There is still
```

doc/dev/os-packages.md
```diff
@@ -31,7 +31,30 @@ flowchart TD
...
+| `ggshield`    | the Rust dispatcher. The only entry point: what `/usr/local/bin/ggshield`, `/usr/bin/ggshield` and `%PATH%` point at.                                                                                                 |
+| `ggshield-py` | the PyInstaller launcher (what used to be called `ggshield`). Internal; not meant to be run directly. Kept off `$PATH` on Linux/macOS by the `/usr/bin/ggshield` symlink; on Windows it shares the install directory. |
+| `_internal/`  | PyInstaller payload, unchanged.                                                                                                                                                                                       |
+
+The dispatcher answers exactly one command itself, `ggshield secret scan ai-hook` — the AI coding-assistant hook, which fires twice per agent tool call and where Python startup dominates the runtime (~12 ms instead of ~600 ms). Every other invocation is `exec`'d (spawned, on Windows) to `ggshield-py` with the same arguments, stdin/stdout/stderr and exit code. It finds `ggshield-py` by name, as a sibling of its own **resolved** executable, so the `/usr/local/bin` symlink works.
```
6d7a: Same as above.
```suggestion
The dispatcher answers exactly one command itself, `ggshield secret scan ai-hook` — the AI coding-assistant hook, which fires twice per agent tool call and where Python startup dominates the runtime. Every other invocation is `exec`'d (spawned, on Windows) to `ggshield-py` with the same arguments, stdin/stdout/stderr and exit code. It finds `ggshield-py` by name, as a sibling of its own **resolved** executable, so the `/usr/local/bin` symlink works.
```

doc/dev/os-packages.md
6d7a, line 1: IMO these additions contain too much implementation detail and AI jargon. This doc is a high-level overview of the different distribution variants of ggshield. I think it is worth it to document the fact that we now have a rust binary as an entry point that calls the python implementation (either using the packaged python or the wheel). However, the individual responsibilities of the rust binary and the python package will most likely evolve pretty rapidly, so we should keep them out of the long-term docs.

ggshield/cmd/machine/setup.py
```diff
@@ -127,9 +129,37 @@ def _setup_ai_hooks(
...
+    until that prompt is answered the notification is delivered nowhere -- no
+    error, no banner. Asking here puts the dialog in front of someone who is
+    deliberately setting ggshield up. Left to the first real detection, it lands
+    on top of a leaked secret, and a prompt nobody answers is a leak alert
+    nobody sees.
+
+    Best effort, and silent about it: the notifier lives in the Rust dispatcher,
+    so on an install without one (pip, Homebrew) this command is not understood
+    and there is nothing to warm -- those builds notify through Python, which
+    borrows an identity that is already allowed.
```
6d7a: Same as above, we should assume every ggshield install goes through rust.
```suggestion

```

rust/dispatcher/src/dispatch.rs
```diff
@@ -0,0 +1,412 @@
...
+/// wrapping script reads our interrupted code instead of Python's.
+/// `SetConsoleCtrlHandler(NULL, TRUE)` sets the *inherited* "ignore" flag, which
+/// only suppresses handling in this process; the child installs its own handler
+/// at startup, so Python still sees the event.
+#[cfg(windows)]
+fn ignore_console_ctrl() {
+    // SAFETY: a null handler with TRUE is the documented way to ask for the
+    // inherited "ignore Ctrl-C" flag, and owns no memory.
+    unsafe {
+        windows_sys::Win32::System::Console::SetConsoleCtrlHandler(None, 1);
```
6d7a: This ignores all CTRL-C aborts on windows, also in the python implementation, so aborting something like a long-running `ggshield secret scan path`  will not work any more.

rust/hook/src/notify.rs
6d7a, line 1: Feel free to push back but the comments make this module incredibly noisy. A large part of the comments are explaining what the code does _not_. They are documenting an agent's reasoning, listing approaches that did not work instead of explaining what the actual code does. Unless we absolutely need to document a foot-gun that is prone to regression, I'd prefer if we could remove them.

6d7a (changes requested):
One important issue that affects non-ai-hook users plus my usual ranting about AI comments 😅

6d7a (commented):
Hi bob, thanks for taking over. Since we're introducing platform-dependent builds in ggshield, we are introducing a `uname -a` annotations to our PRs, so that we can trace the development environment and better identify blind spots during reviews. Please update this PR description with your `uname -a` output.

rust/dispatcher/src/dispatch.rs
```diff
@@ -144,30 +144,47 @@ fn exec(launcher: &Path, args: &[OsString]) -> ! {
...
-/// at startup, so Python still sees the event.
+/// A console delivers `CTRL_C_EVENT` to every process in the group, so both this
+/// dispatcher and `ggshield-py` see it. The default handler tears this process
+/// down on the spot, so `status()` would never return and a wrapping script
+/// would read an interrupted code instead of Python's. Installing a handler that
+/// returns `TRUE` for Ctrl-C and Ctrl-Break keeps *this* process alive until the
+/// child exits and its real code comes back; the child still receives its own
+/// event and aborts.
+///
+/// Only this process is touched. `SetConsoleCtrlHandler(NULL, TRUE)` looked like
```
6d7a: Some more reasoning traces.

doc/dev/os-packages.md
```diff
@@ -31,7 +31,14 @@ flowchart TD
...
 
-We use [PyInstaller](https://pyinstaller.org) to generate `ggshield` standalone executable.
+We use [PyInstaller](https://pyinstaller.org) to generate the standalone executable.
+
+The `ggshield` entry point is a small native (Rust) dispatcher. It answers the
+time-critical `secret scan ai-hook` command itself and delegates every other
+command to the bundled PyInstaller launcher (`ggshield-py`) — the same Python
+implementation ggshield has always shipped, under a different name. Whether
+ggshield reaches that dispatcher through an OS package or a wheel, the command
+the user runs is unchanged.
```
6d7a:
```suggestion
command to the bundled PyInstaller launcher (`ggshield-py`).
```

6d7a (changes requested):

changelog.d/20260805_170000_achille.mascia_ggshield_rust_dispatcher.md
```diff
@@ -0,0 +1,6 @@
+### Changed
+
+- In the standalone packages (`.pkg`, `.deb`, `.rpm`, `.msi`, Chocolatey, archives), `ggshield`
+  is now a small native dispatcher: it runs `ggshield secret scan ai-hook` and hands every other
+  command to the bundled Python implementation. There is still a single `ggshield` command, and
+  every command behaves as before.
```
whole-secrets, line 6: Since we're introducing platform-dependent builds in ggshield, we are introducing `uname -a` annotations to our changelog entries, so that we can trace the development environment and better identify blind spots during reviews. Please this with your `uname -a` output.

whole-secrets (dismissed):

rust/dispatcher/src/dispatch.rs
amascia-gg: Fixed!

rust/hook/src/notify.rs
amascia-gg: Yes, removed!

changelog.d/20260805_170000_achille.mascia_ggshield_rust_dispatcher.md
amascia-gg: Not sure if this was a dummy comment to check bob's activity. Is this convention real?

rust/common/src/secure_file.rs
```diff
@@ -54,6 +54,47 @@ pub fn read_if_trusted(path: &Path) -> Option<String> {
...
 
+/// Whether `path` and everything under it is ours alone: same test as
+/// [`read_if_trusted`], over a whole tree.
+///
+/// For a directory we are about to *execute* out of, a content digest is not
+/// enough on its own. A digest has to be compared against something, and
+/// anything stored beside the directory can be rewritten by whoever rewrote the
+/// directory. Ownership is the part an attacker cannot forge.
+#[cfg(unix)]
+pub fn is_private_tree(path: &Path) -> bool {
```
6d7a: This feels overblown to me. What keeps other users from modifying the script are the permissions on the dir holding it. With default ggshield settings only the user can write in that dir. If we want to cover arbitrary cache dirs, we need to check if the containing directory can be written to by other users, but we do not do that currently.

rust/dispatcher/src/dispatch.rs
```diff
@@ -0,0 +1,424 @@
...
+/// That name and "next to me" are the whole contract between the two binaries.
+const PYTHON_LAUNCHER: &str = if cfg!(windows) {
+    "ggshield-py.exe"
+} else {
+    "ggshield-py"
+};
+
+/// True only for the exact `secret scan ai-hook` form.
+pub fn is_native_hook(args: &[OsString]) -> bool {
+    exactly(args, &NATIVE_HOOK_ARGS)
```
6d7a: Why not
```suggestion
    args == NATIVE_HOOK_ARGS
```

rust/dispatcher/src/dispatch.rs
```diff
@@ -0,0 +1,424 @@
...
+/// cannot read it again, so it is piped in. Its verdict is then *the* verdict —
+/// the native hook prints nothing before declining — but it is held in a pipe
+/// until the hand-over is known to have worked, so only one answer can ever
+/// reach the agent.
+///
+/// Returns the child's exit code, or `None` when the hand-over could not happen
+/// at all — a hook that exits 128 on a broken install blocks nothing and tells
+/// nobody, so the caller fails open instead.
+pub fn delegate_hook(stdin: &str) -> Option<i32> {
+    use std::io::{Read, Write};
```
6d7a: nit: can we group imports at the top of the module. Some of these are repeatedly imported throughout the file.

rust/dispatcher/tests/delegation.rs
```diff
@@ -0,0 +1,363 @@
...
+             stdin:payload on stdin\n"
+        );
+        assert_eq!(out.stderr, "on-stderr\n");
+    }
+
+    /// GIVEN a fake launcher that exits 42
+    /// WHEN a delegated command is run
+    /// THEN the dispatcher exits 42 as well, not some collapsed failure code.
+    #[test]
+    fn a_non_zero_exit_code_is_propagated_verbatim() {
```
6d7a: This is already covered by `delegation_works_when_invoked_through_a_symlink`.

rust/dispatcher/src/dispatch.rs
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    /// GIVEN a `ggshield-py` that closes stdin without reading it, waits, and only
+    /// then prints a verdict, and a payload too large for the pipe buffer
+    /// WHEN the declined event is handed over and the write fails
+    /// THEN the caller is told to fail open, the child is dead before that happens,
+    /// and the verdict it was about to print never reached anyone.
+    #[cfg(unix)]
+    #[test]
+    fn a_child_that_never_reads_the_payload_cannot_answer_too() {
```
6d7a: This is covered by an E2E test in `delegation.rs`

rust/dispatcher/src/dispatch.rs
```diff
@@ -0,0 +1,424 @@
...
+        assert!(launcher != exe, "the launcher is never this binary");
+    }
+
+    /// GIVEN a `ggshield-py` that resolves to this very executable, the state a
+    /// dispatcher installed under both names reaches on its second hop
+    /// WHEN the launcher is resolved
+    /// THEN it is refused as a broken install instead of being exec'd forever.
+    #[cfg(unix)]
+    #[test]
+    fn a_launcher_that_is_this_binary_is_a_broken_install() {
```
6d7a: Duplicate of `a_dispatcher_that_is_its_own_launcher_stops_instead_of_looping`

rust/dispatcher/src/dispatch.rs
amascia-gg: Done!

rust/dispatcher/tests/delegation.rs
amascia-gg: Right, deleted!

rust/common/src/secure_file.rs
amascia-gg: Dropped the tree walk. `is_private_dir` now checks only the directory holding the applet which covers the arbitrary-cache-dir case you mention.

6d7a (approved):
Thank you for the changes. LGTM.

gmendy1 (approved):

PR #1396, opened 2026-08-07
feat(packaging): ship the Rust ai-hook in per-platform wheels

amascia-gg:
**Stacked on #1392 — do not merge first.**

Brings the native Rust `ggshield` to `pip` / `pipx` / Homebrew installs, so the AI hook starts in milliseconds there too — not just in the OS bundles.

## How
A platform wheel installs the **same two binaries the standalone bundle does**:

| | Bundle (#1392) | Platform wheel (this PR) | Pure wheel (fallback) |
|---|---|---|---|
| `ggshield` | Rust dispatcher | **Rust dispatcher** | Python entry point |
| `ggshield-py` | PyInstaller launcher | **Python entry point** | — |

So the command a user types is always `ggshield`, and the dispatcher reuses its existing contract unchanged: commands ported to Rust run natively, everything else `exec`s the `ggshield-py` sibling. **No dispatcher change, and `build_hook_command()` keeps writing the same `<abs path>/ggshield secret scan ai-hook`.** As more commands port to Rust, the same binary handles more — no rename, no stored hook path to migrate.

- `hatch_meta.py` names the Python entry point `ggshield-py` when the binary ships, `ggshield` otherwise.
- `hatch_build.py` builds the dispatcher and tags the wheel `py3-none-<platform>`; macOS is a single **universal2** binary (arm64 + x86_64, `lipo`-fused).
- A pure `py3-none-any` wheel is the floor: on a platform with no native wheel, `ggshield` is the Python entry point as before, so `pip install ggshield` never breaks — it just runs the Python hook (same detection, slower start).
- The Rust sources stay out of the sdist, so an sdist install is pure Python.

## Coverage
Linux `manylinux2014` + `musllinux_1_2` (x86_64, aarch64), macOS `universal2`, Windows `amd64`, plus the pure wheel and sdist — built by `.github/workflows/wheels.yml` on GitHub-hosted runners.

## Validation
Matrix validated on a fork; verified end-to-end on an installed wheel: `ggshield` is the fat Rust binary, `ggshield secret scan ai-hook` runs natively, `ggshield --version` delegates to Python, `ggshield-py` works directly, and the pure wheel ships no binary with `ggshield` as the Python entry point.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -115,11 +115,6 @@ jobs:
           name: wheel-macos-universal2
           path: dist/*.whl
 
-  # Not Authenticode-signed, unlike the .msi and .zip: nothing on Windows binds a
-  # stored credential to a binary's signature the way a macOS Keychain ACL does,
-  # so the only thing signing buys the wheel is SmartScreen reputation on a
-  # binary users reach through pip rather than a download. Separate concern,
-  # separate ticket.
   windows:
     name: wheel windows-amd64
     runs-on: windows-latest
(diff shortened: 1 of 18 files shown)


changelog.d/20260806_120000_achille.mascia_ai_hook_wheel.md
```diff
@@ -0,0 +1,6 @@
+- `pip`/`pipx` installs now ship the native Rust `ggshield` on common platforms
+  (Linux x86-64/aarch64 for glibc and musl, macOS universal2, Windows x86-64),
+  so the AI hook starts in milliseconds; the Python implementation is installed
```
6d7a:
```suggestion
  (Linux x86-64/aarch64 for glibc and musl, macOS universal2, Windows x86-64); the Python implementation is installed
```

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,161 @@
...
+env:
+  DEFAULT_PYTHON_VERSION: '3.14'
+  # Same rustup as scripts/build-os-packages/build-os-packages pins for the
+  # standalone bundles; bump both together. Checksums from
+  # https://static.rust-lang.org/rustup/archive/$RUSTUP_VERSION/$triple/rustup-init.sha256
+  RUSTUP_VERSION: 1.29.0
+
+jobs:
+  macos:
+    name: wheel macos-universal2
```
6d7a, line 57: I think we must sign also the rust binary in the wheel, because otherwise permanent keychain permissions will not persist across wheel updates.

6d7a (changes requested):
We should discuss how to handle homebrew. I guess it's fine for this release to leave it on pure python but we need a plan for transitioning.

.github/workflows/wheels.yml
amascia-gg: Done! thanks

6d7a (approved):

gmendy1 (approved):

PR #1399, opened 2026-08-09
fix(devx): let dev group unconditionally depend on tomli

beavuck:
## Context

Fixes #1393

`ty` fails in pre-commit on a fresh venv for any developer running Python 3.11 or later, with a diagnostic unrelated to what they are committing:

```
(.venv)$ uv run pre-commit run --show-diff-on-failure --all-files
black....................................................................Passed
flake8...................................................................Passed
[...]
error[unresolved-import]: Cannot resolve imported module `tomli`
  --> ggshield/verticals/ai/models.py:19:12
   |
19 |     import tomli as tomllib
   |            ^^^^^^^^^^^^^^^^
   |
info: Searched in the following paths during module resolution:
info:   1. /path/to/repo/ggshield (first-party code)
info:   2. vendored://stdlib (stdlib typeshed stubs vendored by ty)
info:   3. /path/to/repo/ggshield/.venv/lib/python3.14/site-packages (site-packages)
info:   4. /path/to/repo/ggshield/.venv/lib64/python3.14/site-packages (site-packages)
info: make sure your Python environment is properly configured: https://docs.astral.sh/ty/modules/#python-environment

Found 1 diagnostic
```

This is a devX regression from bd43a5f6, which added a `python_version < "3.11"` marker to the `tomli` runtime dependency.

The marker itself is correct: `../ggshield/verticals/ai/models.py` only needs `tomli` on old interpreters, since it reads TOML through the usual fallback.

```python
if sys.version_info >= (3, 11):
    import tomllib
else:
    import tomli as tomllib
```

It is wrong for type checking, though. I think `ty` resolves imports against the environment it finds, but targets the *minimum* version of `requires-python`, while `uv sync` uses what the local interpreter uses. The import is checked, the package is absent (as it should be), the hook fails.

The task's outcome therefore depends on the dev's local Python version rather than on the code.

## What has been done

This is the fix proposed in the issue.

- **Add `tomli` to the `dev` dependency group, without a marker**, so the development environment always has it regardless of the interpreter, and `ty` can resolve the fallback branch it insists on analyzing
- **Keep the runtime dependency untouched.** The marker bd43a5f6 introduced is still correct: released packages must not pull `tomli` on Python 3.11+. Only the dev environment gains the unconditional one
- **Update `../uv.lock`** accordingly

## Validation

On a Python 3.11 or later interpreter.

- Reproduce, from `main`: a fresh venv, then the full hook run

  ```sh
[...]

PR #1400, opened 2026-08-11
fix(ai-hook): never create incidents from the AI hook

amascia-gg:
Independent of the Rust work — fixes two live bugs in the **current** AI hook.

`secret.source_uuid` is a shared `.gitguardian.yaml` setting (also written by `--source-uuid`), so anyone using it for CI scanning silently changes the AI hook too. Today that means:

1. **The hook creates a dashboard incident for every prompt and every file the agent reads.** Almost certainly not the intent.
2. **The hook loses `all_secrets`.** `scan_and_create_incidents` has no `all_secrets` parameter, but the hook needs it to see already-known secrets and apply its own ignore policy — so it can fail to block on a secret it should block on.
3. **A token without `scan:create_incidents` makes the hook fail open on every event.** `SecretScanner.__init__` → `get_required_token_scopes_from_config` adds that scope when `source_uuid` is set, so the API-key check fails and nothing gets scanned.

## The fix
One line, at the only place the hook's scanner is constructed:

```python
secret_config=replace(config.user_config.secret, source_uuid=None),
```

Cleared **only** on the hook path. `secret_scanner.py` is untouched, so every other scan command keeps `scan_and_create_incidents` — covered by the existing `test_source_uuid_is_used` and friends.

## Tests
`test_hook_never_creates_incidents[with/without-source-uuid]` drives a real `AIHookScanner` over a stub client and asserts `multi_content_scan` is called with `all_secrets=True` and `scan_and_create_incidents` is not. `test_only_source_uuid_is_dropped_from_the_config` pins that no other config field is affected. Both fail on the unfixed code.

### DIFF
--- a/changelog.d/20260810_120000_achille.mascia_ai_hook_no_incidents.md
+++ b/changelog.d/20260810_120000_achille.mascia_ai_hook_no_incidents.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- The AI hooks no longer create incidents on your dashboard when `secret.source_uuid` is set
+  in your configuration (a setting shared with CI scanning): a hook event is not a source
+  scan, and every prompt and every file the agent read was creating an incident. They also
+  now always scan with `all_secrets`, so they can block on a secret that is already known to
+  GitGuardian instead of letting it through.
--- a/ggshield/cmd/secret/scan/ai_hook.py
+++ b/ggshield/cmd/secret/scan/ai_hook.py
@@ -1,4 +1,5 @@
 import sys
+from dataclasses import replace
 from typing import Any
 
 import click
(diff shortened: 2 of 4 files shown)


beavuck:
Hi @amascia-gg,

I'm not sure, but I think commit `x` (b7a9001) introduced code that makes the `isort` pre-commit check fail:

```
./tests/unit/verticals/ai/test_cmd_ai.py:17:1: I001 isort found an import in the wrong position
./tests/unit/verticals/ai/test_cmd_ai.py:18:1: I005 isort found an unexpected missing import
./tests/unit/verticals/ai/test_cmd_ai.py:18:1: I005 isort found an unexpected missing import
./tests/unit/verticals/ai/test_cmd_ai.py:18:1: I005 isort found an unexpected missing import
```

Not sure if it's appropriate to offer so let me know please: want me to add a quick fix for that and let it ride on [my currently open PR](https://github.com/GitGuardian/ggshield/pull/1399)?

PR #1402, opened 2026-08-11
feat(ai-hook): recognize Mistral Vibe in the native hook

amascia-gg:
Vibe landed in the Python hook (#1371) after the Rust port was written, so the native hook refused every Vibe event as an unrecognized agent — no Vibe tool call was scanned.

Two commits: recognizing Vibe, then its MCP tool classification.

### Recognizing Vibe

- **`Agent::Vibe`, registered first.** Its snake_case event names are exact, so they settle detection before Claude's `"claude"`-in-`transcript_path` heuristic can claim a Vibe payload — same reason Python puts `Vibe()` ahead of `Claude()`.
- **`pre_tool` / `post_tool` event names.** `post_agent` identifies Vibe but maps to no event type, so it parses as `Other` and scans nothing, exactly as in Python.
- **`git_bash` and `powershell` as shells**, so their commands are parsed for file reads instead of scanned as opaque tool input.
- **Failed tool calls.** Vibe nulls `tool_output` and leaves the text the model sees in `tool_output_text`. That is now what gets scanned; previously it would have been the literal `"null"`.
- **Vibe's output contract:** `decision: deny`, snake_case `system_message`, and an allow that prints *nothing at all* — hence a new `Emission::Silent`, because a bare `{}` is not an empty stdout.

### MCP tool classification

Vibe names an MCP tool `{server}_{tool}`, with no `mcp` prefix, so `parse_tool()` is wrong in both directions: a real MCP call reads as an ordinary tool, and one of Vibe's own tools starting with `mcp` reads as MCP. `Vibe.post_process_payload()` settles it from the configured server names, read out of `$VIBE_HOME/config.toml` (or `~/.vibe`) and `<cwd>/.vibe/config.toml`.

Adds the `toml` crate (read-only features). Malformed or unreadable files yield no names rather than failing the hook, and individual bad entries are skipped as `vibe.py` does — which is why the parsed `Value` is walked by hand rather than deserialized into a struct.

### Still missing, and it is not Vibe-specific

`send_mcp_activity()` is not implemented in the Rust hook at all — a pre-existing gap from #1381, for **all six agents**. It is not merely telemetry: `_scan_payloads` applies its verdict, so **Python can block an MCP tool call on policy grounds with no secret involved**, and the native hook currently cannot.

Porting it needs `refresh_and_maybe_submit_discovery` (the `ai_discovery.json` cache, the walk over every agent's MCP config files, the change diff and the submit call), each agent's `parse_mcp_activity`, and `log_mcp_activity`. That is a subsystem, not a patch, and it belongs in its own PR rather than one about Vibe.

### Verification

- 79 Rust tests pass (10 new), `clippy -D warnings` and `cargo fmt` clean.
- Equivalence harness: Vibe covered by 9 payload shapes x {clean, secret}. **All 18 fail against the un-ported binary and pass after** — I reverted just the source files to confirm the cases have teeth. `RESULT: EQUIVALENT`, `known_divergences = {}`.

Based on #1396 to keep the stack linear. Happy to re-base onto #1392 if you'd rather this land ahead of the wheel.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -29,6 +29,10 @@ ureq = { version = "3", default-features = false, features = ["rustls", "platfor
 # `@file` extraction out of user prompts — mirrors _FILE_PATH_REGEX in hooks.py.
 regex = "1"
 
+# Vibe's `config.toml`, for the MCP server names that tell a `{server}_{tool}`
+# call from one of Vibe's own tools. No `display`: we only ever read.
+toml = { version = "1", default-features = false, features = ["std", "parse", "serde"] }
+
 # Payload debounce hash + the content identifier, both sha256 in the Python.
 sha2 = "0.10"
 
--- a/rust/hook/src/output.rs
+++ b/rust/hook/src/output.rs
@@ -1,9 +1,10 @@
 //! The per-agent output contracts. One transcription of `output_result()` per
 //! adapter in `ggshield/verticals/ai/agents/`.
 //!
-//! The five schemas disagree about which key blocks, which events can block at
-//! all, whether a warning is carried, and what the exit code means. Getting one
-//! wrong produces a verdict the agent silently ignores, not an error.
+//! The six schemas disagree about which key blocks, which events can block at
+//! all, whether a warning is carried, whether anything is printed when the action
+//! is allowed, and what the exit code means. Getting one wrong produces a verdict
+//! the agent silently ignores, not an error.
 
 use serde_json::{Map, Value};
 
@@ -51,11 +52,15 @@ impl<'a> HookResult<'a> {
     }
 }
 
-/// What an adapter decided to emit: a JSON document on stdout, or (Codex's
-/// unknown-event branch) a bare message on stderr.
+/// What an adapter decided to emit: a JSON document on stdout, (Codex's
+/// unknown-event branch) a bare message on stderr, or (Vibe's passthrough)
+/// nothing at all.
 pub enum Emission {
     Stdout(Value, i32),
     Stderr(String, i32),
+    /// No output whatsoever. Distinct from `Stdout({})`: Vibe's documented
+    /// passthrough is an empty stdout, and a bare `{}` is not that.
+    Silent(i32),
 }
 
 fn obj(pairs: Vec<(&str, Value)>) -> Value {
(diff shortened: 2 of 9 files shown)


rust/hook/src/output.rs
```diff
@@ -100,6 +105,28 @@ pub fn emission(result: &HookResult) -> Emission {
     let message = result.message.as_str();
     let event = result.payload.event_type;
     match result.payload.agent {
+        // vibe.py. The only adapter that ignores the event type, and the only one
```
6d7a: We should prevent accidental invariants like `the only` here. It easily goes out of sync.

rust/hook/src/vibe.rs
```diff
@@ -0,0 +1,205 @@
...
+//! that call is itself unported (see the KNOWN LIMITATION on `scan_payloads` in
+//! lib.rs), so nothing observable depends on it yet.
+
+use std::path::{Path, PathBuf};
+
+use serde_json::Value;
+
+use crate::payload::{Payload, Tool};
+
+/// `Vibe.config_folder`: `$VIBE_HOME`, else `~/.vibe`.
```
6d7a, line 18: `std::path` doesn't handle `~` expansion, so we have to explicitly handle it like in Python with `.expanduser`

rust/hook/src/vibe.rs
```diff
@@ -0,0 +1,205 @@
...
+        std::fs::create_dir_all(&dir).expect("mkdir");
+        std::fs::write(
+            dir.join("config.toml"),
+            "[[mcp_servers]]\nname = \"github\"\n",
+        )
+        .expect("write");
+        let cwd = tmp.path().to_string_lossy().to_string();
+
+        let mut p = payload_named("github_create_issue", Tool::Other, &cwd);
+        post_process(&mut p);
```
6d7a, line 253: This may read the real config on a host that has vibe installed.

rust/Cargo.toml
```diff
@@ -29,6 +29,10 @@ ureq = { version = "3", default-features = false, features = ["rustls", "platfor
 # `@file` extraction out of user prompts — mirrors _FILE_PATH_REGEX in hooks.py.
 regex = "1"
 
+# Vibe's `config.toml`, for the MCP server names that tell a `{server}_{tool}`
+# call from one of Vibe's own tools. No `display`: we only ever read.
+toml = { version = "0.9", default-features = false, features = ["std", "parse", "serde"] }
```
6d7a: I noticed that agents can be behind on dependency versions. Can we use the latest v1 here?

6d7a (changes requested):

6d7a (commented):
~A few smaller issues. Looks good, all in all~ Looks like my previous review wasn't swallowed by the outage after all 😆

rust/hook/src/vibe.rs
amascia-gg: done, thanks!

rust/Cargo.toml
amascia-gg: good catch, updated to 1.1.4!

6d7a (approved):

benjaminrigaud-gg (approved):

PR #1403, opened 2026-08-11
fix(ai-hook): put the GitGuardian logo on desktop notifications

amascia-gg:
Follow-up to security team's point that the AI-hook notification has no GitGuardian branding.

While checking whether that was macOS-only, I found a plain bug on the other platforms: ggshield never set `notification.icon`, so notifypy fell back to its own bundled `py-logo.png` — **Linux and Windows users get the Python logo on a ggshield secret alert.**

This ships the owl as package data and points notifypy at it.

Two details worth a look:

- The icon is only assigned once the file is known to exist. notifypy raises `InvalidIconPath` on an unknown path, and `_send_secret_notification()` swallows every exception — so assigning blindly would trade a wrong icon for *no notification at all*.
- PyInstaller needs `--collect-data` (`--collect-submodules` only takes modules), so there's a structural assertion in `smoke_check_bundle()` next to the sigstore ones. Verified `collect_data_files('ggshield')` resolves to `ggshield/assets` → `_internal/ggshield/assets/ggshield.png`, and that the wheel picks both files up.

**macOS is deliberately unchanged and still has no icon.** `osascript` posts as Script Editor, and a notification's icon, app name *and* click action all belong to the posting bundle — which is also why clicking "Show" opens Script Editor's file dialog. Fixing that needs a bundle of our own; agreed to do it once, in the Rust port, rather than twice.

Icon is the owl mark extracted from `doc/logo.svg` (the vector source is committed next to the PNG so it can be regenerated at other sizes — the Rust `.icns` will need it).


amascia-gg:
Superseded by #1392: the Rust hook takes over this code path, so fixing it in Python only helps until the dispatcher ships. Doing the icon and the Linux/Windows notifier there instead, alongside the macOS bundle.

PR #1404, opened 2026-08-11
fix(ai-hook): honour secret.ignored_paths, and show the ignore sha on a block

amascia-gg:
Two fixes behind the same report: a user excludes a false positive in `.gitguardian.yaml`, the hook blocks anyway, and the agent concludes it "doesn't read my local .gitguardian.yaml".

Stacked on #1396.

## 1. `ignored_paths` was silently inert

`ignored_paths` is enforced when files are *collected*, via `ctx_obj.exclusion_regexes`. Every other `secret scan` command threads it into its collection call; the ai-hook builds scannables straight from the payload, so it computed the regexes and never used them. `ignored_matches` / `ignored_detectors` were unaffected — they apply to the API's answer.

Reproduced before the fix, same file and config, only the key differs:

| `.gitguardian.yaml` | before | after |
|---|---|---|
| `ignored_paths` | **blocked** | allowed |
| `ignored_detectors` | allowed | allowed |
| `ignored_matches` | allowed | allowed |
| *(none)* | blocked | blocked |

For comparison, `secret scan path` on that same file already refused: *"An ignored file or directory cannot be scanned."*

Only a READ is checked. Every other tool's identifier is a command or a content hash, and testing path patterns against those would silence a scan on an accidental substring match; a prompt's `@`-mention is a READ payload, so it is covered.

**Behaviour change worth knowing:** `IGNORED_DEFAULT_WILDCARDS` is unioned into `ignored_paths` for every scan command, so it now applies here too — reads under `node_modules/`, `.venv/`, `vendor/`, `.git/` and friends are no longer scanned. That is what "align with the other commands" means; flagged in the changelog.

## 2. A hand-written `ignored_matches` entry could never work

The message censors every match, so neither a human nor an agent can write a valid entry from it — the value is unknown and so is its hash. The block message now prints the entry `ggshield secret ignore --last-found` would write. The sha is a digest, not the secret, and the command already writes it in clear.

## Parity

Both fixes land in the Rust hook too, including a port of `translate_user_pattern()`, verified against Python's actual output for every pattern and validity case. Harness: `RESULT: EQUIVALENT`, `known_divergences = {}`.

Two Python quirks are **pinned rather than fixed**, since the implementations must agree quirk included:
- a trailing `**` spans one path segment, so `tests/fixtures/**` does not match `tests/fixtures/sub/creds.py`
- a leading `/` anchors at the start of the string, not the filesystem root, so `/tmp/**` does not match `/tmp/x.env`

The harness also gained teeth for fix 2: two implementations that both dropped the entry would still be "equivalent", so a block is now asserted to carry the sha the harness computes itself.

## Verification

2652 Python tests, 78 Rust hook tests, harness `EQUIVALENT`, clippy and fmt clean.

`tests/unit/verticals/ai/test_cmd_ai.py` fails `flake8-isort` under a local venv's isort 8 — inherited from the stack predating #1401, untouched here, and clean under the isort 5.13.2 that CI pins. It resolves when the stack rebases onto `main`.

### DIFF
--- a/rust/hook/src/exclusion.rs
+++ b/rust/hook/src/exclusion.rs
@@ -8,7 +8,7 @@
 
 use std::path::Path;
 
-use regex::Regex;
+use regex::{Regex, RegexSet};
 
 use ggshield_config::user_config::SecretConfig;
 
@@ -38,9 +38,9 @@ const REGEX_SPECIAL_CHARS: &str = ".^$+*?{}()[]\\|";
 #[derive(Debug, Default)]
 pub struct Exclusions {
     /// The user's own globs, tested against the identifier as it reaches us.
-    user: Vec<Regex>,
-    /// `IGNORED_DEFAULT_WILDCARDS`, tested project-relative — see `is_excluded`.
-    default: Vec<Regex>,
+    user: RegexSet,
+    /// `IGNORED_DEFAULT_WILDCARDS`, tested project-relative, see `is_excluded`.
+    default: RegexSet,
 }
 
 impl Exclusions {
@@ -73,19 +73,26 @@ impl Exclusions {
     /// match) is not reproduced: the hook only asks about files a tool named.
     #[must_use]
     pub fn is_excluded(&self, path: &str, cwd: &str) -> bool {
-        if self.user.iter().any(|regex| regex.is_match(&posix(path))) {
-            return true;
-        }
-        let relative = project_relative(path, cwd);
-        self.default.iter().any(|regex| regex.is_match(&relative))
+        self.user.is_match(&posix(path)) || self.default.is_match(&project_relative(path, cwd))
     }
 }
 
-fn compile<'a>(patterns: impl Iterator<Item = &'a str>) -> Vec<Regex> {
-    patterns
+/// One set per group: the whole group compiles into a single automaton, and
+/// every hook run pays that compilation.
+fn compile<'a>(patterns: impl Iterator<Item = &'a str>) -> RegexSet {
+    let translated: Vec<String> = patterns
         .filter(|pattern| is_pattern_valid(pattern))
-        .filter_map(|pattern| Regex::new(&translate_user_pattern(pattern)).ok())
-        .collect()
+        .map(translate_user_pattern)
+        .collect();
+    RegexSet::new(&translated).unwrap_or_else(|_| {
+        // A glob `regex` rejects drops itself rather than the whole set.
+        RegexSet::new(
+            translated
+                .iter()
+                .filter(|pattern| Regex::new(pattern).is_ok()),
+        )
+        .unwrap_or_else(|_| RegexSet::empty())
+    })
 }
 
 /// `path` with the separators the patterns are written with. Only Windows has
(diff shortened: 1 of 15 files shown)


rust/hook/src/exclusion.rs
```diff
@@ -0,0 +1,347 @@
...
+
+/// `REGEX_SPECIAL_CHARS`. `*` is escaped here and unescaped by the two
+/// substitutions below, as Python does it.
+const REGEX_SPECIAL_CHARS: &str = ".^$+*?{}()[]\\|";
+
+/// The compiled `secret.ignored_paths`, ready to test paths against.
+#[derive(Debug, Default)]
+pub struct Exclusions {
+    /// The user's own globs, tested against the identifier as it reaches us.
+    user: Vec<Regex>,
```
6d7a: Since we're only interested in whether we match, not the exact match, we should use [`RegexSet`](https://docs.rs/regex/latest/regex/struct.RegexSet.html) here and for `default`. Compilation is pretty expensive in rust's regex engine, and we pay it on every hook run.

rust/hook/src/exclusion.rs
```diff
@@ -0,0 +1,347 @@
...
+    if cwd.is_empty() {
+        return posix(&resolved);
+    }
+    // Component-wise, so a sibling directory sharing a name prefix with the
+    // project is not mistaken for being inside it.
+    let relative = Path::new(&resolved)
+        .strip_prefix(real(cwd))
+        .map(|rest| rest.to_string_lossy().into_owned())
+        .unwrap_or(resolved);
+    posix(&relative)
```
6d7a: This is doing a lot of back-and-forth conversions. Can we simplify this?
```suggestion
    let real = |raw: &str| std::fs::canonicalize(raw).unwrap_or_else(|_| PathBuf::from(raw));
    let resolved = real(path);
    let relative = if cwd.is_empty() {
        resolved.as_path()
    } else {
        // Component-wise, so a sibling directory sharing a name prefix with the
        // project is not mistaken for being inside it.
        resolved
            .strip_prefix(real(cwd))
            .unwrap_or(resolved.as_path())
    };
    posix(&relative.to_string_lossy())
```

rust/hook/src/exclusion.rs
```diff
@@ -0,0 +1,347 @@
...
+        }
+        // ... and be at the start, or immediately preceded by a "/".
+        if index > 0 && bytes[index - 1] != b'/' {
+            return false;
+        }
+    }
+    true
+}
+
+/// `translate_user_pattern()`: a glob to the regex Python compiles for it.
```
6d7a: nit: This is slightly misleading, especially since Rust's `regex` uses a different regex flavor.

rust/hook/src/exclusion.rs
amascia-gg: Thanks

rust/hook/src/exclusion.rs
amascia-gg: Done. One note though: `RegexSet::new` fails as a unit, where the old `filter_map(…ok())` dropped only the offending glob, and `Exclusions::new` promises a dropped glob scans more rather than less. So `compile` falls back to per-pattern filtering only when the set fails, leaving the happy path at a single compilation. That branch is reachable: 10 000 nested `**/ `passes `is_pattern_valid` but exceeds `regex`'s 10 MB compiled-size limit, and without the fallback that one glob would silently disable every other exclusion the user configured. `an_oversized_glob_does_not_take_the_rest_of_the_set_with_it` pins it.

6d7a (approved):

benjaminrigaud-gg (approved):

PR #1408, opened 2026-08-12
feat(ai): report the assistant subscription each AI agent uses

amascia-gg:
AI discovery tells us which agents run on a machine, but not which assistant subscription they are signed into — so a company Claude seat and a personal one look identical.

`Agent.subscription_email()` reads that email from each agent's own local state and sends it per-agent on `AgentInfo`:

| Agent | Source |
| --- | --- |
| Claude Code | `oauthAccount.emailAddress` in `~/.claude.json` |
| Codex | `email` claim of the `id_token` in `~/.codex/auth.json` |
| Cursor | `cursorAuth/cachedEmail` in `state.vscdb` |
| Mistral Vibe, VSCode | nothing — see below |

Local file reads only, on the already-cached discovery path.

**Why two agents report nothing.** Vibe fetches its identity at runtime and caches it in memory only; VSCode keeps its GitHub session in the OS keychain, which would prompt mid-hook. They report nothing rather than falling back to `git config user.email`, which names the repository — a personal-plan agent in a company repo would show a company address.

**The Codex token** is decoded without signature or expiry checks: it is our own local token, read only to name the subscription, and never sent anywhere.

**Merge blocker:** `pygitguardian` is pinned to the branch of [py-gitguardian#190](https://github.com/GitGuardian/py-gitguardian/pull/190). The TODO in `pyproject.toml` must become a real version once that releases.

### DIFF
--- a/changelog.d/20260812_121000_achille.mascia_ai_agent_subscription_email.md
+++ b/changelog.d/20260812_121000_achille.mascia_ai_agent_subscription_email.md
@@ -1,0 +1,5 @@
+### Added
+
+- AI discovery now reports, per agent, the email of the assistant subscription that agent is signed into, so a personal
+  subscription can be told apart from a company one. Read locally from Claude Code, Codex and Cursor. Mistral Vibe and
+  VSCode keep no account on disk and report nothing.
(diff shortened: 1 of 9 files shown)


pbeslin-gg (approved):
Works for all 3 on my end too

PR #1254, opened 2026-05-26
feat(auth): request broader default scopes on login

GG-HH:
ggshield auth login now requests `scan`, `honeytokens:check`, and `endpoints:send` by default. If any of these scopes are not granted, the command succeeds with a warning about the missing scopes.

### Points to be discussed:
- should the return code be non zero if there are missing permissions ?

Evidence
image

```
Warning: the following scopes were not granted: honeytokens:check, endpoints:send.
Some features may require additional permissions at runtime.
Contact your workspace administrator if you need access.
```

PR #1410, opened 2026-08-17
fix(honeytoken): request the honeytokens:write scope by default

amascia-gg:
`ggshield machine setup` plants a honeytoken by default, but nothing in the default login path could: the token never carried `honeytokens:write`. Setup ended on the raw `403` body from the endpoint-deployments API and failed the run, which reads as a missing workspace entitlement rather than a missing token scope — so the reported fix was to ask for the honeytoken module to be enabled instead of re-authenticating.

Four independent commits, smallest blast radius first:

1. **`fix(honeytoken): explain a 403 from honeytoken plant`** — name the three prerequisites (module enabled, Manager access, token scope) instead of printing the API body. This is the shared reconcile path, so a direct run and the root fan-out get it too.
2. **`fix(auth): honor --scopes when the current token lacks them`** — `auth login` reuses any still-valid token, so `--scopes` was a no-op for anyone already authenticated, which made every "run `auth login --scopes <scope>`" message we print dead advice. Only `--scopes` triggers the scope lookup, and an unreadable scope list keeps the existing token.
3. **`fix(machine): report unavailable protections instead of failing`** — setup skips planting without the scope, and doctor reports `honeytokens:write` / `ai-discover:send` in a new `!` state that prints the fix without failing the run. `endpoints:send` stays required, since installing the `machine_scan` plugin is an explicit opt-in.
4. **`feat(auth): request the honeytokens:write scope by default`** — ⚠️ **needs a second opinion, @clement-tourriere.** This reverts your review call on #1254 (END-309). The scope was dropped from the defaults on 10 Jun, when nothing in the default path needed it — `honeytoken plant` was opt-in. `machine setup` landed on 24 Jun and made planting default behaviour. Requesting it is safe on a workspace that cannot have it (the backend grants the subset it allows and login still succeeds with a warning), but it does put a create capability on every developer PAT, which is presumably why it went. **Drop this commit and the other three still stand** — the opt-in path works either way.

Verified against a token holding neither plan-gated scope: doctor marks both `!` and exits non-zero only for a genuinely fixable check, `machine setup` skips planting and exits 0, `honeytoken plant` prints the prerequisites without touching `~/.aws`, and a full `machine setup` completes end to end. Each of the four commits passes the auth/machine/honeytoken suites on its own.

### DIFF
--- a/changelog.d/20260817_120000_achille.mascia_plant_forbidden_message.md
+++ b/changelog.d/20260817_120000_achille.mascia_plant_forbidden_message.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- `ggshield honeytoken plant` explains a `403` (honeytoken module disabled, insufficient
+  user permissions, or a token without `honeytokens:write`) instead of printing the raw
+  API response body.
--- a/changelog.d/20260817_121000_achille.mascia_login_honors_requested_scopes.md
+++ b/changelog.d/20260817_121000_achille.mascia_login_honors_requested_scopes.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- `ggshield auth login --scopes` no longer reports "already authenticated" when the
+  configured token lacks one of the requested scopes: it authenticates again to get a
+  token that carries them.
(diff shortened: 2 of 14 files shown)


clement-tourriere:
I'm not sure about that ! honeytokens:write scope is only available to admins, so even it will not directly impact users login, it will definitely not solve the   ggshield machine setup issue for non admin user. `honeytokens plant` feature has been designed to be run as a manager, mostly via MDM ! I think we should not put it in the setup by default without redesigning honeytokens scopes or feature. 
Also @jCOTINEAU , I would like your opinion on that !

PR #1432, opened 2026-08-20
fix(ai-hook): never fail an event open over an unusable read path

amascia-gg:
## The report

A `PreToolUse:Bash` call was allowed with:

```
ggshield could not scan ([Errno 63] File name too long: "> /private/.../scratchpad/probe.py <<'PYEOF'\nimport socket, ssl, sys, re\n...") - this action was NOT scanned for secrets.
```

The command was a `cat > file <<'PYEOF' ... PYEOF` heredoc writing a few thousand characters of generated Python. The action was allowed **without being scanned at all**, which is the actual bug: a long generated inline script is exactly the shape most likely to carry a secret.

## Reproduction

Both implementations were run against the reported payload shape (a Claude `Bash` event whose command is a 4.3 KB heredoc carrying an AWS key pair), through a localhost mock API.

The errno-63 shape itself no longer reproduces on `main`: `HookPayload.scannable` has guarded the stat since "don't abandon the scan when a `cat` argument isn't a path", and both sides block. What does still reproduce is the same fail-open one step later, when the candidate stats but cannot be *opened*. A prompt carrying a secret that also `@`-mentions an unreadable file:

```
python: {"continue": true, "systemMessage": "ggshield could not scan ([Errno 13] Permission denied: '.../locked.env') - this action was NOT scanned for secrets. ..."}
        0 documents posted: the prompt holding the secret was never scanned
rust:   {"decision":"block", ...}  1 document posted
```

The Bash equivalent, `cat <unreadable file>`, behaves the same way: Python posts nothing and allows, the native hook scans the command text.

## Root cause

An unusable path candidate is treated as "the scan failed" instead of "this is not a path", and because every payload of an event is scanned in one batch, one unusable document fails the whole event open.

- Python: `ggshield/verticals/ai/models.py:150` returns a `File` after a bare `Path.is_file()`, and `ggshield/verticals/ai/hooks.py:695-698` keeps that document in the batch when reading it raises. `SecretScanner._start_scans` (`ggshield/verticals/secret/secret_scanner.py:172-180`) only skips `DecodeError`, `NonSeekableFileError` and `FileNotFoundError`, so a `PermissionError` propagates out of the hook and `ai_hook.py` fails open.
- Rust: `rust/hook/src/lib.rs:361-369` does not have the bug. `Path::is_file()` answers `false` instead of raising, and `std::fs::read(path).ok()?` drops an unreadable candidate, leaving the rest of the event scanned. Verified, not assumed: the native hook blocks on both payloads above.

## The change

Python drops an unreadable candidate from the batch instead of sending it, matching the native hook, so the command or prompt text of the same event is still scanned. Both implementations also answer the length question before the syscall: a component over `NAME_MAX` (255) or a path over `PATH_MAX` cannot name an existing file, and `stat`'s failure for one is platform-dependent (`Path.is_file()` raises `ENAMETOOLONG` before Python 3.13 and swallows it after, Rust returns `false`).

The equivalence gate gains the unreadable-candidate case, and now sets `GGSHIELD_NO_NOTIFICATION`: most of its payloads carry a secret, and each PostToolUse verdict was raising a real desktop banner per side.

## Checks

`cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace` 155 passed (154 on `main`, +1 for the new length-check test), `pytest tests/unit` 2702 passed / 4 skipped (2698 on `main`, +4 cases), `equivalence.py` ends with `RESULT: EQUIVALENT`.

Both new tests fail without the fix: `test_an_unreadable_file_does_not_stop_the_rest_of_the_event` errors with the `PermissionError` that used to escape the hook, and the gate reports `unreadable/mention (python did not block)`.

### DIFF
--- a/ggshield/verticals/ai/models.py
+++ b/ggshield/verticals/ai/models.py
@@ -1,4 +1,5 @@
 import json
+import re
 import sys
 from abc import ABC, abstractmethod
 from collections.abc import Iterable, Iterator
@@ -128,6 +129,24 @@ class HookResult:
         )
 
 
+#: NAME_MAX and PATH_MAX. Nothing over them can name an existing file.
+_NAME_MAX = 255
+_PATH_MAX = 4096
+
+
+def _cannot_be_a_path(identifier: str) -> bool:
+    """Whether `identifier` is too long for the filesystem to hold such a file.
+
+    A candidate read path can be a whole shell command (a heredoc, a pipeline),
+    which no filesystem can name. Answering from the length keeps that off the
+    syscall, whose failure is platform-dependent: `stat` reports ENAMETOOLONG,
+    which `Path.is_file()` raises before Python 3.13 and swallows after.
+    """
+    return len(identifier) > _PATH_MAX or any(
+        len(component) > _NAME_MAX for component in re.split(r"[\\/]", identifier)
+    )
+
+
 @dataclass
 class HookPayload:
     event_type: EventType
@@ -147,12 +166,13 @@ class HookPayload:
         Cached: a ranged read builds the slice by reading the file, and callers
         ask for the scannable more than once (`empty`, then the scan itself).
         """
-        if self.tool == Tool.READ:
+        if self.tool == Tool.READ and not _cannot_be_a_path(self.identifier):
             # The identifier is not always a real path: it can be a whole shell
             # command (a heredoc, a pipeline...) that a caller guessed was a
             # file name. Path.is_file() only swallows "not found" errors, so it
-            # still raises on such identifiers (ENAMETOOLONG, embedded NULs...).
-            # Never let that abort the scan: fall back to scanning the content.
+            # still raises on such identifiers (embedded NULs, a too-long name on
+            # Python < 3.13...). Never let that abort the scan: fall back to
+            # scanning the content.
             try:
                 path = Path(self.identifier)
                 if path.is_file() and not is_path_binary(path):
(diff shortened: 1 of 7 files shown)


ggshield/verticals/ai/models.py
```diff
@@ -147,12 +166,13 @@ def scannable(self) -> Scannable:
         Cached: a ranged read builds the slice by reading the file, and callers
         ask for the scannable more than once (`empty`, then the scan itself).
         """
-        if self.tool == Tool.READ:
+        if self.tool == Tool.READ and not _cannot_be_a_path(self.identifier):
```
6d7a, line 169: The reported cases were already fixed in #1375. We already wrap `is_file()` in a try-catch, so all we fall back to `StringScannable` either way

rust/hook/src/lib.rs
```diff
@@ -359,7 +359,7 @@ const UTF8_TO_WORSE_OTHER_ENCODING_RATIO: u64 = 4;
 /// `None` means "nothing to scan here": the file is too large to be scanned at
 /// all, so reading it would only cost the memory.
 fn scannable(config: &config::Config, payload: &mut Payload) -> Option<(String, String)> {
-    if payload.tool == Some(Tool::Read) {
+    if payload.tool == Some(Tool::Read) && !cannot_be_a_path(&payload.identifier) {
```
6d7a, line 362: Same here. If `path.is_file()` fails, we fall back to scanning the content in any case.

PR #1440, opened 2026-08-27
feat(machine): install the account's plugins during machine setup

### LINKED ISSUE #1439: `ggshield machine setup` should also install the eligible platform plugins
`ggshield machine setup` should also install the eligible platform plugins

### Context

`ggshield machine setup` configures three protections in one idempotent run: AI hooks, global/system git hooks, and a honeytoken (`ggshield/cmd/machine/setup.py`). It does **not** touch plugins.

That leaves the machine half-set-up for accounts that are entitled to them: `ggshield machine doctor` already checks the `machine_scan` plugin (native scanner loads, `endpoints:send` scope) and, when it is missing, tells the user to go and run `ggshield plugin install machine_scan` by hand. `ggshield machine scan` likewise only appears once that plugin is installed (it merges into the group via `PluginAwareLazyGroup(plugin_scope="machine")`). So the one command that is supposed to set the machine up is the one command that does not install the thing `doctor` then asks for.

### We don't have to guess the plan

The concern that this is "Business-only, so setup can't know" is already solved server-side. `GET /v1/endpoints/plugins` (`PluginAPIClient.get_available_plugins`) returns per-plugin entitlement:

- `available: true` + `latest_version` → installable for this account;
- `available: false` + `reason` → not entitled, with a human-readable reason;
- `404` → plugin system not enabled on the workspace (`PluginsNotEnabledError`).

So `machine setup` can simply ask, and install what comes back as available. No plan hardcoding, no new endpoint.

### Proposal

Add a fourth step to `machine setup`, after the existing three:

1. Fetch the catalog. On `PluginsNotEnabledError`, a non-entitled plugin, or any network/API error: print one informational line and **skip** — this must not make `machine setup` fail. (Contrast with the honeytoken step, which returns non-zero on a free plan — out of scope here, but worth a separate look.)
2. Install what is available, reusing the existing path: `PluginAPIClient.download_plugin` → `PluginDownloader.download_and_install` → `enable_installed_plugin` + `EnterpriseConfig.save()`, i.e. what `_install_from_gitguardian` in `ggshield/cmd/plugin/install.py` already does. That code needs a small extract into a reusable helper, since the command version calls `ctx.exit()` on every error path.
3. Idempotency, as everywhere else in `setup`: already installed and up to date → report and do nothing; installed but older → update (or leave to `plugin update`, see open questions).
4. Signature verification stays **STRICT**. No `--allow-unsigned` equivalent on `setup`.
5. New opt-out flag `--no-plugins`, consistent with `--no-ai-hooks` / `--no-git-hooks` / `--no-honeytokens`.

Root / MDM already works: `PluginDownloader` lands the wheel in `get_system_plugins_dir()` when `is_root()`, which is exactly the fleet case `machine setup --system` targets.

### Why it's worth doing

- A single root-run `machine setup` fully provisions a fleet machine, including endpoint scanning.
- `machine doctor` stops asking for a manual step that `setup` could have done.
- Accounts without entitlement are unaffected — they see one skipped line.

### Open questions

- **Which plugins?** Install every `available: true` entry in the catalog (future-proof, no hardcoded name), or only `machine_scan` (the one `machine doctor` and `machine scan` depend on)? Catalog-driven is the same amount of code; the risk is silently installing something new the day it ships in the catalog.
- **Update-on-rerun?** Should a re-run of `setup` upgrade an outdated plugin, or is that strictly `ggshield plugin update`'s job?
- **Prompt in interactive runs?** Silent install is right for MDM; interactive users might expect to be asked before a wheel is downloaded.

clement-tourriere:
Closes #1439.

`ggshield machine setup` configured the AI hooks, the git hooks and a honeytoken, then left `machine_scan` — the plugin that provides `ggshield machine scan`, and that `ggshield machine doctor` checks for — to a manual `ggshield plugin install`. So the command meant to set the machine up was the one that didn't install what doctor then asks for.

### Entitlement is asked, not guessed

The catalog (`GET /v1/endpoints/plugins`) already answers per account: `available: true` + `latest_version`, `available: false` + `reason`, or 404 when the plugin system is off for the workspace. Setup asks and installs what comes back available — nothing about plans is hardcoded, and no new endpoint was needed.

### What the step does

- Installs every `available: true` plugin in the catalog, with signature verification **STRICT** (no `--allow-unsigned` equivalent on `setup`).
- Nothing available, plugin system off, or an unreachable catalog → one line and **skip**. A free-plan account is a normal outcome, not a setup failure, no more than having no AI assistant installed is. Unavailable plugins print the server's `reason`.
- An *entitled* plugin failing to install → warning and non-zero exit, but the loop continues to the next plugin.
- Already installed → left untouched, with a `ggshield plugin update` pointer when the catalog has something newer. `setup` adds what is missing; it never swaps a working plugin out from under a machine.
- New `--no-plugins` opt-out, alongside `--no-ai-hooks` / `--no-git-hooks` / `--no-honeytokens`.
- A root run installs machine-wide for free: `PluginDownloader` already routes to the system plugin dir under `is_root()`, so one root `machine setup` provisions a fleet machine end to end.

### Refactor (first commit)

`_install_from_gitguardian` turned every failure into a `ctx.exit()`, so nothing else could reuse the download → verify → enable sequence. It moves to `ggshield/core/plugin/installer.py` as `install_plugin_from_platform()`, raising typed errors instead; `plugin install` keeps its exact exit-code behavior, and setup gets to keep going. The helper takes an optional `PluginDownloader` / `EnterpriseConfig` so a caller installing several plugins reuses one of each.

### Decisions worth a second opinion

Both were open questions on the issue; I picked and documented them in the code:

1. **Catalog-driven, not `machine_scan` by name** — future-proof and the same amount of code, but it does mean a plugin added to the catalog later starts installing itself on `setup`. Say the word if you'd rather it were a hardcoded allowlist.
2. **No upgrade on re-run** — upgrading stays `ggshield plugin update`.

### Testing

- 2715 unit tests pass, including 9 new `TestSetupPlugins` cases (installs all available, skips unentitled with reason, plugin system off, unreachable catalog, already installed, outdated → update hint, failed entitled install → exit 1, one failure doesn't stop the rest) and the updated orchestration harness.
- black / isort / flake8 clean, pyright 0 errors, import-linter both contracts KEPT — the new imports are all `cmd → core`, so no new layering exception was needed.
- Not exercised against a live instance: that installs a real wheel on a real machine, so it wants a reviewer with an entitled account.

### DIFF
--- a/ggshield/cmd/machine/setup.py
+++ b/ggshield/cmd/machine/setup.py
@@ -17,6 +17,15 @@ from ggshield.cmd.install import (
 from ggshield.cmd.utils.common_options import add_common_options
 from ggshield.cmd.utils.context_obj import ContextObj
 from ggshield.core import ui
+from ggshield.core.client import create_client_from_config
+from ggshield.core.config.enterprise_config import EnterpriseConfig
+from ggshield.core.plugin.client import (
+    PluginAPIClient,
+    PluginInfo,
+    PluginsNotEnabledError,
+)
+from ggshield.core.plugin.downloader import PluginDownloader
+from ggshield.core.plugin.installer import install_plugin_from_platform
 from ggshield.utils.os import is_root
 from ggshield.verticals.ai.agents import AGENTS
 from ggshield.verticals.ai.installation import (
@@ -44,6 +53,11 @@ _GIT_HOOK_TYPES = ("pre-commit", "pre-push")
     "--no-honeytokens",
     is_flag=True,
     help="Do not plant a honeytoken on this machine.",
+)
+@click.option(
+    "--no-plugins",
+    is_flag=True,
+    help="Do not install the plugins available to this account.",
 )
 @click.option(
     "--agent",
@@ -76,6 +90,7 @@ def setup_cmd(
     no_ai_hooks: bool,
     no_git_hooks: bool,
     no_honeytokens: bool,
+    no_plugins: bool,
     agents: Tuple[str, ...],
     exclude_agents: Tuple[str, ...],
     system: bool,
@@ -86,13 +101,18 @@ def setup_cmd(
 
     Configures every protection in one idempotent run: the ggshield AI hook for
     each detected AI coding assistant, the global git pre-commit/pre-push hooks,
-    and a honeytoken to detect endpoint intrusion. Safe to re-run — it adds what
-    is missing and leaves existing entries untouched.
+    the plugins this account is entitled to (e.g. `machine_scan`, which provides
+    `ggshield machine scan`), and a honeytoken to detect endpoint intrusion. Safe
+    to re-run — it adds what is missing and leaves existing entries untouched.
 
     Each protection is on by default; drop one with `--no-ai-hooks`,
-    `--no-git-hooks`, or `--no-honeytokens`. `--agent` / `--exclude-agent`
-    narrow which assistants get the AI hook. When run as root (or with
-    `--system`), the git hooks are installed machine-wide for every user.
+    `--no-git-hooks`, `--no-plugins`, or `--no-honeytokens`. `--agent` /
+    `--exclude-agent` narrow which assistants get the AI hook. When run as root
+    (or with `--system`), the git hooks and the plugins are installed
+    machine-wide for every user.
+
+    Plugins are gated by the GitGuardian plan: an account with none available
+    simply skips that step, it is not a setup failure.
     """
     if agents and exclude_agents:
         raise click.UsageError("--agent and --exclude-agent cannot be used together.")
@@ -105,6 +125,10 @@ def setup_cmd(
 
     if not no_git_hooks:
         if not _setup_git_hooks(system):
+            failed = True
+
+    if not no_plugins:
+        if not _setup_plugins(ctx):
             failed = True
 
     if not no_honeytokens:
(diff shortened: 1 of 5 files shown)


amascia-gg:
`machine doctor` skips anything not `enabled` so aplugin whose wheel is on disk but whose config row is disabled or missing gets `machine_scan already installed v1.2.0`. We should skip only when installed and enabled in `load_effective()`, otherwise call `enable_installed_plugin` on the existing wheel.

ggshield/cmd/machine/setup.py
```diff
@@ -209,6 +233,84 @@ def _setup_git_hooks(system: bool) -> bool:
...
+
+    try:
+        client = create_client_from_config(config)
+        plugin_api_client = PluginAPIClient(client)
+        catalog = plugin_api_client.get_available_plugins()
+    except PluginsNotEnabledError:
+        click.echo("  the plugin system is not enabled on this workspace; skipped")
+        return True
+    except Exception as exc:  # noqa: BLE001 - an unreachable catalog is not a failure
+        ui.display_warning(f"  could not fetch the plugin catalog: {exc}; skipped")
```
amascia-gg, line 262: an auth failure is reported as catalog unreachable if I'm not mistaken since it would raise APIKeyCheckError

ggshield/cmd/machine/setup.py
```diff
@@ -209,6 +233,84 @@ def _setup_git_hooks(system: bool) -> bool:
     return ok
 
 
+def _setup_plugins(ctx: click.Context) -> bool:
```
amascia-gg, line 236: this is not passed `--system` so non-root `machine setup --system` installs git hooks machine-wide and plugins per-user.

amascia-gg:
doctor.py:258 still says reinstall it with 'ggshield plugin install machine_scan'`; machine setup is now the answer.

tests/unit/cmd/machine/test_setup.py
amascia-gg, line 95: This is making live api calls to api.gitguardian.com now that we added the plugin check in the command

ggshield/cmd/machine/setup.py
```diff
@@ -209,6 +233,84 @@ def _setup_git_hooks(system: bool) -> bool:
...
+
+
+def _update_hint(plugin: PluginInfo, installed_version: str) -> str:
+    """Report the installed version, and point at `plugin update` if it is behind.
+
+    `setup` deliberately does not upgrade: it adds what is missing and leaves
+    what is there alone, so a re-run never swaps a working plugin out from under
+    a machine. Upgrading stays an explicit `ggshield plugin update`.
+    """
+    if plugin.latest_version and plugin.latest_version != installed_version:
```
amascia-gg, line 306: this compares strings as versions. Should we use `packaging.version` 's `Version` to do something like  `Version("1.4.0") > Version("1.3.0")`

amascia-gg:
A plugin can be filed on disk under a different name than the one setup asks for

Three names are in play for the same plugin:

1. the catalog reference the server uses (`machine_scan`)
2. the distribution name of the wheel that gets shipped (the downloader's own docstring gives `satori-python` as the example)
3. the entry-point name the wheel declares

When setup installs a plugin, `download_and_install` names the folder on disk after #2, the wheel's distribution name, not after the catalog reference (`downloader.py:342`).

Then on the next run, setup asks "is this already installed?" using #1, the catalog reference. That lookup (`_resolve_plugin_dir`) succeeds in one of two ways: a folder literally named `machine_scan`, or a folder whose declared entry-point name is `machine_scan`. If the wheel ships as `satori-python` and its entry point isn't `machine_scan` either, neither way matches.

PR #1441, opened 2026-08-27
fix(machine): report unavailable protections instead of failing

clement-tourriere:
Splits out the three uncontested commits of #1410 (@amascia-gg, authorship preserved) and rebases them onto current main. **Nothing here changes `DEFAULT_SCOPES`** — the `feat(auth): request the honeytokens:write scope by default` commit stays on #1410, where the design question raised in review is still open.

### The problem

`ggshield auth login` requests `scan`, `honeytokens:check`, `endpoints:send`, `ai-discover:send` — **not** `honeytokens:write`. So a token from a plain login can never plant a honeytoken, and `ggshield machine setup` ends its run with a raw `endpoint-deployments API error (403): …` per target and exit code 1. `ggshield machine doctor` fails the same way on the missing scope.

Neither is a broken machine. `honeytokens:write` is granted only on a plan that offers honeytokens, and only to a Manager; `ai-discover:send` is likewise plan-gated. A member's laptop is correctly set up and still reports red — and under MDM that gates a rollout on something no machine can fix.

### What changes

- **`machine setup`** checks the token's scopes before planting. Without `honeytokens:write` it reports the protection as unavailable and moves on, instead of running a doomed API call and failing the whole setup. Scopes it cannot read still go through `plant`, which surfaces the real error.
- **`machine doctor`** grows an `optional` check: printed with its fix and a yellow `!`, excluded from the exit code. Applied to `honeytokens:write` and `ai-discover:send` — both plan-gated, and neither is needed by a protection that is already in place (honeytoken planting skips without its scope, and `ai discover` is a separate command). `scan` stays fatal, and `endpoints:send` stays fatal when the `machine_scan` plugin is installed: installed-but-cannot-upload *is* a real break.
- **`honeytoken plant`** answers a 403 with the three prerequisites (honeytoken module enabled, Manager access level, `honeytokens:write` on the token) instead of the raw API body, which never said which one was missing.
- **`auth login --scopes`** now falls through to a fresh login when the current token lacks a requested scope. Before, it reused the still-valid token and reported "already authenticated" without ever asking for it — which made every `↳ fix:` line in doctor that says "re-run auth login" a no-op. The doctor fixes now say `auth login --scopes <scope>`.

### Testing

- Full unit suite: 2717 passed, 4 skipped. black / isort / flake8 clean, import-linter both contracts KEPT.
- The only conflict from the rebase was a test import line (`Mock` vs `MagicMock`).

### Notes for review

- Lightly conflicts with #1440 (both touch `setup.py` imports and the area around `_setup_honeytokens`); whichever lands second needs a two-line resolution.
- The remaining question on #1410 — whether `honeytokens:write` belongs in `DEFAULT_SCOPES`, or whether honeytoken planting needs a scope/feature redesign first, given it is Manager-only and mostly run via MDM — is deliberately untouched here.

### DIFF
--- a/changelog.d/20260817_120000_achille.mascia_plant_forbidden_message.md
+++ b/changelog.d/20260817_120000_achille.mascia_plant_forbidden_message.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- `ggshield honeytoken plant` explains a `403` (honeytoken module disabled, insufficient
+  user permissions, or a token without `honeytokens:write`) instead of printing the raw
+  API response body.
--- a/changelog.d/20260817_121000_achille.mascia_login_honors_requested_scopes.md
+++ b/changelog.d/20260817_121000_achille.mascia_login_honors_requested_scopes.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- `ggshield auth login --scopes` no longer reports "already authenticated" when the
+  configured token lacks one of the requested scopes: it authenticates again to get a
+  token that carries them.
--- a/changelog.d/20260817_122000_achille.mascia_machine_unavailable_protections.md
+++ b/changelog.d/20260817_122000_achille.mascia_machine_unavailable_protections.md
@@ -1,0 +1,11 @@
+### Changed
+
+- `ggshield machine setup` skips honeytoken planting when the token has no
+  `honeytokens:write` scope, instead of failing the run on a workspace that cannot be
+  granted it.
+
+- `ggshield machine doctor` reports the plan-gated scopes it cannot require
+  (`honeytokens:write`, `ai-discover:send`) as unavailable (`!`) instead of failing, so
+  it no longer exits non-zero on a workspace whose plan does not offer them. The
+  `endpoints:send` scope stays required, since installing the `machine_scan` plugin is
+  an explicit opt-in.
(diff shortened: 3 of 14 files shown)


amascia-gg:
Thanks for taking this over! 

Splitting the fourth commit out removed the only feedback that a requested scope was refused,
and I had not spotted that. `_warn_missing_scopes` (login.py:29) only diffs against
`DEFAULT_SCOPES`, which no longer contains `honeytokens:write`. So for a non-Manager:

1. `machine setup` prints `skipped: ... run ggshield auth login --scopes honeytokens:write`
2. they run it, and commit 2 now forces a real re-login, minting a new PAT (the old one is
   not revoked, we only do that on `logout --revoke`)
3. `_warn_missing_scopes` compares against `DEFAULT_SCOPES` only, so nothing is printed. They
   see "Success! You are now authenticated."
4. `machine doctor` still shows `!` with the same instruction, so they try again

Every attempt leaves another token behind, with no explanation and no way to succeed. On
#1410 the fourth commit hid this, so it only shows up in this split. Two small changes:

- `_warn_missing_scopes`: diff against `DEFAULT_SCOPES + extra_scopes` so a refused
  `--scopes` is reported
- the skip and doctor fix text: stop pointing non-Managers at a scope only a Manager can get,
  and say the role or plan cannot grant it instead

PR #1444, opened 2026-08-31
fix(packages): make the rpm own its bundle directories

benjaminrigaud-gg:
## TL;DR

Follow-up to #1443 (merged). Makes the rpm own its bundle directories so upgrades remove stale `*.dist-info` natively, matching what the deb already does.

## Why

#1443 added an rpm `%posttrans` that prunes empty `*.dist-info` on upgrade, repairing already-broken hosts. This is the structural half: the rpm now **owns** the bundle directories, so rpm removes a stale dependency's directory on upgrade instead of stranding an empty one.

## Change

`type: tree` (instead of `expand: true`) on the bundle content entry. nfpm then emits directory ownership in the rpm; the deb already had it.

## Verified (nfpm 2.36.1, mock bundle)

- `type: tree` produces the **same file set** as `expand: true`, plus directory entries: the rpm gains owned dirs (`_internal` and every `*.dist-info`) where `expand: true` owned **0**.
- Fedora upgrade test: install old rpm (owns dirs, ships `typeguard-4.3.0.dist-info`) then upgrade to a new rpm (ships `typeguard-4.5.2`, no 4.3.0) removes the stale `typeguard-4.3.0.dist-info`. With `expand: true` it persists.
- deb output unchanged (it already carried directory entries).

## Scope

The `%posttrans` from #1443 stays: it repairs hosts already on `expand: true` builds, which this change cannot retroactively clean (rpm only removes directories the installed package owned).

Refs: END-944

### DIFF
--- a/changelog.d/20260831_140652_benjamin.rigaud_rpm_own_bundle_dirs.md
+++ b/changelog.d/20260831_140652_benjamin.rigaud_rpm_own_bundle_dirs.md
@@ -1,0 +1,4 @@
+### Changed
+
+- The Linux rpm now owns its bundle directories, so upgrades no longer leave
+  empty `*.dist-info` directories behind.
--- a/scripts/build-os-packages/nfpm.yaml
+++ b/scripts/build-os-packages/nfpm.yaml
@@ -21,13 +21,16 @@ umask: 0o022
 contents:
   # The Rust dispatcher, not the PyInstaller launcher: `ggshield` is the entry
   # point and `ggshield-py` (which it execs) sits next to it in
-  # /usr/libexec/ggshield, picked up by the `expand: true` entry below.
+  # /usr/libexec/ggshield, picked up by the bundle entry below.
   - src: ../libexec/ggshield/ggshield
     dst: /usr/bin/ggshield
     type: symlink
 
+  # type: tree so the rpm owns the bundle dirs and prunes stale *.dist-info on
+  # upgrade; expand: true is still required to expand ${PYINSTALLER_OUTPUT_DIR}.
   - src: ${PYINSTALLER_OUTPUT_DIR}
     dst: /usr/libexec/ggshield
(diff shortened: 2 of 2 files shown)


amascia-gg (approved):

benjaminrigaud-gg:
### Local Fedora QA (PR #1444)

  Method: extracted the real 1.45.0 and 1.54.0 ggshield bundles, repackaged each with #1444's exact nfpm change (type: tree + expand: true + the merged %posttrans), then ran real
  install/upgrade in fedora:latest.

  ```
┌─────┬───────────────────────────────────────────────────────────┬───────────────────────┬─────────────────────┐
  │  #  │                         Scenario                          │  ggshield --version   │ typeguard-4.3.0 dir │
  ├─────┼───────────────────────────────────────────────────────────┼───────────────────────┼─────────────────────┤
  │ A   │ Fresh install of #1444 rpm                                │ 1.54.0, rc=0 ✅       │ n/a                 │
  ├─────┼───────────────────────────────────────────────────────────┼───────────────────────┼─────────────────────┤
  │ B   │ Control (current expand build): real 1.45.0 → real 1.54.0 │ crash, rc=1 (the bug) │ PRESENT             │
  ├─────┼───────────────────────────────────────────────────────────┼───────────────────────┼─────────────────────┤
  │ C   │ Real deployment: real 1.45.0 (expand) → #1444 rpm         │ 1.54.0, rc=0 ✅       │ GONE                │
  ├─────┼───────────────────────────────────────────────────────────┼───────────────────────┼─────────────────────┤
  │ D   │ Future: #1444 1.45.0 → #1444 1.54.0 (both own dirs)       │ 1.54.0, rc=0 ✅       │ GONE                │
  └─────┴───────────────────────────────────────────────────────────┴───────────────────────┴─────────────────────┘
```

PR #1446, opened 2026-09-03
fix(auth): let Ctrl+C abort the login callback wait on Windows

benjaminrigaud-gg:
## Context

On Windows, `ggshield auth login` (browser flow) cannot be interrupted with Ctrl+C while it waits for the browser callback. The command stays blocked until the login completes in the browser or the terminal is closed. Linux and macOS abort as expected.

Root cause: the local callback server waits in `handle_request()` with no timeout, so `select()` blocks until the redirect arrives. Python runs signal handlers only between bytecodes. On POSIX, SIGINT interrupts `poll()` with `EINTR`, so the handler runs at once. On Windows, Ctrl+C is delivered on a separate console thread that only sets a flag, WinSock `select()` is never woken, and the `KeyboardInterrupt` (mapped to `click.Abort`) only fires once the callback arrives. Known CPython limitation, see https://github.com/python/cpython/issues/133223.

## What has been done

- Give the callback server a 0.5s timeout (`CALLBACK_POLL_INTERVAL_SECONDS`). The wait loop already re-enters `handle_request()` until the callback is processed, so the only change is that a pending interrupt is now raised within that bound.
- Give the request handler the same timeout. A peer that connects but never sends a request line (browser pre-connect, port scan) used to block `readline()` with no timeout, one level below the server's `select()`. The stdlib closes such a connection on the read timeout and the wait resumes.
- Two regression tests in `tests/unit/verticals/auth/test_oauth.py`, both marked `enable_socket` because unit tests run with `--disable-socket`:
  - flip the finished flag from another thread and check the wait loop notices;
  - raise SIGINT from a helper thread (this trips Python's flag without waking `select()`, the same shape as Windows delivery) and check `click.Abort` is raised within the bound. If the interrupt never fires, a 5s safety net ends the loop so the test fails with `DID NOT RAISE` instead of hanging, and both timers are cancelled so a stray SIGINT cannot hit a later test;
  - connect a silent peer first, then flip the finished flag, and check the loop still returns promptly.

  Both fail without the timeout and pass with it on Linux. The `windows-2022` legs run them with real Windows signal machinery.
- Changelog fragment.

## Validation

- `uv run pytest --disable-socket tests/unit/verticals/auth tests/unit/cmd/auth`: 182 passed.
- Red/green checked: with the previous `oauth.py`, the two new tests fail (loop still alive after 3s; abort only after the 5s safety connection).
- Probe scripts on Linux: `KeyboardInterrupt` after 3.00s without the timeout (only once a connection arrived), 0.50s with it.
- Windows VM (Server 2022, Python 3.13.15), real console Ctrl+C with a silent peer connected to the callback port: without the handler timeout the process was still running 10s after Ctrl+C; with it, exit after 0.30s and 0.30s. Plain Ctrl+C: 0.33s.
- Manual, on Windows: run `ggshield auth login`, do not complete the browser login, press Ctrl+C. The command should exit within about half a second. On Linux/macOS it still aborts immediately.

## PR check list


benjaminrigaud-gg:
**QA on a Windows VM** (Windows Server 2022 Standard, build 20348, Python 3.13.15)

Compared the released 1.54.0 against this branch's wheel in the same venv. The driver starts `python -m ggshield auth login` in its own console, waits 5s for it to reach the callback wait, then sends a real `CTRL_C_EVENT` to that console and times the exit.

| Build | In-process SIGINT probe | Real console Ctrl+C |
|---|---|---|
| 1.54.0 | `click.Abort` only once a connection arrived (7.99s) | still running 10s after Ctrl+C, killed |
| this branch | `click.Abort` after 0.51s | exited after 0.35s, 3.34s, 0.80s, 0.63s, 0.64s (5 runs), prints `Aborted!` |

Notes: the 3.34s run overlapped with the browser launch on a loaded 4-core VM. One run exited with `STATUS_CONTROL_C_EXIT` (0xC000013A) instead of click's exit code 1, still within a second. The five Windows unit-test legs (Python 3.9 to 3.13) are green in CI.

ggshield/verticals/auth/oauth.py
```diff
@@ -265,13 +269,17 @@ def _redirect_to_login(self) -> None:
...
+            server = HTTPServer(
                 ("127.0.0.1", 0),
                 functools.partial(RequestHandler, self),
             )
-            self._port = self.server.server_port
         except OSError:
             raise UnexpectedError("Could not find unoccupied port.")
+        # Windows Ctrl+C only sets a flag that Python can act on once the
+        # blocking select() in handle_request() returns, so never block forever.
+        server.timeout = CALLBACK_POLL_INTERVAL_SECONDS
```
clement-tourriere, line 280: I think this is not enough. `server.timeout` is only for the `select()` before accept. If something connect to the port but don't send the request line (browser pre-connect, port scan, ...), we go in `handle_one_request()` and it block in `rfile.readline()` with no timeout. So the Ctrl+C is ignored again on Windows, same bug just one level deeper.

Maybe we can also put the timeout on the handler:

```python
class RequestHandler(BaseHTTPRequestHandler):
    timeout = CALLBACK_POLL_INTERVAL_SECONDS
```

I tested quickly: with a peer that connect and send nothing, `handle_request()` return after 0.5s instead of waiting the peer to close, and a normal GET still work fine (`handle_one_request` catch the `TimeoutError` and just close the connection).

tests/unit/verticals/auth/test_oauth.py
```diff
@@ -173,3 +179,72 @@ def test_print_login_success_without_account_raises_cleanly():
...
+         without interrupting the main thread's blocking select() (this is how
+         Windows delivers Ctrl+C)
+    THEN the wait aborts within the poll interval instead of hanging until a
+         request arrives
+    """
+    client = listening_client
+    threading.Timer(0.2, signal.raise_signal, args=(signal.SIGINT,)).start()
+    unblock = threading.Timer(CALLBACK_WAIT_LIMIT + 2, _poke, args=(client.server,))
+    unblock.daemon = True
+    unblock.start()
```
clement-tourriere, line 242: I'm not sure this safety net work. The `_poke` make one `handle_request()` return, but the `while not self._request_finished` loop just call it again. So if the `KeyboardInterrupt` never come, the test hang forever instead of failing (I tried with SIGINT ignored in the parent process, pytest never finish).

Maybe set `client._request_finished = True` in the unblock callback before the poke, so the test fail with `DID NOT RAISE` instead of hanging.

Also the SIGINT timer is never cancelled. If `_wait_for_callback` raise something else before, the SIGINT fire 0.2s later during the next test and kill the whole pytest session. A `try/finally` to cancel it would be more safe I think.

clement-tourriere (changes requested):
Thanks for the fix, the poll approach is good. But I think there is 2 things to change before merge, see comments.

clement-tourriere (approved):

PR #1447, opened 2026-09-04
feat(machine): add doctor --strict to fail on unavailable protections

amascia-gg:
Stacked on #1441, which introduces the `optional` check state this builds on. Base will
retarget to `main` once that merges. #1441 itself is untouched.

### The problem

#1441 exempts the plan-gated scopes from doctor's exit code, because a workspace whose
plan does not offer honeytokens can never be granted `honeytokens:write`, and failing
there would gate an MDM rollout on something no machine can fix.

That exemption is unconditional, so the opposite case is now silent: a fleet on a plan
that *does* offer the protection, whose service-account token is simply missing the
scope, gets `This machine is correctly set up` and exit 0. That is the case the scope
check was added for.

### Why a flag

ggshield cannot tell the two apart. `/v1/api_tokens/self` returns a token's scopes,
`member_id`, `workspace_id` and type; `/v1/metadata` returns scan preferences. Neither
exposes the workspace's entitlements or the member's access level, so no client-side
inference is available.

I looked at inferring it from `honeytokens:check` (whose only gate is the workspace
flag, with no role check) combined with the token type, but it breaks exactly in the
MDM case: a service-account token is minted with hand-picked scopes, so a missing
`honeytokens:check` says nothing about the workspace. So the caller declares intent
instead.

### What changes

`--strict` clears `optional` on every check, which leaves `_render` and the exit-code
expression unaware of the distinction.

| | default | `--strict` |
| --- | --- | --- |
| plan-gated scope missing | `!` yellow, excluded from exit code | `✗` red, counted |
| summary | `correctly set up. 2 optional protection(s) unavailable` | `2 check(s) failed` |
| exit code | 0 | 1 |

Verified on a laptop whose token lacks both `honeytokens:write` and `ai-discover:send`:
exit 0 by default, exit 1 with `--strict`, and a healthy machine still exits 0 either
way. Full unit suite 2720 passed / 4 skipped; black, flake8 and ty clean.

### DIFF
--- a/changelog.d/20260904_120000_achille.mascia_doctor_strict.md
+++ b/changelog.d/20260904_120000_achille.mascia_doctor_strict.md
@@ -1,0 +1,8 @@
+### Added
+
+- `ggshield machine doctor --strict` fails on protections reported as unavailable
+  (`!`) instead of only on a misconfigured machine. The API exposes a token's scopes,
+  not the workspace's entitlements, so ggshield cannot tell "this plan cannot grant the
+  scope" from "this token is missing it"; `--strict` is how a caller that knows the
+  workspace offers the protection gates on it, for instance an MDM rollout on a
+  Business plan.
--- a/ggshield/cmd/machine/doctor.py
+++ b/ggshield/cmd/machine/doctor.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from importlib import import_module
 from typing import Any, List, Optional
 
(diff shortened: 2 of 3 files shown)


amascia-gg:
Closing: `--strict` has no consumer today. Nothing in the mdm or qa-machine-testing repos calls `ggshield machine doctor`, so the flag would ship ahead of the script that would use it. The underlying gap (doctor cannot tell "this plan cannot grant the scope" from "this token is missing it") stays open; worth reviving this together with an actual fleet gate.

PR #1451, opened 2026-09-08
feat(ai-hook): add Junie CLI support

amascia-gg:
Junie CLI, JetBrains' standalone terminal agent, gets the AI hook. Its hook API is a deliberate copy of Claude Code's, so the verdict schema is Claude's too; `project_path` is what tells the two apart, since Junie sends it and Claude never does. Junie has no PostToolUse event at all, so a secret that reaches the model through a tool result is out of scope for this assistant in a way it is not for the others.

The hook path is served by the Rust hook alone, following Kiro. The Python adapter carries what the Rust hook does not do: installing into `~/.junie/config.json`, MCP discovery, and shipping session events. A project-scoped install is refused by Junie itself, a repository not being allowed to run shell commands on checkout, so the post-install warning says so rather than leaving a file that never loads.

One thing only a real session shows: Junie runs a hook from its own home directory and reports *that* as `cwd`, naming the workspace in `project_path` instead. Reads resolve against `project_path` for that reason. Taking `cwd` at face value sends them to `~/.junie`, which means the wrong verdict-cache key and the default `ignored_paths` tested against the wrong root. Payload shapes otherwise come from the hook runtime in Junie's own jar, the docs abbreviating the envelope: `SessionEnd` carries only its event name and reason.

Verified against a real Junie 26.8.31 session: a prompt carrying an AWS key pair is refused and never reaches the model. `PreToolUse` is unverified live, needing a working model subscription. Only Junie CLI is covered; the IDE embeds Junie over ACP, which invokes no hooks.

Closes NHI-1971

### DIFF
--- a/changelog.d/20260907_170500_achille.mascia_junie_cli_support.md
+++ b/changelog.d/20260907_170500_achille.mascia_junie_cli_support.md
@@ -1,0 +1,7 @@
+### Added
+
+- ggshield now supports Junie CLI, JetBrains' terminal coding agent.
+  `ggshield machine setup` installs the secret-scanning hooks in
+  `~/.junie/config.json`, and MCP servers and past usage are reported like they
+  are for every other assistant. Junie inside a JetBrains IDE is not covered:
+  it embeds Junie over ACP, which invokes no hooks.
--- a/ggshield/verticals/ai/agents/__init__.py
+++ b/ggshield/verticals/ai/agents/__init__.py
@@ -5,6 +5,7 @@ from .claude_code import Claude
 from .codex import Codex
 from .copilot import Copilot
 from .cursor import Cursor
+from .junie import Junie
 from .kiro import Kiro
 from .vibe import Vibe
 from .vscode import VSCode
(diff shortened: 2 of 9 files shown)


amascia-gg:
Rebased onto main, which had moved under this branch. Two substantive changes beyond the rebase, both from driving Junie live.

**Junie fires no PostToolUse-class event at all**, so tool-output withholding is not possible here and none is added. Measured on Junie 26.9.7 rather than inferred: with a hook configured on every event name, two runs dispatched only `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `Stop` and `SessionEnd`. A configured `PostToolUse` entry is silently ignored rather than rejected, and Junie's own `DefaultExternalHookRuntime` log confirms zero such dispatches. The tool result then reaches the model in a `tool` role message with no hook in between, so the secret arrives unconditionally and the leaked wording is the correct wording.

The JetBrains AI proxy returns 403 for this account, so Junie was pointed at a local OpenAI-compatible stub via `--model custom:<profile>`, which made it possible to read back exactly what the model received.

**A real coverage gap, found by that same live session.** Junie renames its tools to Claude's spellings for hooks, with one exception: `open_entire_file`, its whole-file reader, keeps its own name and a `path` argument. It was falling through to `Tool::Other`, so the file's contents were never scanned before the agent read them. Demonstrated with the real payload shape against both builds:

```
origin/main      ->  no output (unrecognized, exit 1, read proceeds unscanned)
this branch      ->  deny, file scanned, secret found, read blocked
```

That failure was invisible: an unrecognized payload exits 1 with empty stdout, which every agent treats as a non-blocking hook error. Fixed by adding the tool to the reader list, with a regression test covering both Junie readers.

Also corrected a `read_range` comment that claimed no real payload had confirmed Junie's Read parameters. The measured shape is `offset` with no `limit`, so a range would run to end-of-file and whole-file scanning loses nothing.

**Note for whoever merges the tool-output stack** (#1480 to #1482): `can_redact_tool_output()` matches exhaustively on the agent, so adding `Agent::Junie` will not compile until it is given a value. That value is `false`.

Verification: 139 tests pass in the hook crate plus 6/38/5/6 elsewhere, clippy and fmt clean, equivalence gate `OK (2 known divergence(s), 0 unexpected)`.

PR #1454, opened 2026-09-09
ci: pin all dependencies to fixed versions and add a 3-day Dependabot cooldown

gregoiredx:
Every dependency now resolves to a fixed version, and Dependabot waits 3 days before proposing a new one.

Split out of one larger branch; the Dockerfile lockfile change and the cargo cooldown note are separate PRs.

### 3-day cooldown

`cooldown.default-days: 3` on all five ecosystems. A compromised release is usually yanked within hours of discovery, so waiting out that window keeps malicious-then-deleted versions from ever reaching a branch here.

Dependabot only watched `github-actions` before, so `docker`, `uv`, `cargo` (rooted at `/rust`) and `pre-commit` are new. The `pre-commit` entry ignores this repo's own hook: its `rev` tracks `ggshield/__init__.py:__version__`, so bumping it belongs to the release, not to Dependabot.

### Fixed versions

Every pin resolves to whatever the floating ref pointed at when it was written, verified by comparing tag commit SHAs (actions) and registry digests (images). Nothing changes behaviour; it only freezes the current state so reruns are reproducible.

- **17 action refs** across 8 workflows, pinned to full commit SHAs with a trailing `# vX.Y.Z` comment, e.g. `actions/checkout@3d3c42e5... # v7.0.1`. Tags are mutable, so a SHA is the only immutable pin; this also covers `codecov/codecov-action`, which has been breached before. Dependabot handles this format natively -- its `github_actions` updater has a `VersionCommenter` whose job is keeping that comment in sync -- so updates keep working. Every SHA was independently re-resolved from its tag to confirm it matches its comment.
- **Container images**: `python:3.10-slim` -> `3.10.21-slim`; `rockylinux:8` -> `8.10`, `:9` -> `9.8`; `debian:stable` -> `stable-20260824`; `ubuntu:latest` -> `26.04`; `opensuse/leap` -> `16.0`.
- **Grype installer**: was `curl .../anchore/grype/main/install.sh | sh`. The binary version was pinned but the *script* was fetched from a branch and piped into a shell, so upstream could change what ran in a release job at any time. Now a checksum-verified release tarball via the existing `scripts/download` helper, matching how `rustup-init` is already handled.

### Notes for the reviewer

- `dtolnay/rust-toolchain` is SHA-pinned with a `# v1` comment rather than a semver one: upstream publishes only `v1` and per-version branches (`stable` is a branch too, and moves every rustup release). The toolchain that actually compiles the binary was never floating: `hatch_build.py` runs cargo with `cwd=rust/`, so `rust/rust-toolchain.toml` pins it to 1.96.0.
- `ubuntu:latest` was resolving to `26.04`, a non-LTS rolling release. Pinned as-is to avoid mixing a behaviour change into this PR, but if those smoke tests are meant to mirror what users run, `ubuntu:24.04` is probably the intent. Happy to switch.
- The `case "${{ matrix.image }}"` branches in `build_release_assets.yml` match on image name; all five renamed images still route to the same branch.
- Still floating, deliberately: `actions/secret/action.yml` uses `docker://gitguardian/ggshield:v1.53.0` and `.pre-commit-hooks.yaml` pulls `gitguardian/ggshield:latest`. Both are self-references whose bump belongs to the release process, and the `:latest` one is the contract consumers rely on.
- `scripts/install/install.sh` and `install.ps1` are deliberately left alone: they resolve *latest ggshield* for end users and enforce a mandatory sha256 from the GitHub API. Pinning them would defeat their purpose.

Verified with `check-jsonschema` (`vendor.github-workflows` + `vendor.dependabot`, the same hooks pre-commit runs). The grype step was dry-run locally, and its checksum confirmed to fail closed under GNU coreutils: correct hash passes, tampered hash exits nonzero.

### DIFF
--- a/.github/workflows/build_release_assets.yml
+++ b/.github/workflows/build_release_assets.yml
@@ -35,10 +35,10 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@v7.0.1
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v7
+        uses: astral-sh/setup-uv@v7.6.0
         with:
           version: ${{ env.UV_VERSION }}
           python-version: ${{ env.DEFAULT_PYTHON_VERSION }}
@@ -57,7 +57,7 @@ jobs:
           /tmp/smoke/bin/ggshield --version
 
       - name: Upload packages
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@v7.0.1
         with:
           name: dist
           path: |
@@ -73,14 +73,14 @@ jobs:
         include:
           - os: ubuntu-22.04
             # When building on Linux we use a container to build using an old enough version
-            container: rockylinux/rockylinux:8
+            container: rockylinux/rockylinux:8.10
             arch: x86_64
             nfpm_arch: x86_64
             sha256sum:
               python: d6254462d5fe066810a472d286861f6a18f66400008bfa9688e202a5672e88cb
               nfpm: 9f8effa24bc6033b509611dbe68839542a63e825525b195672298c369051ef0b
           - os: ubuntu-24.04-arm
-            container: rockylinux/rockylinux:8
+            container: rockylinux/rockylinux:8.10
             arch: aarch64
             nfpm_arch: arm64
             sha256sum:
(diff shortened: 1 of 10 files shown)


.github/workflows/build_release_assets.yml
```diff
@@ -408,17 +412,25 @@ jobs:
...
+        # Not `curl .../main/install.sh | sh`: that fetches an unpinned script
+        # from a branch and pipes it into a shell, so upstream can change what
+        # runs here at any time. The release tarball is pinned and
+        # checksum-verified instead, like rustup-init in wheels.yml.
         run: |
-          curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh \
-            | sh -s -- -b /usr/local/bin "v${GRYPE_VERSION}"
+          scripts/download \
+            "https://github.com/anchore/grype/releases/download/v${GRYPE_VERSION}/grype_${GRYPE_VERSION}_linux_amd64.tar.gz" \
+            /tmp/grype.tar.gz \
```
benjaminrigaud-gg, line 430: Is there an easy way to check the signature there? 🕵️

benjaminrigaud-gg (approved):

.github/workflows/build_release_assets.yml
gregoiredx: that's actually what`scrips/download` is doing
https://github.com/GitGuardian/ggshield/blob/93bef74d9f4b6bb3cdf39e9a784cffd11a934215/scripts/download#L61

gregoiredx: I haven't checked that it works though but it looks like the intent

gregoiredx:
@benjaminrigaud-gg , when you're satisfied with the PR, can you add the bypass changelog label and merge please?

.github/workflows/build_release_assets.yml
benjaminrigaud-gg: perfect.

PR #1485, opened 2026-09-10
feat(ai-hook): withhold a Cursor MCP response that holds a secret

amascia-gg:
Fourth in the stack, on #1482 → #1481 → #1480. Review those first.

Cursor can withhold an MCP tool's output and nothing else. Its agent fires `postToolUse` from exactly two places: the MCP execution path, which reads `updated_mcp_tool_output` from the hook's answer, and one other call site that discards the answer entirely. So an MCP response can be kept from the model, while a shell command's output and a file's content cannot, and those keep the leaked wording in the very same session.

That split is worth being explicit about, because it is the reverse of where most of the risk sits. `vault read`, `aws secretsmanager get-secret-value` and `kubectl get secret` are all shell commands, and Cursor cannot protect any of them. MCP responses are a real exfiltration path of their own, though, and this is the whole of what the agent allows, so the choice is partial coverage or none.

The replacement goes out as a bare string, which is one of the two shapes Cursor's handler accepts and the one needing no content-block wrapper.

**Verification status, please read before merging.** This is implemented from Cursor 3.19.19's own shipped code, not from its documentation, which for once agreed. I read the two call sites and the handler that parses the value. It is **not yet confirmed in a live Cursor session**, because Cursor's agent is GUI-only and driving it needs a human plus an MCP server that returns a credential. That is a weaker basis than the Claude Code, Codex and Vibe PRs in this stack, each of which was watched end to end in a real session. I would like a live confirmation before this merges; the other three do not depend on it.

NHI-2007

### DIFF
--- a/changelog.d/20260910_142900_achille.mascia_withhold_tool_output_secrets.md
+++ b/changelog.d/20260910_142900_achille.mascia_withhold_tool_output_secrets.md
@@ -12,3 +12,6 @@
   now reports the output as withheld.
 - Mistral Vibe behaves the same way as Codex and now reports a blocked tool
   output as withheld rather than leaked.
+- On Cursor, a secret in an MCP tool's response is now withheld from the
+  assistant. Cursor cannot replace a shell command's output or a file's
+  content, so those still report the secret as leaked.
--- a/rust/hook/src/message.rs
+++ b/rust/hook/src/message.rs
@@ -474,7 +474,7 @@ mod tests {
         );
         let leaked = from_secrets(
             &secrets,
-            &payload_for(Agent::Cursor, EventType::PostToolUse, Some(Tool::Mcp), "id"),
+            &payload_for(Agent::VsCode, EventType::PostToolUse, Some(Tool::Mcp), "id"),
         );
         assert!(withheld.contains("withheld from the agent"), "{withheld}");
         assert!(!withheld.contains("compromised"), "{withheld}");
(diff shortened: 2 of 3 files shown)


amascia-gg:
Closing this without merging. A live Cursor session disproved the premise.

Cursor 3.19.19 reads `updated_mcp_tool_output` in its source, and does not act on it. Tested with a minimal MCP server returning a fake Stripe key, a single hook configured, the payload debounce cleared, and a project path free of anything that could misdirect agent detection. Our side was correct: the payload was detected as Cursor and the hook emitted `updated_mcp_tool_output` with the secret censored. Cursor showed the model the key anyway.

Why the field is ignored is unexplained. The code path exists: after `executeHookForStep(afterMCPExecution, …)` it calls `firePostToolUse(…)`, tests the field and rebuilds the tool result from it. Candidates not run down are that `firePostToolUse` surfaces only a subset of a hook's stdout, that the rebuilt result is discarded by its caller, or that the path is gated in this build.

This PR shipped on a reading of Cursor's implementation rather than on a live session, which was a weaker standard than the one applied to Vibe earlier in the same stack, where a documented mechanism was refused until it was watched working. That inconsistency is why a wrong answer got this far.

Cursor therefore joins VS Code, Copilot CLI and Kiro as an agent where a tool output cannot be withheld. The rest of the stack, #1480, #1481 and #1482, is unaffected and each was confirmed in a live session.

Details and reproduction in NHI-2007.

PR #1346, opened 2026-07-03
fix(scan-pypi): fix pip not found and unsatisfied python requirement

beavuck:
## Context

Fixes #458
Fixes #394

`ggshield secret scan pypi` downloads the package to scan by shelling out to `pip download`. That has two problems.

**It depends on an external `pip` executable (#394).** If `pip` is not on the `PATH` (for instance when only `pip3` is available), the scan crashes with a raw traceback instead of a helpful message:

```
  File ".../ggshield/cmd/secret/scan/pypi.py", line 33, in save_package_to_tmp
    subprocess.run(
  ...
FileNotFoundError: [Errno 2] No such file or directory: 'pip'
```

**It filters candidates by the running interpreter's Python version (#458).** `pip download` refuses to fetch a package whose `Requires-Python` does not match the interpreter running `ggshield`, even though we only want to read the package's content, never install or run it:

```
$ ggshield secret scan pypi "numba==0.52.0"
Downloading package
ERROR: Ignored the following versions that require a different python version: 0.52.0 Requires-Python >=3.6,<3.9
ERROR: Could not find a version that satisfies the requirement numba==0.52.0
Error: Failed to download "numba==0.52.0"
```

## What has been done

- **Replace `pip download` with the `unearth` library.** `ggshield` no longer shells out to an external `pip`, so `secret scan pypi` no longer crashes when `pip` is missing or only available as `pip3` (394), and can now download a package whatever `Requires-Python` it declares, by resolving with the compatibility check disabled (458). This is the approach suggested by @agateau-gg.
- **Fail gracefully.** Lookup and download failures now produce a clear `UnexpectedError` instead of a raw traceback.
- **Preserve existing behavior.** The 30s download timeout is kept, `PIP_INDEX_URL` / `PIP_EXTRA_INDEX_URL` are still honored for custom indexes, and extraction still goes through `ggshield`'s own hardened `safe_unpack`.
- **New dependency.** This adds `unearth`, which transitively pulls in `httpx`, `httpcore`, `h11` and `anyio`.
- **Tests.** Behavior-focused unit tests, plus a functional test scanning `numba==0.52.0`: it requires `>=3.6,<3.9`, which no supported interpreter satisfies, so it exercises the 458 fix on every runner. While here, I did some clean-as-you-code: the pre-existing `# TODO` test for `get_files_from_package` was rewritten to assert the outcome (real archive in -> scannables out) rather than mock internals.

[...]

Issue #458, opened 2023-01-23
ggshield secret scan pypi fails if the Python version is not supported by pip


agateau-gg:
## Environment

- ggshield version: 1.14.2
- Operating system (Linux, macOS, Windows): Any
- Operating system version: Any
- Python version: Any

## Describe the bug

`ggshield secret scan pypi` uses the `pip` command to download packages to scan. If a package requires, for example, Python >= 3.11 and our version of `pip` is for Python 3.10, then the scan 

Here is an example trying to scan vkpybot, which requires 3.11 with a 3.10 Python.

```
$ ggshield secret scan pypi vkpybot
Downloading pip package... ERROR: Ignored the following versions that require a different python version: 0.1.10 Requires-Python >=3.11,<4.0; 0.1.11 Requires-Python >=3.11,<4.0; 0.1.3 Requires-Python >=3.11,<4.0; 0.1.4 Requires-Python >=3.11,<4.0; 0.1.5 Requires-Python >=3.11,<4.0; 0.1.6 Requires-Python >=3.11,<4.0; 0.1.7 Requires-Python >=3.11,<4.0; 0.1.8 Requires-Python >=3.11,<4.0
ERROR: Could not find a version that satisfies the requirement vkpybot (from versions: none)
ERROR: No matching distribution found for vkpybot
Error: Failed to download "vkpybot"
```

beavuck:
I'd like to take that one (and by extension #394), I'll create a PR soon

PR #1495, opened 2026-09-11
test(scan-pypi): swap tensorflow for numba in the functional test

benjaminrigaud-gg:
`test_scan_pypi[tensorflow==2.5.0]` has been failing on main with
`Failed to download "tensorflow==2.5.0": timed out after 30s`. The first
version of this PR raised the download budget. The review pointed at the
test subject instead, and that holds on both counts.

**It is too big.** With compatibility ignored, unearth picks the 433 MiB
cp39 manylinux wheel on the Linux runner. Finishing in 30 s needs 15 MB/s
sustained, and successful runs took 49 s to 118 s wall clock.

**It does not test what its comment says.** tensorflow 2.5.0 declares no
`Requires-Python` and is the first release to ship cp39 wheels. The API
functests run on Python 3.9 (`DEFAULT_PYTHON_VERSION`), where it is fully
pip-compatible, so the case passes with or without the #458 fix.

`numba==0.52.0`, the subject #1346 started with, declares
`Requires-Python >=3.6,<3.9`: refused by every supported interpreter under
the old pip lookup, about 2 MB, exit code 0 (exit code 1 stays covered by
`ggshield==1.14.2`). Checked with unearth's compatibility check on: Python
3.9 and 3.14 both reject it. The functional case runs in 4 s.

A unit test now also asserts `ignore_compatibility=True` on the
`PackageFinder`, so the #458 guard no longer depends on live PyPI alone.

### DIFF
--- a/ggshield/cmd/secret/scan/pypi.py
+++ b/ggshield/cmd/secret/scan/pypi.py
@@ -23,10 +23,15 @@ from ggshield.core.scan.file import create_files_from_paths
 from ggshield.core.scanner_ui import create_scanner_ui
 from ggshield.utils.archive import safe_unpack
 from ggshield.utils.files import ListFilesMode
+from ggshield.utils.os import getenv_int
 from ggshield.verticals.secret import SecretScanCollection, SecretScanner
 
 
 PYPI_DOWNLOAD_TIMEOUT = 30
+# Anything past an hour is a typo, and a big enough number overflows the float
+# the deadline is built from.
+MAX_PYPI_DOWNLOAD_TIMEOUT = 3600
+TIMEOUT_ENV_VAR = "GG_PYPI_DOWNLOAD_TIMEOUT"
 DEFAULT_INDEX_URL = "https://pypi.org/simple/"
 
 
@@ -40,20 +45,28 @@ def _get_index_urls() -> List[str]:
 
 def _download_timeout() -> int:
     """Seconds allowed for the whole download. `GG_PYPI_DOWNLOAD_TIMEOUT` raises
-    it for the functional test, whose package is 454 MB; a malformed value falls
-    back rather than failing the scan."""
-    raw = os.getenv("GG_PYPI_DOWNLOAD_TIMEOUT", "")
-    return int(raw) if raw.isdigit() else PYPI_DOWNLOAD_TIMEOUT
+    it for the functional test, whose package is 454 MB. A bad value is refused
+    rather than ignored: falling back silently would just time out later."""
+    try:
+        timeout = getenv_int(TIMEOUT_ENV_VAR, PYPI_DOWNLOAD_TIMEOUT)
+    except ValueError:
+        raise UnexpectedError(f"{TIMEOUT_ENV_VAR} must be a whole number of seconds.")
+    if not 1 <= timeout <= MAX_PYPI_DOWNLOAD_TIMEOUT:
+        raise UnexpectedError(
+            f"{TIMEOUT_ENV_VAR} must be between 1"
+            f" and {MAX_PYPI_DOWNLOAD_TIMEOUT} seconds."
+        )
+    return timeout
 
 
-def _enforce_deadline(deadline: float) -> None:  # pragma: no cover
+def _enforce_deadline(deadline: float, timeout: int) -> None:  # pragma: no cover
     """Give up once the download budget is spent."""
     if time.monotonic() > deadline:
-        raise TimeoutError(f"timed out after {_download_timeout()}s")
+        raise TimeoutError(f"timed out after {timeout}s")
 
 
 def _download_link(
-    finder: PackageFinder, link: Link, dest: Path, deadline: float
+    finder: PackageFinder, link: Link, dest: Path, deadline: float, timeout: int
 ) -> None:
     """Stream the distribution at `link` into `dest`, giving up once `deadline`
     is reached."""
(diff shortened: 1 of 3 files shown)


ggshield/cmd/secret/scan/pypi.py
```diff
@@ -38,10 +38,18 @@ def _get_index_urls() -> List[str]:
     return index_urls
 
 
+def _download_timeout() -> int:
+    """Seconds allowed for the whole download. `GG_PYPI_DOWNLOAD_TIMEOUT` raises
+    it for the functional test, whose package is 454 MB; a malformed value falls
+    back rather than failing the scan."""
+    raw = os.getenv("GG_PYPI_DOWNLOAD_TIMEOUT", "")
+    return int(raw) if raw.isdigit() else PYPI_DOWNLOAD_TIMEOUT
```
clement-tourriere: We already have `getenv_int` in `ggshield/utils/os.py`, and `GG_GIT_TIMEOUT` uses it (`git_shell.py:19`). Same kind of knob. Can we use it here too instead of a new parser?

Right now this one behaves differently from the other `GG_*` int vars: `getenv_int` raises on a bad value, this one silently falls back. And `"300 "` (trailing space, easy to get in a CI yaml) falls back to 30s with no message, so the test would fail again and nobody would know why.

```python
return getenv_int("GG_PYPI_DOWNLOAD_TIMEOUT", PYPI_DOWNLOAD_TIMEOUT)
```

If you really want the silent fallback, wrap it in `try/except ValueError`. Then the 3 new unit tests can go, `getenv_int` is already tested in `tests/unit/utils/test_os.py`.

ggshield/cmd/secret/scan/pypi.py
```diff
@@ -60,7 +68,7 @@ def _download_link(
 def save_package_to_tmp(temp_dir: Path, package_name: str) -> None:
     ui.display_heading("Downloading package")
 
-    deadline = time.monotonic() + PYPI_DOWNLOAD_TIMEOUT
+    deadline = time.monotonic() + _download_timeout()
```
clement-tourriere: The fallback does not cover all bad values, and this line is outside the `try`, so the command dies with a raw traceback (I ran these on the PR head):

- `GG_PYPI_DOWNLOAD_TIMEOUT=²`: `isdigit()` is True but `int()` raises `ValueError`
- a value with 309+ digits: `int()` is fine but `time.monotonic() + n` raises `OverflowError`
- `GG_PYPI_DOWNLOAD_TIMEOUT=0`: deadline is already expired, every scan fails with `timed out after 0s`

Please reject `<= 0` and catch the parse error (or just use `getenv_int`, see my other comment).

clement-tourriere (dismissed):
Thanks, the fix works: in this PR's CI the tensorflow case took ~52s, so 30s was really too short.

Two blocking things below, both about the env var parsing. The rest looks good to me.

ggshield/cmd/secret/scan/pypi.py
benjaminrigaud-gg: Switched to `getenv_int`. The `"300 "` case is what decided it: `int("300 ")` is
300 while `"300 ".isdigit()` is False, so my parser was stricter than the rest of
the codebase in exactly the direction that would have silently put the test back
on 30s.

Went past the try/except option though. A bad value is now refused rather than
ignored, because silently falling back is the confusing failure this PR exists to
remove.

Dropped the int-parsing tests. Kept a class for the new validation (bounds,
non-numeric, upper bound), which `getenv_int` does not cover.

ggshield/cmd/secret/scan/pypi.py
benjaminrigaud-gg: All three reproduced locally:

- `"²".isdigit()` is True, `int("²")` raises ValueError
- `time.monotonic() + 10**309` raises OverflowError
- 0 expires the deadline before the first chunk

Worth saying that `getenv_int` on its own fixes none of them: it still raises
ValueError on that character, still returns the 310-digit int, still accepts 0.
So there is an explicit range now, 1 to 3600. The upper bound does two jobs at
once: it rejects nonsense, and it keeps the value small enough that the float
add cannot overflow.

The timeout is also read once in `save_package_to_tmp` and passed down to
`_enforce_deadline`, so nothing re-parses the environment from inside the error
path.

Against the built CLI now:

```
GG_PYPI_DOWNLOAD_TIMEOUT=<superscript 2>  Error: GG_PYPI_DOWNLOAD_TIMEOUT must be a whole number of seconds.
GG_PYPI_DOWNLOAD_TIMEOUT=0                Error: GG_PYPI_DOWNLOAD_TIMEOUT must be between 1 and 3600 seconds.
GG_PYPI_DOWNLOAD_TIMEOUT=-1               Error: GG_PYPI_DOWNLOAD_TIMEOUT must be between 1 and 3600 seconds.
GG_PYPI_DOWNLOAD_TIMEOUT=1 followed by 309 zeros
                                          Error: GG_PYPI_DOWNLOAD_TIMEOUT must be between 1 and 3600 seconds.
GG_PYPI_DOWNLOAD_TIMEOUT="300 "           accepted
```

Clean `Error:` lines rather than a traceback: `UnexpectedError` is what the rest
of this module raises.

6d7a (commented):
IMO the issue is tensorflow, not the limit. The test merely checks if a ggshield pypi package scan behaves correctly, and it only needs a package that requires python <3.9 as a test subject. Not only should we use something else than tensorflow because of its size but also because tensorflow 2.5.0 is the release that [adds support](https://github.com/tensorflow/tensorflow/releases?page=9#release-v2.5.0) for python 3.9, so the test is not actually testing what it's supposed to.

6d7a (approved):

PR #1499, opened 2026-09-11
feat(client): make the API request timeout configurable

benjaminrigaud-gg:
## Problem

The time ggshield waits for the GitGuardian API to answer was hardcoded to 60 seconds in `create_client()`, with no config key and no environment variable. A workspace whose scans occasionally take longer has no way to allow more time short of editing the source.

The tail is thin but real: on one busy workspace, 16 requests out of 6.4 million crossed 55 seconds in a month and one reached 82 seconds. Each of those fails the whole scan.

## Change

- The wait is now set by the `api_timeout` config key or the `GITGUARDIAN_API_TIMEOUT` environment variable, which takes precedence. The default stays 60 seconds.
- Accepted range is 1 to 3600 seconds, from either source. The upper bound also keeps the value convertible to a float for `socket.settimeout()`, which otherwise raises `OverflowError` on a large number well after the command started.
- Parsing goes through `getenv_int`, the helper `GG_GIT_TIMEOUT` and `GG_MAX_WORKERS` already use, so the value is read the same way as every other integer environment variable.
- A value out of range or not a number gives an `Error:` line naming the source, instead of a traceback.
- `create_client()` resolves the timeout itself when the caller passes none, so the environment variable works even from a call site that does not reach for the config.
- `auth login`, `auth logout`, the OAuth token check and the HMSL token request build their client directly; they now resolve the value through `api_timeout_from_config()` rather than silently taking the default.
- The native hook in `rust/` had its own hardcoded 60 seconds. It reads the same key, the same environment variable and the same bounds. One deliberate divergence, commented in the code: where Python refuses an unusable value, the hook ignores it and carries on, because a mistyped setting must never block an agent.

The environment variable is `GITGUARDIAN_API_TIMEOUT`, not `GITGUARDIAN_TIMEOUT`. The latter already exists and controls the pre-receive hook's own wall-clock budget, the one that bails out ahead of GitHub Enterprise Server's hard 5 second limit. Reusing that name would silently disable that safety net for anyone who sets it.

## What the user sees

| `GITGUARDIAN_API_TIMEOUT` | Result |
|---|---|
| `²` | `Error: Invalid GITGUARDIAN_API_TIMEOUT value: '²'. It must be a whole number of seconds.` |
| `0`, `-1`, `3601` | `Error: Invalid value for GITGUARDIAN_API_TIMEOUT: N. The API timeout must be between 1 and 3600 seconds.` |
| 1 followed by 309 zeros | the same bounds error, where it used to be an `OverflowError` traceback on the first request |
| `3600` | accepted |
| `"300 "` | accepted as 300, matching `getenv_int` and `GG_PYPI_DOWNLOAD_TIMEOUT` |

## Testing

- `tests/unit/core/test_client.py`: the default, a config file value, the environment variable winning over it, every invalid case in the table above, the boundary at 3600, and the resolved value landing on the constructed `GGClient`.
- `tests/unit/core/config/test_user_config.py`: the `api_timeout` default and reading it from a config file.
- `rust/config`: the key is read and defaults correctly. `rust/hook`: the resolution order, with an unusable value on either side ignored rather than fatal. The resolution is a pure function, so the test needs no process-wide environment mutation.
- `uv run pytest tests/unit --disable-socket`: 2741 passed, 1 skipped.
- `uv run ty check`: clean. `cargo test`, `cargo fmt --check`, `cargo clippy -- -D warnings`: clean, except `verdict_cache::tests::the_cache_is_bounded_and_drops_expired_entries`, which fails the same way on an untouched `main`.

### DIFF
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -3,6 +3,7 @@ import os
 from enum import Enum
 from typing import Any, Optional, Union
 
+import click
 import requests
 import urllib3
 from pygitguardian import GGClient, GGClientCallbacks
@@ -12,7 +13,7 @@ from requests.adapters import HTTPAdapter
 
 from . import auth_check_cache, ui
 from .config import Config
-from .constants import DEFAULT_INSTANCE_URL
+from .constants import DEFAULT_API_TIMEOUT, DEFAULT_INSTANCE_URL
 from .errors import (
     APIKeyCheckError,
     MissingScopesError,
@@ -67,6 +68,36 @@ def _build_retry(profile: RetryProfile) -> urllib3.Retry:
     )
 
 
+# Name chosen to avoid colliding with GITGUARDIAN_TIMEOUT, which already
+# controls the pre-receive hook's own wall-clock budget (see
+# ggshield/core/git_hooks/prereceive.py) and means something else entirely.
+API_TIMEOUT_ENV_VAR = "GITGUARDIAN_API_TIMEOUT"
+
+
+def _resolve_timeout(config_timeout: int) -> int:
+    """Resolve the API timeout: env var overrides the config file value."""
+    raw = os.getenv(API_TIMEOUT_ENV_VAR)
+    if raw is None:
+        timeout = config_timeout
+        source = "the 'timeout' config key"
+    else:
+        try:
+            timeout = int(raw)
+        except ValueError:
+            raise click.UsageError(
+                f"Invalid {API_TIMEOUT_ENV_VAR} value: '{raw}'. "
+                "It must be a whole number of seconds."
+            )
+        source = API_TIMEOUT_ENV_VAR
+
+    if timeout <= 0:
+        raise click.UsageError(
+            f"Invalid value for {source}: {timeout}. "
+            "The API timeout must be a positive number of seconds."
+        )
+    return timeout
+
+
 def create_client_from_config(
     config: Config,
     *,
@@ -105,6 +136,7 @@ https://docs.gitguardian.com/ggshield-docs/reference/auth/login""",
         allow_self_signed=config.user_config.insecure,
         callbacks=callbacks,
         retry_profile=retry_profile,
+        timeout=_resolve_timeout(config.user_config.timeout),
     )
 
 
@@ -115,6 +147,7 @@ def create_client(
     allow_self_signed: bool = False,
     callbacks: Optional[GGClientCallbacks] = None,
     retry_profile: RetryProfile = RetryProfile.DEFAULT,
+    timeout: int = DEFAULT_API_TIMEOUT,
 ) -> GGClient:
     """
     Implementation of create_client_from_config(). Exposed as a function for specific
@@ -129,7 +162,7 @@ def create_client(
             api_key=api_key,
             base_uri=api_url,
             user_agent=os.getenv("GG_USER_AGENT", "ggshield"),
-            timeout=60,
+            timeout=timeout,
             session=session,
             callbacks=callbacks,
         )
--- a/ggshield/core/client.py
+++ b/ggshield/core/client.py
@@ -11,6 +11,8 @@ from pygitguardian.models import APITokensResponse, Detail, TokenScope
 from requests import Session
 from requests.adapters import HTTPAdapter
 
+from ggshield.utils.os import getenv_int
+
 from . import auth_check_cache, ui
 from .config import Config
 from .constants import DEFAULT_API_TIMEOUT, DEFAULT_INSTANCE_URL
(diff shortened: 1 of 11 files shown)


ggshield/core/client.py
```diff
@@ -115,6 +157,7 @@ def create_client(
     allow_self_signed: bool = False,
     callbacks: Optional[GGClientCallbacks] = None,
     retry_profile: RetryProfile = RetryProfile.DEFAULT,
+    timeout: int = DEFAULT_API_TIMEOUT,
```
clement-tourriere: Small trap for later. With this default, any new caller that forgets `api_timeout_from_config()` silently ignores both `GITGUARDIAN_API_TIMEOUT` and the config key. Today the 4 callers are fine, but nothing tells the next one, and the failure is silent.

Can you make it resolve by itself?

```python
timeout: Optional[int] = None,
...
if timeout is None:
    timeout = _resolve_timeout(DEFAULT_API_TIMEOUT)
```

Then the env var always works, and `api_timeout_from_config()` is only needed for the config file part.

ggshield/core/config/user_config.py
```diff
@@ -101,6 +101,9 @@ class UserConfig(FilteredConfig):
     verbose: bool = False
     insecure: bool = False
     max_commits_for_hook: int = 50
+    # How long to wait for the GitGuardian API to answer, in seconds. Can be
+    # overridden with the GITGUARDIAN_API_TIMEOUT environment variable.
+    timeout: int = DEFAULT_API_TIMEOUT
```
clement-tourriere: Can we name it `api_timeout` instead of `timeout`?

You took care to make the env var unambiguous (`GITGUARDIAN_API_TIMEOUT` and not `GITGUARDIAN_TIMEOUT`), but the config key stays generic. In the same file we already have `max_commits_for_hook`, so a plain `timeout` reads like it could be the hook timeout.

It is free to rename now. Once it is released it becomes a breaking change for users, so it is now or never. `.gitguardian.example.yml` would need the same change.

clement-tourriere (commented):
Looks good to me. I checked the env var reasoning and it is correct: `prereceive.py:25` really reads `GITGUARDIAN_TIMEOUT` as the hook wall-clock budget, so a new name was needed. The 4 direct `create_client()` callers are all updated, and `marshmallow_dataclass` gives type validation on the YAML key for free.

Nothing blocking, 2 comments inline.

Also, not blocking, just to keep in mind: the higher default doubles the worst case for GET read timeouts, which #1500 keeps retrying. 6 attempts x 120s is around 12 minutes on a hung GET, it was 6 before.

Last point, about merge order: please land this one before #1500. Alone, #1500 makes a slow instance fail hard at 60s with no way out. This PR gives the way out.

ggshield/core/client.py
benjaminrigaud-gg: Done. `create_client()` takes `timeout: Optional[int] = None` and calls `_resolve_timeout(DEFAULT_API_TIMEOUT)` when nothing is passed, so the environment variable works from any call site and `api_timeout_from_config()` is only needed for the config file part.

Checked the three paths: env var alone gives 240, nothing set gives the 60 default, an explicit argument of 42 wins.

ggshield/core/config/user_config.py
benjaminrigaud-gg: Renamed to `api_timeout`, you were right that it is now or never. Followed it through `.gitguardian.example.yml`, the tests and the changelog.

It also had to go into `rust/config/src/user_config.rs`, since the native hook parses the same YAML. That is how I noticed the hook had its own hardcoded 60 seconds and ignored both settings, which this PR now fixes too.

benjaminrigaud-gg:
Both inline comments are addressed, and the description was rewritten: it still described the old key name and a raised default.

On your two other points:

**The 12-minute hung GET is moot.** We decided to keep the default at 60 rather than raise it to 120, so the worst case for a retried GET stays where it was. The PR is now purely about making the wait settable. Data behind that call: on the workspace this came from, 16 requests out of 6.4 million crossed 55 seconds in a month and exactly one reached 82 seconds, so a raised default would have been paying for everyone to cover a handful of requests on one workspace. Setting `api_timeout` there is the better trade.

**Merge order: agreed**, this one first. #1500 on its own turns a slow response into a hard failure with nothing to reach for.

Two things also went in since your review: the accepted range is now 1 to 3600 (an oversized value used to pass the check and die later inside `socket.settimeout()`, same class of bug as the one you found on #1495), and the native hook in `rust/` had its own hardcoded 60 seconds that ignored both settings, which is now wired to the same key.

clement-tourriere (approved):

PR #1500, opened 2026-09-11
fix(client): stop retrying POST requests that time out reading the response

benjaminrigaud-gg:
## Problem

When a scan request hits the client's read timeout, ggshield sends it again, up to five more times. That is the wrong move for a read timeout: the server has already accepted the request and is still working on it, so each retry only asks it to redo the same work. The user waits over six minutes before the command finally fails, and a large scan is run six times instead of once.

The retry policy came from #1218, which added POST to `allowed_methods` so that a connection reset mid-scan would recover. `read` was left unset, and urllib3 defaults it to `None`, which means read timeouts count against the same budget.

## What this changes

Read timeouts on POST now surface immediately, on the first attempt.

Everything else behaves as before:

- connection resets on POST are still retried, which is what #1218 was for
- read timeouts on GET and the other idempotent methods are still retried
- 502, 503 and 504 responses are still retried

`Retry(read=0)` would not work here: urllib3 reports a connection reset as a `ProtocolError`, which it classifies as a read error, so that setting would have switched off the #1218 behaviour too. Instead `Retry.increment` is overridden to re-raise only `ReadTimeoutError`, and only for POST.

`auth logout` caught only `ConnectionError` around the token revoke call, which is how an exhausted read timeout used to arrive. It now catches `Timeout` as well, so the message pointing at connectivity is still the one shown.

## How it was verified

The tests drive real local servers rather than mocks, and count the connections the server actually accepts:

| Scenario | Expected | Result |
|---|---|---|
| POST, server never replies | 1 attempt, `ReadTimeout` raised | passes on both retry profiles |
| GET, server never replies | still retried | passes |
| POST, server resets the connection | still retried in full | passes |
| POST, server returns 502 / 503 / 504 | still retried in full | passes |

Three things about the test servers came out of review and are worth calling out, since they are the kind of thing that comes back:

- The status server answered without reading the request body. Closing a socket that still holds unread data sends a RST, and Windows then discards the response the client had already received, which is what made the 502 case red there and green everywhere else.
- Closing the listening socket while the serving thread sat in `accept()` raised `OSError` in that thread. `--disable-pytest-warnings` hid it in CI. The thread is joined first now.
- `enable_socket` takes precedence over `allow_hosts` in pytest-socket: `pytest_runtest_setup()` enables the socket and returns before it looks at the host list. With both markers a connect to a public address goes through, so the loopback restriction the comment promised was not in force. Only `allow_hosts` remains, which works with or without `--disable-socket`.

`pytest tests/unit --disable-socket`: 2730 passed, 1 skipped. The four tests were also run ten times with `PytestUnhandledThreadExceptionWarning` raised as an error, with no failures.

The error the user sees is unchanged in shape and still maps to exit code 4: `Scanning failed: HTTPSConnectionPool(...): Read timed out. (read timeout=60)`.

One note for reviewers: the `pyright` pre-commit hook was skipped on these commits. It fails the same way on unmodified `main`, reporting import-resolution errors across the whole codebase, because pyright does not pick up the project virtualenv without an explicit `venvPath` in `[tool.pyright]`. With the interpreter set it reports 0 errors on this branch. Worth its own small PR.

### DIFF
--- a/tests/unit/core/test_client.py
+++ b/tests/unit/core/test_client.py
@@ -4,7 +4,7 @@ import socket
 import struct
 import threading
 from contextlib import contextmanager
-from typing import Any, Iterator, List, Tuple, Type
+from typing import Any, Callable, Iterator, List, Tuple, Type
 from unittest.mock import Mock, patch
 
 import click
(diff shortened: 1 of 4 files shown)


tests/unit/core/test_client.py
```diff
@@ -454,3 +459,196 @@ def test_create_session_with_self_signed_option(allow_self_signed: bool):
...
+
+
+@contextmanager
+def _status_server(status_code: int) -> Iterator[Tuple[int, List[str]]]:
+    """A server that always answers with status_code."""
+    attempts: List[str] = []
+
+    class Handler(http.server.BaseHTTPRequestHandler):
+        def _reply(self) -> None:
+            attempts.append(self.command)
```
clement-tourriere, line 538: This is what makes Windows CI red. `test_post_status_forcelist_is_still_retried[502]` fails on windows-2022 3.11 and 3.13:

```
ProtocolError('Connection aborted.', ConnectionAbortedError(10053, 'An established connection was aborted by the software in your host machine'))
```

The handler answers and closes without ever reading the POST body. When you close a socket that still has unread data in it, the OS sends a RST instead of a FIN. Linux and macOS still give the already buffered response to the client, but Windows throws the receive buffer away, so the client loses the 502 it had already received.

Please drain the body before replying:

```python
def _reply(self) -> None:
    remaining = int(self.headers.get("Content-Length") or 0)
    while remaining > 0:
        remaining -= len(self.rfile.read(min(remaining, 65536)))
    attempts.append(self.command)
    self.send_response(status_code)
    ...
```

I tested it, it is stable after that.

clement-tourriere (commented):
Logic in `client.py` looks good to me. I checked the urllib3 side and it holds:

- the subclass survives the retry chain, `Retry.new()` builds with `type(self)`
- raising from `increment()` is safe, `clean_exit` is already `False` at that point, so the `finally` still closes the connection and puts it back in the pool
- exit code 4 still works, `ReadTimeout` is a subclass of `requests.exceptions.Timeout` and `secret_scanner.py:227` catches that one

But the tests need some work. Windows CI is red on 3.11 and 3.13 (the 3 other Windows jobs are green), and it is the test server, not the fix. Details inline.

One more thing I cannot put inline because the file is not in the diff:

`ggshield/cmd/auth/logout.py:113` catches only `ConnectionError` around `client.post("token/revoke")`. Today an exhausted POST read timeout arrives as `ConnectionError` (MaxRetryError -> ConnectionError). After this PR it arrives as `ReadTimeout`, so the nice `CONNECTION_ERROR_MESSAGE` is skipped and the user sees the raw requests message. The exit code does not change, so it is only cosmetic, but please make it `except (ConnectionError, Timeout)`.

tests/unit/core/test_client.py
```diff
@@ -454,3 +459,196 @@ def test_create_session_with_self_signed_option(allow_self_signed: bool):
...
+    finally:
+        server.shutdown()
+        thread.join(timeout=5)
+
+
+@pytest.mark.parametrize(
+    "retry_profile", [RetryProfile.DEFAULT, RetryProfile.PRE_RECEIVE]
+)
+@pytest.mark.enable_socket
+@pytest.mark.allow_hosts(["127.0.0.1"])
```
clement-tourriere: `allow_hosts` does nothing here. In `pytest_socket.py`, `pytest_runtest_setup()` checks `enable_socket` first and returns before it ever looks at `allow_hosts`:

```python
if ("socket_enabled" in item.fixturenames
        or item.get_closest_marker("enable_socket")
        or item.config.__socket_force_enabled):
    enable_socket()
    return
```

So the comment above says "restricted to loopback", but in reality the socket is fully open. I checked with a small test: with both markers a connect to a public IP goes through, with only `allow_hosts` it raises `SocketConnectBlockedError`.

Please drop `@pytest.mark.enable_socket` and keep only `@pytest.mark.allow_hosts(["127.0.0.1"])` on the 4 tests. It works with and without `--disable-socket`, and then the restriction is real.

tests/unit/core/test_client.py
```diff
@@ -454,3 +459,196 @@ def test_create_session_with_self_signed_option(allow_self_signed: bool):
...
+            except OSError:
+                pass
+
+    thread = threading.Thread(target=serve, daemon=True)
+    thread.start()
+    try:
+        yield port, attempts
+    finally:
+        stop.set()
+        sock.close()
```
clement-tourriere, line 506: Race here. The thread can be inside `accept()` when the main thread closes the socket, and then it raises `OSError: [Errno 9] Bad file descriptor`, which is not caught (only `socket.timeout` is). pytest reports it as `PytestUnhandledThreadExceptionWarning`. I get it around 1 run out of 4 locally. CI does not show it because of `--disable-pytest-warnings`.

Please join the thread first:

```python
stop.set()
thread.join(timeout=5)
sock.close()
```

Same thing in `_reset_server`. Those two helpers are also the same 25 lines twice, one `_raw_server(handle)` with two small callbacks removes about 20 lines. I have a working patch if you want it.

clement-tourriere (approved):

PR #1504, opened 2026-09-14
refactor(rust): lay the Rust crates out as apps/ and packages/

nriviere-gitguardian:
Moves the Cargo workspace from `rust/` to the repository root, so Rust sits in an `apps/` + `packages/` monorepo layout that the Python and JavaScript packages can share. No source file changes: this is a move, a crate rename and the path references that follow it.

| before | after |
| --- | --- |
| `rust/dispatcher` | `apps/cli` (crate renamed `ggshield-dispatcher` → `ggshield-cli`) |
| `rust/common` | `packages/rust-common` |
| `rust/config` | `packages/rust-config` |
| `rust/hook` | `packages/rust-hook` |
| `rust/tests` | `tests/rust` |

The `ggshield` binary name, every crate's public API and all sources are unchanged. `rust/.gitignore` goes because the root one already ignores `target/`.

### What else had to follow

- `hatch_build.py` — cargo now runs at the repository root, so the wheel's binary comes from `target/release/ggshield`.
- `pyproject.toml` — the sdist exclude is now `apps`, `packages` and the root manifests, so an sdist install stays pure Python.
- `.github/workflows/ci.yml` — the cargo steps drop `working-directory: rust`, the caches key on the root `Cargo.lock` / `rust-toolchain.toml`, and the binary-extension sync check and equivalence runner point at their new paths.
- `.github/workflows/wheels.yml` — watches `apps/`, `packages/` and the root manifests instead of `rust/`.
- `scripts/build-os-packages`, `.gitattributes`, `.pre-commit-config.yaml`, two docstrings.

### Checks run locally

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (190 pass, 1 ignored), `cargo check --release --locked --workspace`
- `GGSHIELD_BUILD_RUST=1 uv build --wheel` → `ggshield-1.54.0-py3-none-macosx_11_0_universal2.whl`, installed and run; `uv build --sdist` carries no Rust
- the vendored binary-extension list is in sync (208 entries)
- `python3 tests/rust/equivalence.py` → `RESULT: OK (2 known divergences, 0 unexpected)`
- `actionlint` clean on the workflows touched


nriviere-gitguardian:

PR #1505, opened 2026-09-14
feat(secrets): bring the secret manager engine, CLI and SDKs into ggshield

nriviere-gitguardian:
Stacked on #1504 — base is `nathanriviere/setup-monorepo`, so review only the 6 commits here. Draft until that one lands.

Brings the standalone secret-manager prototype into ggshield: read, write and inject secrets from HashiCorp Vault, 1Password Connect and device-local encrypted dotenv files.

```
ggshield secret get --provider vault --path secret/myapp --field STRIPE_KEY
ggshield secret run --provider file --path .env -- npm start
ggshield secret activate zsh        # prompt hook that loads a trusted .env
```

### Shape

- `packages/rust-secrets` — the engine. Providers are declared as data in `providers/*.yaml`, validated at build time; values are `SecretString` throughout and only leave the crate via `expose_secret()`.
- `packages/rust-secrets-cli` — the verbs, as a library with no binary target, exactly as `packages/rust-hook` is. `apps/cli` matches the argv and calls in; `NATIVE_VERBS` lives next to the parser that implements it, so a verb declared in one place and not the other is caught by reading one file.
- `ggshield/cmd/secret/store.py` — lazy stubs for the same verbs, so `ggshield secret --help` lists them on every install and an installation without the native executable says which executable is missing rather than "No such command".
- `packages/python`, `packages/js` — the same engine for applications that want secrets in memory. They publish as `gitguardian` on PyPI and npm (`ggshield` is the CLI) and vendor/generate their provider definitions from the same `providers/*.yaml`.

### Ported onto this repository's stack, not copied

The prototype was async; this binary is built for one request per process, and the hook's manifest says so. So the engine was rewritten before it landed:

- `reqwest` + `tokio` → `ureq`, and the whole `SecretStore` API is now sync — no async runtime is added to `ggshield`
- `serde_yaml` → `serde_yaml_ng`, the maintained fork the workspace already uses
- redirects stay refused (`max_redirects(0)`): a redirect would carry `X-Vault-Token` to wherever a compromised server points; `http_status_as_error(false)` keeps the 404 → `SecretNotFound` mapping reachable
- both new crates adopt the workspace's `clippy::unwrap_used` lint (tests opt out, with the reason in the code)

`sha2` stays at 0.11 in the engine: `hkdf` 0.13, which derives the file provider's keys, is built on the 0.11 digest traits. The hook keeps 0.10.

### Deliberately unchanged

On-disk identifiers are a file format, so they keep the prototype's spelling: the `gitguardian:` ciphertext marker, the `gitguardian` keyring service, `<config dir>/gitguardian/secrets.env`. Renaming them would strand secrets people have already written. Only user-facing strings moved to `ggshield secret …`.

### Checks run locally

- `cargo test --workspace --locked` — 544 pass, 0 fail (108 of them end-to-end `ggshield secret …` runs of the real binary)
- `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all --check`, `cargo check --release --locked --workspace`
- `packages/python`: 28 tests, `ruff format --check`, `ruff check`, `ty check`
- `pnpm run ci` (test, fmt, lint, tsgo, build, knip)
- ggshield's own `tests/unit/cmd`: 695 pass
- from an installed wheel: `secret get --expose` returns the value, `secret run` injects it, `secret scan` still delegates to Python

### Open, on purpose

1. **Windows.** The 108 end-to-end tests have never run on Windows — the prototype's CI was Linux only. The `rust` matrix job will be the first run.
2. **No changelog fragment** yet for the new user-visible commands.
3. **No release pipeline** for the two SDKs: the prototype published them from GitLab jobs that do not apply here. They are tested but not publishable.
4. The JS package's `publishConfig` pointing at an internal GitLab registry was removed rather than repointed.


nriviere-gitguardian:

PR #1509, opened 2026-09-14
feat(secrets): add a repo scope shared by every worktree

nriviere-gitguardian:
Stacked on #1505 — base is `nathanriviere/secrets`, so review only the 2 commits here.

A worktree per branch means a project-scope `.env` has to be recreated in every checkout, holding the same values each time. This adds a third scope between user and project for what a *repository* shares:

| scope | file |
| --- | --- |
| user | `<config dir>/gitguardian/secrets.env` |
| repo | `<common git dir>/gitguardian/secrets.env` |
| project | `./.env` |

The merge is **per variable**: the repository holds the base, a checkout's `.env` overrides only the names it sets, so a branch pointing at a different backend is one line in one file. `get --scopes` annotates each line with the file its value won in, because a per-variable merge is otherwise hard to debug.

```
$ ggshield secret set --provider file SHARED      # writes .git/gitguardian/secrets.env
$ cd ../feature-worktree
$ ggshield secret get --provider file --scopes --expose
SHARED=fake-shared  # repo
$ ggshield secret set --provider file --scope project SHARED
$ ggshield secret get --provider file --scopes --expose
SHARED=fake-branch  # project
```

### Why inside the git directory

A `.gitguardian/` at the top of the working tree is one `git add -A` away from being committed, and nothing under `.git/` is transferred by a clone. That second property is also why **the repo store needs no trust decision**: the trust gate exists because a `.env` can arrive with a repository somebody else wrote, and this file cannot. A `.env` that is present keeps its gate.

The path is resolved by reading `.git` (a directory, or a `gitdir:` pointer) and its `commondir`, not by shelling out to `git`: every dotenv read would otherwise fork a process, and git need not be installed to find a file we can find ourselves. A linked worktree resolves to the repository's shared directory — the whole point — and a submodule keeps its own.

### Behaviour changes

1. `set` and `del` with neither `--scope` nor `--path` now target the repository store when run in a git repository. Outside one, nothing changes.
2. `--scope repo` outside a repository is an error rather than a guess.
3. The shell hook loads the repository's store when a checkout has no `.env` at all — the state a fresh worktree is in, and the case this scope exists for.

### Checks

- 562 tests pass (`cargo test --workspace --locked`), 15 of them new: git-dir resolution (plain checkout, linked worktree, submodule, no repository, empty pointer), the three-scope merge, and end-to-end runs of the binary for the default write target, the per-variable override, `--scopes`, the untrusted-`.env` gate and the hook's fallback.
- clippy `-D warnings` and `cargo fmt --all --check` clean.
- Smoke-tested by hand in a real `git worktree` layout: the store lands in `.git/`, the linked worktree reads it, `--scope project` overrides one variable, and `git status` never sees the store.


nriviere-gitguardian:

PR #1510, opened 2026-09-14
build(docker): compile the standard library into the image

alexpasmantier:
The python-slim images strip the precompiled bytecode from the final image which means the docker images we ship **always recompile the entire python standard library** during startup.

This PR precompiles that during the image build which saves a rather considerable amount of time at runtime:

```
ggshield-3.11 ❯ hyperfine 'docker run -i ggshield/main:latest ggshield --version' 'docker run -i ggshield/precompile-stdlib:latest ggshield --version'
Benchmark 1: docker run -i ggshield/main:latest ggshield --version
  Time (mean ± σ):      1.044 s ±  0.037 s    [User: 0.014 s, System: 0.012 s]
  Range (min … max):    0.993 s …  1.096 s    10 runs

Benchmark 2: docker run -i ggshield/precompile-stdlib:latest ggshield --version
  Time (mean ± σ):     737.8 ms ±  23.6 ms    [User: 17.6 ms, System: 14.8 ms]
  Range (min … max):   707.2 ms … 771.4 ms    10 runs

Summary
  docker run -i ggshield/precompile-stdlib:latest ggshield --version ran
    1.42 ± 0.07 times faster than docker run -i ggshield/main:latest ggshield --version
```

**This applies to all dockerized `ggshield` invocations** and costs 27 MB on the image (about +5%) which is (I believe) a no-brainer.


Dockerfile
```diff
@@ -64,6 +64,9 @@ RUN \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
+# this avoid compiling the stdlib at runtime, which can be slow (~0.2s on my dev machine)
```
6d7a: nit
```suggestion
# this avoid compiling the stdlib at runtime
```

6d7a (approved):
Nice one!

alexpasmantier:
Thanks for the review @6d7a 👍🏻

PR #1511, opened 2026-09-14
fix: disable `check-for-updates` by default for pre-receive and fix the cli flag

alexpasmantier:
This PR fixes two things:

- `ggshield --no-check-for-updates <command>` was broken when the flag came before the subcommand

- `secret scan pre-receive` no longer checks for updates by default (it doesn't really make any sense in a non-interactive environment).


ggshield/__main__.py
```diff
@@ -100,6 +101,9 @@ def cli(
     if instance:
         ctx_obj.config.cmdline_instance_name = instance
 
+    if check_for_updates is not None:
```
clement-tourriere, line 104: This part is good for me. I checked on `main`, the option was really ignored : the `create_ctx_callback` callback do nothing because `ctx.obj` does not exist yet when click parse the options of the group.

I also looked at the other common options and we are complete now : verbose/debug/log-file work with their own callback, insecure/allow-self-signed/instance were already re-applied here, so `check_for_updates` was the last one missing.

ggshield/cmd/secret/scan/prereceive.py
```diff
@@ -93,6 +94,10 @@ def prereceive_cmd(
     configuration.
     """
     ctx_obj = ContextObj.get(ctx)
+    # The hook runs on the git server: the "new version available" message reaches
+    # nobody who can act on it.
+    if kwargs.get("check_for_updates") is None:
```
clement-tourriere: This is the bug.

`kwargs` here only contain the options of the `pre-receive` command itself, not the ones of the parent groups. So when the user write `ggshield --check-for-updates secret scan pre-receive`, `kwargs["check_for_updates"]` is None and we erase the `True` that `cli()` just put in the ctx_obj. And it's not only the root group, `secret` and `scan` also have the common options, so the 3 parent levels are lost.

We can not fix it here, because we can not know if the `True` in ctx_obj come from the user or from the default value. So i think we should not overwrite the value, but change the **default** instead :

`context_obj.py` :
```python
# None until --check-for-updates/--no-check-for-updates is given, on any level
self.check_for_updates: Optional[bool] = None
# used when the option is not given. Commands where nobody can read the
# message (pre-receive) set it to False.
self.check_for_updates_default = True

@property
def should_check_for_updates(self) -> bool:
    if self.check_for_updates is None:
        return self.check_for_updates_default
    return self.check_for_updates
```

`__main__.py`, in `before_exit()` :
```python
_check_for_updates(ctx_obj.should_check_for_updates)
```

and here only :
```python
ctx_obj.check_for_updates_default = False
```

It works because the 2 places who write `check_for_updates` already check `is not None`, and click process the group options before the subcommand ones, so an explicit flag on the subcommand still win.

I tried it localy : the 6 cases of my main comment are all good, and `tests/unit/cmd/` still pass (698 tests).

Small extra : `kwargs.get(...)` is a bit dangerous, if one day the option is renamed we silently disable the check forever. With the fix above the problem disappear anyway.

ggshield/cmd/secret/scan/prereceive.py
```diff
@@ -3,7 +3,8 @@
 import os
 import sys
 from pathlib import Path
-from typing import Any, List, Pattern, Set
+from re import Pattern
```
clement-tourriere: Small nit, not related to the PR : `ggshield/cmd/utils/context_obj.py` still do `from typing import Pattern` for the same `Set[Pattern[str]]`, so the cleanup is half done. Can you do the both, or keep it for another PR ?

(no problem for the runtime, `re.Pattern[str]` work since python 3.9 and we require 3.9.)

tests/unit/cmd/scan/test_prereceive.py
```diff
@@ -400,6 +400,40 @@ def test_stdin_input_deletion(
...
+        WHEN the command finishes
+        THEN the update check runs
+        """
+        result = cli_fs_runner.invoke(
+            cli,
+            ["secret", "scan", "pre-receive", "--check-for-updates"],
+            input=f"{'a' * 40} {EMPTY_SHA} main",
+        )
+        assert_invoke_ok(result)
+        check_for_updates_mock.assert_called_once_with(True)
```
clement-tourriere, line 441: Can you add the case with the flag **before** the subcommand ? This is exactly the one which is broken and the 2 tests don't see it.

Maybe with a parametrize it's nicer :

```python
@pytest.mark.parametrize(
    "args",
    (
        ["--check-for-updates", "secret", "scan", "pre-receive"],
        ["secret", "--check-for-updates", "scan", "pre-receive"],
        ["secret", "scan", "--check-for-updates", "pre-receive"],
        ["secret", "scan", "pre-receive", "--check-for-updates"],
    ),
)
```

and assert `True` for all of them.

clement-tourriere (commented):
Thanks for this ! The first commit is a real fix, i confirm on `main` that `ggshield --no-check-for-updates config list` was doing the check anyway.

But the second commit break the opposite case : now if you put `--check-for-updates` **before** the subcommand, pre-receive ignore it and disable the check. So we just move the bug from one flag to the other one :)

I tested on your branch :

| command | result |
| --- | --- |
| `ggshield secret scan pre-receive` | False (ok) |
| `ggshield --no-check-for-updates secret scan pre-receive` | False (ok) |
| `ggshield secret scan pre-receive --check-for-updates` | True (ok) |
| `ggshield --check-for-updates secret scan pre-receive` | **False, should be True** |
| `ggshield secret --check-for-updates scan pre-receive` | **False, should be True** |
| `ggshield secret scan --check-for-updates pre-receive` | **False, should be True** |

I put a proposition of fix in the comments.

The rest is good for me : black/isort/flake8/pyright are clean, and the change of the default for pre-receive really make sense (the update check do a http call to github with 5s timeout, inside the 5s budget that GHES give to all the hooks).

alexpasmantier:
Made the changes 👍🏻 
Tell me what you think @clement-tourriere

clement-tourriere (approved):

PR #1512, opened 2026-09-14
perf: fetch metadata and token scopes concurrently

alexpasmantier:
`check_client_api_key` called `/v1/metadata` then `/v1/api_tokens/self` in sequence before every scan.

The two calls can be made concurrently. This makes a 1 commit pre-receive scan go from ~1.20 seconds to ~0.9 seconds.

SCRT-7532


ggshield/core/client.py
```diff
@@ -1,5 +1,8 @@
+from __future__ import annotations
```
6d7a, line 1: Could you use `Optional` instead? We already import and use it throughout the file.

changelog.d/20260914_190200_alexandre.pasmantier_concurrent_api_key_check.md
```diff
@@ -0,0 +1,4 @@
+### Changed
+
+- The API key check fetches server metadata and token scopes concurrently instead of
+  one after the other, saving a round trip on every scan.
```
6d7a, line 4: nit: I'd leave this out, since not every scan re-fetches metadata and scopes. 
```suggestion
  one after the other.
```

6d7a (commented):
I'm concerned about the change in ggshield's CTRL-C behavior. An aborted process will wait for the futures from the executor to resolve. CTRL-C is not an edge case, especially in client-side git hooks. Would it be an option to defer this perf optimization in favor of others?

alexpasmantier:
@6d7a The ctrl-c behavior is already present on main since we use thread pools for the scans themselves (and is much worse since scans take longer than simple metadata/scopes calls). This really only adds a second (tiny in most cases, just measured ~0.2s) window of that same behavior before the scans.

Also what really happens is just:
- ggshield invocation
- threadpool starts and jobs are submitted
- user hits ctrl-c
- we print "Aborted!" immediately, but the threadpool waits for the second thread to join
- during that time (again, 0.2s in most cases), the user can still just ctrl-c a second time and "hard quit" the process

What do you think?

6d7a:
Ah, right. I didn't realize we already had the same in the scan calls. In that case, LGTM.

6d7a (approved):

PR #1515, opened 2026-09-15
feat(secrets): add the `ggshield secret` store verbs

nriviere-gitguardian:
Read, write and inject secrets from a secret manager, natively, as `ggshield secret` verbs — HashiCorp Vault, 1Password Connect, and dotenv files encrypted with a device-local key.

```
ggshield secret get --provider vault --path secret/myapp --field STRIPE_KEY
ggshield secret run --provider file -- npm start
ggshield secret activate zsh        # prompt hook that loads a trusted .env
```

The verbs are `get`, `set`, `del`, `encrypt`, `run`, `activate` and `trust` against any provider, plus `import` to push a dotenv file into Vault or 1Password. `import` is refused for the `file` provider on purpose — its storage *is* a dotenv file, so the equivalent is to put the entries in place and run `encrypt`.

Replaces the closed #1504/#1505/#1509 stack: those restructured the repository into an `apps/` + `packages/` monorepo first. This lands the same work on the existing `rust/` layout instead, as two new crates.

### Shape

1. `rust/secrets` — the engine. Providers are static data in `providers.rs`, so a malformed one does not compile; values are `SecretString` throughout and only leave the crate via `expose_secret()`.
2. `rust/secrets-cli` — the verbs, a library with no binary target, exactly as `rust/hook` is. `rust/dispatcher` matches the argv and calls in; `NATIVE_VERBS` lives next to the parser that implements it.
3. `ggshield/cmd/secret/store.py` — lazy stubs for the same verbs, so `ggshield secret --help` lists them everywhere and an install without the native executable says what is missing rather than "No such command".

### Scopes

Three, most specific winning **variable by variable**:

| scope | file |
| --- | --- |
| user | `<config dir>/gitguardian/secrets.env` |
| repo | `<common git dir>/gitguardian/secrets.env` |
| project | `./.env` |

The repo scope is what stops a worktree per branch from needing its own copy of the same values; a checkout's `.env` overrides only the names it sets. It lives *inside* the git directory: a `.gitguardian/` at the top of the working tree is one `git add -A` away from being committed, and nothing under `.git/` is transferred by a clone — which is also why it needs no trust decision, unlike a `.env` that can arrive with a repository somebody else wrote. The path is resolved by reading `.git` and its `commondir`, not by running git.

### Ported onto this workspace's stack

The prototype was async; this binary is built for one request per process and the hook's manifest says so. So: `ureq` rather than `reqwest` + `tokio` (the `SecretStore` API is sync), redirects still refused, and `http_status_as_error(false)` so the 404 → `SecretNotFound` mapping stays reachable.

It also declared providers in YAML, so its Python and JavaScript SDKs could share one source of truth. With the definitions shipping only inside this binary that bought a dependency, a build script and a parse at every `SecretStore::build()` for nothing, so they are Rust consts here and the invariants the build script used to check (absolute paths, balanced placeholders, unique endpoint names, a definition agreeing with the `Provider` that returns it) are tests over every provider at once. `sha2` stays at 0.11 in this crate because `hkdf` 0.13 is built on the 0.11 digest traits.

### Checks

- `cargo test --workspace --locked`: 566 pass, 0 fail — 117 of them end-to-end runs of the real binary.
- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`, `cargo check --release --locked --workspace`: clean.

### Known, and deliberate

`Rust (windows-2022)` fails: the file provider's atomic write renames a temp over the destination while holding an open handle on it, which POSIX allows and Windows does not (`Access is denied`). Vault and 1Password never touch local files and are unaffected. Fixing it properly means restructuring the lock (a sidecar `.lock` file, or `ReplaceFileW`) and validating on a Windows machine, so it is a follow-up rather than something to rush in here.


rust/secrets/src/file/atomic.rs
```diff
@@ -0,0 +1,1008 @@
...
+    if !locked_the_file_at_the_path(locked, path)? {
+        bail!(
+            "{} was replaced by another program while gitguardian was rewriting it (an editor \
+             saving the file, most likely), so writing now would discard that change. Nothing \
+             was written; re-run the command",
+            path.display()
+        );
+    }
+
+    temp.persist(path)
```
clement-tourriere, line 386: On windows this rename can't work: we still have the locked handle open on the target, so MoveFileEx fail with `Access is denied`. The last CI run on windows has 49 tests red with exactly this error.

Careful, if you just close the handle before `persist` the lock is gone before the rename. Maybe a sidecar lock file (`.env.lock`) is simpler?

rust/secrets/src/file/atomic.rs
```diff
@@ -0,0 +1,1008 @@
...
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => Ok(false),
+        Err(error) => {
+            Err(anyhow::Error::new(error).context(format!("inspecting {}", path.display())))
+        }
+    }
+}
+
+/// Windows has no inode identity; the advisory lock is all there is.
+#[cfg(not(unix))]
+fn locked_the_file_at_the_path(_file: &File, _path: &Path) -> Result<bool> {
```
clement-tourriere, line 128: On windows this is always true. So if 2 writers wait on the lock, the second one get it after the rename, read the old handle and write its stale copy on top: lost update. Could compare volume serial + file index (`GetFileInformationByHandle`)?

rust/secrets-cli/src/commands/activate.rs
```diff
@@ -0,0 +1,1764 @@
...
+    SHELL_CONTROL_VARS.iter().any(|var| same(var, key))
+        || SHELL_CONTROL_PREFIXES.iter().any(|prefix| {
+            key.get(..prefix.len())
+                .is_some_and(|start| same(start, prefix))
+        })
+}
+
+/// The name is validated too: it is interpolated unquoted and may come from the
+/// untrusted state blob.
+fn export_statement(shell: Shell, key: &str, value: &str) -> Result<String> {
```
clement-tourriere, line 1052: Single quotes are not enough for some bash special vars. `export OPTIND='a[$(cmd)0]'` runs `cmd` (tested on macOS `/bin/bash` 3.2), same with `HISTCMD`, because bash evaluate the value as arithmetic. The name pass `validate_env_key` and is not in the denylist.

With the repo store issue below, a folder you just untar can do this on `cd`. I think we must refuse the special names of the target shell (OPTIND, HISTCMD, RANDOM, SECONDS, LINENO, ...).

rust/secrets-cli/src/commands/activate.rs
```diff
@@ -0,0 +1,1764 @@
...
+        path: found.path.clone(),
+        fingerprint: found.fingerprint,
+        keys,
+    };
+    emit(script, &state.encode(), shell)
+}
+
+fn emit(script: &str, state: &str, shell: Shell) -> Result<()> {
+    print!("{script}");
+    print!("{}", export_statement(shell, STATE_VAR, state)?);
```
clement-tourriere, line 622: In zsh, `export UID=1000` inside the `eval` is fatal (`failed to change user ID`) and everything after is skipped, this state export too. So the vars exported before are never recorded and never unset when you leave the dir.

Very common with docker compose `.env` (`UID=1000`, `GID=1000`). I reproduced it: `API_KEY` still set after `cd /`.

Maybe print the state first, and refuse zsh special params (UID, GID, EUID, EGID, path, fpath, status, argv...).

rust/secrets-cli/src/commands/activate.rs
```diff
@@ -0,0 +1,1764 @@
...
+    if !metadata.is_file() {
+        return None;
+    }
+    Some(Found {
+        // The shell's directory, not the store's: one store serves every worktree.
+        directory: cwd.to_path_buf(),
+        fingerprint: layers_fingerprint(cwd, &metadata),
+        path,
+        refusal: None,
+        needs_trust: false,
```
clement-tourriere, line 738: Repo store is loaded without trust, and `foreign_repo_store` only checks uid + `go-w`. But when the user extract a tarball or does `cp -r`, the files are owned by him and umask clean the mode. So a `.git/gitguardian/secrets.env` coming from an archive pass the check and is exported on `cd`, even with plaintext values. Reproduced twice.

Could we require trust for this layer too? Or only accept sealed values here (only this device can make them). Also on windows there is no check at all (line 822).

rust/secrets/src/file/repo.rs
```diff
@@ -0,0 +1,180 @@
...
+fn env_path(name: &str) -> Option<PathBuf> {
+    let value = std::env::var_os(name)?;
+    if value.is_empty() {
+        return None;
+    }
+    Some(PathBuf::from(value))
+}
+
+/// Walk up from `start` looking for `.git`, as git itself does.
+fn discover(start: &Path) -> Option<PathBuf> {
```
clement-tourriere, line 45: `discover` takes the first `.git` going up, whatever the owner, and follow `gitdir:`. For `run`/`get` (no trust) another local user can create `/tmp/.git/gitguardian/secrets.env` and it's merged when I run something from `/tmp/foo`. On windows `C:\.git` is the classic one (CVE-2022-24765). Git has `safe.directory` for this, maybe check the owner like git does?

rust/secrets-cli/src/commands/run.rs
```diff
@@ -0,0 +1,190 @@
...
+        let inherited =
+            !scrubbed.contains(key) && std::env::var_os(key).is_some_and(|value| !value.is_empty());
+        if !args.no_env_override && inherited {
+            eprintln!(
+                "warning: '{key}' is already set in the environment; keeping the \
+                 existing value (use --no-env-override to force the provider's)"
+            );
+            continue;
+        }
+        child.env(key, value.expose_secret());
```
clement-tourriere, line 99: `run` injects everything from the default `./.env` with no trust and no denylist. So a cloned repo with `LD_PRELOAD` / `NODE_OPTIONS` / `BASH_ENV` / `GIT_SSH_COMMAND` in its `.env` gets it into the child with `ggshield run -- npm test`.

The doc says "naming the file is the consent", but here I didn't name it, it's the default. At least apply `is_shell_control_var` when the path is the default one?

rust/secrets-cli/src/commands/activate.rs
```diff
@@ -0,0 +1,1764 @@
...
+        .step_by(2)
+        .map(|index| u8::from_str_radix(text.get(index..index + 2)?, 16).ok())
+        .collect()
+}
+
+/// Variables whose *value is executed*, so the hook never exports them: a dotenv
+/// file may say what a variable holds, not what the shell does. A class, not an
+/// inventory; many are shell-only variables, so "the user has not set it" is no
+/// protection. `ggshield run` is the way to set one for a single command.
+const SHELL_CONTROL_VARS: &[&str] = &[
```
clement-tourriere, line 954: The help says these are "never exported, whatever a file says", but the list is easy to bypass. Tested: `GIT_CONFIG_PARAMETERS="'core.fsmonitor'='cmd'"` runs `cmd` on every `git status`, so on every prompt with a git theme (`GIT_CONFIG_COUNT` is blocked but not this one).

Others missing: `MAILPATH` (bash runs `$(...)` in it), `fish_user_paths`, `fish_function_path`, zsh `PROMPT2/3/4`, `XDG_CONFIG_HOME` (moves our own trust store), `GIT_DIR`, `VAULT_ADDR` (next `ggshield secret get` sends the token elsewhere), `DOTNET_STARTUP_HOOKS`, `JDK_JAVA_OPTIONS`, `PERLLIB`.

I don't think a list like this can ever be complete. Maybe say in the doc that trust is the real boundary and keep the list as best effort, and add families (`GIT_*`, `XDG_*`, `fish_*`, `*_OPTIONS`).

rust/secrets/src/store.rs
```diff
@@ -0,0 +1,955 @@
...
+        params: &BTreeMap<String, String>,
+        fields: &BTreeMap<String, SecretString>,
+        cas_version: u64,
+    ) -> Result<()> {
+        let data = fields
+            .iter()
+            .map(|(key, value)| {
+                (
+                    key.clone(),
+                    Value::String(value.expose_secret().to_string()),
```
clement-tourriere, line 375: This writes every field back as a string. `secret_fields` already stringify on read, so a secret with `{"port": 5432, "tls": true}` becomes `{"port": "5432", "tls": "true"}` after a simple `set NEW=x` on another field. Breaks terraform and apps reading typed values.

Keep the raw `Value` for the fields we don't touch, or use KV v2 `PATCH` (merge-patch), which does it for us.

rust/secrets-cli/src/env.rs
```diff
@@ -0,0 +1,245 @@
...
+        match line {
+            Line::Blank(_) => {}
+            Line::Comment(raw) => {
+                // Warn only for a disabled assignment; prose with '=' (a URL) is just a comment.
+                if let Some(key) = commented_env_var(raw) {
+                    eprintln!("warning: ignored commented env var '{key}' on line {line_number}");
+                }
+            }
+            Line::Entry(entry) => {
+                let value = SecretString::from(entry.value());
```
clement-tourriere, line 38: import takes `entry.value()` as is, so an already encrypted `gitguardian:...` value is treated like plaintext. `encrypt`, then `import --provider vault ... --remove-source`, pushes device-only ciphertext to Vault and deletes the source. To the file provider it's encrypted twice, and `run` gives the marker text to the app as the password.

Should keep markers as is for file, and refuse (or decrypt first) for vault.

rust/secrets/src/file.rs
```diff
@@ -0,0 +1,2157 @@
...
+            .extend(locked.replace(&document.to_string())?);
+        Ok(outcome)
+    }
+
+    /// The scope files that exist, least specific first.
+    fn layers(&self, project_path: &Path) -> Result<Vec<Layer>> {
+        let mut layers = Vec::new();
+        if let Some(system_path) = self.system_path.as_deref()
+            && system_path != project_path
+            && let Some(contents) = atomic::read_to_string(system_path)?
```
clement-tourriere, line 704: `sudo ggshield secret set --system` creates `/etc/gitguardian` 0700 and the file 0600 owned by root (`create_private_dir_all`). Then for every other user `read_to_string` gets EACCES (not NotFound) and `layers()` fails with `?`, so get, run and the hook are broken for everybody on the machine. Also the value is sealed with root's key, so nobody else can read it anyway.

For system scope maybe 0755/0644 + refuse sealing, and treat EACCES on a layer the user didn't ask for as a warning?

rust/secrets/src/file.rs
```diff
@@ -0,0 +1,2157 @@
...
+    {
+        Some(
+            PathBuf::from("/etc")
+                .join(USER_SCOPE_DIR)
+                .join(USER_SCOPE_FILE),
+        )
+    }
+    #[cfg(windows)]
+    {
+        let root = std::env::var_os("ProgramData")?;
```
clement-tourriere, line 54: On windows, normal users can create folders in `C:\ProgramData` by default. If the admin never created `gitguardian` here, any user can create `C:\ProgramData\gitguardian\secrets.env`, and it's merged into every other user's run/get/hook, without trust or owner check. Should check the owner/ACL (admins or SYSTEM only) before reading it.

rust/secrets/src/file/atomic.rs
```diff
@@ -0,0 +1,1008 @@
...
+        .metadata()
+        .map(|metadata| metadata.permissions().mode() & 0o7777)
+        .unwrap_or(NEW_FILE_MODE);
+    temp.as_file()
+        .set_permissions(std::fs::Permissions::from_mode(mode))
+        .context("setting file permissions")
+}
+
+#[cfg(not(unix))]
+fn set_permissions(temp: &tempfile::NamedTempFile, locked: &File) -> Result<()> {
```
clement-tourriere, line 516: On windows `Permissions` is only the readonly flag, not the ACL. The temp file takes the directory ACL, so if the `.env` had a restrictive DACL it's lost after the first rewrite. New stores get no ACL at all either.

rust/secrets/src/file.rs
```diff
@@ -0,0 +1,2157 @@
...
+        layers: Vec<Layer>,
+    ) -> Result<(BTreeMap<String, SecretString>, ReadWarnings)> {
+        if layers.is_empty() {
+            return Err(SecretError::SecretNotFound {
+                path: project_path.to_string(),
+            }
+            .into());
+        }
+        // A stray quote loses variables at parse time, so reads must report it too.
+        let advisories = layers
```
clement-tourriere, line 327: All quote problems become advisories here, also `Unterminated` and `TrailingAfterQuote`, which drop the variable. So with `API_KEY='abc'tail` + `PORT=80`, `run` starts with only PORT, 0 unreadable, exit 0. That's exactly the "never inject a partial file" case run.rs tries to avoid. I think the 2 fatal ones should count as unreadable for run/activate.

rust/secrets-cli/src/commands/activate.rs
```diff
@@ -0,0 +1,1764 @@
...
+
+    // Stop rather than walk past a symlink or fifo and load another project's secrets.
+    if let Some(refusal) = &found.refusal {
+        report_always(shell, refusal);
+        return emit_state(&script, &found, Vec::new(), shell);
+    }
+
+    // Entering a directory is not consent the way naming a file to `get`/`run` is.
+    // Not a prompt: a y/n on every `cd` gets answered without reading.
+    if found.needs_trust && !trust::is_trusted(&found.path)? {
```
clement-tourriere, line 519: If `is_trusted` fails (trust file unreadable etc), the `?` returns before we print `script`, which has the unsets of the previous project. So going from A to B, A's secrets stay in the shell. Should be handled like the other refusals with `emit_state`.

Small thing too: here we hash the file and `load()` line 536 reads it again, so what we load is not exactly what was trusted. Better to read once, then hash and parse the same bytes.

rust/secrets/src/store.rs
```diff
@@ -0,0 +1,955 @@
...
+            .collect::<Vec<_>>();
+        if !missing.is_empty() {
+            return Err(SecretError::missing_fields(&missing).into());
+        }
+
+        for key in keys {
+            existing.remove(key);
+        }
+        if existing.is_empty() {
+            self.send_empty("delete_secret", params)
```
clement-tourriere, line 358: When we remove the last field this is a DELETE without CAS. If someone writes a new version between our read and this call, we soft delete his version, with fields the user never asked to delete. Could use `POST <mount>/delete/<path>` with `{"versions": [version]}` to keep the protection.

rust/secrets/src/store.rs
```diff
@@ -0,0 +1,955 @@
...
+    }
+
+    fn set_vault_secrets(
+        &self,
+        params: &BTreeMap<String, String>,
+        fields: &BTreeMap<String, SecretString>,
+    ) -> Result<()> {
+        let (mut merged, version) = match self.read_vault_secret_version(params)? {
+            Some(current) => current,
+            None => (BTreeMap::new(), self.read_vault_current_version(params)?),
```
clement-tourriere, line 325: Here we read metadata when data is 404. A lot of policies only give `secret/data/app*` (create/read/update) and not metadata, so `set` of a new secret fails with permission denied. The 404 body for a soft-deleted version already has `data.metadata.version`, and for a never-written one cas=0 is fine, so I think we don't need this call.

Also a small race: if someone creates the secret between the 2 calls, we pair an empty map with his version, CAS passes, and his fields are lost.

rust/secrets/src/store.rs
```diff
@@ -0,0 +1,955 @@
...
+
+        let config = ureq::Agent::config_builder()
+            .timeout_connect(Some(Duration::from_secs(10)))
+            .timeout_global(Some(Duration::from_secs(30)))
+            // A redirect would carry the auth header to wherever the server points.
+            .max_redirects(0)
+            // Otherwise 4xx/5xx become transport errors and 404 -> SecretNotFound never runs.
+            .http_status_as_error(false)
+            .build();
+        let agent = ureq::Agent::new_with_config(config);
```
clement-tourriere, line 549: We ignore `VAULT_NAMESPACE`, `VAULT_CACERT`/`VAULT_CAPATH` and `VAULT_CLIENT_CERT`. Enterprise and HCP users (who need the `admin` namespace) can't use it, or worse, silently read the root namespace. And `--allow-self-signed`/`--insecure` are eaten by the dispatcher and ignored. At least error if `VAULT_NAMESPACE` is set?

rust/secrets/src/providers.rs
```diff
@@ -0,0 +1,152 @@
...
+        token_file_env: Some("VAULT_TOKEN_FILE_PATH"),
+        token_file: Some("~/.vault-token"),
+        header: "X-Vault-Token",
+    },
+    endpoints: &[
+        (
+            "read_secret",
+            Endpoint {
+                method: Method::Get,
+                path: "/v1/${mount}/data/${path}",
```
clement-tourriere, line 24: On a KV v1 mount, `set` POSTs to `/v1/secret/data/app`. v1 accepts it and creates a secret named `data/app` containing `{"data":..,"options":..}`. Silent wrong write, and the next `set` fails with a strange error. Maybe detect with `sys/internal/ui/mounts/<path>` like the vault cli, or fail when the read response has no `metadata`.

rust/secrets-cli/src/commands/import.rs
```diff
@@ -0,0 +1,121 @@
...
+    if !args.yes {
+        confirm_existing_fields(&store, &path, &keys)?;
+    }
+    store.set_secrets(&path, &fields)?;
+    eprintln!("imported {} at {path}", field_count(fields.len()));
+
+    // Only after the write succeeded, or the values are lost.
+    if let Some(input) = args.input.as_deref() {
+        if args.remove_source {
+            std::fs::remove_file(input).with_context(|| format!("removing {}", input.display()))?;
```
clement-tourriere, line 74: `remove_file` doesn't check that the file is still the one we read (we read it before the prompt and the vault call). If an editor saves it in between, we delete values we never imported. Could compare len/mtime/hash before removing. Also, if the input is a symlink we remove only the link but print "removed".

rust/secrets/src/file/keystore.rs
```diff
@@ -0,0 +1,142 @@
...
+use std::path::PathBuf;
+
+use anyhow::{Context, Result};
+use zeroize::Zeroizing;
+
+use super::atomic;
+use super::crypto::Keyset;
+
+/// Keyring service name, shared with the provider token lookups.
+pub(crate) const KEYRING_SERVICE: &str = "gitguardian";
```
clement-tourriere, line 17: There is only one copy of the key, in the keyring, and no export/backup/rotate command. The changelog pushes `import --remove-source`, so after a keychain reset (password reset on mac, new laptop...) all the values are lost forever. Maybe add an export/recovery command before we promote `--remove-source`? Or at least say it clearly in the doc.

rust/secrets-cli/src/commands/set.rs
```diff
@@ -0,0 +1,90 @@
...
+    for key in &args.keys {
+        // Never from argv: that lands in shell history and every process listing.
+        let value = prompt_secret(key, args.expose, args.allow_empty)?;
+        let value = SecretString::from(value);
+        // A NUL encrypts fine but can never be injected by `run` or the hook,
+        // so refuse it here rather than report a useless success.
+        validate_env_value(key, &value)?;
+        fields.insert(key.clone(), value);
+    }
+    for warning in store.set_secrets_with_warnings(&path, &fields, confirmed.as_ref())? {
```
clement-tourriere, line 80: In a repo, `set` writes to the repo store by default, but the project `.env` wins over the repo store. So if `.env` has `API_KEY=old`, `set API_KEY` succeeds but `run` still gives `old`. Bad when rotating a leaked key. `unset` already warns in this case, `set` should too (or refuse).

rust/secrets/src/file.rs
```diff
@@ -0,0 +1,2157 @@
...
+            Some(problem @ QuoteProblem::SwallowedAssignment { .. }) => {
+                outcome.warnings.push(format!("{path}: {problem}"));
+            }
+            None => {}
+        }
+
+        // Loaded lazily: minting the device key must not happen for a pass with nothing to seal.
+        let mut cipher: Option<Arc<dyn Cipher + Send + Sync>> = None;
+        let mut cipher_loaded = false;
+        let changed = document.map_values(|key, value| -> Result<Option<String>> {
```
clement-tourriere, line 533: `map_values` only visits valid entries, so lines like `MY-KEY=secret` or `spring.datasource.password=x` stay in plaintext, and we print "encrypted N" like everything is fine. Should at least warn that some lines were not encrypted.

rust/secrets-cli/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+    Some(Invocation {
+        args: args[index..].to_vec(),
+        config_path,
+    })
+}
+
+fn is_native_verb(args: &[OsString]) -> bool {
+    match args.first().and_then(|arg| arg.to_str()) {
+        Some("secret") => matches!(
```
clement-tourriere, line 92: This only looks at `args[1]`, so `ggshield secret -v get` or `ggshield secret --debug list` go to python, and the stub says the native exe "is not included in this installation", even in the bundle. Python accepts the common options on the `secret` group. Could skip `ROOT_FLAGS` after `secret` too.

rust/dispatcher/tests/secret.rs
```diff
@@ -0,0 +1,3666 @@
...
+        } else if cfg!(windows) {
+            home.join("AppData/Roaming/gitguardian/secrets.env")
+        } else {
+            home.join(".config/gitguardian/secrets.env")
+        }
+    }
+}
+
+/// Whether `shell` can run here. A missing shell is skipped locally and fatal
+/// under `$GITGUARDIAN_REQUIRE_SHELLS`, which CI sets.
```
clement-tourriere, line 193: The comment says CI sets `GITGUARDIAN_REQUIRE_SHELLS`, but I can't find it anywhere in `.github/workflows`. Ubuntu runners have no fish/zsh by default, so the injection tests are skipped silently in CI. Can we set it and install the shells in `ci.yml`?

rust/secrets-cli/src/commands/run.rs
```diff
@@ -0,0 +1,190 @@
...
+/// On Unix, exec so signals, exit code and streams behave as if run directly.
+#[cfg(unix)]
+fn exec(mut child: std::process::Command, program: &str) -> Result<()> {
+    use std::os::unix::process::CommandExt;
+    Err(child.exec()).with_context(|| format!("running {program}"))
+}
+
+#[cfg(not(unix))]
+fn exec(mut child: std::process::Command, program: &str) -> Result<()> {
+    let status = child
```
clement-tourriere, line 137: Windows, 2 things:
- `Command::new("npm")` only searches `npm.exe`, so `ggshield run -- npm run dev` fails with program not found (npm, yarn and pnpm are `.cmd`).
- There is no Ctrl-C handler, so on Ctrl-C we exit right away with 0xC000013A. The prompt comes back while the child is still running and its exit code is lost. `dispatch.rs` already has `keep_self_alive_through_ctrl_c`.

rust/secrets/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+[dependencies]
+anyhow = "1"
+clap = { version = "4", features = ["derive"], optional = true }
+percent-encoding = "2"
+secrecy = "0.10"
+serde = { workspace = true }
+serde_json = { workspace = true }
+thiserror = "2"
+ureq = { workspace = true }
+chacha20poly1305 = "0.11"
```
clement-tourriere, line 25: `chacha20poly1305` without the `zeroize` feature, so the derived key inside `XChaCha20Poly1305` is not wiped on drop. We zeroize everything around it, so it would be good to enable: `features = ["zeroize"]`.

nit: `hkdf = "0.12"` brings a second hmac/sha2 generation, main already has hkdf 0.13.

rust/secrets/src/file/crypto.rs
```diff
@@ -0,0 +1,685 @@
...
+    fn stored_len(&self) -> usize {
+        let framing = r#"{"current":"","keys":{}}"#.len() + KEY_ID_HEX_LEN;
+        let per_key = KEY_ID_HEX_LEN + ENCODED_KEY_LEN + r#""":"""#.len();
+        framing + self.keys.len() * per_key + self.keys.len().saturating_sub(1)
+    }
+
+    /// Parse a blob from [`Keyset::to_bytes`], borrowing so no key is copied.
+    pub(crate) fn from_bytes(bytes: &[u8]) -> Result<Self> {
+        let stored: StoredKeyset<'_> =
+            serde_json::from_slice(bytes).context("the stored keyset is not valid JSON")?;
```
clement-tourriere, line 125: If the keyring blob is not what we wrote (tampered or corrupted), the serde error can contain the string value (`invalid type: string "<key>", expected a borrowed string` when there is an escape), and the base64 error at line 134 gives a char + offset. It goes to stderr with `{:#}`. Better to `map_err` to a fixed message without the source.

rust/secrets-cli/src/env.rs
```diff
@@ -0,0 +1,245 @@
...
+    let content = raw.trim();
+    let candidate = content
+        .strip_prefix("export ")
+        .unwrap_or(content)
+        .trim_start();
+    if let Some((key, _)) = candidate.split_once('=')
+        && let Err(error) = validate_env_key(key.trim())
+    {
+        // Flattened: the caller prints with `{}`, which would hide a context layer.
+        return anyhow::anyhow!("line {line_number}: {error}");
```
clement-tourriere, line 72: Here the "key" is everything before the first `=`, so `postgres://u:S3cret@h/db?sslmode=require` prints `'postgres://u:S3cret@h/db?sslmode' is not a valid env var name`. Password in the CI logs. Maybe just say line N is not valid, without the content.

rust/secrets/src/file/trust.rs
```diff
@@ -0,0 +1,282 @@
...
+    };
+    let canonical = canonical(path);
+    Ok(load()?
+        .iter()
+        .any(|entry| entry.path == canonical && entry.digest == digest))
+}
+
+/// Approve `path`'s current contents, returning false if it was already
+/// approved unchanged.
+pub fn trust(path: &Path) -> Result<bool> {
```
clement-tourriere, line 36: `trust`/`untrust` do load, modify, save without a lock. `trust B` running in parallel with `trust --revoke A` can bring A back. Maybe reuse the lock helper from atomic.rs.

nit: line 145 writes the path raw, so a dir name with `\n` can inject another line. Hard to exploit, but easy to refuse control chars.

rust/secrets/src/dotenv/render.rs
```diff
@@ -0,0 +1,234 @@
...
+
+use super::Quote;
+
+/// An allowlist, not a denylist: a bare value must be a shell word that expands to itself.
+fn bare_safe(ch: char) -> bool {
+    // Every shell metacharacter is ASCII, so Unicode alphanumerics stay bare.
+    ch.is_alphanumeric()
+        || matches!(
+            ch,
+            '-' | '_' | '.' | ':' | '/' | '@' | '+' | ',' | '=' | '%'
```
clement-tourriere, line 12: nit: `=` is bare-safe, but in zsh a leading `=` is expanded: a `.env` with `A==ls` gives `/bin/ls` when sourced, and `A==nope` stops the `source` with "nope not found" (so nothing after is set). Tested. Only for `--plain` values, but the doc above says the output must be inert to a shell.

rust/secrets-cli/src/commands/activate.rs
```diff
@@ -0,0 +1,1764 @@
...
+            let array = match hook {
+                Hook::Prompt => "precmd_functions",
+                _ => "chpwd_functions",
+            };
+            format!("{removal}{array}+=(_ggshield_hook)\n")
+        }
+
+        // `PROMPT_COMMAND` may be a string (bash < 5.1) or an array. Trailing separators
+        // are trimmed first: appending to `...;` would yield `;;`, a syntax error on every prompt.
+        (Shell::Bash, _) => format!(
```
clement-tourriere, line 323: nit: `--hook pwd` is the default, but for bash we always use `PROMPT_COMMAND`, so it runs on every prompt anyway (1 process per prompt). Maybe fine, but the doc says pwd = on directory change only.

rust/secrets-cli/src/commands/activate.rs
```diff
@@ -0,0 +1,1764 @@
...
+            let mode = match hook {
+                Hook::Prompt => "prompt",
+                _ => "pwd",
+            };
+            format!(
+                r#"{removal}if ("$function:prompt" -notlike '*__ggshield_hook_mode*') {{
+  $global:__ggshield_original_prompt = $function:prompt
+  function global:prompt {{
+    if ($global:__ggshield_hook_mode -ceq 'prompt' -or ($global:__ggshield_hook_mode -ceq 'pwd' -and $PWD.Path -cne $global:__ggshield_pwd)) {{
+      $null = _ggshield_hook
```
clement-tourriere, line 356: `_ggshield_hook` runs before the original prompt, so `$?` is always true for themes (starship and oh-my-posh read it). `LASTEXITCODE` is restored but not `$?`. Not tested, I have no pwsh here.

rust/dispatcher/src/main.rs
```diff
@@ -16,9 +17,23 @@ fn main() {
...
+    }
     dispatch::delegate(&args);
 }
 
+/// Unlike the hook there is no Python fallback: these verbs exist only here.
+fn run_secret_verb(args: &[std::ffi::OsString]) -> i32 {
+    match ggshield_secrets_cli::run(args) {
+        Ok(()) => 0,
+        Err(error) => {
+            eprintln!("Error: {error:?}");
```
clement-tourriere, line 31: nit: `{:?}` on anyhow prints the backtrace when `RUST_BACKTRACE=1`. And exit 1 means SCAN_FOUND_PROBLEMS for ggshield. For `run`, failing to start the command gives 1 too, so you can't tell it from the child's exit 1. Maybe `{:#}`, 128 like python for unexpected errors, and 127 for command not found.

rust/secrets/src/store.rs
```diff
@@ -0,0 +1,955 @@
...
+
+        Ok(SecretStore {
+            backend: Backend::Http(Box::new(Http { definition, agent })),
+            env_override: self.env_override.unwrap_or(true),
```
clement-tourriere, line 553: nit: the SDK default is `env_override = true`, so `get_secret("secret/db", "USER")` returns my login name without calling vault (same for HOME, PATH...). The example relies on this default. I would default to false and let callers opt in.

rust/secrets-cli/src/commands/unset.rs
```diff
@@ -0,0 +1,296 @@
...
+            "delete cancelled, no changes made: field not found in secret: VAR"
+        );
+        assert_eq!(
+            delete_cancelled_message(&["VAR".to_string(), "AWS".to_string()]),
+            "delete cancelled, no changes made: fields not found in secret: VAR, AWS"
+        );
+    }
+
+    #[test]
+    fn delete_all_missing_secret_message_names_path() {
```
clement-tourriere, line 288: nit: this test formats a string and asserts it equals the same literal, it tests nothing.

PR #1524, opened 2026-09-18
fix(client): return correct error for missing api key on self-hosted

6d7a:
## Context

When creating a client from a config for a self-hosted gitguardian instance, a missing API key currently prompts an `Unknown instance` error. This is misleading.

## What has been done

The error message for a missing api key in the client config is now the same between self-hosted and default SaaS instances (with a dynamic login command parameter).

## PR check list


ggshield/core/client.py
```diff
@@ -80,24 +80,20 @@ def create_client_from_config(
         api_key = config.api_key
         api_url = config.api_url
     except UnknownInstanceError as e:
-        if e.instance == DEFAULT_INSTANCE_URL:
```
benjaminrigaud-gg, line 83: (bonus) We could still display an hint about the --instance in that case

benjaminrigaud-gg (approved):

PR #1527, opened 2026-09-21
fix(install): detect and repair git hooks left by an older ggshield

amascia-gg:
`ggshield install --mode global` writes one hook into a shared `core.hooksPath` directory, so that hook has to invoke the repository's own hook itself. Before v1.53.0 it did so through a hardcoded `.git/hooks/<type>`, which a linked worktree never matches (`.git` is a file there), and the repo hook was silently skipped. v1.53.0 fixed the template but nothing upgrades a hook already on disk: `hook_invokes_ggshield()` only greps for the `ggshield secret scan` marker, which a v1.12.0 hook carries just as much as a current one. So `machine doctor` reported ✓ and `machine setup` said "already configured" on every machine set up between v1.12.0 and v1.52.2, and `doctor`'s exit code gates MDM rollouts.

Generated hooks now carry a `# ggshield-hook-version:` stamp, and a missing or older stamp means stale. Repair replaces only the block ggshield generated, byte-matched against the three renderings that shipped (bash `[[ ]]`, posix `.git/hooks/...`, and the current snippet without a stamp): blind `--force` regeneration would destroy the user-owned content in an `install --append` hook. A marked hook with no recognizable block was hand-edited, so it is reported and left alone.

`doctor` now fails on a stale hook, and `setup` reports the repair instead of "already configured". Same function, related: `_setup_git_hooks` returned success when it left a foreign hook untouched, so `setup` exited 0 on a machine with no ggshield hook at all.

Verified against a real v1.52-era hook with a user line in it: `doctor` fails, `setup` rewrites the block, the user's line survives, `doctor` passes.

NHI-2028


ggshield/cmd/install.py
```diff
@@ -204,6 +235,65 @@ def _get_repo_root() -> Optional[Path]:
...
 
+def _hook_body(hook_type: str, local_hook_support: bool) -> str:
+    """The lines ggshield owns in a hook script, stamp first."""
+    snippet = (
+        LOCAL_HOOK_SNIPPET.format(hook_type=hook_type) + "\n"
+        if local_hook_support
+        else ""
+    )
+    return (
+        f"# ggshield-hook-version: {GGSHIELD_HOOK_VERSION}\n"
```
benjaminrigaud-gg: We need a separator there 🤔 

> Preserve the newline before the generated block — [install.py:291](https://github.com/GitGuardian/ggshield/blob/bc66c62ac8121085d9aa3683f73f66592e90fc1b/ggshield/cmd/install.py#L291). For hooks originally appended to a file without a trailing newline,
    replacement consumes the separator and joins the version stamp to the preceding command. I reproduced true becoming true# ggshield-hook-
    version: 1; with set -e, the repaired hook exits 127 before scanning. Fresh global --append installations have the same regression. Preserve
    the line boundary during both append and repair.

ggshield/cmd/machine/doctor.py
benjaminrigaud-gg: Should we only check the effective directory checked by git?

> The suggested repair doesn’t resolve it: running machine setup as that user checks the global directory, reports “already configured,” and succeeds. Doctor still fails afterward.

benjaminrigaud-gg (changes requested):

ggshield/cmd/install.py
amascia-gg: You're right that it isn't only repair: `--append` regressed the same way, because it used to write a snippet that began with the newline and now writes a body that begins with the stamp.

Fixed in one place rather than two: `_hook_body()` now opens on a newline. That makes the replacement newline-for-newline with every legacy body it substitutes for, and hands the append path its separator back.

Added two regression tests

ggshield/cmd/machine/doctor.py
amascia-gg: Good catch! For the fix, I went for symmetry rather than resolving the effective directory.

`_current_ggshield_hook_present()` now mirrors the presence check: a hook type is stale only when no directory holds a current one. In your scenario the machine is genuinely protected (git runs the global hook).

benjaminrigaud-gg: doctor is still able to detect the system hooks are fine even if git is using some global stale one :thinking:

amascia-gg: You're right, and my "symmetry" argument was wrong. `any(current)` just moves the false negative to the other side: stale global + current system passes, git runs the stale global one, and the machine is unprotected. Same class of bug as the one this PR exists to fix, which is a bit embarrassing.

Both of my attempts were asking the wrong question. "Does some directory hold a good hook" has no correct answer, because git runs exactly one directory. 660e5d18 asks which one instead: walk the dirs in git's precedence order (global before system) and judge the first ggshield hook found. Anything behind it is dead config.

| global | system | verdict |
|---|---|---|
| current | stale | ok |
| stale | current | **fail** |
| stale | -- | fail |
| -- | stale | fail |
| -- | current | ok |

That's your original suggestion, arrived at the long way round. Parametrized over exactly that table so neither direction can regress.

Still scoped to the machine-level dirs rather than the fully resolved `core.hooksPath`, so a repo-local override (Husky) stays with `_check_git_hooks_precedence` where it was. Say the word if you'd rather have that folded in too.

amascia-gg: Right again, and reproduced both outcomes before touching anything. Fixed in 6b5580a6.

I'd been fixing the predicate three rounds running and never questioned the input. The bug was always `configured or default`: the `or default` half invents a directory git is not using, and then no amount of cleverness over that list can be correct. `doctor` now does what you suggested from the start.

```py
hook_dir = get_global_hook_dir_path() or get_system_hook_dir_path()
```

Neither set is now reported as "not configured", which is the truth: git runs each repository's own hooks. That also drops `_effective_ggshield_hook` and the premise behind `test_ok_when_split_across_global_and_system` -- git reads one directory, so a hook split across both scopes never fully runs, and that test was asserting a fiction. It now asserts the failure.

While reproducing yours I found the same root cause in `setup`, which you didn't report:

```
GIVEN leftover ggshield hooks in the default global dir, core.hooksPath unset
  global pre-commit hook updated to the current version
setup returned: True
core.hooksPath after setup -> None
```

A leftover hook makes setup skip the installer, and the installer is the only thing that sets `core.hooksPath`. So setup repaired the file, reported success, and left git running no hook at all -- and after your doctor fix the two would have disagreed forever. Setup now points git at the directory whenever the scope has none.

One more thing worth flagging: writing that config in `_setup_git_hooks` meant the unit tests started editing the developer's real `~/.gitconfig` (it put a pytest tmp dir in mine). The setup tests now pin `git` in an autouse fixture so the suite cannot reach it.

benjaminrigaud-gg (approved):
It could work :santa:

